PR #1, opened 2025-11-13
lib: start adding lib helpers

ellemouton:
(this adds a subset of what was in the og lib code)


ellemouton:
@gemini-code-assist review this

ellemouton:
@gemini-code-assist review this

lib/scripts/anchors.go
```diff
@@ -0,0 +1,23 @@
...
+
+import "github.com/btcsuite/btcd/wire"
+
+var (
+	// ANCHOR_PKSCRIPT a standard P2A script. This is a keyless output.
+	// This corresponds to address: bc1pfeessrawgf.
+	// The script path for this output requires no witness elements, meaning
+	// that it can be spent by anyone who provides the script and control block.
+	ANCHOR_PKSCRIPT = []byte{
+		0x51,       // OP_1
```
Roasbeef: Will we retain the same rules from lnd re no inline comments?

Roasbeef: Also if you help me review https://github.com/btcsuite/btcd/pulls. Then we can start to use all the new utilities for p2a, etc. 

Same with https://github.com/btcsuite/btcd/pull/2432.

lib/scripts/vtxo.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/lightningnetwork/lnd/lntypes"
+)
+
+const (
+	// CollabMultisigLeafWitnessSize is the estimated witness size for the
+	// collaborative spend path of a boarding or vtxo output.
+	// The size is calculated as:
+	// 1 + 64 bytes for client signature + length byte
+	// 1 + 64 bytes for server signature + length byte
+	CollabMultisigLeafWitnessSize = lntypes.WeightUnit(1 + 64 + 1 + 64)
```
Roasbeef, line 51: Didn't realize we had this stuff in `lntypes`, any reason to prefer it over what we have in `chainfee`? https://github.com/lightningnetwork/lnd/blob/f6005ed350352d25494d1ecf4775b3266318bbc4/lnwallet/chainfee/rates.go#L59-L60

lib/scripts/vtxo.go
```diff
@@ -0,0 +1,296 @@
...
+func NewVTXOSpendInfo(tapscript *waddrmgr.Tapscript,
+	leaf VTXOLeafType) (*VTXOSpendData, error) {
+
+	leafIndex, err := leaf.LeafIndex()
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the leaf.
+	targetLeaf := tapscript.Leaves[leafIndex]
```
Roasbeef, line 172: Should have a defensive check here re the length to avoid a panic.

lib/scripts/vtxo.go
```diff
@@ -0,0 +1,296 @@
...
+	builder := txscript.NewScriptBuilder()
+
+	// Ensure the proper party can sign for this output.
+	builder.AddData(schnorr.SerializePubKey(exitKey))
+	builder.AddOp(txscript.OP_CHECKSIG)
+
+	// Assuming the above passes, then we'll now ensure that the CSV delay
+	// has been upheld, dropping the int we pushed on. If the sig above is
+	// valid, then a 1 will be left on the stack.
+	builder.AddInt64(int64(csvDelay))
```
Roasbeef: Non-blocking, but you might want to consider using this new script creation function: https://pkg.go.dev/github.com/btcsuite/btcd/txscript#ScriptTemplate. 

You can see it in action here in this PR: https://github.com/lightningnetwork/lnd/pull/9639/files#diff-5417cb1a81f342c8a87e557eb6d82431a0f451123c73cb638c361bf55ee41201L167-L174. It's a bit more terse/flexible. Was inspired by some of the script utilities the devs making STARKS on Bitcoin were using earlier this year.

lib/scripts/vtxo_test.go
```diff
@@ -0,0 +1,362 @@
...
+			assertEngineExecution(t, i, testCase.valid, newEngine)
+		})
+	}
+}
+
+// assertEngineExecution executes the VM returned by the newEngine closure,
+// asserting the result matches the validity expectation. In the case where it
+// doesn't match the expectation, it executes the script step-by-step and
+// prints debug information to stdout.
+func assertEngineExecution(t *testing.T, testNum int, valid bool,
```
Roasbeef: 👍

lib/tree/signing.go
```diff
@@ -0,0 +1,515 @@
...
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/input"
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+type Musig2PubNonce [musig2.PubNonceSize]byte
+
+// TxSignerCoordinator represents a signing session for a single VTX transaction.
+type TxSignerCoordinator struct {
```
Roasbeef: With the split we're going for rn, I wonder how much of this should be in the other repo? 

Re client vs server.

lib/tree/signing.go
```diff
@@ -0,0 +1,515 @@
...
+	keys        []*btcec.PublicKey
+	nonces      map[string]Musig2PubNonce
+	partialSigs map[string]*musig2.PartialSignature
+
+	sigHash  [32]byte
+	aggNonce Musig2PubNonce
+	aggKey   *btcec.PublicKey
+}
+
+func NewTxSignerCoordinator(node *Node, sweepTweek []byte,
```
Roasbeef: Blanket `godoc` comment. Won't leave it anywhere else in the PR.

Roasbeef: What's `sweepTweak` here? Should we break it out into its own type alias or somethign like that?

Roasbeef: Ah ok it's just the tapscript root. Let's call it maybe `sweepTapscriptRoot`?

lib/tree/signing.go
```diff
@@ -0,0 +1,515 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	sigHash, err := txscript.CalcTaprootSignatureHash(
+		txscript.NewTxSigHashes(tx, prevOutFetcher),
+		txscript.SigHashDefault, tx,
+		0,
+		prevOutFetcher,
```
Roasbeef: Can prob go on the line above. Won't litter w/ style comments tho.

lib/tree/signing.go
```diff
@@ -0,0 +1,515 @@
...
+// TxSignerCoordinator represents a signing session for a single VTX transaction.
+type TxSignerCoordinator struct {
+	// cosigners is the list of public keys of the cosigners involved in
+	// signing this transaction's input.
+	cosigners   map[string]*btcec.PublicKey
+	sweepTweek  []byte
+	keys        []*btcec.PublicKey
+	nonces      map[string]Musig2PubNonce
+	partialSigs map[string]*musig2.PartialSignature
+
```
Roasbeef: Seems like we can entirely use: https://github.com/lightningnetwork/lnd/blob/f6005ed350352d25494d1ecf4775b3266318bbc4/input/musig2.go#L42-L86 here? 

Concrete version being: https://github.com/lightningnetwork/lnd/blob/f6005ed350352d25494d1ecf4775b3266318bbc4/input/musig2_session_manager.go#L34-L43. 

If we don't find the multi-session stuff useful, then there's also: https://pkg.go.dev/github.com/btcsuite/btcd/btcec/v2@v2.3.6/schnorr/musig2#Session. Kon's latest changes adds the ability to accept the final aggregated nonces, etc.

lib/tree/signing.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+	t.aggNonce = aggNonce
+
+	return aggNonce, nil
+}
+
+func (t *TxSignerCoordinator) AddPartialSignature(signer *btcec.PublicKey,
+	sig *musig2.PartialSignature) error {
+
+	if _, ok := t.cosigners[toHexKey(signer)]; !ok {
```
Roasbeef: Re above, this can just call `session.CombineSig` which handles the validation, etc.

lib/tree/signing.go
```diff
@@ -0,0 +1,515 @@
...
+				txid, err)
+		}
+
+		sigs[txid] = sig
+	}
+
+	return sigs, nil
+}
+
+type TxSignerSession struct {
```
Roasbeef: Ah ok I see this partially uses the stuff linked above. I also see a clearer client vs server line here too.

lib/tree/signing.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+
+	if !ok {
+		return fmt.Errorf("not all nonces registered successfully")
+	}
+
+	return nil
+}
+
+// TODO: need to update the musig2 lib  & input lib to allow setting aggregate
```
Roasbeef: 👍

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr/musig2"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// Node represents a node in a transaction tree. It is made up of a
+// transaction along with the children transactions that spend its outputs.
+type Node struct {
```
Roasbeef, line 20: Very nice 😎

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+	}
+
+	return nil
+}
+
+// ForEach traverses the tree and applies the given callback function to each
+// node. The traversal is done in depth-first order (pre-order: current node,
+// then children). If the callback returns an error, traversal stops and the
+// error is returned.
+func (n *Node) ForEach(callback func(*Node) error) error {
```
Roasbeef, line 349: Nice

Roasbeef: One thing to consider for both `ForEach` and `ForEachLeaf` is making a complementary iterator (`iter.Seq`), or adding `*Iter` methods along side.

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+	if n == nil {
+		return nil
+	}
+
+	// Apply callback to current node.
+	if err := callback(n); err != nil {
```
Roasbeef, line 351: Re the above, this stays mostly the same, you just `yield(n)` instead.

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+func (n *Node) GetLeafForCosigner(targetKey *btcec.PublicKey) (*Node, error) {
+	if n == nil {
+		return nil, nil
+	}
+
+	var foundLeaf *Node
+
+	err := n.ForEachLeaf(func(leaf *Node) error {
+		if ContainsCosigner(leaf.CoSigners, targetKey) {
+			foundLeaf = leaf
```
Roasbeef: Above comment means this could just be a break from the loop Once the target key has been found.

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+		return nil, fmt.Errorf("failed to create base transaction: %w", err)
+	}
+
+	// Add the signature as witness data for the single input
+	// For taproot keyspend, the witness consists of just the signature
+	if len(tx.TxIn) != 1 {
+		return nil, fmt.Errorf("expected exactly 1 input, got %d", len(tx.TxIn))
+	}
+
+	tx.TxIn[0].Witness = wire.TxWitness{n.Signature.Serialize()}
```
Roasbeef: Also needs to consider the witness script and the control block?

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+
+// VerifySigned verifies that all nodes in the tree have valid signatures.
+// It checks that:
+// 1. Every node has a signature
+// 2. Each signature is valid for its transaction
+//
+// Parameters:
+//   - sweepTweak: The taproot tweak bytes used in the musig2 signing process
+//   - prevOutFetcher: Fetcher for previous outputs needed for signature verification
+func (n *Node) VerifySigned(sweepTweak []byte,
```
Roasbeef: Hmm, I thought `sweepTweak` would differ for each node as there may be a diff set of signers?

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+
+	// Find the operator key (appears in all nodes)
+	operatorKey := n.findOperatorKey(allKeys)
+
+	// Assign aliases: operator first, then participants
+	for keyStr := range allKeys {
+		if _, exists := keyMap[keyStr]; !exists {
+			var alias string
+			if keyStr == operatorKey {
+				alias = "OP"
```
Roasbeef: Not sure where this is used, but perhaps we should use some enum or two maps instead to distinguish the operator from all the other participants?

lib/tree/node.go
```diff
@@ -0,0 +1,877 @@
...
+	}
+
+	for _, child := range n.Children {
+		child.collectAllKeys(allKeys)
+	}
+}
+
+// findOperatorKey identifies the operator key as the one that appears in all
+// leaf nodes
+func (n *Node) findOperatorKey(allKeys map[string]bool) string {
```
Roasbeef: Re the above, if we have an enum or a dual map structure, then we don't need a method like this.

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+
+// Tree represents a complete transaction tree with all the context needed
+// for signing operations. It wraps a Node root with additional metadata
+// required for musig2 signing sessions.
+type Tree struct {
+	// Root is the root Node of the transaction tree
+	Root *Node
+
+	// SweepKey is the public key used in the sweep script tweak
+	SweepKey *btcec.PublicKey
```
Roasbeef: From who's PoV?

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+	// Create the tree signer session
+	return NewTreeSignerSession(
+		wallet, signerKeyDesc, sweepTapLeaf.Script,
+		prevOutFetcher, t.Root,
+	)
+}
+
+// NewTreeSignerCoordinator creates a new TreeSignerCoordinator for this tree.
+// This encapsulates the setup logic for the coordinator.
+func (t *Tree) NewTreeSignerCoordinator() (*TreeSignerCoordinator, error) {
```
Roasbeef: Client server split.

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+	return &Tree{
+		Root:       root,
+		SweepKey:   nil, // Connector trees don't have sweep keys
+		SweepDelay: 0,   // Connector trees don't have sweep delays
+		PrevOut:    prevOut,
+	}, nil
+}
+
+func BuildVTXOTree(
+	// input is the outpoint in the commitment transaction that the
```
Roasbeef: Similar comment here re comment style.

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+
+func buildTree(input *wire.OutPoint, leaves []Leaf, radix int,
+	tweak []byte, extraCosigner fn.Option[*btcec.PublicKey]) (*Node,
+	error) {
+
+	if len(leaves) == 0 {
+		return nil, fmt.Errorf("empty leaves")
+	}
+
+	// Sort receivers by amount (descending) using LPT heuristic for better
```
Roasbeef: LPT?

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+	// balance.
+	indexed := make([]indexedLeaf, len(leaves))
+	for i, r := range leaves {
+		indexed[i] = indexedLeaf{
+			Leaf:          r,
+			originalIndex: i,
+		}
+	}
+	sort.Slice(indexed, func(i, j int) bool {
```
Roasbeef: Does the sort need to be stable?

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+		return indexed[i].originalIndex < indexed[j].originalIndex
+	})
+
+	sortedLeaves := make([]Leaf, len(indexed))
+	for i, il := range indexed {
+		sortedLeaves[i] = il.Leaf
+	}
+
+	// Initialize queue with root work item
+	queue := []workItem{{
```
Roasbeef: Could consider copying this: https://github.com/Roasbeef/btcd/blob/0fcb3feb219adc05bfa41d3c98759e772ae3fd94/mempool/txgraph/collections.go#L80-L238.

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+}
+
+// indexedLeaf wraps a Leaf with its original index for stable sorting.
+type indexedLeaf struct {
+	Leaf
+	originalIndex int
+}
+
+// workItem represents a unit of work in the tree building queue
+type workItem struct {
```
Roasbeef: Can declare this before it's used.

lib/tree/tree.go
```diff
@@ -0,0 +1,601 @@
...
+		outputs = append(outputs, &wire.TxOut{
+			Value:    amount,
+			PkScript: inputScript,
+		})
+
+		allCosigners = append(allCosigners, groupCosigners...)
+	}
+
+	// Add anchor output.
+	outputs = append(outputs, scripts.AnchorOutput())
```
Roasbeef: Do we have an implicit sort order here? If not, then how do all parties get the same tree?

lib/tree/tree_test.go
```diff
@@ -0,0 +1,2551 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+// TODO(elle): re-write these claude-written tests.
```
Roasbeef: Code cov can be a good guide. Then property tests to make sure test cases can't be weaseled out of. Then mutation tests to make sure the tests are actually effective.

Roasbeef (commented):
Looking like a very solid base!

No major blocking comments, mainly wanting to update my mental model with the key primitives. 

I merged in some of the template repo code, so when you rebase, you can drop some comments re the makefile, linter, etc.

Roasbeef:
@claude review this

Roasbeef:
@codex review this

lib/scripts/vtxo.go
ellemouton: i guess just cause we use this type for the `input.TxWeightEstimator` 🤷

ellemouton: `func (twe *TxWeightEstimator) AddWitnessInput(
	witnessSize lntypes.WeightUnit) *TxWeightEstimator {`

lib/scripts/vtxo.go
ellemouton: very cool. updated

lib/tree/node.go
ellemouton: it's spending the keypath

lib/tree/node.go
ellemouton: sweepTweek is the script that pays to the operator. It is the same for each vtx transaction. It is only for the leaves (ie, the vtxos) that the tweak pays to the client

lib/tree/tree.go
ellemouton: we send parties the tree. For each vtxo they want, we send the tx packet from root to leaf

lib/tree/signing.go
ellemouton: the reason i could not use that is cause it seems like to create that, we need to call musig2.NewContext which takes a raw private key?

so iiuc, i tink we first want to expose that new change from kon in the MuSig2Signer?

lib/tree/node.go
ellemouton: removed this

ellemouton (commented):
thanks for review @Roasbeef  🙏 

split out the tree sign coordinator to the server side: https://github.com/lightninglabs/darepo/pull/36

also added some basic unit-type integration tests there to test some of the flows. 

Added `ark.go` here too... but havent actually used that in any flows anywhere so that is still quite rough. So can either leave and adapt in future as needed, or remove for now

.gitignore
```diff
@@ -2,3 +2,35 @@
...
+*.exe~
+*.dll
+*.so
+*.dylib
+
+# Test binary, built with 'go test -c'
+*.test
+
+# Test artifacts.
+**/*artifacts*/
```
bhandras: nit: not needed as it's already set above.

lib/scripts/anchors.go
```diff
@@ -0,0 +1,27 @@
...
+
+var (
+	// AnchorPkScript a standard P2A script. This is a keyless output.
+	// This corresponds to address: bc1pfeessrawgf.
+	// The script path for this output requires no witness elements, meaning
+	// that it can be spent by anyone who provides the script and control
+	// block.
+	AnchorPkScript = []byte{
+		// OP_1.
+		0x51,
```
bhandras: nit: this could be `txscript.OP_1, txscript.OP_DATA_2, 0x4e, 0x73`

lib/scripts/nums.go
```diff
@@ -0,0 +1,35 @@
...
+	// ARKNUMSKey is a NUMS key (nothing up my sleeves number) that has
+	// no known private key. This was generated using the following script:
+	// https://github.com/lightninglabs/lightning-node-connect/tree/
+	// master/mailbox/numsgen, with the seed phrase "Ark Protocol NUMS".
+	ARKNUMSKey = MustParsePubKey(ARKNUMSHex)
+)
+
+// MustParsePubKey parses a hex encoded public key string into a public key and
+// panic if parsing fails.
+func MustParsePubKey(pubStr string) btcec.PublicKey {
```
bhandras: nit: we could make this unexported so it won't be accidentally called elsewhere.

lib/scripts/vtxo.go
```diff
@@ -0,0 +1,350 @@
...
+//
+//	<timeout_key> OP_CHECKSIG
+//	<exit_delay>  OP_CHECKSEQUENCEVERIFY OP_DROP
+func UnilateralCSVTimeoutTapLeaf(timeoutKey *btcec.PublicKey,
+	csvDelay uint32) (txscript.TapLeaf, error) {
+
+	// Use ScriptTemplate to construct the timeout script. This script
+	// ensures the proper party can sign for this output, and that the CSV
+	// delay has been upheld.
+	secondLevelLeafScript, err := txscript.ScriptTemplate(`
```
bhandras, line 302: TIL 💡

lib/scripts/vtxo_test.go
```diff
@@ -0,0 +1,452 @@
...
+					clientSig.Serialize(),
+					collabSpendInfo.WitnessScript,
+					collabSpendInfo.ControlBlock,
+				}
+			},
+			valid:       false,
+			useSequence: 0,
+		},
+		{
+			name: "invalid collaborative spend signature order " +
```
bhandras, line 210: nit: we could add another test case to test for bogus signature (eg. one party sets a fake sig).

lib/scripts/vtxo_test.go
```diff
@@ -0,0 +1,452 @@
...
+			assertEngineExecution(t, i, testCase.valid, newEngine)
+		})
+	}
+}
+
+// assertEngineExecution executes the VM returned by the newEngine closure,
+// asserting the result matches the validity expectation. In the case where it
+// doesn't match the expectation, it executes the script step-by-step and
+// prints debug information to stdout.
+func assertEngineExecution(t *testing.T, testNum int, valid bool,
```
bhandras: Very nice!

lib/scripts/vtxo.go
```diff
@@ -2,6 +2,11 @@ package scripts
 
 // VTXO Taproot Tree Structure:
 //
+// - The "Owner" is the participant who is able to unilaterally recover funds
```
bhandras, line 5: Nice, let's squash 🎃

lib/tree/node.go
```diff
@@ -0,0 +1,898 @@
...
+	}
+
+	// Create a new node with the same basic info.
+	extracted := &Node{
+		Input:     n.Input,
+		CoSigners: n.CoSigners,
+		Outputs:   n.Outputs,
+		Signature: n.Signature,
+		FinalKey:  n.FinalKey,
+		Children:  make(map[uint32]*Node),
```
bhandras, line 478: nit: Alternatively we could use a slice here because then the iteration order will always be deterministic (vs with maps we might loop through the children where this is used in any order).

lib/tree/node_test.go
```diff
@@ -0,0 +1,1547 @@
...
+	t.Run("prints deep tree", func(t *testing.T) {
+		deepRoot := createDeepTree(t)
+		output := deepRoot.PrettyPrint()
+
+		// Count branches and leaves in output.
+		require.Contains(t, output, "Branch")
+		require.Contains(t, output, "Leaf")
+
+		// Should have tree connectors.
+		require.Contains(t, output, "├──")
```
bhandras, line 1498: 🤓

lib/tree/build_test.go
```diff
@@ -0,0 +1,711 @@
...
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPartitionLeaves tests the partitionLeaves function.
+func TestPartitionLeaves(t *testing.T) {
```
bhandras, line 14: Very nicely tested!

lib/tree/tree.go
```diff
@@ -0,0 +1,261 @@
...
+	SweepTapscriptRoot []byte
+}
+
+// NewTree constructs a transaction tree from the given leaves using BFS.
+// The radix parameter controls the branching factor - each branch node will
+// have up to 'radix' children. A radix of 2 creates a binary tree, radix of 4
+// creates a quad-tree, etc. Higher radix values reduce tree depth but increase
+// transaction size.
+func NewTree(rootOutpoint wire.OutPoint, rootOutput *wire.TxOut,
+	leaves []LeafDescriptor,
```
bhandras: nit: we could reformat this to be more LL style.

lib/tree/tree.go
```diff
@@ -0,0 +1,261 @@
...
+		BatchOutpoint:      t.BatchOutpoint,
+		BatchOutput:        t.BatchOutput,
+		SweepTapscriptRoot: t.SweepTapscriptRoot,
+	}, nil
+}
+
+// Verify recursively verifies that the tree structure is consistent.
+// It checks that each child's input correctly references the parent's
+// transaction at the expected output index.
+func (t *Tree) Verify() error {
```
bhandras, line 125: nit: I think we could do iterative here vs recursive just to avoid any possible stack overflows.

lib/tree/signing.go
```diff
@@ -0,0 +1,261 @@
...
+	return s.signSession.PublicNonce, nil
+}
+
+// RegisterNonces registers other participants' nonces with this session.
+// It filters out the signer's own nonce before registration.
+//
+// TODO: once lnd's MuSig2RegisterNonces supports submitting the aggregate nonce
+// instead of having to submit all individual nonces, we can update this method
+// to do the same.
+func (s *TxSignerSession) RegisterNonces(nonces [][66]byte) error {
```
bhandras: nit: musig2.PubNonceSize

lib/tree/signing.go
```diff
@@ -0,0 +1,261 @@
...
+	noncesSet map[string][]Musig2PubNonce) error {
+
+	for txid, txSession := range s.txs {
+		nonces, ok := noncesSet[txid]
+		if !ok {
+			return fmt.Errorf("nonces for tx %s not found", txid)
+		}
+
+		// Convert to [][66]byte format.
+		ns := make([][66]byte, len(nonces))
```
bhandras: nit: musig2.PubNonceSize

lib/tree/signing.go
```diff
@@ -0,0 +1,261 @@
...
+		return fmt.Errorf("not all nonces registered successfully")
+	}
+
+	return nil
+}
+
+// Sign generates the partial signature for this transaction.
+func (s *TxSignerSession) Sign() (*musig2.PartialSignature, error) {
+	return s.signer.MuSig2Sign(
+		s.signSession.SessionID, s.sigHash, true,
```
bhandras: nit: do we always want to clean up the session here? It'll crash if called again iirc.

lib/tree/batch.go
```diff
@@ -0,0 +1,432 @@
...
+	// CoSignerKey is the public key of the VTXO owner who must co-sign
+	// along with the operator to spend this VTXO collaboratively.
+	CoSignerKey *btcec.PublicKey
+}
+
+// NewVTXODescriptor constructs a VTXODescriptor by building the VTXO taproot
+// script using the scripts package. This is a convenience helper for clients to
+// easily create VTXO descriptors for tree construction.
+func NewVTXODescriptor(
+	amount btcutil.Amount,
```
bhandras: nit: formatting could be LLized (here and elsewhere).

lib/tx/forfeit_test.go
```diff
@@ -0,0 +1,248 @@
...
+			return out
+		}
+	}
+
+	return nil
+}
+
+// CreateKey returns a deterministically generated key pair. It returns the
+// public key and mock signer for the key. This is useful for testing.
+func CreateKey(index int32) (*btcec.PublicKey, input.Signer) {
```
bhandras: nit: iirc i've seen this before, maybe worth extracting these common helpers.

lib/tx/ark.go
```diff
@@ -0,0 +1,875 @@
...
+	return signDesc, spendInfo, nil
+}
+
+// VerifyArkTxPackage verifies that an ARK transaction package is valid.
+// This checks:
+// - Checkpoint txs correctly spend input VTXOs.
+// - ARK tx correctly spends checkpoint outputs.
+// - Amounts balance.
+// - Client signatures on ARK tx inputs are valid.
+func VerifyArkTxPackage(pkg *ArkTxPackage, inputs []VTXOInput,
```
bhandras: Super nice!

lib/tx/ark.go
```diff
@@ -0,0 +1,875 @@
...
+		}
+	}
+
+	return nil
+}
+
+// parseSchnorrSig parses a schnorr signature from bytes, stripping the
+// sighash byte if present.
+func parseSchnorrSig(sigBytes []byte) (*schnorr.Signature, error) {
+	// Signatures from lnd may have sighash byte appended.
```
bhandras: TIL 💡

bhandras (approved):
LGTM, excellent work! ✨

lib/tree/signing.go
ellemouton: just going off the docs: 
```
 If this node isn't responsible for
	// combining all the partial signatures, then the cleanup parameter
	// should be set, indicating that the session can be removed from memory
	// once the signature was produced.
```

what do you recommend?

ellemouton: i think we would not call it again right? cause it will ideally be wrapped in a state machine that prevents that

bhandras: Yeah agree, likely will only be called once. Alternatively we can just expose an argument to cleanup the sess.

lib/tree/node.go
ellemouton: the reason i did it like this is cause when we extract a path for a client, it does not necessarily contain all the children for that tx. So we need to keep track of what it's index is (note that the same was done in arklib for the same reason)

lib/tree/tree.go
ellemouton: i guess: how big do we expect these trees to get? The arklib also uses recursion.. happy to update this to do iterative - but then would you say we should also update all the other helpers that use recursion like ForEach?

ellemouton:
Thanks @bhandras 🙏 

For now, ive removed the last 2 commits (ark & checkpoint txs since those are not quite ready i think and not used properly anywhere).

ellemouton:
need to still fix some stuff

lib/tree/signing.go
ellemouton: ok, added the argument

lib/scripts/vtxo_test.go
ellemouton: added

lib/tree/tree.go
bhandras: I guess the tree itself isn't super big but it could get very deep. We can also just do the iterative conversion in a follow up since its non critical.

bhandras (approved):
LGTM 🚢

lib/scripts/nums.go
```diff
@@ -0,0 +1,35 @@
...
+// ARKNUMSHex is the hex encoded version of the ARK NUMs key.
+const ARKNUMSHex = "02372f225b3caee8213096de3229ee4335306b0" +
+	"7c3c169438461b5d4749884ec65"
+
+var (
+	// ARKNUMSKey is a NUMS key (nothing up my sleeves number) that has
+	// no known private key. This was generated using the following script:
+	// https://github.com/lightninglabs/lightning-node-connect/tree/
+	// master/mailbox/numsgen, with the seed phrase "Ark Protocol NUMS".
+	ARKNUMSKey = mustParsePubKey(ARKNUMSHex)
```
sputn1ck, line 24: as the numsgen is deterministic, couldn't we include a test which does the hashing with the seed phrase and the found out NUMS, just so it's publicly verifiable in this repo?

lib/scripts/vtxo.go
```diff
@@ -0,0 +1,356 @@
...
+	vtxoCollabPathLeafIndex = 0
+
+	// vtxoTimeoutPathLeafIndex is the index of the timeout path leaf in the
+	// VTXO tapscript tree.
+	vtxoTimeoutPathLeafIndex = 1
+)
+
+// VTXOLeafType is an enum-like type to identify the different leaves in a VTXO
+// tapscript tree.
+type VTXOLeafType int
```
sputn1ck, line 71: Is this where we'd add more leaf types for scripting support?

sputn1ck (approved):
LGTM! 2 minor questions

lib/scripts/nums.go
ellemouton: good idea. added!

bhandras (approved):
lgtm no.3

PR #3, opened 2025-11-14
chainbackend+chainsource: add chainsource actor and backend impl

Roasbeef:
In this PR, we make an initial actor wrapper around the routine chain notifications. This includes: block height notification, spend, conf, etc. This sub-system enables us to decouple from the existing daemon events, as any state machine can request a new event to be sent once a txid gets confirmed. 

Fixes https://github.com/lightninglabs/darepo/issues/11


chainbackends/doc.go
```diff
@@ -0,0 +1,64 @@
...
+//
+// # Available Backends
+//
+// LNDBackend: Full-node backend that wraps lnd's chain notification and fee
+// estimation interfaces. Suitable for production deployments where lnd is
+// available. Provides real-time notifications via lnd's chainntnfs package.
+//
+// ExplorerBackend: Light client backend using HTTP APIs from block explorers
+// (such as mempool.space). Suitable for lightweight deployments without a full
+// node. Uses polling with exponential backoff for eventual consistency.
```
ellemouton: currently only the LNDBackend is defined in this pr

chainsource/conf_actor.go
```diff
@@ -0,0 +1,247 @@
...
+	if req.TargetConfs == 0 {
+		return fn.Err[ConfResp](fmt.Errorf(
+			"target confirmations must be greater than zero"))
+	}
+
+	a.txid = req.Txid
+	a.pkScript = req.PkScript
+	a.targetConfs = req.TargetConfs
+	a.heightHint = req.HeightHint
+	a.notifyActor = req.NotifyActor
```
ellemouton, line 111: what stops the caller just overriding the previous subscription?

chainsource/conf_actor.go
```diff
@@ -0,0 +1,247 @@
...
+		ref.Tell(a.ctx, event)
+	})
+}
+
+// failConfirmation completes the promise with an error (Future mode) or does
+// nothing (Actor mode - errors are not delivered in async mode).
+func (a *ConfActor) failConfirmation(err error) {
+	a.promise.WhenSome(func(p actor.Promise[ConfirmationEvent]) {
+		p.Complete(fn.Err[ConfirmationEvent](err))
+	})
```
ellemouton, line 206: does the actor not get notified of errors?

chainsource/chainsource.go
```diff
@@ -0,0 +1,403 @@
...
+	keyPart, err := txidOrScriptKey(req.Txid, req.PkScript)
+	if err != nil {
+		return fn.Err[ChainSourceResp](fmt.Errorf(
+			"failed to generate service key: %w", err))
+	}
+
+	// We'll then use that "key part"  to generate a unique actor ID and
+	// also service key.
+	actorID := fmt.Sprintf(
+		"conf.%s.%s.%d", req.CallerID, keyPart, req.TargetConfs,
```
ellemouton, line 172: where does the responsibility lie for ensuring that callers dont use colliding IDs?

chainsource/chainsource.go
```diff
@@ -0,0 +1,403 @@
...
+
+	// Unregister the actor using the service key.
+	unregisterByServiceKey(a, serviceKey)
+
+	return fn.Ok[ChainSourceResp](&UnregisterConfResponse{})
+}
+
+// handleUnregisterSpend cancels a spend subscription by finding and stopping
+// the dedicated actor.
+func (a *ChainSourceActor) handleUnregisterSpend(ctx context.Context,
```
ellemouton, line 258: what if a caller forgets to unregister? how do they even know they need to unregister?

chainsource/chainsource.go
```diff
@@ -0,0 +1,403 @@
...
+// unregisterByServiceKey is a generic helper that finds and unregisters all
+// actors registered with the given service key. This eliminates code
+// duplication across the three unregister handler methods.
+func unregisterByServiceKey[Req, Resp actor.Message](
+	a *ChainSourceActor, serviceKey actor.ServiceKey[Req, Resp]) {
+
+	refs := actor.FindInReceptionist(a.system.Receptionist(), serviceKey)
+	for _, ref := range refs {
+		serviceKey.Unregister(a.system, ref)
+	}
```
ellemouton, line 309: i dont really get why we would expect 1 call to "UnregisterX" to cancel multiple actors? i thought "serviceKey" is like a description of the type of service offered and then an ActorRef it a ref to the specific actor that we want to cancel. but here it looks like service key is also keyed by individual id? so then i dont really get the diff between service key and actor ref

chainbackends/doc.go
Roasbeef: Ah yeah can update. I removed the explorer as I think we have a version of it somewhere.

chainsource/conf_actor.go
Roasbeef: Good eye, this should reject to have just one per instance. 

Before I had it s.t once you constructed the actor, it would start the dispatch, but I wanted to have more control over it.

chainsource/conf_actor.go
Roasbeef: Ah to notify the error? I can add that as well. Depends on how the consumer starts to work. 

The other thing is that even in notify mode, the caller gets a future response which will include this error.

chainsource/chainsource.go
Roasbeef: Right now the responsibility is on the caller side. 

Here's where the mapping is created in the framework: https://github.com/lightningnetwork/lnd/blob/9319e4c8e60ff57103bcf82b98aeea70b7fc7c8a/actor/system.go#L124-L128


You can have many actors with the same service key, but he actor ID must be unqiue. 

The caller ID here is meant to force that unique mapping, tho it's done in a weak way rn. The idea is that this is typically something like `subsystem || outpoint`.

chainsource/chainsource.go
Roasbeef: Good question. 

One way would be to have the actors die when work is done. But if you look at a recent PR I have for lnd, you actually want to keep the subscription in order to be able to handle re-orgs. 

They would unregister if they didn't need the conf at all anymore.

You can imagine the unroll state machine, where maybe another version (eg: future RBF or w/e) confirms and they need to cancel a subscription. 

This is a job of a caller rn, just like if you had to do a `defer` to cancel as normal. 

I don't think this is the _ideal_ version, but I was able to put this out there to build on top of it.

chainsource/chainsource.go
Roasbeef: So right now we actually make 1:1 service keys, as they're tied to a given caller ID. 

You're right though, that you typically wouldn't want to do this, depending on the final model. As mentioned you can have many actors with the same service key, and then sort of load balance across them. 

Re service key vs ref:
  * The service key is basically the address. In order to actually interact with the actor you need a ref (access to the mailbox). 
  * Once you make an actor you get a ref you can pass round. The ref is actually an interface, so you can mock it if needed on. 


There's another API that lets you function _only_ with a service key. 

Here's an example of a fragment of that from the README: 

```
## Routers: Distributing Work

Sometimes, you might have multiple actors performing the same kind of task, and
you want to distribute messages among them. A `Router` can do this. It's not an
actor itself but acts as a dispatcher.

A `Router` uses a `RoutingStrategy` to pick one actor from a group registered
under a `ServiceKey`.

```go
// Assume 'system' and 'myServiceKey' are set up, and multiple actors
// are registered with 'myServiceKey'.

// Create a round-robin routing strategy.
roundRobinStrategy := actor.NewRoundRobinStrategy[*MyRequest, *MyResponse]()

// Create a router for 'myServiceKey' using this strategy.
// Messages sent to this router will be forwarded to one of the actors
// registered under 'myServiceKey'.
// The router also needs a DLO for messages it can't route (e.g., if no actors are available).
serviceRouter := actor.NewRouter(
    system.Receptionist(),
    myServiceKey,
    roundRobinStrategy,
    system.DeadLetters(),
)

// Now, interact with the router as if it were an ActorRef:
serviceRouter.Tell(context.Background(), &MyRequest{Data: "Message via router"})

futureReplyFromRouter := serviceRouter.Ask(context.Background(), &MyRequest{Data: "Ask via router"})
// ... await futureReplyFromRouter ...

If the router cannot find any available actors for the `ServiceKey` (e.g., none
are registered or running), `Tell` operations will typically send the message to
the router's configured DLO, and `Ask` operations will return a `Future`
completed with `ErrNoActorsAvailable`.
```

Notice how you _skip_ the ref, and just do `tell/ask` on a router.

Roasbeef: From here: https://github.com/lightningnetwork/lnd/blob/actor/v0.0.1-alpha/actor/README.md

Roasbeef:
@codex review this

ellemouton (approved):
LGTM! very nice seeing how this is used in https://github.com/lightninglabs/darepo-client/pull/15 🙏

PR #4, opened 2025-11-14
multi: stage boarding actor client 

Roasbeef:
This PR implements the boarding process from the client's PoV. 

Right now it's built on a older version of `lib`, I'll remove it from draft once I update the call sites with the new APIs. Some functions like signing for the boarding input don't seem to be in the latest version of `lib` yet either. 

Any messages that need to be set to the round-specific state machine flow through the main actor. This will be typically sent from the server, but also the wallet can ask for a new boarding address. 

### Boarding Addresses + Boarding Intents 

Boarding addresses track addresses that we've given out to 3rd parties to deposit into the system. Once a party sends to a boarding address, it becomes a boarding intent (has an outpoint, and chain info). 

One thing we need to resolve is that right now, with the way the chain notifier works in lnd, we'll only be notified of the _first_ time a `pkScript` is sent to. For now, to partially handle this, I track the current height of the last send to a boarding address, and use that as a height hint. This ofc, isn't fool proof. 

Instead I think we may need to import the boarding addresses into the underlying wallet, then rely on some sort of transaction notification stream (eg: `SubscribeTransactions`) like in ind. 

### Server Messages + Expectations 

The actor as implemented today is intended to be able to handle multiple pending rounds. We'll track a primary round, attempt to board a batch, then effectively seal it. Upon restart, we register for confirmation for the batches, to be able to finalize them a manifest the new VTXOs. 

One thing we need to iron out is exactly what the server message flow is. The state machine supports boarding multiple inptus at once, which means multiple sets of nonces, musig2 sessions, partial sigs, etc. As implemented right now, the client will send the nonces one by one, then the partial sigs in a similar fashion. 

The current `ToProto()` methods are meant to map the client events to proto messages. However, it's possible that we want this to live elsewhere instead. 

### Persistence 

Right now some basic persistence interfaces are in place. The one that likely needs to be revised the most is the round state, as right now it just returns the exact state, but we likely want that instead to be a more generic type, as it'll be used elsewhere.


Roasbeef:
Will split this up into multiple PRs.

PR #12, opened 2025-11-19
assets: add comprehensive asset tx builder

bhandras:
Adds a comprehensive tx builder for TAP asset transactions with the ability to add btc only inputs/outputs and ephemeral anchors for zero fee transactions. The tx builder is meant to implement the necessary plumbing before we can add the asset ARK tree.


assets/client.go
```diff
@@ -0,0 +1,753 @@
...
+
+	t.assetNameMutex.Lock()
+	defer t.assetNameMutex.Unlock()
+
+	assetIDStr := hex.EncodeToString(assetID)
+	if name, ok := t.assetNameCache[assetIDStr]; ok {
+		return name, nil
+	}
+
+	assetStats, err := t.UniverseClient.QueryAssetStats(
```
Roasbeef, line 159: This'll be a super slow query in practice. As it scans over this deprecated time series DB. 

Instead, this should just use `QueryLeaf` or `QueryProof` directly.

assets/client.go
```diff
@@ -0,0 +1,753 @@
...
+
+	return nil
+}
+
+// PrepareAndCommitVirtualPsbts prepares and commits virtual psbt to a BTC
+// template so that the underlying wallet can fund the transaction and add the
+// necessary additional input to pay for fees as well as a change output if the
+// change keydescriptor is not provided.
+func (t *TapdClient) PrepareAndCommitVirtualPsbts(ctx context.Context,
+	vpkt *tappsbt.VPacket, feeRateSatPerVByte chainfee.SatPerVByte,
```
Roasbeef, line 265: Perhaps this should should take an encapsulating struct instead for arg, and alsor eturn value? I count maybe 15 combined input+output params at a glance.

assets/client.go
```diff
@@ -0,0 +1,753 @@
...
+	if customLockID != nil {
+		commitRequest.CustomLockId = (*customLockID)[:]
+	}
+
+	if feeRateSatPerVByte == 0 {
+		commitRequest.SkipFunding = true
+	}
+
+	if changeKeyDesc != nil {
+		err := addP2WPKHOutputToPsbt(
```
Roasbeef, line 322: Seems to assume the change will be p2wkh. Pretty sure it's actually P2TR. In which case, the logic above needs to be updated to also include the taproot bip32 fingerprints.

assets/client.go
```diff
@@ -0,0 +1,753 @@
...
+			return nil, nil, nil, nil, err
+		}
+	}
+
+	return fundedPacket, activePackets, passivePackets, commitResponse, nil
+}
+
+// LogAndPublish logs and publishes a psbt with the given active and passive
+// assets.
+func (t *TapdClient) LogAndPublish(ctx context.Context, btcPkt *psbt.Packet,
```
Roasbeef, line 390: Same here re param length.

assets/client.go
```diff
@@ -0,0 +1,753 @@
...
+	leafKey := &universerpc.AssetKey{
+		Outpoint: &universerpc.AssetKey_Op{
+			Op: outpoint,
+		},
+		ScriptKey: &universerpc.AssetKey_ScriptKeyBytes{
+			ScriptKeyBytes: scriptKey.SerializeCompressed(),
+		},
+	}
+
+	_, err = t.InsertProof(ctx, &universerpc.AssetProof{
```
Roasbeef, line 595: Would need to double check if the backing wallet will see this as their proof or not. IIRC there's another RPC that's intended to be used in this setting.

assets/client.go
```diff
@@ -0,0 +1,753 @@
...
+
+	var lastProof *proof.Proof
+
+	for i := 0; i < proofFile.NumProofs(); i++ {
+		lastProof, err = proofFile.ProofAt(uint32(i))
+		if err != nil {
+			return nil, err
+		}
+
+		err = t.ImportProof(ctx, lastProof)
```
Roasbeef: Same here re if this is the preferred route for this.

harness/tap_test_utils.go
```diff
@@ -0,0 +1,328 @@
...
+	_, err := tc.TapdClient.AddFederationServer(
+		ctx, &universerpc.AddFederationServerRequest{
+			Servers: []*universerpc.UniverseFederationServer{
+				{
+					Host: universeHost,
+				},
+			},
+		},
+	)
+	// Ignore duplicate/unusable errors when the peer already has
```
Roasbeef, line 228: Lacking new line above.

assets/onboarding.go
```diff
@@ -0,0 +1,396 @@
...
+type OnboardingMode int
+
+const (
+	// OnboardingModeMuSig2 uses MuSig2 keyspend for cooperative path and
+	// CSV scriptspend for timeout.
+	//
+	// Anchor structure:
+	//   - InternalKey: MuSig2(userInternal, operatorInternal)
+	//   - Tapscript: CSV timeout path
+	OnboardingModeMuSig2 OnboardingMode = iota
```
Roasbeef: Do we need _both_ of these? Thought we'd just pick one.

assets/onboarding.go
```diff
@@ -0,0 +1,396 @@
...
+//  2. Script-only mode: NUMS anchor with CHECKSIGADD cooperative + CSV timeout
+//
+// The mode is determined at creation time and affects the anchor structure.
+type OnboardingKit struct {
+	// Mode determines the anchor structure (MuSig2 or script-only).
+	Mode OnboardingMode
+
+	// UserKey is the user's public key used for both cooperative signing
+	// and timeout sweep path.
+	UserKey *btcec.PublicKey
```
Roasbeef: Can likely re-use some of the internal types to embed in the `lib` package. Recently we added boarding related types.

assets/onboarding.go
```diff
@@ -0,0 +1,396 @@
...
+// BIP86 taproot tweak, following the standard MuSig2 protocol.
+func NewOnboardingKit(userKey, operatorKey *btcec.PublicKey,
+	keyLocator keychain.KeyLocator, assetID asset.ID, amount uint64,
+	csvExpiry uint32, chainParams *address.ChainParams) (
+	*OnboardingKit, error) {
+
+	// Combine the user and operator keys using MuSig2.
+	sortKeys := true
+	muSig2Key, err := input.MuSig2CombineKeys(
+		input.MuSig2Version100RC2,
```
Roasbeef: Can prob pack these in a bit more on each line. We can experiment with some steering in the agent files.

assets/onboarding.go
```diff
@@ -0,0 +1,396 @@
...
+	}, nil
+}
+
+// NewScriptOnlyOnboardingKit creates a new onboarding kit in script-only mode.
+//
+// This mode uses NUMS as the internal key (forcing scriptspend) with two
+// spending paths:
+//  1. Cooperative: CHECKSIGADD requiring both user and operator signatures
+//  2. Timeout: CSV allowing user to recover after expiry
+func NewScriptOnlyOnboardingKit(userKey, operatorKey *btcec.PublicKey,
```
Roasbeef: Also I assume this is for the assets level? 

I'm not sure if we _have_ to replicate the scripts on the asset level in addition to the normal Bitcoin level for boarding scripts. For any sort of other off-chain spends like OOR unroll, etc -- I think it can make more sense there, as you don't want the asset to be able to "escape early".

assets/onboarding.go
```diff
@@ -0,0 +1,396 @@
...
+func (k *OnboardingKit) GenCooperativePathScript() ([]byte, error) {
+	if k.Mode != OnboardingModeScriptOnly {
+		return nil, fmt.Errorf("cooperative script only available in " +
+			"script-only mode")
+	}
+
+	builder := txscript.NewScriptBuilder()
+
+	// Add user key + CHECKSIG.
+	builder.AddData(schnorr.SerializePubKey(k.UserKey))
```
Roasbeef: There's a script template based builder in `txscript` now if you're interested in it. Can cut down the amount of lines to specify a script.

assets/onboarding.go
```diff
@@ -0,0 +1,396 @@
...
+		TweakedScriptKey: &asset.TweakedScriptKey{
+			RawKey: keychain.KeyDescriptor{
+				PubKey: asset.NUMSPubKey,
+			},
+			Tweak: merkleRoot,
+		},
+	}
+
+	expectedScriptKey := tapScriptKey.PubKey.SerializeCompressed()
+	// Normalize to even prefix.
```
Roasbeef: Space above. 

Missing complete sentence further above.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+
+// SignerRole identifies a signer participating in MuSig2 aggregation.
+type SignerRole string
+
+// AnchorKeyMode describes the admissible spend path on the Bitcoin key path.
+type AnchorKeyMode uint8
+
+const (
+	AnchorKeyModeMuSig2 AnchorKeyMode = iota
+	AnchorKeyModeStatic
```
Roasbeef, line 46: As in p2a anchor? Why should we have two diff versions?

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+		return c.WitnessFunc(controlBlock, args)
+	}
+
+	return c.Closure.Witness(controlBlock, args)
+}
+
+// InputConfig associates a proof chain with anchor metadata.
+type InputConfig struct {
+	// ProofFile is the raw proof blob exported from tapd.
+	ProofFile []byte
```
Roasbeef, line 156: `proof.File`?

Roasbeef: Also do you actually need the entire file? Or just the last state transition.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+// output.
+type BtcOutputPlan struct {
+	// Description echoes the spec label.
+	Description string
+
+	// ValueSat mirrors the satoshi amount assigned to the output.
+	ValueSat int64
+
+	// PkScript is the Bitcoin script committed to the output.
+	PkScript []byte
```
Roasbeef, line 365: So how does the plan differ from the spec here? Looks like we can just embed the spec in the plan. With the only unique field being the output index.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+
+	// OutputIndex records the final PSBT output index assigned after
+	// Commit().
+	OutputIndex int
+}
+
+// OpTrueScriptDetails exposes the pre-built artifacts for OP_TRUE scripts.
+type OpTrueScriptDetails struct {
+	Artifacts *OpTrueArtifacts
```
Roasbeef, line 374: Any instance where this should be valid and the pointer nil? If not, then perhaps we can drop the pointer.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+	}
+
+	controlBlockBytes, err := artifacts.ControlBlock.ToBytes()
+	if err != nil {
+		return BTCAnchorSpec{}, fmt.Errorf("serialize control block: "+
+			"%w", err)
+	}
+
+	tapLeaf := *artifacts.TapLeaf
+	tapLeaf.Script = append([]byte(nil), artifacts.TapLeaf.Script...)
```
Roasbeef: Why nil append? I think you want to be explicit as possible here on the script level.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+		}
+
+		proofCount := pfFile.NumProofs()
+		if proofCount == 0 {
+			return nil, fmt.Errorf("proof file %d contains no "+
+				"proofs", idx)
+		}
+
+		lastProofIdx := uint32(proofCount - 1)
+		pr, err := pfFile.ProofAt(lastProofIdx)
```
Roasbeef, line 1227: If we only actually need the last proof, then we can get away with just storing a `proof.Proof`.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+	return plan, nil
+}
+
+// Commit invokes CommitVirtualPsbts and stores anchor metadata.
+//
+//nolint:funlen
+func (b *AssetTxBuilder) Commit(ctx context.Context, wallet assetWalletClient,
+	opts CommitOptions) error {
+
+	if b.vpacket == nil {
```
Roasbeef, line 1387: I think we can break this function down into some smaller re-useable components.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("publish transfer: %w", err)
+	}
+
+	return resp, nil
+}
+
+// BuildAnchorChild assembles a CPFP child transaction that spends every zero
+// value BTC anchor associated with the builder.
+func (b *AssetTxBuilder) BuildAnchorChild(ctx context.Context,
```
Roasbeef: > Spends every zero value BTC anchor associated with the builder.

We'll need to take care to not violate the BIP 431 constraints.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2773 @@
...
+		PreviousOutPoint: prev,
+		Sequence:         sequence,
+	})
+}
+
+const (
+	// probeSequence is the sequence value we attach to ephemeral "fee
+	// probe" inputs when estimating the weight impact of yet-to-be-added
+	// anchors.
+	probeSequence = wire.MaxTxInSequenceNum
```
Roasbeef: Why is this necessary at all? The witness at this range shouldn't affect the weight of a transaction.

assets/closures.go
```diff
@@ -0,0 +1,162 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+)
+
+type Closure interface {
```
Roasbeef, line 15: So a closure is "just" a witness script basically?

assets/helpers.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/taproot-assets/commitment"
+	"github.com/lightninglabs/taproot-assets/proof"
+)
+
+// GenTaprootAssetRootFromProof generates the taproot asset root hash from a
+// proof. This root hash is used as the inclusion proof in the taproot control
+// block for timeout sweep transactions. It proves the asset commitment is part
+// of the taproot tree.
+func GenTaprootAssetRootFromProof(prf *proof.Proof) ([]byte, error) {
```
Roasbeef, line 26: The proof already has the tapscript root in it. As this is part of the inclusion proofs.

assets/musig2_signer.go
```diff
@@ -0,0 +1,231 @@
...
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr/musig2"
+)
+
+// MuSig2Signer represents one party in an N-of-N MuSig2 signing session. Each
+// signer only knows their own private key and the other parties' public keys.
+type MuSig2Signer struct {
```
Roasbeef: Should use `musig2.Session`.

Roasbeef (changes requested):

GeorgeTsagk:
Still groking the details of this PR, but wanted to also drop a note that the asset tx builder may be something that we can outsource to https://github.com/lightninglabs/tap-sdk

Down that path, all that will be needed from this PR is import the sdk and do the top level tailoring
cc @darioAnongba

Can always do this in a future iteration in order to not stall current PR

darioAnongba:
> Still groking the details of this PR, but wanted to also drop a note that the asset tx builder may be something that we can outsource to https://github.com/lightninglabs/tap-sdk
> 
> Down that path, all that will be needed from this PR is import the sdk and do the top level tailoring cc @darioAnongba
> 
> Can always do this in a future iteration in order to not stall current PR

Yes that's exactly the plan. The ultimate goal is for this repo to import a single dependency as `github.../tap-sdk` to support taproot-assets, the `taprc` being encapsulated there and only accessible through SDK exposed functions that abstract it, same for the transaction builder.

assets/asset_tx_builder.go
GeorgeTsagk: +1 This can have great performance impact, some proofs can end up being multiple MBs. Given a deep enough tree we'd be wasting lots of resources on this.

Should be fairly easy to unfold the proofs over the tree path leading to the leaf

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2323 @@
...
+// OpTrueScriptSpec implements the AssetScriptSpec interface for the simple
+// OP_TRUE asset script.
+type OpTrueScriptSpec struct {
+	// InternalKey is used for both the OP_TRUE script key derivation and
+	// as the anchor internal key. Use asset.NUMSPubKey for a global script
+	// key, or a unique key per output for distinct script keys.
+	InternalKey *btcec.PublicKey
+}
+
+func (s *OpTrueScriptSpec) Prepare(anchorKey AnchorKeySpec,
```
GeorgeTsagk, line 409: missing godoc

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2323 @@
...
+	return &AssetScriptArtifacts{
+		ScriptKey:      s.ScriptKey,
+		AnchorInternal: s.ScriptKey.PubKey,
+		OutputKey:      s.ScriptKey.PubKey,
+	}, nil
+}
+
+// CommitOptions capture the configurable parameters when embedding the
+// assembled virtual transaction into a BTC-level anchor PSBT.
+type CommitOptions struct {
```
GeorgeTsagk, line 462: Another thing to keep in mind is verifying node/leaf output commitments before signing the inputs.

You'd need to check that to make sure your assets are being carried over, by validating their inclusion in the new outpoint.

I assume this is left for a follow-up PR, as we're mostly focusing on the asset tx **building** here.

assets/asset_tx_builder.go
```diff
@@ -0,0 +1,2323 @@
...
+	Closures []ScriptClosure
+
+	// Sequence optionally sets the input sequence number. A zero value
+	// uses the default (MaxTxInSequenceNum). For CSV script path spends,
+	// this must be set to the CSV delay value.
+	Sequence uint32
+}
+
+// OutputConfig defines the anchor we are assembling.
+type OutputConfig struct {
```
GeorgeTsagk, line 171: It's not super clear how the clients will influence the output key of each node in the tree.

One approach would be to let the clients guide the server into crafting the correct tap commitment.

I think a more flexible/snappy approach would be to have the server craft all the tap commitments for each node in the tree, given that it knows exactly which assets each leaf holds. Given that we're using op_true for each intermediate node this should be fairly easy from the server's perspective as no client input is needed for the assets level spend.

Clients would then also verify inclusion of their assets in the tap commitment of each input/output while verifying the tree path. This I believe is a case where `OP_TRUE` asset scripts significantly simplify server/client interaction.

GeorgeTsagk: This is a lean approach as the client signing remains as is, we only add an additional slim layer of tap commitment verification.

Of course the client would always need to fully verify the assets entering the batch tx -- no compromises there.

GeorgeTsagk (commented):
The current asset tx builder is good enough for the things it aims to deliver.

Most of my comments are about where we go from there in the ark integration. Most of them are non-blocking for the current code but would like to have at least one back and forth on them to make sure there's a mental alignment.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

bhandras:
Closing as superseded by Dario's current Taproot Assets implementation.

The transaction-builder boundary moved to [tap-sdk #158](https://github.com/lightninglabs/tap-sdk/pull/158), which landed the advanced custom-anchor transaction builder. The Ark-facing tree implementation then landed through [Wavelength #1084](https://github.com/lightninglabs/wavelength/pull/1084). Active onboarding and recovery work continues in [Wavelength #1309](https://github.com/lightninglabs/wavelength/pull/1309).

This older prototype no longer matches the selected architecture and still has requested changes. We should not rebase or continue it.

PR #13, opened 2025-11-20
lib/types: define boarding types

ellemouton:
Define the various types that both client and server will need during boarding.


bhandras (approved):

Roasbeef (approved):
LGTM 🪣

PR #15, opened 2025-11-22
wallet+db: add initial version of wallet.Ark focused on boarding address/intents

Roasbeef:
In this PR, we add an initial version of `wallet.Ark`. This partially fixes: https://github.com/lightninglabs/darepo/issues/24. 

The main focus of this PR is to better carve out the pipe line of creating boarding addresses, and then boarding intents (confirmed addr UTXOs). The chain notifier by default will only notify for the first instance of an address, so that won't work here. Instead, we import the addrs into the wallet (we should pick another wallet account here), then register for a block epoch, and poll `listunspent` each time a block is mined. Once we find newly confirmed UTXOs, we import them into the wallet. 

An initial version of the DB is in place as well. 

PR is 60% slop to be refined. 

The main TODOs here are:
  

- [x] Remove the usage of JSON in the database. 
- [x] Update the boarding actor to use this new flow where it's notified of confirmed boarding inents


Roasbeef:
@codex review

/gemini review

Roasbeef:
@claude review this

lib/scripts/vtxo.go
```diff
@@ -114,8 +114,17 @@ func VTXOTapScript(ownerKey, cosignerKey *btcec.PublicKey,
...
-		nil
+	tapscript := input.TapscriptFullTree(
+		&ARKNUMSKey, collabLeaf, timeoutLeaf,
+	)
+
+	// Compute and set the root hash since TapscriptFullTree doesn't
+	// populate it. Callers need this to construct taproot addresses.
+	tree := txscript.AssembleTaprootScriptTree(tapscript.Leaves...)
+	rootHash := tree.RootNode.TapHash()
+	tapscript.RootHash = rootHash[:]
```
ellemouton, line 125: is this something that should be fixed upstream in `AssembleTaprootScriptTree`?

db/sqlc/migrations/000002_boarding_tables.up.sql
```diff
@@ -0,0 +1,125 @@
...
+-- Create index on creation_time for chronological queries.
+CREATE INDEX IF NOT EXISTS idx_boarding_addresses_creation_time
+    ON boarding_addresses(creation_time DESC);
+
+-- Create table to store boarding intents.
+-- Boarding intents track the lifecycle of a specific boarding attempt from
+-- on-chain confirmation through round completion. Intents are only created
+-- after the boarding UTXO has been confirmed on-chain, so conf_height and
+-- conf_hash are always present.
+CREATE TABLE IF NOT EXISTS boarding_intents (
```
ellemouton, line 74: nice 🫡

db/boarding_wallet.md
```diff
@@ -0,0 +1,258 @@
+# Boarding Wallet Database Schema
```
ellemouton, line 1: very nice

wallet/messages.go
```diff
@@ -0,0 +1,256 @@
+package wallet
+
+import (
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/lightninglabs/darepo-client/baselib/actor"
+	"github.com/lightninglabs/darepo-client/chainsource"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+)
+
```
ellemouton, line 10: something i miss from the classic interface is having a quick-view of "these are the available requests/responses" (proto service style) - maybe we can still have something like that? even if just in an md file?

wallet/messages.go
```diff
@@ -0,0 +1,256 @@
...
+// RegisterConfirmationNotifierRequest registers an actor to receive
+// BoardingUtxoConfirmedEvent messages when new boarding UTXOs are detected and
+// confirmed. This enables the round actor to be notified of new boarding
+// opportunities.
+type RegisterConfirmationNotifierRequest struct {
+	actor.BaseMessage
+
+	// NotifierID uniquely identifies this notifier for later
+	// unregistration. Typically this is the actor's name or service key.
+	NotifierID string
```
ellemouton, line 145: maybe MinConf too? so that we only board with confirmed intents - the operator will probably set this minimum value in its "terms"

wallet/wallet.go
```diff
@@ -0,0 +1,463 @@
...
+)
+
+const (
+	// BoardingKeyFamily is the BIP32 key family used for deriving boarding
+	// address keys.
+	BoardingKeyFamily = 42
+
+	// MinBoardingConfs is the minimum number of confirmations required
+	// before notifying about a boarding UTXO.
+	MinBoardingConfs = 1
```
ellemouton, line 26: ah yes, i think we should let the caller set this as it will be dependent on the operator's terms

wallet/wallet.go
```diff
@@ -0,0 +1,463 @@
...
+			return BlockEpochNotification{BlockEpoch: epoch}
+		},
+	)
+	req := &chainsource.SubscribeBlocksRequest{
+		CallerID:    "boarding-wallet",
+		NotifyActor: fn.Some(epochRef),
+	}
+
+	future := a.chainSource.Ask(ctx, req)
+	result := future.Await(ctx)
```
ellemouton, line 138: should we have a `Stop` method that then un-subscribes?

ellemouton (dismissed):
very cool!

lib/scripts/vtxo.go
Roasbeef: So The issue isn't `AssembleTaprootScriptTree`, it's actually the lnd utility used here `input.TapscriptFullTree`, it just doesn't set `RootHash` at all: https://github.com/lightningnetwork/lnd/blob/c746aeed31c7f528e9cff3f57d5044892b2a02c9/input/taproot.go#L102-L111

db/boarding_wallet.md
Roasbeef: Yup, in the future we can even do things like use the CC set up in CI to auto update this stuff so it's never out of date.

wallet/messages.go
Roasbeef: Makes sense, will add it on.

wallet/messages.go
Roasbeef: Hmm yeah you have a good point. I added a new `MessageSpec` in a follow up commit, it serves as mainly type level documentation, but gives you an idea on the request/response mappings. We ofc can always just make a normal interface def around things as well.

wallet/wallet.go
Roasbeef: Good idea! Will add.

PR #17, opened 2025-11-24
Support multi-leaf tree extraction and expand client boarding info

ellemouton:
- add forfeit script field and map connector leaves directly to forfeited VTXOs so clients know which leaf/output to spend
- update tree extraction helpers to accept multiple cosigners/leaf indices, enabling multi-path subtree delivery with expanded tests and usages
- extend Makefile unit targets with per-package filters, logging/timeout options, and a debug test shortcut plus updated help examples

### DIFF
--- a/lib/tree/node.go
+++ b/lib/tree/node.go
@@ -555,16 +555,20 @@ func (n *Node) ExtractPathForIndices(targetIndices ...int) (*Node, error) {
 	targetSet := fn.NewSet(targetIndices...)
 
 	// Find the paths to all target leaves.
-	extracted, _ := n.extractPathForIndicesRecursive(targetSet, 0)
+	extracted, _, found := n.extractPathForIndicesRecursive(targetSet, 0)
+	if !found {
+		return nil, fmt.Errorf("no paths found for target indices")
+	}
 
 	return extracted, nil
 }
 
 // extractPathForIndicesRecursive recursively extracts paths to multiple target
-// leaf indices. Returns the extracted node and the number of leaves consumed in
-// this subtree.
-func (n *Node) extractPathForIndicesRecursive(
-	targetSet fn.Set[int], currentIndex int) (*Node, int) {
+// leaf indices. Returns the extracted node, the number of leaves consumed in
+// this subtree, and a boolean indicating whether any target indices were found
+// in this subtree.
+func (n *Node) extractPathForIndicesRecursive(targetSet fn.Set[int],
+	currentIndex int) (*Node, int, bool) {
 
 	// If this is a leaf node.
 	if n.IsLeaf() {
@@ -577,10 +581,10 @@ func (n *Node) extractPathForIndicesRecursive(
 				Signature: n.Signature,
 				FinalKey:  n.FinalKey,
 				Children:  make(map[uint32]*Node),
-			}, currentIndex + 1
+			}, currentIndex + 1, true
 		}
 
-		return nil, currentIndex + 1
+		return nil, currentIndex + 1, false
 	}
 
 	// This is a branch node, check children.
@@ -603,10 +607,11 @@ func (n *Node) extractPathForIndicesRecursive(
 			continue
 		}
 
-		childExtracted, newLeafIndex :=
-			child.extractPathForIndicesRecursive(targetSet, leafIndex)
-
-		if childExtracted != nil {
+		childExtracted, newLeafIndex, found :=
+			child.extractPathForIndicesRecursive(
+				targetSet, leafIndex,
+			)
+		if found {
 			extracted.Children[i] = childExtracted
 			foundAnyTarget = true
 		}
@@ -615,10 +620,10 @@ func (n *Node) extractPathForIndicesRecursive(
 	}
 
 	if foundAnyTarget {
-		return extracted, leafIndex
+		return extracted, leafIndex, true
 	}
 
-	return nil, leafIndex
+	return nil, leafIndex, false
 }
 
 // PrevOutputFetcher creates a PrevOutputFetcher that can provide transaction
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -42,10 +42,47 @@ DB_CONNECTIONSTRING ?= sqlite://./darepo.db
 
 # Build tags.
 DEV_TAGS := dev
-LOG_TAGS := nolog
 TEST_FLAGS :=
 RELEASE_TAGS :=
 
+# If specific package is being unit tested, construct the full name of the
```
Roasbeef: Just added some (but not all?) of these to the master branch. IIUC, in order to actually be able to use the logging build tag stuff, we also need to get in https://github.com/lightninglabs/darepo-client/pull/24

lib/tree/node.go
```diff
@@ -458,13 +458,21 @@ func (n *Node) NumTx() int {
...
+	// Check if any of the target keys are in this node's cosigners.
+	hasMatch := false
+	for _, targetKey := range targetKeys {
+		if ContainsCosigner(n.CoSigners, targetKey) {
+			hasMatch = true
+			break
+		}
+	}
+
+	if !hasMatch {
```
Roasbeef, line 479: Error instead of nil perhaps? May save some nil pointer defers in the future.

lib/tree/node.go
```diff
@@ -585,6 +573,105 @@ func (n *Node) extractPathForIndexRecursive(
...
+	// Validate all indices are non-negative.
+	for _, idx := range targetIndices {
+		if idx < 0 {
+			return nil, fmt.Errorf("leaf index must be "+
+				"non-negative, got %d", idx)
+		}
+	}
+
+	// Create a set of target indices for efficient lookup.
+	targetSet := make(map[int]bool, len(targetIndices))
```
Roasbeef: Could be `fn.Set` here.

lib/tree/node.go
```diff
@@ -585,6 +573,105 @@ func (n *Node) extractPathForIndexRecursive(
...
+	for _, idx := range targetIndices {
+		targetSet[idx] = true
+	}
+
+	// First, count total leaves to validate indices.
+	totalLeaves := n.countLeaves()
+	for _, idx := range targetIndices {
+		if idx >= totalLeaves {
+			// At least one index is out of bounds, return nil (no error).
```
Roasbeef: Can do this validation check before we make `targetSet` as if it fails then we can exit earlier.

lib/tree/node.go
```diff
@@ -585,6 +573,105 @@ func (n *Node) extractPathForIndexRecursive(
...
+	// Find the paths to all target leaves.
+	extracted, _ := n.extractPathForIndicesRecursive(targetSet, 0)
+
+	return extracted, nil
+}
+
+// extractPathForIndicesRecursive recursively extracts paths to multiple target
+// leaf indices. Returns the extracted node and the number of leaves consumed in
+// this subtree.
+func (n *Node) extractPathForIndicesRecursive(
```
Roasbeef: Cool func 😎

lib/tree/tree.go
```diff
@@ -75,17 +75,24 @@ func NewTree(rootOutpoint wire.OutPoint, rootOutput *wire.TxOut,
...
 
-// ExtractPathForCoSigner extracts the path relevant for a given cosigner and
-// returns a new Tree containing only the nodes where the cosigner's key is
-// present.
-func (t *Tree) ExtractPathForCoSigner(targetKey *btcec.PublicKey) (*Tree,
+// ExtractPathForCoSigners extracts the path relevant for one or more cosigners
+// and returns a new Tree containing only the nodes where any of the provided
+// cosigner keys are present. This is useful when a client has multiple VTXOs
+// with different keys in the same tree.
+func (t *Tree) ExtractPathForCoSigners(targetKeys ...*btcec.PublicKey) (*Tree,
```
Roasbeef, line 82: We have this habit of passing around pointers to pub keys everywhere....

In the past, that was needed since the pub key actually contained an underlying `*big.Int`. But these days, it just uses fixed int libs, with the entire struct being static. As a result, we don't _always_ need to bass pointer to pubkyes any longer. Not passing them can also save some nil defef checks as well. 

Just some food for thought.

lib/types/boarding.go
```diff
@@ -23,6 +23,11 @@ type OperatorTerms struct {
 	// delay will give the server time to respond to unilateral spends of
 	// a VTXO that has been forfeit or spent.
 	VTXOExitDelay uint32
+
+	// ForfeitScript is the output script that clients must use for the
```
Roasbeef, line 30: What use does the server have re a custom forfeit script here? Is this related to OOR?

lib/types/boarding.go
```diff
@@ -181,8 +181,8 @@ type ClientBatchInfo struct {
...
 	BatchOutputs []*BatchOutputInfo
 
-	// ConnectorOutputs contains the connector output info for each
-	// connector output that is relevant to the client. The number of
-	// connector leaves should match the number of connector requests.
-	ConnectorOutputs []*ConnectorOutputInfo
+	// ConnectorLeafMap maps each forfeited VTXO outpoint to its assigned
+	// connector leaf information. This allows the client to determine which
+	// connector leaf corresponds to each of their forfeit requests.
+	ConnectorLeafMap map[string]*ConnectorLeafInfo
```
Roasbeef: What's the key meant to be for this map? Can we get some stronger typing vs just `string` here? To make harder for a caller to misuse.

Roasbeef (changes requested):

lib/tree/node.go
ellemouton: added a boolean return value.

lib/types/boarding.go
ellemouton: nothing to do with OOR.

how else must the server communicate what the client should use in a forfeit tx output script? just direct payment to the server's main operator key?

Roasbeef: Ahh ok, it's the output script that the forfeit transaction sends to. Makes sense.

Roasbeef (approved):
LGTM 🔌

bhandras (approved):
LGTM ✅

PR #18, opened 2025-11-24
baselib/actor: add abstract Mailbox interface

Roasbeef:
This PR ports the abstract mailbox interface from lnd PR 10142, replacing direct channel usage in actors with a pluggable Mailbox interface. The implementation uses Go 1.23's iter.Seq pattern for clean message iteration and includes atomic synchronization to prevent race conditions.

The refactoring simplifies the Actor's process loop from ~50 lines to ~30 lines while enabling future mailbox strategies like priority queues or durable storage. All 15 new mailbox tests pass with the race detector, and the change is backward compatible.


Roasbeef:
@codex review this

PR #20, opened 2025-11-24
baselib/actor: request deadlines and resource cleanup

Roasbeef:
This PR enables actors to respect caller-provided deadlines and implement resource cleanup hooks. The envelope now carries caller context which is merged with the actor's context, allowing behaviors to detect both system shutdown and request timeouts.

The optional Stoppable interface provides an OnStop hook for cleaning up external resources like database connections or file handles during shutdown. These features enable proper backpressure handling and guaranteed resource release.


Roasbeef:
@claude review this

@codex review this

/gemini review

PR #24, opened 2025-11-24
build: add context-based logging infrastructure

Roasbeef:
PR #7 moved us away from global logging toward per-instance loggers, which is the right direction for testability and avoiding shared mutable state. However, threading a logger through every function parameter can become tedious, especially in deep call stacks where most functions just pass the logger along without using it.

This PR introduces a compromise: context-based logging utilities that let us attach a logger to a context and extract it later. Functions that already take a context can access the logger without additional parameters, while still maintaining explicit control over which logger instance is used. The pattern works naturally with request-scoped contexts, where each request can carry its own logger with relevant metadata.

The PR also ports lnd's build-tag system for controlling log levels and output destinations at compile time. This makes it easy to enable verbose logging during test debugging with `make unit log="stdlog trace" pkg=db` while keeping tests quiet by default.

### DIFF
--- a/build/context_logger_test.go
+++ b/build/context_logger_test.go
@@ -1,0 +1,102 @@
+package build
+
+import (
+	"os"
+	"testing"
+
+	"github.com/btcsuite/btclog/v2"
+	"github.com/stretchr/testify/require"
+)
+
+// TestContextWithLoggerRoundTrip tests that a logger can be stored in a context
+// and retrieved successfully.
+func TestContextWithLoggerRoundTrip(t *testing.T) {
+	t.Parallel()
+
+	// Create a real logger for testing.
+	backend := btclog.NewDefaultHandler(os.Stdout)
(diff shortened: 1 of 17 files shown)


build/context_logger_test.go
```diff
@@ -0,0 +1,103 @@
...
+// and retrieved successfully.
+func TestContextWithLoggerRoundTrip(t *testing.T) {
+	t.Parallel()
+
+	// Create a real logger for testing.
+	backend := btclog.NewDefaultHandler(os.Stdout)
+	logger := btclog.NewSLogger(backend.SubSystem("TEST"))
+
+	// Store the logger in a context.
+	ctx := ContextWithLogger(context.Background(), logger)
```
bhandras: nit: use `t.Context()` here and elsewhere.

bhandras (approved):
Nice! LGTM, pending linter fixes 🎉

sputn1ck (approved):
LGTM!

#NoGlobals

PR #27, opened 2025-11-26
docs+build: introduce ast-grep to encode some of the style guide elements not captured by the linter

Roasbeef:
This PR introduces ast-grep (sg) for automated style enforcement, encoding many of the formatting patterns documented in `docs/development_guidelines.md` as machine-checkable rules. The rules cover multi-line function call wrapping, struct literal formatting, switch/select case spacing, inline comments, and function definition wrapping. Log and error calls are correctly excluded from the symmetric wrapping rules since they follow different compact formatting guidelines.

Two new Makefile targets are added: make ast-lint to check for style issues and make ast-grep-fix to auto-fix safe violations. Both support a pkg parameter to focus on specific directories (e.g., make ast-lint pkg=wallet). Generated files like *.pb.go and sqlc output are automatically excluded from scanning.

The PR also updates CLAUDE.md and AGENTS.md to document ast-grep usage for both code search (structural pattern matching that understands Go syntax) and linting. Several existing style violations in the db package are fixed as part of this change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,43 @@
 - Spacing between switch/select cases
 - When wrapping function calls, put closing paren on its own line with all args on new lines
 
+### ast-grep for Code Search and Linting
+
+This project uses ast-grep (`sg`) for AST-level code search and style enforcement.
+
+**Code Search (prefer over grep for Go patterns):**
+- Use `sg run -p 'pattern' -l go` for structural code search
+- ast-grep understands Go syntax, so `sg run -p 'func $NAME($$$ARGS)' -l go` finds all functions
+- For simple text search, grep is fine; for code patterns, use ast-grep
+
+**Pattern examples:**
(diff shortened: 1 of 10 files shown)


Roasbeef:
Note that we don't add it to CI for now, as there're existing violations, but we can use this to extend things going forward.

bhandras (approved):
I think this is great, later when we have a formatter that covers all edge cases we can also just switch to that. LGTM!

sputn1ck (approved):
very cool! LGTM

Roasbeef:
> I think this is great, later when we have a formatter that covers all edge cases we can also just switch to that. LGTM!

Yep, the other thing is that you can put it in "fix" mode, and it'll _auto fix_ everything that's flagged. What I do now is I find something I didn't like formatting wise that was generated, fix it, then ask it to try to generate a new rule.

PR #29, opened 2025-12-02
chainsource: add IncludeBlock option for confirmation registration

Roasbeef:
This commit extends the confirmation registration API to optionally include the full block containing the confirmed transaction. This capability is needed for constructing merkle proofs, which require access to all transactions in the block to compute the merkle root and proof path.

This commit also fixes ast-grep style violations in test files.

### DIFF
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -138,11 +138,18 @@ func (b *LNDBackend) BroadcastTx(ctx context.Context, tx *wire.MsgTx,
 // receiving confirmation events.
 func (b *LNDBackend) RegisterConf(ctx context.Context,
 	txid *chainhash.Hash, pkScript []byte, numConfs uint32,
-	heightHint uint32) (*chainsource.ConfRegistration, error) {
+	heightHint uint32,
+	includeBlock bool) (*chainsource.ConfRegistration, error) {
 
-	// Register with lnd's notifier.
+	// Build options for lnd's notifier. If includeBlock is true, we use
+	// WithIncludeBlock() to request the full block in the confirmation.
+	var opts []chainntnfs.NotifierOption
+	if includeBlock {
+		opts = append(opts, chainntnfs.WithIncludeBlock())
+	}
+
 	event, err := b.notifier.RegisterConfirmationsNtfn(
-		txid, pkScript, numConfs, heightHint,
+		txid, pkScript, numConfs, heightHint, opts...,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to register confirmation: %w",
@@ -152,7 +159,6 @@ func (b *LNDBackend) RegisterConf(ctx context.Context,
 	// Create a channel to convert lnd's TxConfirmation to our type.
 	confChan := make(chan *chainsource.TxConfirmation, 1)
 
-	// Start a goroutine to convert and forward the confirmation.
 	go func() {
 		defer close(confChan)
 		defer event.Cancel()
@@ -163,12 +169,12 @@ func (b *LNDBackend) RegisterConf(ctx context.Context,
 				return
 			}
 
-			// Convert to our type.
 			conf := &chainsource.TxConfirmation{
 				BlockHash:   lndConf.BlockHash,
 				BlockHeight: lndConf.BlockHeight,
 				TxIndex:     lndConf.TxIndex,
 				Tx:          lndConf.Tx,
+				Block:       lndConf.Block,
 			}
 
 			confChan <- conf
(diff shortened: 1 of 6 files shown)


sputn1ck (approved):

chainsource/chainsource_test.go
```diff
@@ -1163,3 +1179,132 @@ func TestOutpointOrScriptKey(t *testing.T) {
...
+	require.Len(t, event.Block.Transactions, 2)
+
+	// Verify other fields.
+	require.Equal(t, blockHash, event.BlockHash)
+	require.Equal(t, int32(200), event.BlockHeight)
+	require.NotNil(t, event.Tx)
+}
+
+// TestConfActorNoIncludeBlock verifies that when IncludeBlock=false (default),
+// the Block field in the confirmation event is nil even if the backend provides
```
ellemouton: "even if backend provided it" i think is incorrect  cause i think this test mocks the backend not providing it

ellemouton (approved):
logic looks good - my main question is around why we need - we dont need to provide merkle proofs to the server since the server has a full-chain backend. So is this more for providing proofs directly to other peers?

chainbackends/lnd.go
```diff
@@ -138,11 +138,18 @@ func (b *LNDBackend) BroadcastTx(ctx context.Context, tx *wire.MsgTx,
 // receiving confirmation events.
 func (b *LNDBackend) RegisterConf(ctx context.Context,
 	txid *chainhash.Hash, pkScript []byte, numConfs uint32,
-	heightHint uint32) (*chainsource.ConfRegistration, error) {
+	heightHint uint32,
+	includeBlock bool) (*chainsource.ConfRegistration, error) {
```
bhandras, line 142: nit: formatting

chainsource/chainsource_test.go
```diff
@@ -1163,3 +1179,132 @@ func TestOutpointOrScriptKey(t *testing.T) {
 		})
 	}
 }
+
+// TestConfActorIncludeBlock verifies that when IncludeBlock=true, the Block
+// field in the confirmation event is populated from the backend.
```
bhandras: nit: the test really just tests the mock backend so this comment is misleading.

chainsource/chainsource_test.go
```diff
@@ -1163,3 +1179,132 @@ func TestOutpointOrScriptKey(t *testing.T) {
...
+func TestConfActorIncludeBlock(t *testing.T) {
+	t.Parallel()
+
+	ctx := t.Context()
+	backend := newMockBackend()
```
bhandras: Alternatively we can use the harness to test this functionality with the real backend. I think relying too much on fakes/mocks can backfire eventually as we're not really testing the validity of our system but that our system behaves as expected given mock implementation.

chainsource/chainsource_test.go
```diff
@@ -1163,3 +1179,132 @@ func TestOutpointOrScriptKey(t *testing.T) {
...
+}
+
+// TestConfActorNoIncludeBlock verifies that when IncludeBlock=false (default),
+// the Block field in the confirmation event is nil even if the backend provides
+// it.
+func TestConfActorNoIncludeBlock(t *testing.T) {
+	t.Parallel()
+
+	ctx := t.Context()
+	backend := newMockBackend()
```
bhandras: Same comments as above. Also the two tests could be unified into a single table driven one.

Roasbeef:
> logic looks good - my main question is around why we need - we dont need to provide merkle proofs to the server since the server has a full-chain backend. So is this more for providing proofs directly to other peers?

The goal is to make things somewhat asymmetric. The server can checks these without needing to fetch blocks, granted it has direct access to headers. This also lets the server run with just neutrino as well. This can also serve as a type of authentication proof in the future as well.

chainbackends/lnd.go
Roasbeef: How would you wrap this instead? If `includeBlock` is on the line above it, then the line would be too long. 

Something we can try to add to `make ast-lint` perhaps?

chainsource/chainsource_test.go
Roasbeef: I think both approaches are valid. Here we're zooming in on just basically parsing logic in the `chainsource` actor. As this is a unit test, it runs super fast compared to spinning up an entire node.

Roasbeef:
It's used in https://github.com/lightninglabs/darepo-client/pull/34 during the round join phase.

chainsource/chainsource_test.go
Roasbeef: Nice suggestion, squashed them into a single TDD.

PR #30, opened 2025-12-03
lib+wallet: add SigningKey descriptor and TxProof field

Roasbeef:
This PR adds two related changes to support boarding verification:

1. **SigningKey descriptor change**: Updates `VTXORequest.SigningKey` to use `keychain.Descriptor` instead of a raw public key, enabling proper key derivation for signing.

2. **TxProof field**: Adds an optional `TxProof` field to `BoardingRequest` and `BoardingChainInfo` types. This allows clients to provide SPV proof that their boarding transaction is confirmed, reducing server-side chain queries.

The taproot-assets v0.7.0 dependency is added to support the `proof.TxProof` type.


wallet/interfaces.go
```diff
@@ -207,6 +209,13 @@ type BoardingChainInfo struct {
 
 	// Amount is the boarding output amount.
 	Amount btcutil.Amount
+
+	// TxProof is the SPV proof that the boarding transaction exists in a
+	// confirmed block. This includes the merkle proof, block header, and
+	// output construction details needed for server verification without
+	// querying its own chain source. None if the proof hasn't been
+	// constructed yet (e.g., block data not available).
+	TxProof fn.Option[proof.TxProof]
```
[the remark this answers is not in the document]
Roasbeef: It's intentional for now, as it can be re-created by obtaining the block header, then fetching the block to reconstruct.

Roasbeef:
> Is the full taproot-assets package needed, or just the proof subpackage?


I had it like this initially, but we'll soon bring in the `tapd` project from the assets tree work.

sputn1ck (approved):

lib/types/boarding.go
```diff
@@ -79,9 +80,11 @@ type VTXORequest struct {
 	// construction of the collaborative spend path of the VTXO.
 	OperatorKey *btcec.PublicKey
 
-	// SigningKey is the public key that the client will use in the building
-	// of the VTXO tree during Musig2 signing sessions.
-	SigningKey *btcec.PublicKey
+	// SigningKey is the key descriptor that the client will use in the
+	// building of the VTXO tree during Musig2 signing sessions. We use
+	// keychain.KeyDescriptor instead of just *btcec.PublicKey because we
+	// need the key locator for signing operations.
```
ellemouton, line 88: this `VTXORequest` is meant to represent the info known to both server & client. Ie, the client sends this to the server & the server uses it to build the tree. The server should not know the descriptor that the client is using imo. I think it is worth separating these types out & keeping things very explicit else i think we will end up adding more things here that are client/server specific

Roasbeef: Re the server's view, the descriptor can be empty in terms of the key family, no? 

> this VTXORequest is meant to represent the info known to both server & client. Ie, the client sends this to the server & the server uses it to build the tree.

There'll be a proto translation somewhere there right? 

Understood that the types may not be being used as you originally intended, I didn't have full context when working on this series of changes. I added this in so I can pass the key desc directly into the set of relevant signer interfaces that typically take a key descriptor. 

So how about we make that split distinction once you start to integrate the client+server together in a future PR?

ellemouton: > There'll be a proto translation somewhere there right?

Yeah def, but imo nice to have a 1:1 mapping to a golang type that we expect the other side to use.

> So how about we make that split distinction once you start to integrate the client+server together in a future PR?

cool yeah sounds good 👍

ellemouton (approved):

ellemouton (approved):

PR #31, opened 2025-12-03
chainsource: add Tx field to ConfirmationEvent

Roasbeef:
Enhances the `ConfirmationEvent` structure by adding a `Tx` field that contains the complete confirmed transaction. This allows consumers of confirmation events to inspect transaction outputs directly without making additional chain queries.

This optimization is particularly useful for the boarding flow where the client needs to validate that the confirmed transaction matches expected outputs and amounts. By including the full transaction in the confirmation event, we eliminate an extra round trip to the chain backend.


Roasbeef:
This PR should be reviewed instead: https://github.com/lightninglabs/darepo-client/pull/29

PR #33, opened 2025-12-03
baselib/protofsm: add transition table type

Roasbeef:
Introduces a generic `TransitionTable` type that provides compile-time documentation of state machine behavior. The table captures all valid state transitions including source state, triggering event, destination state, emitted outbox messages, and human-readable descriptions.

The table serves multiple purposes:
- **Executable documentation** that stays in sync with implementation
- **Test validation target** ensuring all transitions are covered
- **Diagram generation** via RenderMermaid for Mermaid syntax output

Query methods allow looking up transitions by source state or event type using reflection for type matching. This enables tooling to generate visual diagrams that accurately reflect the implemented state machine.

### DIFF
--- a/baselib/protofsm/transition_table.go
+++ b/baselib/protofsm/transition_table.go
@@ -1,0 +1,209 @@
+package protofsm
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// TransitionEntry represents a single state transition with type safety.
+// This provides at-a-glance visibility into how states transition in response
+// to events, including any outbox messages emitted.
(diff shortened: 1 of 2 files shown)


sputn1ck (approved):
I guess this requires a user generated table right? An AST style thing should be fairly straightforward in the future right?

Roasbeef:
> I guess this requires a user generated table right? An AST style thing should be fairly straightforward in the future right?

Today correct. Yes possbile to use `go generate` to auto generate by parsing the AST, then we can add CI to make sure it's properly generated. I had a version of something similar to that, but just wanted to check this in first.

PR #34, opened 2025-12-03
round: implement boarding FSM state machine

Roasbeef:
Implements the client-side boarding round state machine for coordinating Bitcoin deposits into the Ark VTXO tree. The architecture strictly separates protocol logic from side effects: a deterministic FSM handles all state transitions and validation.

## Core Components

- **States** (`states.go`): Immutable state types from Idle through BoardingComplete
- **Events** (`events.go`, `common_events.go`): Event definitions driving the FSM
- **Transitions** (`transitions.go`): Core protocol logic for each state's event handling
- **Validation** (`validation.go`): Commitment transaction verification (boarding inputs, VTXO leaves)
- **Outbox Messages** (`outbox_messages.go`): Server communication types with ToProto methods

## Security Model

Security-critical validation ensures clients verify complete VTXO tree signatures before releasing boarding input signatures. This prevents malicious operators from capturing funds without providing promised VTXOs. See `round/README.md` for the full protocol flow.

## Testing

Test coverage includes:
- Validation tests with real MuSig2 signing infrastructure
- FSM test harness for state transition testing
- Multi-intent test scenarios

### DIFF
--- a/lib/tree/tree.go
+++ b/lib/tree/tree.go
@@ -6,7 +6,9 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/lib/scripts"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
 )
@@ -313,4 +315,163 @@ func (t *Tree) NewTreeSignerSession(wallet input.MuSig2Signer,
 		wallet, signerKey, t.SweepTapscriptRoot, prevOutFetcher,
 		t.Root,
 	)
+}
+
+// ValidatePath validates the tree path from the batch output to the client's
+// VTXO for a given signing key. Each signing key maps to exactly one VTXO,
+// providing a unique identifier for the VTXO and keeping MuSig2 signing simple.
+//
+// The client MUST validate the complete path before signing the boarding UTXO.
+// This ensures they can recover their funds even if the operator disappears.
+//
+// Returns the extracted client sub-tree on success.
+func (t *Tree) ValidatePath(signingKey *btcec.PublicKey,
+	expectedLeaf LeafDescriptor,
+	operatorKey *btcec.PublicKey) (*Tree, error) {
+
+	if t == nil {
+		return nil, fmt.Errorf("tree is nil")
+	}
+
+	// Use VerifyVTXOPath for core validation: script match, exactly 1 leaf
+	// per signing key, anchor output, and cosigner presence in all nodes.
+	if err := t.VerifyVTXOPath(
+		signingKey, expectedLeaf.PkScript,
+	); err != nil {
+		return nil, fmt.Errorf("VTXO path verification failed: %w", err)
+	}
+
+	// Extract the client's sub-tree for additional validation and to
+	// return to caller.
+	clientTree, err := t.ExtractPathForCoSigners(signingKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract client path: %w", err)
+	}
+
+	// Get the single leaf node (VerifyVTXOPath already enforced exactly 1).
+	leaves := clientTree.Root.GetLeafNodes()
+	leaf := leaves[0]
+	vtxoOutput := leaf.Outputs[0]
+
+	// Ensure the VTXO amount matches the request to prevent value
+	// extraction by the operator.
+	if vtxoOutput.Value != int64(expectedLeaf.Amount) {
+		return nil, fmt.Errorf("VTXO output value %d != expected %d",
+			vtxoOutput.Value, expectedLeaf.Amount)
+	}
+
+	// Confirm operator key is present as co-signer to enable collaborative
+	// spending path.
+	if !ContainsCosigner(leaf.CoSigners, operatorKey) {
+		return nil, fmt.Errorf("leaf does not include operator key " +
+			"in co-signers")
+	}
+
+	return clientTree, nil
+}
+
+// ValidateAndSubmitSignatures validates and submits the complete VTXT
+// signatures to the tree. This must be called BEFORE the client signs the
+// boarding UTXO input.
+//
+// The signatures are provided as a map from transaction ID to raw signature
+// bytes. Each entry corresponds to a transaction in the VTXT.
+//
+// The client MUST NOT sign the boarding UTXO until the VTXT is fully signed
+// and validated. Otherwise, the operator could include the boarding UTXO in a
+// commitment tx without providing valid VTXOs.
+func (t *Tree) ValidateAndSubmitSignatures(
+	signatures map[chainhash.Hash][]byte) error {
+
+	if t == nil {
+		return fmt.Errorf("tree is nil")
+	}
+	if len(signatures) == 0 {
+		return fmt.Errorf("no signatures provided")
+	}
+
+	// Parse each raw signature into a Schnorr signature structure.
+	sigMap := make(map[TxID]*schnorr.Signature, len(signatures))
+	for txid, sigBytes := range signatures {
+		sig, err := schnorr.ParseSignature(sigBytes)
+		if err != nil {
+			return fmt.Errorf("failed to parse signature for "+
+				"tx %s: %w", txid, err)
+		}
+		sigMap[txid] = sig
+	}
+
+	// Submit and validate all signatures atomically to ensure the entire
+	// tree is properly signed.
+	if err := t.SubmitTreeSigs(sigMap); err != nil {
+		return fmt.Errorf("failed to submit tree signatures: %w", err)
+	}
+
+	// Perform cryptographic verification of all signatures to guarantee
+	// the operator has properly co-signed the VTXT.
+	if err := t.VerifySigned(); err != nil {
+		return fmt.Errorf("tree signature verification failed: %w",
+			err)
+	}
+
+	return nil
+}
+
+// ValidateAnchors validates that all transactions in the tree have valid
+// ephemeral anchor outputs for CPFP fee bumping (BIP 431).
+//
+// Without valid anchors, the client cannot broadcast the VTXT chain for
+// unilateral exit, resulting in fund loss.
+func (t *Tree) ValidateAnchors() error {
+	if t == nil {
+		return fmt.Errorf("tree is nil")
+	}
+
+	// Validate anchors in all tree nodes recursively.
+	return t.Root.ForEach(func(node *Node) error {
+		if node == nil {
+			return fmt.Errorf("tree node is nil")
+		}
+
+		// Convert node to transaction to check version and outputs.
+		tx, err := node.ToTx()
+		if err != nil {
+			return fmt.Errorf("failed to convert node "+
+				"to tx: %w", err)
+		}
+
+		// Verify transaction version is 3 for BIP 431 ephemeral
+		// anchors.
+		if tx.Version != 3 {
+			return fmt.Errorf("transaction version is "+
+				"%d, expected 3 for BIP 431 ephemeral anchors",
+				tx.Version)
+		}
+
+		// All virtual transactions must have at least one output
+		// (anchor).
+		if len(node.Outputs) == 0 {
+			return fmt.Errorf("transaction has no outputs")
+		}
+
+		// The last output must be the ephemeral anchor.
+		anchorIdx := len(node.Outputs) - 1
+		anchorOutput := node.Outputs[anchorIdx]
+
+		// Anchor must have zero value.
+		if anchorOutput.Value != 0 {
+			return fmt.Errorf("anchor output at index %d "+
+				"has value %d, expected 0", anchorIdx,
+				anchorOutput.Value)
+		}
+
+		// Anchor script must match the standard ephemeral anchor
+		// script.
+		if !bytes.Equal(anchorOutput.PkScript, scripts.AnchorPkScript) {
+			return fmt.Errorf("anchor output at index %d has "+
+				"invalid script", anchorIdx)
+		}
+
+		return nil
+	})
 }
--- a/lib/tree/tree_test.go
+++ b/lib/tree/tree_test.go
@@ -978,3 +978,237 @@ func TestTreePrettyPrint(t *testing.T) {
 		require.Contains(t, output, "Transaction Tree")
 	})
 }
+
+// TestValidatePath tests the ValidatePath method.
+func TestValidatePath(t *testing.T) {
+	t.Parallel()
+
+	t.Run("nil tree returns error", func(t *testing.T) {
+		t.Parallel()
+
+		var tree *Tree
+		_, err := tree.ValidatePath(nil, LeafDescriptor{}, nil)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "tree is nil")
+	})
+
+	t.Run("nil signing key returns error", func(t *testing.T) {
+		t.Parallel()
+
+		_, operatorKey := createTestKey(t)
+		_, ownerKey := createTestKey(t)
+
+		leaves := []LeafDescriptor{
+			{
+				PkScript:    []byte("script"),
+				Amount:      1000,
+				CoSignerKey: ownerKey,
+			},
+		}
+
+		tree, err := NewTree(
+			wire.OutPoint{}, &wire.TxOut{Value: 1000}, leaves,
+			operatorKey, make([]byte, 32), 2,
+		)
+		require.NoError(t, err)
+
+		expectedLeaf := LeafDescriptor{
+			PkScript:    []byte("script"),
+			Amount:      1000,
+			CoSignerKey: ownerKey,
+		}
+		_, err = tree.ValidatePath(nil, expectedLeaf, operatorKey)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "cosigner key cannot be nil")
+	})
+
+	t.Run("valid path returns client tree", func(t *testing.T) {
+		t.Parallel()
+
+		_, operatorKey := createTestKey(t)
+		_, ownerKey := createTestKey(t)
+		vtxoScript := []byte("vtxo_script")
+
+		leaves := []LeafDescriptor{
+			{
+				PkScript:    vtxoScript,
+				Amount:      1000,
+				CoSignerKey: ownerKey,
+			},
+		}
+
+		tree, err := NewTree(
+			wire.OutPoint{Hash: chainhash.HashH([]byte("batch"))},
+			&wire.TxOut{Value: sumLeafAmounts(leaves)}, leaves,
+			operatorKey, make([]byte, 32), 2,
+		)
+		require.NoError(t, err)
+
+		expectedLeaf := LeafDescriptor{
+			PkScript:    vtxoScript,
+			Amount:      1000,
+			CoSignerKey: ownerKey,
+		}
+		clientTree, err := tree.ValidatePath(
+			ownerKey, expectedLeaf, operatorKey,
+		)
+		require.NoError(t, err)
+		require.NotNil(t, clientTree)
+
+		// Verify client tree has exactly one leaf.
+		leafNodes := clientTree.Root.GetLeafNodes()
+		require.Len(t, leafNodes, 1)
+	})
+
+	t.Run("amount mismatch returns error", func(t *testing.T) {
+		t.Parallel()
+
+		_, operatorKey := createTestKey(t)
+		_, ownerKey := createTestKey(t)
+		vtxoScript := []byte("vtxo_script")
+
+		leaves := []LeafDescriptor{
+			{
+				PkScript:    vtxoScript,
+				Amount:      1000,
+				CoSignerKey: ownerKey,
+			},
+		}
+
+		tree, err := NewTree(
+			wire.OutPoint{Hash: chainhash.HashH([]byte("batch"))},
+			&wire.TxOut{Value: sumLeafAmounts(leaves)}, leaves,
+			operatorKey, make([]byte, 32), 2,
+		)
+		require.NoError(t, err)
+
+		// Use wrong amount to trigger validation error.
+		expectedLeaf := LeafDescriptor{
+			PkScript:    vtxoScript,
+			Amount:      2000,
+			CoSignerKey: ownerKey,
+		}
+		_, err = tree.ValidatePath(ownerKey, expectedLeaf, operatorKey)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "VTXO output value")
+	})
+
+	t.Run("missing operator key returns error", func(t *testing.T) {
+		t.Parallel()
+
+		_, operatorKey := createTestKey(t)
+		_, ownerKey := createTestKey(t)
+		_, wrongOperatorKey := createTestKey(t)
+		vtxoScript := []byte("vtxo_script")
+
+		leaves := []LeafDescriptor{
+			{
+				PkScript:    vtxoScript,
+				Amount:      1000,
+				CoSignerKey: ownerKey,
+			},
+		}
+
+		tree, err := NewTree(
+			wire.OutPoint{Hash: chainhash.HashH([]byte("batch"))},
+			&wire.TxOut{Value: sumLeafAmounts(leaves)}, leaves,
+			operatorKey, make([]byte, 32), 2,
+		)
+		require.NoError(t, err)
+
+		expectedLeaf := LeafDescriptor{
+			PkScript:    vtxoScript,
+			Amount:      1000,
+			CoSignerKey: ownerKey,
+		}
+
+		// Use wrong operator key to trigger validation error.
+		_, err = tree.ValidatePath(
+			ownerKey, expectedLeaf, wrongOperatorKey,
+		)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "operator key")
+	})
+}
+
+// TestValidateAndSubmitSignatures tests the ValidateAndSubmitSignatures method.
+func TestValidateAndSubmitSignatures(t *testing.T) {
+	t.Parallel()
+
+	t.Run("nil tree returns error", func(t *testing.T) {
+		t.Parallel()
+
+		fakeTxid := chainhash.HashH([]byte("fake-tx"))
+		var tree *Tree
+		err := tree.ValidateAndSubmitSignatures(
+			map[chainhash.Hash][]byte{fakeTxid: {0x01}},
+		)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "tree is nil")
+	})
+
+	t.Run("empty signatures returns error", func(t *testing.T) {
+		t.Parallel()
+
+		_, operatorKey := createTestKey(t)
+		_, ownerKey := createTestKey(t)
+
+		leaves := []LeafDescriptor{
+			{
+				PkScript:    []byte("script"),
+				Amount:      1000,
+				CoSignerKey: ownerKey,
+			},
+		}
+
+		tree, err := NewTree(
+			wire.OutPoint{}, &wire.TxOut{Value: 1000}, leaves,
+			operatorKey, make([]byte, 32), 2,
+		)
+		require.NoError(t, err)
+
+		emptySigs := map[chainhash.Hash][]byte{}
+		err = tree.ValidateAndSubmitSignatures(emptySigs)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "no signatures provided")
+	})
+
+	t.Run("nil signatures returns error", func(t *testing.T) {
+		t.Parallel()
+
+		_, operatorKey := createTestKey(t)
+		_, ownerKey := createTestKey(t)
+
+		leaves := []LeafDescriptor{
+			{
+				PkScript:    []byte("script"),
+				Amount:      1000,
+				CoSignerKey: ownerKey,
+			},
+		}
+
+		tree, err := NewTree(
+			wire.OutPoint{}, &wire.TxOut{Value: 1000}, leaves,
+			operatorKey, make([]byte, 32), 2,
+		)
+		require.NoError(t, err)
+
+		err = tree.ValidateAndSubmitSignatures(nil)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "no signatures provided")
+	})
+}
+
+// TestValidateAnchors tests the ValidateAnchors method.
+func TestValidateAnchors(t *testing.T) {
+	t.Parallel()
+
+	t.Run("nil tree returns error", func(t *testing.T) {
+		t.Parallel()
+
+		var tree *Tree
+		err := tree.ValidateAnchors()
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "tree is nil")
+	})
+}
--- a/round/events.go
+++ b/round/events.go
@@ -1,0 +1,248 @@
+package round
+
+import (
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/lib/tree"
+	"github.com/lightninglabs/darepo-client/wallet"
+	"github.com/lightninglabs/taproot-assets/proof"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+)
+
+// ClientEvent is a sealed interface for all events that can be processed by
+// the client round interaction state machine. The sealed interface pattern
+// prevents external packages from implementing this interface, ensuring type
+// safety and exhaustive pattern matching in state transitions.
+type ClientEvent interface {
+	// clientEventSealed is an unexported method that marks this interface
+	// as sealed, preventing external implementations.
+	clientEventSealed()
+}
+
+// ClientOutMsg is a sealed interface for messages emitted via the FSM outbox.
+// These are typically sent to the server or other actors. Separating this from
+// ClientEvent improves readability by distinguishing internal FSM events from
+// outgoing messages.
+type ClientOutMsg interface {
+	// clientOutMsgSealed is an unexported method that marks this interface
+	// as sealed, preventing external implementations.
+	clientOutMsgSealed()
+}
+
+// ResumeBoardingIntents is emitted to instruct the FSM to resume monitoring
+// boarding intents that were in progress. The intents are provided as a
+// parameter (pre-filtered by the actor) rather than fetched from storage.
+type ResumeBoardingIntents struct {
+	// Intents are the boarding intents to resume, keyed by their outpoint.
+	// These are pre-filtered to include only Confirmed-but-not-Adopted
+	// intents.
+	Intents map[wire.OutPoint]BoardingIntent
+}
+
+func (e *ResumeBoardingIntents) clientEventSealed() {}
+
+// BoardingUTXOConfirmed is emitted when the boarding UTXO has received
+// sufficient confirmations and is ready to be used for boarding.
+type BoardingUTXOConfirmed struct {
+	// Outpoint identifies the confirmed boarding UTXO.
+	Outpoint wire.OutPoint
+
+	// Address is the boarding address for this UTXO. Contains the keys,
+	// tapscript, and exit delay needed to build the BoardingRequest.
+	Address wallet.BoardingAddress
+
+	// BlockHeight is the height at which the UTXO was confirmed.
+	BlockHeight int32
+
+	// BlockHash is the hash of the block containing the transaction.
+	BlockHash chainhash.Hash
+
+	// Confirmations is the number of confirmations the UTXO has.
+	Confirmations int32
+
+	// Tx is the confirmed transaction containing the boarding UTXO. This
+	// allows the FSM to extract output details without additional chain
+	// queries.
+	Tx *wire.MsgTx
+
+	// TxProof is the optional SPV proof for this boarding UTXO. Includes
+	// merkle proof, block header, and output construction details. None if
+	// the proof hasn't been constructed yet.
+	TxProof fn.Option[proof.TxProof]
+}
+
+func (e *BoardingUTXOConfirmed) clientEventSealed() {}
+
+// RegistrationRequested is emitted when the FSM is ready to join a round with
+// the currently confirmed set of boarding intents. The actor should treat this
+// as a batch request containing every confirmed intent.
+type RegistrationRequested struct {
+	Intents []BoardingIntent
+
+	// RoundID allows resuming a previously assigned round when rejoining.
+	RoundID string
+}
+
+func (e *RegistrationRequested) clientEventSealed() {}
+
+// RoundJoined is emitted when the server accepts the client's registration
+// and assigns them to a round. This event arrives via the Outbox from the
+// server FSM.
+type RoundJoined struct {
+	// RoundID is the unique identifier for the round.
+	RoundID string
+}
+
+func (e *RoundJoined) clientEventSealed() {}
+
+// CommitmentTxBuilt is emitted when the server sends the commitment
+// transaction and VTXT path to the client. This event arrives via the Outbox
+// from the server FSM after building the transaction that commits all boarding
+// UTXOs.
+type CommitmentTxBuilt struct {
+	// RoundID identifies which round this commitment transaction belongs
+	// to.
+	RoundID string
+
+	// Tx is the unsigned commitment transaction as a PSBT. Using PSBT
+	// allows the server to include WitnessUtxo for all inputs, which is
+	// required for correct Taproot sighash computation (BIP341).
+	Tx *psbt.Packet
+
+	// VTXTTree contains the client's extracted sub-tree from the virtual
+	// transaction tree. The server sends only the minimal path containing
+	// the transactions needed to reach this client's VTXO leaves, not the
+	// full tree (which may contain hundreds of transactions for all
+	// participants). This sub-tree is sufficient for the client to verify
+	// their VTXOs and perform unilateral exit if needed.
+	VTXTTree *tree.Tree
+}
+
+func (e *CommitmentTxBuilt) clientEventSealed() {}
+
+// CommitmentTxValidated is emitted after the client successfully validates
+// the commitment transaction and VTXT path. This is a critical security
+// checkpoint - the client must verify:
+//  1. Boarding UTXO is an input to commitment tx.
+//  2. VTXT path is valid and leads to expected VTXOs.
+//  3. VTXO amounts and scripts match requests.
+type CommitmentTxValidated struct {
+	// VTXTTree is the validated tree.
+	VTXTTree *tree.Tree
+}
+
+func (e *CommitmentTxValidated) clientEventSealed() {}
+
+// GenerateNonces is emitted when the client has validated the vtxt tree, and
+// nonces should be generated for all the client trees.
+type GenerateNonces struct {
+}
+
+func (e *GenerateNonces) clientEventSealed() {}
+
+// NoncesAggregated is emitted when the server sends back the aggregated
+// nonces from all participants. This event arrives via the Outbox from the
+// server FSM. The server computes aggregated nonces from all participants and
+// sends them back to clients, who use them to generate partial signatures in
+// the next phase of the MuSig2 signing protocol.
+type NoncesAggregated struct {
+	// RoundID identifies which round these aggregated nonces belong to.
+	RoundID string
+
+	// AggregatedNonces maps transaction IDs to their aggregated MuSig2
+	// nonces. Each entry corresponds to a transaction in the VTXT that
+	// requires signing.
+	AggregatedNonces map[chainhash.Hash][]byte
+}
+
+func (e *NoncesAggregated) clientEventSealed() {}
+
+// GeneratePartialSigs is emitted when the client has generated partial
+// signatures for all transactions in their VTXT path using the aggregated
+// nonces.
+type GeneratePartialSigs struct {
+	// PartialSigs maps transaction IDs to their MuSig2 partial signatures.
+	// Each entry corresponds to a transaction in the client's VTXT path.
+	PartialSigs map[chainhash.Hash][]byte
+
+	// SigningKey is the key used for signing.
+	SigningKey *btcec.PublicKey
+}
+
+func (e *GeneratePartialSigs) clientEventSealed() {}
+
+// OperatorSigned is emitted when the server sends the complete VTXT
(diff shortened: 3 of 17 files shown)


round/interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+	// boarding intent.
+	BoardingChainInfo = wallet.BoardingChainInfo
+
+	// WalletBoardingIntent is the wallet's boarding intent type.
+	WalletBoardingIntent = wallet.BoardingIntent
+)
+
+// BoardingStatus type alias and constants from wallet package. The wallet
+// owns the lifecycle of boarding intents; round just reacts to confirmations.
+type BoardingStatus = wallet.BoardingStatus
```
ellemouton, line 83: right now there are type aliases here but then one in `types.go` - feels like we should choose one location

round/interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+// ClientStateTransition is a type alias for the verbose protofsm.StateTransition
+// type used throughout the client boarding FSM. The baselib protofsm uses 3
+// type parameters: InternalEvent, OutboxEvent, and Env. In our case:
+//   - InternalEvent = ClientEvent (events that drive the FSM).
+//   - OutboxEvent = ClientEvent (outbox messages emitted by transitions).
+//   - Env = *ClientEnvironment.
+//
+//nolint:ll
+type ClientStateTransition = protofsm.StateTransition[
+	ClientEvent, ClientEvent, *ClientEnvironment,
```
ellemouton: are there events that are both incoming (Internal) events and outbox events? if not, i think it's nice for readability to make these distinct

round/interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// BoardingRequest is the original boarding request details. It targets
+	// a boarding address by outpoint and includes additional metadata.
+	BoardingRequest types.BoardingRequest
+
+	// VtxoTemplate is the template for the VTXO(s) requested as part of
+	// boarding.
+	//
+	// TODO(roasbeef): add auxleaf for tap here.
+	VtxoTemplate []types.VTXORequest
```
ellemouton, line 127: ok yeah so i know you said we perhaps have different mental models of how this is done & that i can edit this after merge. but just for early visibility: in my model, a boarding request is not really linked to the requested vtxos. instead, the user submist a JoinRequest as a whole which contains a set of: inputs (boarding requests and forfeits) and outputs (leave requests and vtxos). This makes things very flexible in that you can combine any number of boarding utxos with a set of vtxo descriptors (could have one boarding input funding multiple vtxos or multiple boarding inputs funding a single vtxo for example). 

I also like this cause then when we start doing "swap requests" etc, nothing changes in the round flow. all that changes is that the request now contains forfeit(s) and vtxo(s). It's also nice cause a single round can contain: leaves, new boarding inputs, new vtxos and forfeits
Another thing it allows would be: use both boarding utxo & forfeit to create a bigger single vtxo.

round/interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+type ChainBackend interface {
+	// BroadcastTx broadcasts a transaction to the Bitcoin network. Returns
+	// an error if the transaction is rejected by the network (e.g., double
+	// spend, invalid signature, insufficient fee).
+	BroadcastTx(tx *wire.MsgTx) error
+
+	// GetBestBlock returns the current best block hash and height. This is
+	// used for fee estimation and determining height hints for monitoring.
+	GetBestBlock() (*chainhash.Hash, int32, error)
+}
```
ellemouton: afaict, this isnt used at all in the `round` package atm (even in the actor pr)

round/interfaces.go
ellemouton: i think good to get on the same page with this part at least cause i think it will impact quite a lot of the design in terms of persistence etc

round/interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+	RoundID string
+
+	// CreationHeight is the block height when this round was created.
+	CreationHeight uint32
+
+	// CommitmentTx is the signed commitment transaction that anchors all
+	// VTXOs. None until the server constructs and broadcasts it.
+	CommitmentTx fn.Option[*wire.MsgTx]
+
+	// VTXTTree is the complete Virtual Transaction Tree. This is ephemeral
```
ellemouton: "complete vtx tree" - i think we will not send the client the full vtx tree (as this may be hundreds of transactions!) we instead send them the sub-tree that contains vtxs that they need to know to get to each of their requested vtxo leaves. 

also: a round may contain multiple batch/vtxo outputs each of which has a separate tree. so we should have some indication of which output is which tree and which vtxo is in which tree (although this is easy to tell since the vtxo signer key will be in the metadata of the first root tx of the vtx tree)

round/interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+// Round represents a complete Ark round from the client's perspective. A round
+// coordinates one or more actions (boarding, refresh, offboard) that are
+// batched together in a single commitment transaction.
+type Round struct {
+	// RoundID is the unique identifier assigned by the server when the
+	// client joins this round.
+	RoundID string
+
+	// CreationHeight is the block height when this round was created.
+	CreationHeight uint32
```
ellemouton: ConfirmedHeight?

round/interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+	ListActiveRounds() ([]*Round, error)
+
+	// FinalizeRound marks a round as complete and archives it.
+	FinalizeRound(roundID string, txid chainhash.Hash) error
+}
+
+// ClientVTXO represents a Virtual UTXO owned by the client, including all
+// information needed to spend it either collaboratively (with operator) or
+// unilaterally (after timelock expires).
+type ClientVTXO struct {
```
ellemouton, line 208: think we should also have some indication of which output in the round/batch tx is the tree output that this vtxo falls under

round/common_events.go
```diff
@@ -0,0 +1,90 @@
...
+	TxID chainhash.Hash
+
+	// VTXTTree contains the virtual transaction tree. For clients, this
+	// is their extracted path showing only their VTXOs. For the server,
+	// this is the complete tree for all participants.
+	VTXTTree *tree.Tree
+
+	// RoundInfo contains additional metadata about the round such as
+	// participant counts, timeouts, and operator parameters.
+	RoundInfo map[string][]byte
```
ellemouton: shall re remove unused fields like this for now?

round/common_events.go
```diff
@@ -0,0 +1,90 @@
...
+	// UTXOs and commits to the VTXT tree.
+	Tx *wire.MsgTx
+
+	// TxID is the transaction identifier computed from the unsigned
+	// commitment transaction.
+	TxID chainhash.Hash
+
+	// VTXTTree contains the virtual transaction tree. For clients, this
+	// is their extracted path showing only their VTXOs. For the server,
+	// this is the complete tree for all participants.
```
ellemouton: i think this sentence makes it seem like this is an event that both sides receive but i think this is an event received by the client but produced by the server (ie, it doesnt ever contain the full tree for all participants (unless there really is only this one participant)

round/common_events.go
```diff
@@ -0,0 +1,90 @@
...
+	// to.
+	RoundID string
+
+	// Tx is the unsigned commitment transaction that locks all boarding
+	// UTXOs and commits to the VTXT tree.
+	Tx *wire.MsgTx
+
+	// TxID is the transaction identifier computed from the unsigned
+	// commitment transaction.
+	TxID chainhash.Hash
```
ellemouton: derivable from Tx above no?

round/common_events.go
```diff
@@ -0,0 +1,90 @@
...
+// back to clients, who use them to generate partial signatures in the next
+// phase of the MuSig2 signing protocol.
+type NoncesAggregatedEvent struct {
+	// RoundID identifies which round these aggregated nonces belong to.
+	RoundID string
+
+	// AggregatedNonces contains the combined nonces for each transaction
+	// in the VTXT that requires MuSig2 signing. The order corresponds to
+	// the transaction ordering in the VTXT tree.
+	AggregatedNonces [][]byte
```
ellemouton: i think can instead just be a map from txid to nonce (although i guess that takes more space than necessary...)

round/states.go
```diff
@@ -0,0 +1,322 @@
...
+	"fmt"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/baselib/protofsm"
+	"github.com/lightninglabs/darepo-client/lib/tree"
+)
+
+// ClientState is a sealed interface for all states in the client boarding
+// state machine. Each state implements ProcessEvent to handle events and
```
ellemouton: linked to my previous comment, i think calling it "boarding fsm" makes it sound like it is only for boarding utxos but i think this FSM should more generally just handle "round interaction" and should be able to handle a flow that, for example, only has forfeit inputs & vtxo/leave outputs

round/states.go
```diff
@@ -0,0 +1,322 @@
...
+
+	CommitmentTx *wire.MsgTx
+
+	VTXTTree *tree.Tree
+
+	Intents []BoardingIntent
+
+	ClientTrees map[string]*tree.Tree
+
+	Musig2Sessions map[string]*tree.SignerSession
```
ellemouton: what are the keys here?

round/fsm_environment.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/lightninglabs/darepo-client/lib/types"
+)
+
+// OperatorTerms wraps the lib OperatorTerms with additional policy knobs the
+// client actor layer needs (sweep keys, fee targets, confirmation thresholds).
+type OperatorTerms struct {
+	*types.OperatorTerms
```
ellemouton: let's rather edit the OG struct?

round/validation.go
```diff
@@ -0,0 +1,387 @@
...
+	"github.com/lightninglabs/darepo-client/lib/types"
+)
+
+// ValidateVTXTPath validates the VTXT tree path from the batch output to
+// the client's VTXOs. This ensures the client can unilaterally unroll their
+// VTXOs if needed.
+//
+// The client MUST validate the complete VTXT path before signing the boarding
+// UTXO. This ensures they can recover their funds even if the operator
+// disappears.
```
ellemouton: i think a couple of these should instead be methods on tree.Tree

ellemouton (commented):
cool to see this!
I think take a look at my main comment re what constitutes a "round" (ie, i dont think we should think of boarding as a completely separate flow from forfeit/leave etc) as i think we should at least agree on that as i think that could change quite a lot here.

round/interfaces.go
Roasbeef: So you mean another interface type here? I added this as I rebased to use the `baselib` version, which added the new type param for the outbox types. Before this I just modeled the outbox types as normal client events.

round/interfaces.go
Roasbeef: Makes sense, can remove `types.go` and just combine here.

round/interfaces.go
Roasbeef: > in my model, a boarding request is not really linked to the requested vtxos.
> instead, the user submist a JoinRequest as a whole which contains a set of: inputs (boarding requests and forfeits) and outputs (leave requests and vtxos). 

I think that makes sense, I purposefully chose to only implement boarding in the confines of this round. All the other components you mentioned are indeed relevant, but would further increase the size of an already large PR. 

Agree that the join request would have all those components, but wouldn't a boarding input _still_ be mapped to a unique VTXO in the created tree? 

I guess you're thinking of the case where we combine some forfeits and also the boarding input into a single VTXO? 

I think that model can work as well. My goal here though is to implement a purposefully restrcited model to start to incrementally check in code that can partially be extended or modified later. 

Re persistence if you look if you look at the `Round` struct I have here, I think it encompasses the model you have in mind?

round/interfaces.go
Roasbeef: Yeah added it, but not used atm, as the server is the one that's able to broadcast the round as is. 

Will remove.

round/interfaces.go
Roasbeef: Make sense, in any case I always extract the sub-path specific to a client here. 

This is separated from the RPC layer right now, and the actual `tree.Tree` type is the same (IIUC?) for the entire tree,, or just a client sub-tree.

Roasbeef: > also: a round may contain multiple batch/vtxo outputs each of which has a separate tree. so we should have some indication of which output is which tree and which vtxo is in which tree (although this is easy to tell since the vtxo signer key will be in the metadata of the first root tx of the vtx tree)

Makes sense. I chose to implement the simplest version here where there's only a single round output in a commitment transaction. We can add another type at some point to indicate which output/tree the VTXOs live in.

round/interfaces.go
Roasbeef: Intention here is that this is used as a height hint. Not used atm, so will remove it.

round/interfaces.go
Roasbeef: Sure, I think that's slightly out of the scope of this PR as is. This implements a simpler version where there's only a single round output in a commitment txn. 

In either case, I think that would mean just adding a output index somewhere?

round/common_events.go
Roasbeef: Will do!

round/common_events.go
Roasbeef: I put it in this file as it's an event that maps to a proto to be sent by the server. 

Agree the comment isn't entirely accurate, can remove it.

round/common_events.go
Roasbeef: Indeed will remove.

round/common_events.go
Roasbeef: I think that makes sense. This was one of my key questions above re: the format we expect to send/recv the nonces in.

round/states.go
Roasbeef: Sure. I wouldn't get too caught up in this comment. Today it implements just the boarding flow. Tomorrow it will implement generic round interaction.

round/states.go
Roasbeef: Right now the public keys of each of the signers. Will use a better type here.

round/fsm_environment.go
Roasbeef: Sure.

round/validation.go
Roasbeef: Makes sense, so `ValidateVTXTSignatures` and this one?

round/common_events.go
Roasbeef: Modified to send just a single message with the txid map as you suggest.

round/common_events.go
Roasbeef: Removed.

round/interfaces.go
Roasbeef: Spun out a new type for this.

lib/tree/tree.go
```diff
@@ -314,3 +316,239 @@ func (t *Tree) NewTreeSignerSession(wallet input.MuSig2Signer,
...
+// ValidatePath validates the tree path from the batch output to the client's
+// VTXOs for a given signing key. It ensures the client can unilaterally unroll
+// their VTXOs if needed.
+//
+// The client MUST validate the complete path before signing the boarding UTXO.
+// This ensures they can recover their funds even if the operator disappears.
+//
+// Returns the extracted client sub-tree on success.
+func (t *Tree) ValidatePath(signingKey *btcec.PublicKey,
+	expectedVTXOs []VTXOExpectation) (*Tree, error) {
```
ellemouton: i think this is incorrect. I think we want one signing key per VTXO descriptor. That way there is a unique identifier for the VTXO. This also keeps the musig2 signing of the vtxs simple as if there are 4 leaves down stream of a vtxo, then there are 4 keys/sigs for that node. 

if we dont do it this way, there is a significant change in client & server side... im worried about the parallel work just not being compatible

lib/tree/tree_test.go
```diff
@@ -978,3 +978,120 @@ func TestTreePrettyPrint(t *testing.T) {
 		require.Contains(t, output, "Transaction Tree")
 	})
 }
+
+// TestValidatePath tests the ValidatePath method.
+func TestValidatePath(t *testing.T) {
```
ellemouton, line 983: dont think this actually tests anything useful?

lib/tree/tree_test.go
```diff
@@ -978,3 +978,120 @@ func TestTreePrettyPrint(t *testing.T) {
...
+		require.NoError(t, err)
+
+		_, err = tree.ValidatePath(nil, nil)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "signing key is nil")
+	})
+}
+
+// TestValidateAndSubmitSignatures tests the ValidateAndSubmitSignatures method.
+func TestValidateAndSubmitSignatures(t *testing.T) {
```
ellemouton, line 1135: same here

lib/tree/tree.go
```diff
@@ -314,3 +316,239 @@ func (t *Tree) NewTreeSignerSession(wallet input.MuSig2Signer,
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to build signature map: %w", err)
+	}
+
+	// Submit and validate all signatures atomically to ensure the entire
+	// tree is properly signed.
+	if err := t.SubmitTreeSigs(sigMap); err != nil {
+		return fmt.Errorf("failed to submit tree signatures: %w", err)
+	}
```
ellemouton, line 408: yeah - so i think this should just be called directly?

lib/tree/tree.go
```diff
@@ -314,3 +316,239 @@ func (t *Tree) NewTreeSignerSession(wallet input.MuSig2Signer,
 		t.Root,
 	)
 }
+
+// VTXOExpectation describes expected VTXO leaf outputs for path validation.
+type VTXOExpectation struct {
```
ellemouton: cant we use the existing `LeafDescriptor`?

lib/tree/tree.go
ellemouton: also see existing `VerifyVTXOPath`

round/interfaces.go
ellemouton: >  but wouldn't a boarding input still be mapped to a unique VTXO in the created tree?

not necessarily no - could be that multiple boarding inputs utxos are used to create a single VTXO or any number of VTXOs _or_ could be that no boarding inputs are involved at all and only forfeits are the inputs. 

> I guess you're thinking of the case where we combine some forfeits and also the boarding input into a single VTXO?

yes - i think this flexibility is very useful 

> Re persistence if you look if you look at the Round struct I have here, I think it encompasses the model you have in mind?

indeed 👍

round/interfaces.go
```diff
@@ -0,0 +1,274 @@
...
+	// CommitmentTx is the signed commitment transaction that anchors all
+	// VTXOs. None until the server constructs and broadcasts it.
+	CommitmentTx fn.Option[*wire.MsgTx]
+
+	// VTXTTree is the client's extracted sub-tree from the Virtual
+	// Transaction Tree. This contains only the minimal path needed to
+	// reach the client's VTXO leaves, not the full tree. This is ephemeral
+	// and only kept in memory during signing. After confirmation, only
+	// per-VTXO paths are persisted for unilateral exit.
+	VTXTTree fn.Option[*tree.Tree]
```
ellemouton, line 152: think we want a map from output index to tree as there may be multiple trees. but can also add later i guess

round/interfaces.go
```diff
@@ -0,0 +1,274 @@
...
+
+	// BoardingGroup contains all boarding intents for this round.
+	BoardingGroup *BoardingGroup
```
ellemouton: both Round and BoardingGroup have a RoundID. should this maybe just be []BoardingIntent? guess i dont really get the point of the `*Group` structs

round/interfaces.go
```diff
@@ -0,0 +1,274 @@
+package round
```
ellemouton, line 1: think the commit message is outdated

round/common_events.go
```diff
@@ -0,0 +1,81 @@
+package round
```
ellemouton: commit message doesnt match commit contents

round/common_events.go
```diff
@@ -0,0 +1,81 @@
...
+// that commits all boarding UTXOs, while clients receive it to validate the
+// transaction structure and their VTXT paths.
+type CommitmentTxBuiltEvent struct {
+	// RoundID identifies which round this commitment transaction belongs
+	// to.
+	RoundID string
+
+	// Tx is the unsigned commitment transaction that locks all boarding
+	// UTXOs and commits to the VTXT tree.
+	Tx *wire.MsgTx
```
ellemouton: psbt right? 

since we need the prev out info of all the other inputs too in order to sign for our boarding inputs afaiu

round/common_events.go
```diff
@@ -0,0 +1,81 @@
...
+// transaction in the VTXT. Clients must validate these signatures before
+// proceeding to sign the boarding input.
+type OperatorSignedEvent struct {
+	// RoundID identifies which round these signatures belong to.
+	RoundID string
+
+	// Signatures contains the complete aggregated Schnorr signatures for
+	// each transaction in the VTXT. The order corresponds to the
+	// transaction ordering in the VTXT tree.
+	Signatures [][]byte
```
ellemouton: i think rather do the same as done for AggNonces & use a map[txid]-> bytes

round/common_events.go
```diff
@@ -0,0 +1,81 @@
...
+// and timestamps for tracking purposes.
+//
+// TODO(boarding): Evaluate if we need a custom wrapper struct or if lib types
+// are sufficient. If we need to add boarding-specific metadata (e.g., round
+// ID, boarding UTXO reference, creation timestamp), create a BoardingVTXO
+// wrapper struct that embeds the lib VTXO type.
+type VTXOInfo struct {
+	// TODO: Add boarding-specific fields here if needed. For now, we'll
+	// use lib types directly.
+}
```
ellemouton: unused afaict

round/events.go
```diff
@@ -0,0 +1,222 @@
...
+	TxProof fn.Option[proof.TxProof]
+}
+
+func (e *BoardingUTXOConfirmed) clientEventSealed() {}
+
+// RegistrationRequested is emitted when the FSM is ready to join a round with
+// the currently confirmed set of boarding intents. The actor should treat this
+// as a batch request containing every confirmed intent.
+type RegistrationRequested struct {
+	Intents []BoardingIntent
```
ellemouton, line 82: its a bit confusing to me that this is a set of boarding intents which then maps to another set of vtxos inside. 

i know we keep raising this point but am a bit concerned that our ideas wont be compatible. So i'd here have a set of boarding intents and then along side that a set of vtxo intents.

round/events.go
```diff
@@ -0,0 +1,222 @@
...
+
+// RoundJoined is emitted when the server accepts the client's registration
+// and assigns them to a round. This event arrives via the Outbox from the
+// server FSM.
+type RoundJoined struct {
+	// RoundID is the unique identifier for the round.
+	RoundID string
+
+	// SessionInfo contains any session-specific data from the operator.
+	SessionInfo map[string][]byte
```
ellemouton: remove unused?

round/events.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (e *RoundJoined) clientEventSealed() {}
+
+// CommitmentTxBuilt is emitted when the server sends the commitment
+// transaction and VTXT path to the client. This event arrives via the Outbox
+// from the server FSM. It embeds the common event type.
+type CommitmentTxBuilt struct {
+	CommitmentTxBuiltEvent
```
ellemouton: wondering if we shouldnt just keep things explicit for now since they arent really shared/common types at the moment?

round/events.go
```diff
@@ -0,0 +1,222 @@
...
+// checkpoint - the client must verify:
+//  1. Boarding UTXO is an input to commitment tx.
+//  2. VTXT path is valid and leads to expected VTXOs.
+//  3. VTXO amounts and scripts match requests.
+type CommitmentTxValidated struct {
+	// VTXTTree is the validated tree.
+	VTXTTree *tree.Tree
+
+	// ValidationInfo contains details about what was validated.
+	ValidationInfo map[string]interface{}
```
ellemouton: unsure of what this is for

round/events.go
```diff
@@ -0,0 +1,212 @@
...
+	// BlockHeight is the height at which the transaction was confirmed.
+	BlockHeight int32
+
+	// Confirmations is the number of confirmations.
+	Confirmations int32
+
+	// VTXOs are the virtual UTXOs the client now owns. TODO: Evaluate if
+	// we need to wrap lib VTXO types with boarding-specific metadata. For
+	// now, storing as opaque data.
+	VTXOs []byte
```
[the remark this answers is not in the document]
ellemouton: yeah maybe remove?

round/states.go
```diff
@@ -0,0 +1,343 @@
+package round
```
ellemouton, line 1: lots of godocs missing here

round/states.go
```diff
@@ -0,0 +1,343 @@
...
+
+// PendingRoundAssembly tracks all active boarding intents that have been
+// funded on-chain but not yet fully confirmed. Intents are keyed by their
+// on-chain outpoint for efficient lookup when confirmation events arrive. Once
+// all intents reach the required confirmations, the FSM transitions to round
+// registration.
+type PendingRoundAssembly struct {
+	// Intents maps outpoint to boarding intent. Only intents with on-chain
+	// UTXOs (i.e., ChainInfo.OutPoint set) are included in this map.
+	Intents map[wire.OutPoint]BoardingIntent
```
ellemouton, line 52: ok nice, so then we can  just get a set of boarding inputs for a given round 👍

round/states.go
```diff
@@ -0,0 +1,343 @@
...
+
+	TxID chainhash.Hash
+
+	VTXTTree *tree.Tree
+
+	Intents []BoardingIntent
+
+	// ClientTrees maps signer keys (compressed pubkeys) to the client's
+	// extracted sub-tree for that VTXO.
+	ClientTrees map[SignerKey]*tree.Tree
```
ellemouton, line 123: i think this may store lots of duplicate tree nodes potentially. See `ExtractPathForCoSigners` which returns a single tree but with a leaf for each signer given so that a single tree object can be used for a set of signers

round/states.go
```diff
@@ -0,0 +1,343 @@
...
+
+	// Musig2Sessions maps signer keys (compressed pubkeys) to the MuSig2
+	// signing session for that VTXO.
+	Musig2Sessions map[SignerKey]*tree.SignerSession
```
ellemouton, line 229: nice

round/states.go
```diff
@@ -0,0 +1,343 @@
...
+
+	VTXTTree *tree.Tree
+
+	Intents []BoardingIntent
+
+	// ClientTrees maps signer keys (compressed pubkeys) to the client's
+	// extracted sub-tree for that VTXO.
+	ClientTrees map[SignerKey]*tree.Tree
+
+	PartialSigs [][]byte
```
ellemouton: id thin that the Musig2Sessions signer sessions would carry this data?

round/states.go
```diff
@@ -0,0 +1,343 @@
...
+
+	// Musig2Sessions maps signer keys (compressed pubkeys) to the MuSig2
+	// signing session for that VTXO.
+	Musig2Sessions map[SignerKey]*tree.SignerSession
+
+	// BoardingInputIndices maps each boarding intent's outpoint to its
+	// position in the commitment transaction inputs. Used for signing.
+	BoardingInputIndices map[wire.OutPoint]int
+}
+
```
ellemouton, line 276: are we missing an `AggSigsReceived` state?

round/fsm_environment.go
```diff
@@ -0,0 +1,51 @@
+package round
```
ellemouton, line 1: commit message mentions errorReporter

round/validation.go
```diff
@@ -0,0 +1,162 @@
+//nolint:ll
```
ellemouton: cheeky agent?

round/validation.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// Sweep delay must be greater than VTXO exit delay for security.
+	// This ensures the operator has time to respond to griefing attacks.
+	if sweepDelay <= vtxoExitDelay {
+		return fmt.Errorf("sweep delay (%d) must be greater than "+
+			"VTXO exit delay (%d)", sweepDelay, vtxoExitDelay)
+	}
+
+	// Sanity check: Delays should be reasonable (less than ~1 year).
+	const maxReasonableDelay = 52560 // ~1 year in blocks (10 min blocks)
```
ellemouton: make global?

round/outbox_messages.go
```diff
@@ -0,0 +1,193 @@
...
+type JoinRoundRequest struct {
+	actor.BaseMessage
+
+	// BoardingRequests contains all boarding UTXO details for this
+	// session. Each confirmed intent contributes exactly one boarding
+	// request so the server can register them in a single batch.
+	BoardingRequests []types.BoardingRequest
+
+	// VTXORequests specifies the VTXOs the client wants to receive.
+	VTXORequests []types.VTXORequest
```
ellemouton, line 21: ok yes i like this :)

round/outbox_messages.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// RoundID is optional; when empty it instructs the server to assign
+	// a new round. When non-empty, the request is for the specified round.
+	RoundID string
```
ellemouton: i think we'll never know the roundID at this point?

round/outbox_messages.go
```diff
@@ -0,0 +1,193 @@
...
+// SubmitNoncesRequest is sent from client to server with MuSig2 nonces.
+// This implements ClientOutMsg and is emitted via Outbox.
+type SubmitNoncesRequest struct {
+	actor.BaseMessage
+
+	// RoundID identifies the round.
+	RoundID string
+
+	// ParticipantKey identifies the participant.
+	ParticipantKey *btcec.PublicKey
```
ellemouton: in what way does this identify the participant? should it have been in JoinReq too?

round/outbox_messages.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// Nonces maps transaction IDs to their MuSig2 public nonces. Each
+	// entry corresponds to a transaction in the client's VTXT path that
+	// requires signing.
+	Nonces map[chainhash.Hash][]byte
```
ellemouton: think we need `map[txid]map[signingKey]nonce` since for each vtxo, we'll have a different signing session

round/outbox_messages.go
ellemouton: and then perhaps some sig over the message?

round/transitions.go
```diff
@@ -0,0 +1,1093 @@
...
+			{
+				Amount: btcutil.Amount(
+					confirmedOutput.Value,
+				),
+				PkScript:    confirmedOutput.PkScript,
+				ClientKey:   evt.Address.KeyDesc.PubKey,
+				OperatorKey: evt.Address.OperatorKey,
+				Expiry:      evt.Address.ExitDelay,
+				SigningKey:  evt.Address.KeyDesc,
+			},
```
ellemouton, line 157: ok cool, so we can expand on this later on to make it a "set" that was collected by the FSM perhaps

round/transitions.go
```diff
@@ -0,0 +1,1093 @@
...
+		intentMap[evt.Outpoint] = intent
+
+		return &ClientStateTransition{
+			NextState: &PendingRoundAssembly{
+				Intents: intentMap,
+			},
+		}, nil
+
+	default:
+		return nil, fmt.Errorf("idle: unexpected event: %T", event)
```
ellemouton: do we want to handle this in a "non error" way as you mentioned for server side fsm?

round/transitions.go
```diff
@@ -0,0 +1,1093 @@
...
+// ProcessEvent for PendingRoundAssembly tracks confirmed boarding intents and
+// transitions to registration once all are ready.
+func (s *PendingRoundAssembly) ProcessEvent(
+	_ context.Context, event ClientEvent, env *ClientEnvironment,
+) (*ClientStateTransition, error) {
+
+	switch evt := event.(type) {
+	// A new boarding UTXO was confirmed. The wallet handles persistence;
+	// we just add to our internal map.
+	case *BoardingUTXOConfirmed:
```
ellemouton, line 192: can we extract the duplicate handling logic into a helper?

ellemouton: also, my concern with this FSM approach is:

imo iiuc, we should really always be able to handle the `BoardingUTXOConfirmed` event since it could really come at any time & we should just collect these and then occationally check if we have a bunch of ready boarding reqs at which point it then makes sense to start a round join FSM.

ellemouton: so in otherwords, in my mental model, the FSM should just handle the process of actually working through things with the server. but the collection of intents to me is sort of separate

round/transitions.go
```diff
@@ -0,0 +1,1093 @@
...
+			NextState: &ClientFailedState{
+				Reason:      evt.Reason,
+				Error:       evt.Error,
+				Recoverable: evt.Recoverable,
+			},
+		}, nil
+
+	default:
+		return nil, fmt.Errorf("registration_sent: unexpected "+
+			"event: %T", event)
```
ellemouton: what happens to all the collected intents when things error?

round/transitions.go
```diff
@@ -0,0 +1,1093 @@
...
+			expectation := tree.VTXOExpectation{
+				Amount:      vtxoReq.Amount,
+				PkScript:    vtxoReq.PkScript,
+				OperatorKey: env.OperatorTerms.PubKey,
+			}
+
+			clientTree, err := s.VTXTTree.ValidatePath(
+				vtxoReq.SigningKey.PubKey,
+				[]tree.VTXOExpectation{expectation},
+			)
```
ellemouton, line 481: think this confirms my previous comment that it only makes sense to have a single expectation per signing key

round/transitions.go
```diff
@@ -0,0 +1,1093 @@
...
+) (*ClientStateTransition, error) {
+
+	switch event.(type) {
+	case *GeneratePartialSigs:
+		// At this stage, the nonces have been aggregated for each
+		// client, so now we'll generate and send our partial
+		// signatures.
+		//
+		// TODO(roasbeef); how should the partial sigs actually be
+		// assembled?
```
ellemouton: the SigningSessions should handle this for youi think?

round/validation_test.go
```diff
@@ -0,0 +1,493 @@
+package round
```
ellemouton, line 1: think we may want to consolidate these, add godocs etc

ellemouton (commented):
looking good! leaving some comments

round/events.go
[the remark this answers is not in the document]
Roasbeef: Yep, later on in the PR series there's a concrete VTXO type (once we get to persistence) will remove.

lib/tree/tree.go
Roasbeef: > I think we want one signing key per VTXO descriptor. That way there is a unique identifier for the VTXO. This also keeps the musig2 signing of the vtxs simple as if there are 4 leaves down stream of a vtxo, then there are 4 keys/sigs for that node.

That makes sense, I can make this change in this PR. I wasn't sure which route we wanted to go with.

lib/tree/tree_test.go
Roasbeef: Wouldn't disagree, it's here just so the reported test coverage doesn't drop. If we don't want it at all, I can remove it.

lib/tree/tree.go
Roasbeef: Looking at `VerifyVTXOPath`, that looks to verify a subset of what this does. It's missing: the amount check, and operator key. 

I'll update this to use `VerifyVTXOPath` as a helper function.

lib/tree/tree.go
Roasbeef: Can you elaborate?

Roasbeef: This is called directly, but we add some additional parsing logic to extract the signatures map them to txid, etc.

lib/tree/tree.go
Roasbeef: Also even with this change, nothing needs to change with the core logic, it just becomes more restricted (so only a single vtxo expectation passed iyeah n).

lib/tree/tree.go
Roasbeef: Sure, the extra thing we expectation has rn is the operator key. I'll leaf this as is, then add it as a param to the validation method.

round/interfaces.go
Roasbeef: Client logic doesn't support this yet (as this is a scaling concern we don't have yet rn), but it won't be a stretch to add it. 

I'll make an issue right now to track it.

round/interfaces.go
Roasbeef: Good simplification, will just use the intents directly.

round/common_events.go
Roasbeef: Sure, this'll save us some other logic then to traverse the tree to construct this map.

round/common_events.go
Roasbeef: Yup, will update to use a PSBT here instead.

Roasbeef: Is your vision that the server side also sends PSBTs for each of the round transactions? In theory we know the prev out info for al the inputs there already as they're created from the tree, no?

round/common_events.go
Roasbeef: Removed!

round/events.go
Roasbeef: No worries at all, code is malleable!  

I can just have it be 1:1 if you'd like. My thinking is that I can board a 1 BTC input, but then produce smaller 10 million sat VTXOs, so I do a fan out during boarding. Does that make sense? 

So I don't think this is incompatible, as for now the server can just basically assume that each boarding input can only get a single VTXO.

round/events.go
Roasbeef: You mean drop the struct embed? Sure.

round/events.go
Roasbeef: Dropped.

round/states.go
Roasbeef: Ah I think you added that after this was written initially. I'll switch over to that for now. As far as the many of nodes, in the short term it'll be negligible IMO. 

I think I'll also update some of the persistence code due to this. So now for a given batch output, we'll store a single VTXO tree which is a projection of the nodes that we need for the set of VTXOs we own. Then in memory we can either point into that, or have some other mapping.

round/states.go
Roasbeef: So this is the signatures that we generated using the tre signer session. Right now it doesn't cache them or anything, so I store them here by themselves.

round/states.go
Roasbeef: So we just receive `OperatorSIgnedEvent` which contains those signatures. I'll switch this to use a map by txid once again. 

Once we get those signatures, we validate, sign for out boarding input, then go to wait for confirmation.

round/validation.go
Roasbeef: Done!

round/outbox_messages.go
Roasbeef: Right you are, removed.

round/outbox_messages.go
Roasbeef: I assume that by this point, any sort of identify challenge has already happened over the RPC layer.

Roasbeef: Good point re if it identifies for not. Will add to the join request for now for symmetry's sake.

round/outbox_messages.go
Roasbeef: Makes sense, will update

round/transitions.go
Roasbeef: Yup, I'll add this now in this PR.

round/transitions.go
Roasbeef: Have you seen how this hooks up into the wallet? 

Agree that we need to be able to handle this in other states. My idea is that the actor controller will suppress these as needed, then batch them in to send when needed. I had implemented a simpler version of this, left to do the wallet, but didn't fully update the design here. 

Will update it in this PR (I have the actor + fsm stacked locally).

round/transitions.go
Roasbeef: So they're all still persistent on disk, and their state hasn't been modified yet. I'll clear up the error handling here with the move to make a proper error state. 

In the context of the actor driving this, we rn see exactly which messages are sent in, so I think we can actually just make this a noop to make things a bit more robust.

round/transitions.go
Roasbeef: Yup, have updated that in this diff.

round/transitions.go
Roasbeef: Yep

round/validation_test.go
Roasbeef: What do you mean by consolidate? They all contribute to test coverage. So you mean instead I should make them all sub tests?

round/states.go
Roasbeef: For each VTXO, do you think it's useful to store _just_ the relevant nodes that pertain to that on disk? That way if I want to go on chain for a single VTXO, I end up reading _less_ data instance.

round/states.go
Roasbeef: Actually this isn't even used at all, so I'll drop it.

round/outbox_messages.go
Roasbeef: Decided to just remove it for now. For the e2e test we can add some light pubkey layer for demux.

round/interfaces.go
ellemouton: still the case

round/interfaces.go
```diff
@@ -0,0 +1,267 @@
...
+	MarkVTXOSpent(outpoint wire.OutPoint) error
+}
+
+// ClientWallet defines the interface for client wallet operations used by the
+// round actor. This provides MuSig2 signing capabilities needed for round
+// participation. Boarding address creation is handled by the wallet actor.
+type ClientWallet interface {
+	input.MuSig2Signer
+
+	input.Signer
```
ellemouton, line 267: i think input.Signer already embeds input.MuSig2Signer?

round/events.go
ellemouton: > My thinking is that I can board a 1 BTC input, but then produce smaller 10 million sat VTXOs, so I do a fan out during boarding.

Yes - the server side is ready for this as is. i just mean, right now with this, it only allows fanning out whereas with the server logic, it is also possible for a client to fan _in_. ie, to map a set of boarding inputs into a smaller set of vtxos.

round/states.go
ellemouton: > I'll switch over to that for now.

just checking: i think this wasnt updated yeah?

but yeah - understood that at the start it will be negligible. so non-blocking

round/validation.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Ensure the internal key is unspendable (ARK NUMS key) to force
+	// script path spending only.
+	if !tapscript.ControlBlock.InternalKey.IsEqual(&scripts.ARKNUMSKey) {
+		return fmt.Errorf("internal key is not ARK NUMS key")
+	}
+
+	// For full tree types, verify we have both the collaborative and
+	// timeout script paths by constructing the expected script.
+	if tapscript.Type == waddrmgr.TapscriptTypeFullTree {
```
ellemouton: should we error if not this type?

round/validation.go
```diff
@@ -0,0 +1,115 @@
...
+					"found in actual boarding script", i)
+			}
+		}
+	}
+
+	return nil
+}
+
+// Maximum reasonable delay in blocks (~1 year).
+const MaxReasonableDelay = 52560
```
ellemouton: maybe define at top or in environment even?

round/validation.go
```diff
@@ -0,0 +1,115 @@
+package round
```
ellemouton, line 1: commit message does not match diff. those validation functions are not here

round/transitions.go
ellemouton: cool - just checking if this will be updated here or in follow up?

round/transitions.go
```diff
@@ -0,0 +1,1083 @@
...
+				if err := sig.Encode(&buf); err != nil {
+					return nil, fmt.Errorf("failed to "+
+						"encode partial sig for "+
+						"tx %s: %w", txid, err)
+				}
+				partialSigBytes[txid] = buf.Bytes()
+			}
+
+			submitPartialSigs = append(
+				submitPartialSigs, &SubmitPartialSigRequest{
```
ellemouton, line 757: we can adjust in a follow up but just noting: server side has been updated to expect a client to send all its nonces in a single message (same for sigs). rather than a message per nonce set

round/transitions.go
```diff
@@ -0,0 +1,1113 @@
...
+	env *ClientEnvironment) (*ClientStateTransition, error) {
+
+	switch event.(type) {
+	case *RoundComplete:
+		// Boarding is complete for this round. Transition back to Idle
+		// to process new confirmations for existing boarding addresses
+		// or start new rounds.
+		return &ClientStateTransition{
+			NextState: &Idle{},
+		}, nil
```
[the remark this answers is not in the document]
ellemouton: true story

round/validation_test.go
ellemouton: yeah subtests & add godocs. 

to me, this doesnt do/say much:

```
func TestValidateDelayParametersValid(t *testing.T) {
	t.Parallel()

	err := ValidateDelayParameters(1008, 144)
	require.NoError(t, err)
}
```

ellemouton (approved):
noice!

no blocking comments - i think we can instead iterate once we can properly start hooking things up. 

Biggest question remains: is this flexible enough to easily update to also allow fan _in_? as currently it strictly only allows fanning out and i just want to make sure we dont too tightly couple the direction of "a given vtxo defs is tied to a specific boarding input" and that it can fairly easily be updated to allow fanning in where multiple boarding utxos can basically be consolidated into a single vtxo

Roasbeef:
>  is this flexible enough to easily update to also allow fan in? as currently it strictly only allows fanning out and i just want to make sure we dont too tightly couple the direction of "a given vtxo defs is tied to a specific boarding input" and that it can fairly easily be updated to allow fanning in where multiple boarding utxos can basically be consolidated into a single vtxo

Yeah I don't think we're blocking of any path here permanently. Will make this out into an issue.

round/transitions.go
Roasbeef: Yup I'm planning a PR to first align the client+server flow before we kick off the e2e tests over there.

round/validation.go
Roasbeef: yup, added tests here too

round/validation.go
Roasbeef: done

round/interfaces.go
Roasbeef: Indeed, swapped in

PR #38, opened 2025-12-05
vtxo: start to implement vtxo actor, focusing on forfeit, and expiry behavior 

Roasbeef:
In this PR, we start to implement the VTXO actor as sketched out in the issue linked below. 

The VTXO actor is the output of a succesful round. The round actor sends a message to the vtxo factory actor which then spawns a vtxo actor (at this point the details have already been written to disk). 

The VTXO actor+fsm as the following responsibilities:
  * It watches the expiry of the VTXo is manages:
     * Based on the current height, it decides if needs to attempt to refresh the VTXO or actually go to chain. 
     * If it's a refresh it sends a message to the round actor to attempt to join a round. 
     * Otherwise it sends a message to the chain resolver to start to unravel things on chain. 
  * It receives requests to sign a forfeit txn:
     * This may come in response to a refresh message sent to the round, or be triggered by the round itself as it's carrying out a swap. 
      * It'll then take the connector information (at this point the tree is assumed to be validated), signs the forfeit transaction, then shifts states. 

One could argue that the expiry watching could be more centralized. My initial idea was to have the wallet maintain a pqueue of the expiry heights of each VTXO, then trigger the process of refresh or force close. However, after thinking about it more, I felt like it was better to minimize the responsibilities of the wallet, and further encapsulate the key vtxo logic. 

Opening as a draft now. Mainly showing some early work. 

Partially resolves https://github.com/lightninglabs/darepo/issues/19.

### DIFF
--- a/lib/tree/tree.go
+++ b/lib/tree/tree.go
@@ -6,7 +6,9 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/lib/scripts"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
 )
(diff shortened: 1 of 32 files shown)


Roasbeef:
Closing this in favor of new series.

PR #41, opened 2025-12-09
lib: refactor tree building to do two phase assembly

bhandras:
With this commit we refactor the tree building to a two phase approach, where the first phase is the assembly of the tree structure, and the second phase is the materialization of the tree transactions. This allows for easier extension of the tree building process for assets later on while enabling reuse of the existing tree structure assembly logic.

# Asset/BTC Tree Architecture

This document describes the current `lib/tree` design after the two‑pass
refactor. The tree package now supports both BTC‑only trees and asset‑aware
trees while keeping the signing API consistent.

## Overview

Tree construction is split into two passes:

1) **Structure pass** (`PlanStructure`)
   - Inputs: `[]LeafDescriptor`, radix, weight function, operator key.
   - Outputs: `TreeStructure{Root, AssetContext, LeafScriptMap}`.
   - Populates the tree shape (cosigners, children), sets `Node.Amount` to the
     leaf BTC amount and aggregates subtree totals for branches.
   - Asset data is stored in the AssetContext; BTC leaf scripts are stored in
     the `LeafScriptMap`.

2) **Materialization pass** (`Materialize`)
   - Inputs: the root node, root params (`MaterializeParams`), and a materializer.
   - Outputs: each node is filled with its `Input`, `Outputs`, and `FinalKey`.
   - Implementations:
     - `BTCMaterializer`: builds BTC transactions using `Node.Amount` and the `LeafScriptMap`; ignores asset fields.
     - `AssetMaterializer`: builds asset transactions via tapd and uses the AssetContext (proofs, tweaks, metadata).

Signing remains unchanged: `NewTreeSignerSession` takes a tweak lookup (asse trees use `TweakLookupFromAssetContext`; BTC trees pass `nil`).

## Key Data Structures

- `Node`
  - Core fields: `Input`, `Outputs`, `CoSigners`, `Children`, `Amount` (subtree BTC total), `Signature`, `FinalKey`.
  - Asset‑specific state is not stored on `Node`.

- `TreeStructure`
  - `Root`: root node of the structured tree (no outputs yet).
  - `AssetContext`: asset‑only state (asset amounts, proofs, leaf metadata, tweaks); used by the asset materializer and signing tweaks.
  - `LeafScriptMap`: BTC‑only map of leaf node pointers to pkScripts; used by the BTC materializer.

- `MaterializeParams`
  - Common: `Input`.
  - Asset‑only: `ParentProof`, `ParentPlan` (ignored by BTC).

## Assemblers

- `BTCTreeAssembler`
  - `BuildTree(ctx, rootInput, rootOutput, leaves)`:
    - Runs `PlanStructure` with BTC weighting/radix defaults.
    - Sanity checks that the sum of leaf amounts (`root.Amount`) does not exceed `rootOutput.Value`.
    - Materializes with `BTCMaterializer` using `LeafScriptMap`.
    - Returns `*Tree` (no asset context needed for signing).

- `AssetTreeAssembler`
  - `BuildTree(ctx, rootInput, rootPlan, rootProof, rootOutput, leaves, radix)`:
    - Runs `PlanStructure` with asset weighting.
    - Materializes with `AssetMaterializer`, passing the AssetContext.
    - Returns `*Tree` and the AssetContext; callers must keep them paired for
      signing (`TweakLookupFromAssetContext`) and finalization.

## Value Semantics

- BTC amounts:
  - `LeafDescriptor.Amount` is copied to `Node.Amount` at leaves.
  - Branch `Node.Amount` is the sum of descendant leaf amounts.
  - BTC materialization uses these amounts directly; no top‑down splitting.
  - `BTCTreeAssembler` errors if leaf sums exceed the root output value (root may exceed leaf sums if external fees/change are expected).

- Asset data:
  - Stored only in the AssetContext (proofs, asset amounts, labels, funding, tweaks). BTC code does not touch it.

## Signing

- BTC trees: `NewTreeSignerSession` with `tweakLookup=nil`.
- Asset trees: `TweakLookupFromAssetContext(assetCtx)` supplies per‑node tweaks for MuSig2 sessions.

## Compatibility Notes

- `NewTreeWithConfig` now delegates to `BTCTreeAssembler` (BTC path). It uses `context.Background()` internally; use the assemblers directly if you need an explicit context.
- Legacy `buildTreeBFS`/single‑pass construction is superseded by the two‑pass structure/materialize flow.

### DIFF
--- a/lib/tree/batch.go
+++ b/lib/tree/batch.go
@@ -61,7 +61,16 @@ func NewVTXODescriptor(amount btcutil.Amount, ownerKey *btcec.PublicKey,
 
 // ToLeafDescriptor converts a VTXODescriptor to a generic LeafDescriptor.
 func (v VTXODescriptor) ToLeafDescriptor() LeafDescriptor {
-	return LeafDescriptor(v)
+	pkScript := make([]byte, len(v.PkScript))
+	copy(pkScript, v.PkScript)
+
+	cosignerKey := *v.CoSignerKey
+
+	return LeafDescriptor{
+		PkScript:    pkScript,
+		Amount:      v.Amount,
+		CoSignerKey: &cosignerKey,
+	}
 }
 
 // ConnectorDescriptor defines the specification for a connector tree.
@@ -165,17 +174,12 @@ func (c ConnectorDescriptor) ToLeafDescriptors(
 }
 
 // BuildVTXOTree constructs a complete transaction tree from VTXO descriptors
-// using a BFS (breadth-first search) algorithm. It returns a Tree with all
-// transactions fully constructed and linked.
-func BuildVTXOTree(
-	batchOutpoint wire.OutPoint,
-	batchOutput *wire.TxOut,
-	vtxos []VTXODescriptor,
-	operatorCoSignKey *btcec.PublicKey,
-	sweepKey *btcec.PublicKey,
-	sweepDelay uint32,
-	radix int,
-) (*Tree, error) {
+// using a two-phase approach (structure building + materialization). It returns
+// a Tree with all transactions fully constructed and linked.
+func BuildVTXOTree(batchOutpoint wire.OutPoint, batchOutput *wire.TxOut,
+	vtxos []VTXODescriptor, operatorCoSignKey *btcec.PublicKey,
+	sweepKey *btcec.PublicKey, sweepDelay uint32, radix int) (*Tree,
+	error) {
 
 	// Validate inputs.
 	if err := ValidateVTXODescriptors(vtxos); err != nil {
@@ -223,16 +227,21 @@ func BuildVTXOTree(
 	}
 	sweepTapRoot := sweepTapLeaf.TapHash()
 
-	return NewTree(
-		batchOutpoint, batchOutput, leaves,
-		operatorCoSignKey, sweepTapRoot[:], radix,
-	)
+	// Use BTCTreeAssembler for two-phase tree construction.
+	assembler := NewTreeAssembler(TreeConfig{
+		OperatorKey:        operatorCoSignKey,
+		SweepTapscriptRoot: sweepTapRoot[:],
+		Radix:              radix,
+	})
+
+	return assembler.BuildTree(batchOutpoint, batchOutput, leaves)
 }
 
-// BuildConnectorTree constructs a connector tree from a connector descriptor.
-// Unlike VTXO trees, connector trees have identical leaves (same script, same
-// amount) and are signed only by the operator. This is used for forfeit
-// transaction inputs.
+// BuildConnectorTree constructs a connector tree from a connector descriptor
+// using a two-phase approach (structure building + materialization). Unlike
+// VTXO trees, connector trees have identical leaves (same script, same amount)
+// and are signed only by the operator. This is used for forfeit transaction
+// inputs.
 func BuildConnectorTree(
 	batchOutpoint wire.OutPoint,
 	batchOutput *wire.TxOut,
@@ -257,10 +266,15 @@ func BuildConnectorTree(
 	// Convert to leaf descriptors (creates unique keys for each).
 	leaves := connector.ToLeafDescriptors(operatorKey)
 
-	return NewTree(
-		batchOutpoint, batchOutput, leaves,
-		operatorKey, nil, radix,
-	)
+	// Use the Bitcoin only TreeAssembler for two-phase tree construction.
+	// Connector trees have no sweep script (nil SweepTapscriptRoot).
+	assembler := NewTreeAssembler(TreeConfig{
+		OperatorKey:        operatorKey,
+		SweepTapscriptRoot: nil,
+		Radix:              radix,
+	})
+
+	return assembler.BuildTree(batchOutpoint, batchOutput, leaves)
 }
 
 // BuildBatchOutput computes the pkscript and output amount for a batch output
--- a/lib/tree/btc_tree_assembler.go
+++ b/lib/tree/btc_tree_assembler.go
@@ -1,0 +1,142 @@
+package tree
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/wire"
+)
+
+const (
+	// DefaultRadix is the default branching factor for tree construction.
+	DefaultRadix = 2
+)
+
+// TreeConfig carries the parameters needed to build a BTC-only tree.
+type TreeConfig struct {
+	// OperatorKey is the operator's public key (included in all cosigner
+	// sets).
+	OperatorKey *btcec.PublicKey
+
+	// SweepTapscriptRoot is the tapscript root used for the sweep script.
+	// This provides the unilateral exit path for tree outputs.
+	SweepTapscriptRoot []byte
+
+	// Radix is the maximum number of children per branch node. If zero,
+	// defaults to 2.
+	Radix int
+
+	// WeightFn determines how leaves are weighted for partitioning.
+	// If nil, defaults to WeightByBtcAmount().
+	WeightFn PartitionWeightFunc
+}
+
+// TreeAssembler builds a BTC-only tree using the two-pass approach:
+// structure building (pass 1) followed by materialization (pass 2).
+type TreeAssembler struct {
+	cfg TreeConfig
+}
+
+// NewTreeAssembler constructs an assembler with the given configuration.
+func NewTreeAssembler(cfg TreeConfig) *TreeAssembler {
+	return &TreeAssembler{cfg: cfg}
+}
+
+// BuildTree assembles a BTC-only tree using a two-pass approach:
+//
+// Pass 1 (bottom-up): Build tree structure from leaf descriptors, computing
+// cosigners at each level. No transactions are built yet.
+//
+// Pass 2 (top-down): Materialize transactions starting from the root, filling
+// in Input, Outputs, and FinalKey as we traverse down the tree.
+//
+// Parameters:
+//   - rootInput: The outpoint being spent by the root transaction
+//   - rootOutput: The output being spent (provides BTC value)
+//   - leaves: Leaf descriptors defining the tree leaves
+//
+// Returns the assembled tree.
+func (a *TreeAssembler) BuildTree(rootInput wire.OutPoint,
+	rootOutput *wire.TxOut, leaves []LeafDescriptor) (*Tree, error) {
+
+	if len(leaves) == 0 {
+		return nil, fmt.Errorf("no leaves supplied")
+	}
+	if rootOutput == nil {
+		return nil, fmt.Errorf("root output cannot be nil")
+	}
+	if a.cfg.OperatorKey == nil {
+		return nil, fmt.Errorf("operator key cannot be nil")
+	}
+	if a.cfg.Radix < DefaultRadix && a.cfg.Radix != 0 {
+		return nil, fmt.Errorf("radix must be at least %d, got %d",
+			DefaultRadix, a.cfg.Radix)
+	}
+
+	// Note: SweepTapscriptRoot can be nil for connector trees which have
+	// no sweep script (they're signed only by the operator).
+
+	// Apply defaults.
+	radix := a.cfg.Radix
+	if radix == 0 {
+		radix = DefaultRadix
+	}
+
+	weightFn := a.cfg.WeightFn
+	if weightFn == nil {
+		weightFn = WeightByBtcAmount()
+	}
+
+	// Pass 1: Build tree structure bottom-up. This determines the tree
+	// shape, partitions leaves into groups, and computes cosigners at
+	// each level.
+	structCfg := StructureConfig{
+		OperatorKey: a.cfg.OperatorKey,
+		Radix:       radix,
+		WeightFn:    weightFn,
+	}
+
+	result, err := BuildStructure(leaves, structCfg)
+	if err != nil {
+		return nil, fmt.Errorf("build tree structure: %w", err)
+	}
+	root := result.Root
+
+	// Sanity check: the sum of all leaf BTC amounts (stored in
+	// root.Amount) must be equal to the root output value.
+	rootValue := btcutil.Amount(rootOutput.Value)
+	if root.Amount != rootValue {
+		return nil, fmt.Errorf("leaf amounts sum (%v) must equal the "+
+			"output value (%v)", root.Amount, rootValue)
+	}
+
+	// Pass 2: Materialize transactions top-down. The BTCMaterializer
+	// uses the leaf scripts map from structure building.
+	leafScriptFn := ScriptLookupFromMap(result.LeafScriptMap)
+	mat := NewBTCMaterializer(
+		a.cfg.OperatorKey, a.cfg.SweepTapscriptRoot, leafScriptFn,
+	)
+
+	// BTC values are now carried by each node's Amount field (set during
+	// structure building), so we only need to provide the root input.
+	matParams := MaterializeParams{
+		Input: rootInput,
+	}
+
+	// Bitcoin-only tree materialization doesn't use the context so we can
+	// just pass background.
+	if err := Materialize(
+		context.Background(), root, matParams, mat,
+	); err != nil {
+		return nil, fmt.Errorf("materialize tree: %w", err)
+	}
+
+	return &Tree{
+		Root:               root,
+		BatchOutpoint:      rootInput,
+		BatchOutput:        rootOutput,
+		SweepTapscriptRoot: a.cfg.SweepTapscriptRoot,
+	}, nil
+}
--- a/lib/tree/build_test.go
+++ b/lib/tree/build_test.go
@@ -34,7 +34,7 @@ func TestPartitionLeaves(t *testing.T) {
 
 		// 3 leaves with radix 5 should create 3 groups of 1 leaf each.
 		leaves := createLeaves(3)
-		groups := partitionLeaves(leaves, 5)
+		groups := partitionLeaves(leaves, 5, nil)
 
 		require.Len(t, groups, 3)
 		for i, group := range groups {
@@ -50,7 +50,7 @@ func TestPartitionLeaves(t *testing.T) {
 
 		// 4 leaves with radix 4 should create 4 groups of 1 leaf each.
 		leaves := createLeaves(4)
-		groups := partitionLeaves(leaves, 4)
+		groups := partitionLeaves(leaves, 4, nil)
 
 		require.Len(t, groups, 4)
 		for i, group := range groups {
@@ -67,7 +67,7 @@ func TestPartitionLeaves(t *testing.T) {
 		// 8 leaves with radix 4 should create 4 groups of 2 leaves
 		// each.
 		leaves := createLeaves(8)
-		groups := partitionLeaves(leaves, 4)
+		groups := partitionLeaves(leaves, 4, nil)
 
 		require.Len(t, groups, 4)
 		for i, group := range groups {
@@ -85,7 +85,7 @@ func TestPartitionLeaves(t *testing.T) {
 		// - 2 groups with 3 leaves (10 % 4 = 2 extra).
 		// - 2 groups with 2 leaves (10 / 4 = 2 base).
 		leaves := createLeaves(10)
-		groups := partitionLeaves(leaves, 4)
+		groups := partitionLeaves(leaves, 4, nil)
 
 		require.Len(t, groups, 4)
 
@@ -114,7 +114,7 @@ func TestPartitionLeaves(t *testing.T) {
 		t.Parallel()
 
 		leaves := createLeaves(1)
-		groups := partitionLeaves(leaves, 4)
+		groups := partitionLeaves(leaves, 4, nil)
 
 		require.Len(t, groups, 1)
 		require.Len(t, groups[0], 1)
@@ -124,7 +124,7 @@ func TestPartitionLeaves(t *testing.T) {
 		t.Parallel()
 
 		leaves := createLeaves(2)
-		groups := partitionLeaves(leaves, 4)
+		groups := partitionLeaves(leaves, 4, nil)
 
 		// Should create 2 groups with 1 leaf each.
 		require.Len(t, groups, 2)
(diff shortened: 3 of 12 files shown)


lib/tree/build_test.go
```diff
@@ -660,11 +751,11 @@ func TestBuildTreeBFSEdgeCases(t *testing.T) {
...
 		sweepRoot := make([]byte, 32)
 
-		_, err := buildTreeBFS(
-			input, leaves, operatorKey, sweepRoot, 4,
+		// Nil cosigner keys should cause an error during tree building.
+		_, err := runBuild(
+			t, input, leaves, operatorKey, sweepRoot, 4, nil,
 		)
 		require.Error(t, err)
-		require.Contains(t, err.Error(), "cosigner key cannot be nil")
```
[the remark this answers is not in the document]
bhandras: done

lib/tree/batch.go
```diff
@@ -223,17 +234,23 @@ func BuildVTXOTree(
...
-// Unlike VTXO trees, connector trees have identical leaves (same script, same
-// amount) and are signed only by the operator. This is used for forfeit
-// transaction inputs.
+// BuildConnectorTree constructs a connector tree from a connector descriptor
+// using a two-phase approach (structure building + materialization). Unlike
+// VTXO trees, connector trees have identical leaves (same script, same amount)
+// and are signed only by the operator. This is used for forfeit transaction
+// inputs.
 func BuildConnectorTree(
+	ctx context.Context,
```
ellemouton: i dont think we need a context here. for pure helper functions, we rarely should need to pass in a context

lib/tree/batch.go
```diff
@@ -223,17 +234,23 @@ func BuildVTXOTree(
 	}
 	sweepTapRoot := sweepTapLeaf.TapHash()
 
-	return NewTree(
-		batchOutpoint, batchOutput, leaves,
-		operatorCoSignKey, sweepTapRoot[:], radix,
-	)
+	// Use BTCTreeAssembler for two-phase tree construction.
+	assembler := NewBTCTreeAssembler(BTCTreeConfig{
```
ellemouton: maybe just "TreeAssembler"? think the BTC part isn't relevant right ?

lib/tree/build_test.go
```diff
@@ -614,6 +676,35 @@ func TestBuildTreeBFSEdgeCases(t *testing.T) {
 		}
 	}
 
+	// runBuild is a helper to build a tree using BTCTreeAssembler.
+	runBuild := func(t *testing.T, input wire.OutPoint,
```
ellemouton, line 678: doenst look like any of the callers actually pass in a non-nil weight?

ellemouton: i think same for the other `runBuild`

lib/tree/tree.go
```diff
@@ -40,6 +41,29 @@ func NewTree(rootOutpoint wire.OutPoint, rootOutput *wire.TxOut,
...
+		rootOutpoint, rootOutput, leaves, operatorKey,
+		sweepTapscriptRoot, TreeBuildConfig{
+			Radix: radix,
+		},
+	)
+}
+
+// TreeBuildConfig configures tree construction.
+type TreeBuildConfig struct {
+	Radix int
```
ellemouton, line 54: nit: missing godoc

lib/tree/batch.go
```diff
@@ -61,7 +62,16 @@ func NewVTXODescriptor(amount btcutil.Amount, ownerKey *btcec.PublicKey,
...
+	pkScript := make([]byte, len(v.PkScript))
+	copy(pkScript, v.PkScript)
+
+	cosignerKey := *v.CoSignerKey
+
+	return LeafDescriptor{
+		PkScript:    pkScript,
+		Amount:      v.Amount,
+		CoSignerKey: &cosignerKey,
+	}
```
ellemouton, line 73: ah, could maybe have been in its own commit since it's a bug fix. 

i wonder though, what made this necessary? cause i dont think we ever change the values right?

lib/tree/batch.go
```diff
@@ -223,17 +234,23 @@ func BuildVTXOTree(
...
+
+	return assembler.BuildTree(ctx, batchOutpoint, batchOutput, leaves)
 }
 
-// BuildConnectorTree constructs a connector tree from a connector descriptor.
-// Unlike VTXO trees, connector trees have identical leaves (same script, same
-// amount) and are signed only by the operator. This is used for forfeit
-// transaction inputs.
+// BuildConnectorTree constructs a connector tree from a connector descriptor
+// using a two-phase approach (structure building + materialization). Unlike
```
ellemouton, line 241: ok so back to the old way. 

> this allows for easier extension of the tree building process for assets
later on

maybe worth slightly expanding on this just to help understand what is to come 🙏

lib/tree/batch.go
ellemouton: oh i see... 

how about like BaseTree/DefaultTree maybe? 

but i guess also not a huge deal..

lib/tree/batch.go
```diff
@@ -165,9 +175,10 @@ func (c ConnectorDescriptor) ToLeafDescriptors(
 }
 
 // BuildVTXOTree constructs a complete transaction tree from VTXO descriptors
-// using a BFS (breadth-first search) algorithm. It returns a Tree with all
-// transactions fully constructed and linked.
+// using a two-phase approach (structure building + materialization). It returns
+// a Tree with all transactions fully constructed and linked.
 func BuildVTXOTree(
```
ellemouton: while we are here, maybe we should fix the formatting of the params (my bad!!)

lib/tree/btc_tree_assembler.go
```diff
@@ -0,0 +1,132 @@
...
+	// OperatorKey is the operator's public key (included in all cosigner
+	// sets).
+	OperatorKey *btcec.PublicKey
+
+	// SweepTapscriptRoot is the tapscript root used for the sweep script.
+	// This provides the unilateral exit path for tree outputs.
+	SweepTapscriptRoot []byte
+
+	// Radix is the maximum number of children per branch node. If zero,
+	// defaults to 4.
```
ellemouton: shall we make a constant? and while we are at it, maybe add a minimum constant too? (2)

lib/tree/btc_tree_assembler.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	if a.cfg.OperatorKey == nil {
+		return nil, fmt.Errorf("operator key cannot be nil")
+	}
+
+	// Note: SweepTapscriptRoot can be nil for connector trees which have
+	// no sweep script (they're signed only by the operator).
+
+	// Apply defaults.
+	radix := a.cfg.Radix
```
ellemouton, line 82: maybe sanity check that it isnt below 2

lib/tree/btc_tree_assembler.go
```diff
@@ -0,0 +1,132 @@
...
+	result, err := PlanStructure(leaves, structCfg)
+	if err != nil {
+		return nil, fmt.Errorf("build tree structure: %w", err)
+	}
+	root := result.Root
+
+	// Sanity check: the sum of all leaf BTC amounts (stored in
+	// root.Amount) must not exceed the root output value. The sum may be
+	// less than the root output to allow for fees and anchor outputs.
+	// This catches mis-sized leaves early before we attempt to materialize.
```
ellemouton: is this true? i think they need to be exactly equal no?

lib/tree/tree_structure.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+)
+
+// TreeStructureConfig contains configuration for building tree structure.
+type TreeStructureConfig struct {
```
ellemouton: shall we remove the stuttering from these structs? (ie, remove the `Tree` prefixes?)

lib/tree/tree_structure.go
```diff
@@ -0,0 +1,164 @@
...
+		Children:  make(map[uint32]*Node),
+		Amount:    leaf.Amount,
+		Signature: nil,
+	}
+
+	// Also store the leaf pkscript in the dedicated map for BTC path.
+	// This keeps BTC and asset concerns separate.
+	if leafScripts != nil && len(leaf.PkScript) > 0 {
+		leafScripts[node] = leaf.PkScript
+	}
```
ellemouton, line 162: can you explain this a bit? not sure i follow why we need this map

lib/tree/tree_structure.go
```diff
@@ -0,0 +1,164 @@
...
+	}, nil
+}
+
+// buildStructureRecursive recursively builds the tree structure. For a single
+// leaf, it creates a leaf node. For multiple leaves, it partitions them and
+// creates a branch node with children built recursively.
+func buildStructureRecursive(leaves []LeafDescriptor,
+	operatorKey *btcec.PublicKey, radix int,
+	weightFn PartitionWeightFunc,
+	leafScripts map[*Node][]byte) (*Node, error) {
```
ellemouton, line 86: maybe instead of passing in a mutable map, how about instead having a `structureBuilder` struct or something that holds the state and then mutates its state?

lib/tree/tree_structure.go
```diff
@@ -0,0 +1,164 @@
...
+// The returned TreeStructure contains:
+//   - Root: The root node with CoSigners and Children populated
+//   - LeafScriptMap: Map of leaf nodes to output scripts (for BTC trees only)
+//
+// The following Node fields are NOT set (filled in by Materialize):
+//   - Input (zero outpoint)
+//   - Outputs (nil)
+//   - FinalKey
+//   - Signature
+func PlanStructure(leaves []LeafDescriptor,
```
ellemouton: BuildStructure?

lib/tree/tree_structure.go
ellemouton: related question is: why cant we get this leaf script info from the node itself?

lib/tree/materializer_btc.go
```diff
@@ -0,0 +1,175 @@
...
+	if len(leafPkScript) == 0 {
+		return nil, fmt.Errorf("leaf node missing pkscript")
+	}
+
+	// Build outputs: [leaf output, anchor]. Use node.Amount which was set
+	// during structure building from the leaf descriptor.
+	node.Outputs = []*wire.TxOut{
+		wire.NewTxOut(int64(node.Amount), leafPkScript),
+		scripts.AnchorOutput(),
+	}
```
ellemouton, line 109: cant this part just be done in the planning stage? then we dont need the whole LeafScript map thing

lib/tree/materializer_btc.go
```diff
@@ -0,0 +1,175 @@
...
+	if node.IsLeaf() {
+		return m.materializeLeaf(node, params)
+	}
+
+	return m.materializeBranch(node, params)
+}
+
+// materializeLeaf builds outputs for a leaf node.
+func (m *BTCMaterializer) materializeLeaf(node *Node,
+	_ MaterializeParams) (map[uint32]MaterializeParams, error) {
```
ellemouton: remove unused param?

ellemouton: same below

ellemouton (commented):
nice! got some questions

lib/tree/btc_tree_assembler.go
```diff
@@ -0,0 +1,132 @@
...
+// in Input, Outputs, and FinalKey as we traverse down the tree.
+//
+// Parameters:
+//   - rootInput: The outpoint being spent by the root transaction
+//   - rootOutput: The output being spent (provides BTC value)
+//   - leaves: Leaf descriptors defining the tree leaves
+//
+// Returns the assembled tree.
+func (a *BTCTreeAssembler) BuildTree(ctx context.Context,
+	rootInput wire.OutPoint, rootOutput *wire.TxOut,
```
sputn1ck: How can we have a rootOutput here? doesn't the tree need to be build before we can have an output?

lib/tree/batch.go
bhandras: Removed these contexts. We only really need them for the asset case.

lib/tree/batch.go
bhandras: Yeah naming could be better I agree. Changed it up a bit so we don't have "BTC" explicitly everywhere:

- `NewTreeAssembler` returning `TreeAssembler`
- `TreeConfig`

lib/tree/build_test.go
bhandras: Good catch, fixed.

lib/tree/tree.go
bhandras: Added.

lib/tree/batch.go
bhandras: Done :)

lib/tree/btc_tree_assembler.go
bhandras: Done.

lib/tree/btc_tree_assembler.go
bhandras: Done.

lib/tree/tree_structure.go
bhandras: Done.

lib/tree/tree_structure.go
bhandras: Done.

lib/tree/materializer_btc.go
bhandras: Removed.

lib/tree/btc_tree_assembler.go
bhandras: Yes you're right ofc. Sorry, this is where the agent overgeneralized from the asset case to the bitcoin case and i failed to catch it. Fixed.

bhandras: Fixed up the tests too a bit.

lib/tree/materializer_btc.go
bhandras: Unfortunately this is where the btc/asset trees differ so i wanted to keep it separate (hence the map and the script lookup fn added).

bhandras: Right now the options are all a bit awkward (imo):

- Stick the pkscript on the `Node` (nil for branches). Easy to grab later, but it means the shared type carries btc only context forever (doesn't really fit with asset/btc separation).
- We can attach the whole `LeafDescriptor` to leaf nodes. Feels similarly ugly, plus it exposes more leaf-only details at the type level.
- We can split the types (`LeafNode`/`BranchNode` or say a `LeafData` interface). Much cleaner on paper, but it’s a bigger refactor: lots of call sites suddenly need assertions/generics, tests shift around, and the two-pass flow gets more complex (imo not worth it).
- We can use this external map keyed by a stable "id" (the pointer makes sense imo, but agree it seems ugly at first). This keeps `Node` asset-agnostic, leaves branches alone, and lets us carry btc only leaf data from planning to materialization without splitting types.

Imo the map is the least painful option, but curious if you think we can have something nicer. See also https://github.com/lightninglabs/darepo-client/pull/39/commits/b8fdecf27f0c7109551910094115d072899c110c for more context on how the asset materializer is implemented.

lib/tree/btc_tree_assembler.go
bhandras: The root output is already known when the tree is constructed (eg. in the proposed boarding sweep), even if it hasn’t been broadcast yet. We materialize the tree by walking downward from the root.

lib/tree/tree_structure.go
bhandras: Added a comment here in the code as well as see my above review comment for more details.

lib/tree/tree_structure.go
bhandras: > maybe instead of passing in a mutable map, how about instead having a structureBuilder struct or something that holds the state and then mutates its state?

The only shared state right now is this map (see other comment for more details on why we need it), so having a struct with the map field is maybe just an indirection we don't need. Happy to add it if you think it'd still be useful (iiuc you meant to hide map mutation behind a dedicated setter right?)

> related question is: why cant we get this leaf script info from the node itself?

We can ofc, see my other comment why i think it's a bit awkward: https://github.com/lightninglabs/darepo-client/pull/41#discussion_r2606860623

lib/tree/batch.go
bhandras: Technically this is a copy, but you're right we don't change these values. Happy to just remove it (was an auto correction by the agent).

lib/tree/batch.go
bhandras: If you think it's useful i'm happy to add the design doc (see in the description) to this PR too. Alternatively we can add a very detailed comment somewhere.

sputn1ck (approved):

lib/tree/btc_tree_assembler.go
bhandras: Okay chatted a bit more about this, we could make the tree building phases more explicitly used within the round state machine and plan the root output from the leaves as the result of `PlanStructure` instead receiving it and making sure that the structure conforms when materializing. Perhaps we can do this as a next step.

lib/tree/batch.go
```diff
@@ -165,17 +174,12 @@ func (c ConnectorDescriptor) ToLeafDescriptors(
...
-	sweepKey *btcec.PublicKey,
-	sweepDelay uint32,
-	radix int,
-) (*Tree, error) {
+// using a two-phase approach (structure building + materialization). It returns
+// a Tree with all transactions fully constructed and linked.
+func BuildVTXOTree(batchOutpoint wire.OutPoint, batchOutput *wire.TxOut,
+	vtxos []VTXODescriptor, operatorCoSignKey *btcec.PublicKey,
+	sweepKey *btcec.PublicKey, sweepDelay uint32, radix int) (
+	*Tree, error) {
```
ellemouton: nit: Tree can be on previous line

lib/tree/batch.go
ellemouton: maybe im missing exactly where this is in the description but would just like a bit more clarity on exactly what makes things build-able via 1shot with btc-only vs 2-shot with assets. I think maybe a bit more of a peek into what is in the assets context? 

also, realised that currently this is sort of already (before this pr) a 2 shot thing cause: 1) compute batch output pkscript (which requires leaf knowledge) and then 2) inject that into the tree to materialise each tx. 
So im wondering if we shouldnt somehow merge that 1st shot with the assets 1st shot? ie, would the assets stuff not also affect the batch output passed in here too? ie BuildBatchOutput?

ellemouton: so yeah, maybe a doc to visualise would be helpful 🙏

lib/tree/btc_tree_assembler.go
```diff
@@ -0,0 +1,142 @@
...
+	if len(leaves) == 0 {
+		return nil, fmt.Errorf("no leaves supplied")
+	}
+	if rootOutput == nil {
+		return nil, fmt.Errorf("root output cannot be nil")
+	}
+	if a.cfg.OperatorKey == nil {
+		return nil, fmt.Errorf("operator key cannot be nil")
+	}
+	if a.cfg.Radix < DefaultRadix && a.cfg.Radix != 0 {
```
ellemouton, line 73: nit: technically default separate from minimum so can be separate constants i think even though they happen to be same value. Non blocking though

lib/tree/tree_structure.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Radix is the maximum number of children per branch node.
+	Radix int
+
+	// WeightFn determines how leaves are weighted for partitioning.
+	// If nil, defaults to WeightByBtcAmount().
+	WeightFn PartitionWeightFunc
+}
+
+// Structure contains the outputs from PlanStructure.
```
ellemouton: s/PlanStructure/BuildStructure

lib/tree/tree_structure.go
```diff
@@ -0,0 +1,165 @@
...
+type Structure struct {
+	// Root is the root node of the tree structure.
+	Root *Node
+
+	// LeafScriptMap maps leaf node pointers to their output scripts
+	// (pkscript). This is populated during structure building and used by
+	// the BTC materializer. We keep BTC leaf data out of Node to keep the
+	// shared structure asset-agnostic and avoid touching branch nodes.
```
ellemouton, line 32: tbh i dont fully understand the sentence `We keep BTC leaf data out of Node to keep the shared structure asset-agnostic and avoid touching branch nodes.`  what will assets add that changes things here? sorry if im missing obvious things here. like, wouldnt asset additions either change the pk script or add _other outputs_ to the leaf? i think a visualisation of how assets affect the tree would really help

lib/tree/materializer_btc.go
ellemouton: i think a visualisation of what an asset-compatible tree looks like would really help. Really want to avoid making the tree logic more complex than needed. Realise it is my lack of knowledge in the area 🙈  sorry

ellemouton (approved):
@bhandras , do you think you can maybe get the AI to produce a visual of the tree with assets? 

Would really help me understand & then i can more easily help review the asset materialiser too.

lib/tree/batch.go
bhandras: Chatted about this a bit more offline, but ofc will try to add a doc in a follow up, potentially when the asset tree lands.

lib/tree/materializer_btc.go
bhandras: Will try, but perhaps best to do this along the next PR which adds the asset tree.

lib/tree/btc_tree_assembler.go
bhandras: Lets change this default once we know if we want radix 2 or 4 (or other value).

lib/tree/tree_structure.go
bhandras: So assets add other metadata such as proof, asset amount, etc. which we don't want to add to the structure or `Node` either to keep things separate. Ptal at this more detailed explanation on the choice of the map for the btc case: https://github.com/lightninglabs/darepo-client/pull/41#discussion_r2606860623

lib/tree/tree_structure.go
bhandras: done

lib/tree/batch.go
bhandras: done

PR #43, opened 2025-12-10
lib/tree: Musig2 signing updates

ellemouton:
Here, we update the tree signer to make use of the new methods on the
LND Musig2Signer API. The new method lets us pass the aggregated nonce
of a session in rather than requiring that each individual nonce gets
passed in.


sputn1ck (approved):
LGTM! Thanks

bhandras (approved):

PR #47, opened 2025-12-12
systest: add new systest framework build on top of existing harness, add wallet boarding tests

Roasbeef:
In this PR, we add a new `systest` framework built on top of the existing harness. It wraps the harness with an active actor system, and some other useful utilities. 

Along the way we fix some subtle bugs related to cotnexts (deciding when to make a new one vs pass on the parent context). 

We also revamp the way logging is set up for the actors, going back to the old style of relying primarily on a logger passed in as config. We then supplement that as needed with a logger passed in via a context. 

Finally we update CI to run the new systests using both postgres and sqlite.

### DIFF
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -10,9 +10,18 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/darepo-client/chainsource"
 	"github.com/lightningnetwork/lnd/chainntnfs"
-	"github.com/lightningnetwork/lnd/lnwallet"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 )
+
+// TxBroadcaster is a minimal interface for broadcasting transactions. This
+// allows LNDBackend to work with both lnwallet.WalletController (in-process
+// lnd) and lndclient wrappers (remote lnd via gRPC).
+type TxBroadcaster interface {
+	// PublishTransaction broadcasts the given transaction to the network.
+	// The label parameter is optional and may be used for wallet tracking.
+	PublishTransaction(ctx context.Context, tx *wire.MsgTx,
+		label string) error
+}
 
 // LNDBackend implements the chainsource.ChainBackend interface by wrapping
 // lnd's chain notification and fee estimation interfaces. This backend provides
@@ -28,20 +37,20 @@ type LNDBackend struct {
 	// feeEstimator provides fee estimation services from lnd.
 	feeEstimator chainfee.Estimator
 
-	// wallet provides transaction broadcasting capabilities.
-	wallet lnwallet.WalletController
+	// broadcaster provides transaction broadcasting capabilities.
+	broadcaster TxBroadcaster
 }
 
 // NewLNDBackend creates a new LNDBackend instance with the given lnd
 // components. All parameters must be non-nil.
 func NewLNDBackend(notifier chainntnfs.ChainNotifier,
 	feeEstimator chainfee.Estimator,
-	wallet lnwallet.WalletController) *LNDBackend {
+	broadcaster TxBroadcaster) *LNDBackend {
 
 	return &LNDBackend{
 		notifier:     notifier,
 		feeEstimator: feeEstimator,
-		wallet:       wallet,
+		broadcaster:  broadcaster,
 	}
 }
 
@@ -120,12 +129,12 @@ func (b *LNDBackend) TestMempoolAccept(ctx context.Context,
 		"LND backend")
 }
 
-// BroadcastTx broadcasts a transaction to the network using lnd's wallet
-// controller.
+// BroadcastTx broadcasts a transaction to the network using the configured
+// broadcaster.
 func (b *LNDBackend) BroadcastTx(ctx context.Context, tx *wire.MsgTx,
 	label string) error {
 
-	err := b.wallet.PublishTransaction(tx, label)
+	err := b.broadcaster.PublishTransaction(ctx, tx, label)
 	if err != nil {
 		return fmt.Errorf("failed to broadcast transaction: %w", err)
 	}
--- a/chainbackends/lndclient_adapters.go
+++ b/chainbackends/lndclient_adapters.go
@@ -1,0 +1,366 @@
+package chainbackends
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btclog/v2"
+	"github.com/lightninglabs/darepo-client/build"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightningnetwork/lnd/chainntnfs"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// LndClientTxBroadcaster implements TxBroadcaster using
+// lndclient.WalletKitClient.
+type LndClientTxBroadcaster struct {
+	walletKit lndclient.WalletKitClient
+}
+
+// NewLndClientTxBroadcaster creates a new broadcaster backed by lndclient.
+func NewLndClientTxBroadcaster(
+	walletKit lndclient.WalletKitClient) *LndClientTxBroadcaster {
+
+	return &LndClientTxBroadcaster{
+		walletKit: walletKit,
+	}
+}
+
+// PublishTransaction broadcasts the given transaction to the network via lnd.
+func (b *LndClientTxBroadcaster) PublishTransaction(ctx context.Context,
+	tx *wire.MsgTx, label string) error {
+
+	return b.walletKit.PublishTransaction(ctx, tx, label)
+}
+
+// Ensure LndClientTxBroadcaster implements TxBroadcaster.
+var _ TxBroadcaster = (*LndClientTxBroadcaster)(nil)
+
+// LndClientFeeEstimator implements chainfee.Estimator using lndclient.
+type LndClientFeeEstimator struct {
+	walletKit lndclient.WalletKitClient
+}
+
+// NewLndClientFeeEstimator creates a new fee estimator backed by lndclient.
+func NewLndClientFeeEstimator(
+	walletKit lndclient.WalletKitClient) *LndClientFeeEstimator {
+
+	return &LndClientFeeEstimator{
+		walletKit: walletKit,
+	}
+}
+
+// Start is a no-op for lndclient-backed estimators since the connection is
+// already established.
+func (e *LndClientFeeEstimator) Start() error {
+	return nil
+}
+
+// Stop is a no-op for lndclient-backed estimators.
+func (e *LndClientFeeEstimator) Stop() error {
+	return nil
+}
+
+// EstimateFeePerKW returns the estimated fee rate in satoshis per kilo-weight
+// unit for the given confirmation target.
+func (e *LndClientFeeEstimator) EstimateFeePerKW(
+	numBlocks uint32) (chainfee.SatPerKWeight, error) {
+
+	satPerVByte, err := e.walletKit.EstimateFeeRate(
+		context.Background(), int32(numBlocks),
+	)
+	if err != nil {
+		return 0, fmt.Errorf("estimate fee rate: %w", err)
+	}
+
+	satPerKw := chainfee.SatPerVByte(satPerVByte).FeePerKWeight()
+
+	return satPerKw, nil
+}
+
+// RelayFeePerKW returns the minimum fee rate required for transactions to be
+// relayed. For remote lnd, we return a reasonable default.
+func (e *LndClientFeeEstimator) RelayFeePerKW() chainfee.SatPerKWeight {
+	// Default relay fee of 1 sat/vbyte = 250 sat/kw.
+	return chainfee.SatPerKWeight(250)
+}
+
+// Ensure LndClientFeeEstimator implements chainfee.Estimator.
+var _ chainfee.Estimator = (*LndClientFeeEstimator)(nil)
+
+// LndClientChainNotifierConfig holds configuration for LndClientChainNotifier.
+type LndClientChainNotifierConfig struct {
+	// LND is the lndclient services connection.
+	LND *lndclient.LndServices
+
+	// Log is an optional logger for this notifier. If None, the notifier
+	// falls back to extracting a logger from context via
+	// LoggerFromContext, or uses btclog.Disabled if no logger is found.
+	Log fn.Option[btclog.Logger]
+}
+
+// WithLogger returns a new config with the given logger set.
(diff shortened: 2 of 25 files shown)


Roasbeef:
@claude review this

chainbackends/lnd.go
```diff
@@ -10,10 +10,20 @@ import (
...
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 )
 
+// TxBroadcaster is a minimal interface for broadcasting transactions. This
+// allows LNDBackend to work with both lnwallet.WalletController (in-process
+// lnd) and lndclient wrappers (remote lnd via gRPC).
+type TxBroadcaster interface {
+	// PublishTransaction broadcasts the given transaction to the network.
+	// The label parameter is optional and may be used for wallet tracking.
+	PublishTransaction(
```
bhandras: nit: formatting

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+	walletKit lndclient.WalletKitClient) *LndClientTxBroadcaster {
+
+	return &LndClientTxBroadcaster{
+		walletKit: walletKit,
+	}
+}
+
+// PublishTransaction broadcasts the given transaction to the network via lnd.
+func (b *LndClientTxBroadcaster) PublishTransaction(
+	ctx context.Context, tx *wire.MsgTx, label string,
```
bhandras: nit: formatting.

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+	return nil
+}
+
+// EstimateFeePerKW returns the estimated fee rate in satoshis per kilo-weight
+// unit for the given confirmation target.
+func (e *LndClientFeeEstimator) EstimateFeePerKW(
+	numBlocks uint32) (chainfee.SatPerKWeight, error) {
+
+	satPerVByte, err := e.walletKit.EstimateFeeRate(
+		context.Background(), int32(numBlocks),
```
bhandras, line 74: Shall we extend the interface to take a context so we can pass that in?

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+	for _, opt := range opts {
+		opt(notifierOpts)
+	}
+
+	var lndOpts []lndclient.NotifierOption
+	if notifierOpts.IncludeBlock {
+		lndOpts = append(lndOpts, lndclient.WithIncludeBlock())
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
bhandras, line 173: Same comment re: context use, we should really not use embedded background contexts.

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+
+		select {
+		case err, ok := <-errChan:
+			if ok && err != nil {
+				// Log error but don't propagate - the caller
+				// should handle timeout via context.
```
bhandras: Comment doesn't really make sense as the context isn't surfaced to the caller.

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+	}, nil
+}
+
+// RegisterSpendNtfn registers for spend notifications using lndclient's
+// ChainNotifier.
+func (n *LndClientChainNotifier) RegisterSpendNtfn(
+	outpoint *wire.OutPoint, pkScript []byte,
+	heightHint uint32) (*chainntnfs.SpendEvent, error) {
+
+	ctx, cancel := context.WithCancel(context.Background())
```
bhandras, line 213: Just flagging re: context use.

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+		Confirmed: confChan,
+		Cancel:    cancel,
+	}, nil
+}
+
+// RegisterSpendNtfn registers for spend notifications using lndclient's
+// ChainNotifier.
+func (n *LndClientChainNotifier) RegisterSpendNtfn(
+	outpoint *wire.OutPoint, pkScript []byte,
+	heightHint uint32) (*chainntnfs.SpendEvent, error) {
```
bhandras, line 211: Why not return an error chan (applies to all these notifiers).

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+		Cancel: cancel,
+	}, nil
+}
+
+// RegisterBlockEpochNtfn registers for block epoch notifications using
+// lndclient's ChainNotifier.
+func (n *LndClientChainNotifier) RegisterBlockEpochNtfn(
+	bestBlock *chainntnfs.BlockEpoch) (*chainntnfs.BlockEpochEvent, error) {
+
+	ctx, cancel := context.WithCancel(context.Background())
```
bhandras, line 249: Same same.

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,369 @@
...
+				//
+				// TODO(rosabeef): why doesn't lndclient also
+				// give the block height?
+				chainKit := n.cfg.LND.ChainKit
+				blockHash, err := chainKit.GetBlockHash(
+					ctx, int64(height),
+				)
+				if err != nil {
+					log.WarnS(ctx, "Failed to get block hash",
+						err, slog.Int("height", int(height)))
```
bhandras, line 290: So we'd silently fail here skipping the notification.

chainsource/chainsource.go
```diff
@@ -17,6 +19,26 @@ const (
...
+	// Log is an optional logger for this actor instance. If None, the actor
+	// falls back to extracting a logger from context via LoggerFromContext,
+	// or uses btclog.Disabled if no logger is found.
+	Log fn.Option[btclog.Logger]
+}
+
+// WithLogger returns a new config with the given logger set.
+func (c ChainSourceConfig) WithLogger(log btclog.Logger) ChainSourceConfig {
+	c.Log = fn.Some(log)
+	return c
```
bhandras, line 40: nit: add nl before return

systest/main_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+	"sync"
+	"testing"
+
+	"github.com/lightninglabs/darepo-client/harness"
+)
+
+var (
+	// sharedHarness is the shared Docker harness for all systests.
+	sharedHarness *harness.Harness
```
bhandras: Why use a shared harness btw? It's pretty cheap to get a harness running and the cost goes even lower if we go parallel (see ParallelN helper)

systest/main_test.go
```diff
@@ -0,0 +1,59 @@
...
+	// harnessStarted tracks whether the harness was ever started.
+	harnessStarted bool
+)
+
+// TestMain is the entry point for the systest suite. It manages the shared
+// harness lifecycle - the harness is started lazily on the first test that
+// needs it, and stopped after all tests complete.
+func TestMain(m *testing.M) {
+	// Run all tests.
+	code := m.Run()
```
bhandras: The testmain setup makes it very hard to debug individual tests.

bhandras (commented):
My main comments are around context and error handling.

chainbackends/lndclient_adapters.go
Roasbeef: Makes sense, will extend.

Roasbeef: Actually the issue is that we're using `chainfee.Estimator` and `chainntfsn.ChainNotifier` directly. These lnd-level interfaces don't take a context. 

So we either keep with the current version, or we make new interfaces that _do_ pass in contexts.

chainbackends/lndclient_adapters.go
Roasbeef: This is intended to implement `chaintnfs.ChainNotifier` as is. Its meant to be light glue around the `lndclient` version of the respective methods. 

Looking at the `lndclient` source, an error is sent in 4 cases:
  1. Issue reading off the stream. 
  2. Spend event has an unknown types 
  3. Spend event is nil (should only happen if there's an err on recv, so something w/ the stream). 
  4. Failure when processing 

Agree we shouldn't just swallow the errors, but stepping back re the current architecture: it isn't clear how some of these errors would make it back to the client in the first place. 

Consider that this is used mainly via the actor message passing. Right now we only have messages for request and response, and not some of the error cases. So in order to implement this properly we need to extend the various chain actors to also handle these error state messages, which makes a lot of sense. 

So with that said, I agree we shouldn't swallow these errors, and make sure contexts are passed down, etc (though care needs to be taken there as well, as sometimes a child context _should_ be independent, I fixed one such bug in this PR). 

I propose we make an issue and do this in another PR, as it'll require a larger set of changes, as all the callers need to decide what to do and how to handle these errors.

chainbackends/lndclient_adapters.go
Roasbeef: Do you know why historically `lndclient` doesn't just send the header _and_ hash? If it did, then we wouldn't need to do this extra call to fetch the block hash, and API consumption would be more atomic. 

I'm inclined to fix it there instead, though this would be a breaking change in the API type-wise. 

Otherwise our options to handle the failure at this level are something like:
  1. Retry with back off to fetch the block hash. If it fails, this may mean that a re-org occurred, etc. 
  2. Register for a fresh block epoch notification, but this time specify the last block was the starting height to replay forward.

systest/main_test.go
Roasbeef: No fundamental reason, just was focusing on a simple case of serial execution. 

Happy to make this change if we think we want the parallelism from the get go.

Roasbeef: Based on your comment below, decided to just do this now.

systest/main_test.go
Roasbeef: Hmm, and why is that the case exactly? Stepping through the extra frames of `m.Run()` I think you can set breakpoints, etc as normal? `TestMain` is also used in `harness/harness_test.go`, that's where I actually learned of the pattern. 

IIUC, `TestMain` is needed to populate the channel semaphore.

systest/main_test.go
Roasbeef: I think there might be some lingering issues related to making sure non-conflicting ports are used. After updating to  run in parallel, I see this occasionally: 
```
                Error:          Received unexpected error:
                                API error (500): driver failed programming external connectivity on endpoint bitcoin-TestBoardingWalletAddressReuse (047c415d82c9d48403502b043c6a9a9c1dc8ed07e12e81b09c862fa33d23308d): listen tcp4 127.0.0.1:33590: bind: address already in use
                Test:           TestBoardingWalletAddressReuse
                Messages:       failed to start bitcoind
``` 

So I think before we can do this here, we may need to update the harness to handle colliding ports for parallel tests, similar to `lntest` for lnd. 

I don't think this is blocking for this PR though, as the tests work reliably without the parallel execution. I propose we punt this to an issue to update `harness` to handle port allocation more carefully.

bhandras: Interesting. Original design was supposed to take care of this already. See eg.: https://github.com/lightninglabs/darepo-client/blob/a49d740fcc52a83c6f9c77f09b0ba5aae977d237/harness/harness.go#L980 

Seems like Docker's solution isn't perfect either. Added some hardening here: https://github.com/lightninglabs/darepo-client/pull/49

chainbackends/lndclient_adapters.go
bhandras: I think it’s likely that these are simply features the older LND design never implemented because there wasn’t a perceived need at the time. I agree that the APIs should optionally return block metadata as well, and this could potentially be addressed on the LND side.

For Ark clients that use LND, the current behavior is sufficient. That said, I do see the value in having `chainkit` return this information directly. The main unresolved concern remains clean and consistent error propagation.

chainbackends/lndclient_adapters.go
bhandras: More broadly speaking, the question is whether an actor should retry on error, or whether the receiver should handle some subset of errors. In this specific case, I think it’s reasonable to assume that any error is non-transient and should trigger a system-wide failure that is surfaced to the user in some form.

For example, if LND misbehaves, what would a mobile user actually see if the issue is only logged as a warning? If the error is properly surfaced, at least they would know that something is wrong.

I also think errors can be part of the response itself rather than being handled separately. I don’t see a strong need to strictly separate the two.

>I propose we make an issue and do this in another PR, as it'll require a larger set of changes, as all the callers need to decide what to do and how to handle these errors.

Please add TODOs to the code so we don't forget.

systest/main_test.go
bhandras: Sorry, I didn’t explain this clearly earlier. I wasn’t referring to debugging with a debugger, that’s obviously possible.

The real issue is that using a single harness instance means all parallel tests share the same infrastructure, which causes test artifact logs to get mixed together. This makes it much harder to identify flaky tests from CI run artifacts.

More generally, global objects tend to cause problems over time unless they are explicitly used for synchronization within the process.

chainbackends/lndclient_adapters.go
bhandras: Fair point.  Let's keep it for now.

systest/main_test.go
Roasbeef: Cool, will rebase on top of that to see if it resolves the issue.

systest/main_test.go
Roasbeef: Ah sure, makes sense, I'll re-apply the `ParallelN` helper func on top of your latest PR.

systest/main_test.go
Roasbeef: So I tried this, and it didn't work out of the box. I had an issue where some of the created resources to virtualize the network weren't properly cleaned up, which caused the tests to fail. 

An example failure: 
```
     ark-harness-yckeojvf (id=44a60bafc41ce045db9db1f8119508f77ee5f83916b1c12abf9383c83fae34bb)
     2025-12-17 17:21:42.604 [harness/harness.go:480] Docker network verified
     2025-12-17 17:21:42.604 [harness/harness.go:512] Starting bitcoind...
     2025-12-17 17:21:42.606 [harness/harness.go:812] [DEBUG] Failed to kill container
     16e8ff72faa7: Container not running:
```

It doesn't really fail, so much as it hangs actually ~forever (until the actual test timeout).

Roasbeef: I was able to remove the stale network info with `docker network prune -f`.

Roasbeef: I restarted Docker Desktop, and now it seems to work now! Updated the branch to run in parallel and now it branches off your docker fix.

bhandras: > I was able to remove the stale network info with `docker network prune -f`. 

Yeah for some reason docker networks can get stuck on mac and despite our best efforts in the harness they aren't fully pruned, so locally we have to prune unused networks from time to time (there's also a make target for this). iiuc this only on mac though.

PR #48, opened 2025-12-17
MailboxRPC 1/7: durability foundation

Roasbeef:
In this PR, we implement a fault-tolerant durable actor layer. This is intended to slot right into the existing actors that we already use today. It handles a lot of book keeping under the hood to make sure that things are done safe.

This is quite a large PR (even granted it's mostly tests, the auto gen sql code, and the docs), so I'll split it up, but I wanted to put it all up at once so it was easy to navigate, etc. 

## TLV Codec Layer

In order to use this new framework, all messages that need to be written persistently will implement a new TLV-based interface and register that with the codec. The codex is used to encode/decode messages to/from the persistent outbox that all actors now maintain. 

This may also cause a shift in the way we make the struct hierarchy of messages. Right now each package has their own message type/interface. We may end up actually merging this into a single package, then use type aliases in each actor package so it's easy to find which messages a given actor actually cares out. 

See t[his getting started guide re the updates](https://github.com/lightninglabs/darepo-client/blob/072b777a3370898a4e40fee54cf772c74424ac13/docs/durable_actor_quickstart.md) needed to existing actors to start using this new framework, and items that devs need to keep in mind. 

## `DurableActor` Transactional Message Processing 

A `DurableActor`'s message processing will take place in a single database transaction, so we'll want to think carefully about any network access to blocking that takes place. Expansive changes in existing actors (that need to be durable) shouldn't be required. 

The underlying actor framework does a lot of the heavy lifting here. It handles reading from the persistent mailbox, ACK'ing messages after processing has finished, suppressing duplicate messages, handling retransmission, etc, etc. 

The main call site that'll change slightly is when an actor is processing the outbox messages from the FSM. As mentioned above, as everything is run in a single db transaction, if the actor crashes at any point, post restart it'll consume the last un ACK'd message, resuming as normal. 

There's also a special `Restart` message sent by the runtime when an actor is restarted. The actor should process this, then do any sort of init activities such as requesting for confirmations, etc. 

Actors can also optionally checkpoint some state. We may want to use this to checkpoint actual attributes in the actor, leaving the FSM state to dedicated DB methods (as the check pointed state is just TLV atm). 

Along the way, `Tell` now returns an error, as it's possible we fail to write to the db for w/e reason. 

# `OutboxPublisher`

The `OutboxPublisher` is a key component of the entire architecture. There's a single global instance of this atm, but it's designed s.t we can have many instances of them. It handles taking a message from the outbox of an actor (in reality there's actually a global outbox) and writing it to the mailbox/inbox of the target actor. It'll continue to delver on start up until a message is fully ACK'd by the destination actor. This is also used to implement durable asks as mentioned below. 

# `DurableAsks`

The final gap is to make sure that asks are durable. Today, the outbox abstraction is basically a `Tell`. It's set and forget, and doesn't allow for responses w/o a custom flow. 

Durable acks fill this gap. When you need to reliably hear back about the response to a message (eg: asking an actor actor for a signature of something, or transaction publish, etc) you should reach for this. It's like a normal ask, but you also provide the response ID of the actor. A durable ask doesn't return a future like a normal ask. Instead, the response comes back across as a normal message for the actor. 

This reuses the `OutboxPublisher` layer to handle delivery until the response is ack'd, etc, etc. 

# `actortest` package 

A new `internal/actortest` package has been added to test all the relevant scenarios and invariants with a real database. It can also double as some example code re using all the functionality mentioned above. 

Fixes https://github.com/lightninglabs/darepo/issues/4


db/sqlc/migrations/000004_durable_mailbox.down.sql
```diff
@@ -0,0 +1,29 @@
+-- Rollback durable mailbox migration.
```
sputn1ck, line 1: couldn't these migrations live in the actor package itself? golang-migrate supports reading from multiple filesystems and versions via the migrations table config.

db/sqlc/migrations/000004_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+    -- payload contains the TLV-encoded message data.
+    payload BLOB NOT NULL,
+
+    -- promise_id is set for Ask messages to track the response.
+    -- NULL for Tell (fire-and-forget) messages.
+    promise_id TEXT,
+
+    -- callback_actor_id is set for DurableAsk messages to route the response.
+    -- The response will be delivered to this actor's mailbox via outbox.
+    -- NULL for regular Ask/Tell messages.
```
sputn1ck, line 32: `NULL for regular Ask/Tell messages`. wouldn't these not be persisted though?

db/sqlc/migrations/000004_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+
+    -- message_type is the type name for deserialization dispatch.
+    message_type TEXT NOT NULL,
+
+    -- payload contains the TLV-encoded message data.
+    payload BLOB NOT NULL,
+
+    -- promise_id is set for Ask messages to track the response.
+    -- NULL for Tell (fire-and-forget) messages.
+    promise_id TEXT,
```
sputn1ck, line 28: I wonder if we really need to persist ask messages or if they could be dropped in general. As a durable tell can just retrigger a new tell in the receiving actor and the tell would carry the callers id?

db/actor_delivery_store.go
```diff
@@ -0,0 +1,1196 @@
+package db
```
sputn1ck, line 1: similiar to a prior comment, I think we can move all of this to the baselib itself

baselib/actor/tx_environment.go
```diff
@@ -0,0 +1,51 @@
+package actor
+
+// TxEnvironment is an interface that environments can implement to support
```
sputn1ck, line 3: This really is a cool feature!

baselib/actor/ask_response.go
```diff
@@ -0,0 +1,182 @@
...
+import (
+	"fmt"
+	"io"
+
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+// AskResponseMsgType is the TLV type identifier for AskResponse messages.
+// This is a well-known type used by the DurableAsk pattern.
+const AskResponseMsgType tlv.Type = 0xFFFF // 65535 - reserved for system messages
```
bhandras, line 12: Gemini also flagged just re-flagging so it doesn't get lost: this colludes with `RestartTLVType`.

db/sqlc/migrations/000004_durable_mailbox.down.sql
Roasbeef: Sure, just was following the pattern of putting all the `db` stuff in a single location. I kinda like them to live here as the actor package just declares interfaces that we create/satisfy elsewhere, then pass them in.

db/sqlc/migrations/000004_durable_mailbox.up.sql
Roasbeef: Tell messages are still persisted in the outbox to be delivered. This is for cases where the actor that 's processing needs to write another message into the mailbox for a response.

db/sqlc/migrations/000004_durable_mailbox.up.sql
Roasbeef: That's a good question, we may not actually need them at all depending on exactly what the actor needs. I added it as I realized there was a gap where on restart, the destination would process the ask, but then the response would never make its way back to the sender. 

I can remove it if we don't think we need it at all. 

The case where it's helpful is if after writing a message to an outbox, the actor reliably expects a response to come back across. You can emulate this with a double tell basically, but you'd need to include the "sender address" in the tell, then for the sender to know that this is a response to a request it sent, and not just any other event.

baselib/actor/tx_environment.go
Roasbeef: Yeah this is a big part of the "magic". It's what lets the actors be written basically as normal, ensuring that all their db queries/modifcations run in the same context as the outbox interaction.

baselib/actor/ask_response.go
Roasbeef: Ah, assuming we go with some of this, we can add CI/linter checks to make sure all the message types are unique. 

We can also do protobuf or some other system as suggested elsewhere.

baselib/actor/restart.go
```diff
@@ -0,0 +1,194 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+// RestartTLVType is the TLV type identifier for RestartMessage.
+// Uses a high value to avoid conflicts with application message types.
+const RestartTLVType tlv.Type = 0xFFFF
```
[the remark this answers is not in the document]
bhandras: Good catch.

Fixed by assigning `RestartTLVType` a different reserved system type from `AskResponseMsgType` (now `0xFFFE`), and updating the associated unit test. This avoids codec ambiguity when decoding persisted messages.

baselib/actor/delivery.go
```diff
@@ -0,0 +1,253 @@
...
+		// Save the result to the database.
+		var resultBlob []byte
+		var errorText string
+
+		if err := result.Err(); err != nil {
+			errorText = err.Error()
+		} else {
+			// TODO: Serialize the result value if needed.
+			// For now, we just store nil for success.
+			resultBlob = nil
```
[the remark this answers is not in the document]
bhandras: Acked.

Clarified the `Delivery.Ack` doc/comment to explicitly state that for standard `Ask` we currently persist only success/error status (error text), not the successful result value, and that crash-safe request/response should use the `DurableAsk` pattern.

baselib/actor/durable_actor.go
```diff
@@ -0,0 +1,894 @@
...
+	} else if processed {
+		log.DebugS(processCtx, "Skipping duplicate message",
+			"actor_id", a.id,
+			"delivery_id", delivery.ID)
+
+		// Already processed - Ack immediately.
+		if err := delivery.Ack(processCtx, fn.Ok(*new(R))); err != nil {
+			log.WarnS(processCtx, "Failed to ack duplicate", err,
+				"delivery_id", delivery.ID)
+		}
```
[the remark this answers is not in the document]
bhandras: Good catch. Fixed the duplicate-delivery path so we don't call `Delivery.Ack` for
already-processed messages (which can re-touch `ask_results`). Instead we now
only ack the mailbox row via `AckMessage`.

Also tightened the non-transactional Ask flow to only `MarkProcessed` after a
successful ack, avoiding a state where `IsProcessed` is true while the mailbox
row (and/or Ask result) is still pending.

db/actor_delivery_store.go
bhandras: Makes sense. I kept `ActorDeliveryStore` in `db/` for now to keep the baselib
API surface minimal in this PR, but agree the durable mailbox store and SQLC
plumbing could move into baselib as a follow-up once the durability primitives
settle.

bhandras:
@codex review pls

baselib/actor/durable_actor.go
```diff
@@ -0,0 +1,941 @@
...
+
+	err := a.txAwareStore.ExecTx(ctx, false, func(
+		txCtx context.Context, store DeliveryStore,
+	) error {
+
+		// Execute behavior with panic recovery.
+		result := a.executeBehaviorSafely(txCtx, delivery)
+
+		// Mark as processed within the transaction.
+		if err := store.MarkProcessed(
```
bhandras: Should we unpack the result here to check if it's okay before we persist it processed? I assume there's two ways to think about acking:

- we ack everything we we at least tried to process
- we ack only if processing was successful <= this to feels more correct

baselib/actor/durable_mailbox.go
```diff
@@ -0,0 +1,363 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/lightningnetwork/lnd/clock"
+	"github.com/lightningnetwork/lnd/fn/v2"
+)
+
+// generateID generates a random 16-byte hex-encoded ID.
+func generateID() string {
```
bhandras, line 21: Note that this isn't a ULID so if there's a "tie" between two messages, ie both have the same priority and available_at then the order will be undefined. Maybe we could generate a v7 uuid?

db/sqlc/queries/mailbox.sql
```diff
@@ -0,0 +1,276 @@
...
+    lease_token = $2,
+    lease_until = $3,
+    attempts = attempts + 1
+WHERE mailbox_messages.id = (
+    SELECT m.id FROM mailbox_messages m
+    WHERE m.mailbox_id = $1
+      AND m.available_at <= $4
+      AND (m.lease_until IS NULL OR m.lease_until < $4)
+      AND m.attempts < m.max_attempts
+    ORDER BY m.priority DESC, m.available_at ASC
```
bhandras: Should we take into account created_at as well? It'd help with ordering to some extent.

bhandras: I think ordering could be very important and actually may turn into a subtle correctness issue if the order of events is important and not commutative (such as a reordered update/delete type operation). Perhaps just adding a real ULID and adding created_at to strengthen the ordering would be enough.

db/sqlc/queries/mailbox.sql
```diff
@@ -0,0 +1,276 @@
...
+INSERT INTO dead_letters (id, source, actor_id, message_type, payload, failure_reason, attempts, created_at)
+SELECT o.id, 'outbox', o.source_actor_id, o.message_type, o.payload, $2, o.delivery_attempts, $3
+FROM outbox_messages o
+WHERE o.id = $1;
+
+-- name: DeleteOutboxMessage :exec
+-- Delete an outbox message by ID (cleanup after completion or dead letter).
+DELETE FROM outbox_messages WHERE id = $1;
+
+-- name: CleanupCompletedOutbox :exec
```
bhandras: This is currently not called from anywhere (nor is DeleteOutboxMessage called). As a follow up i think it's important we hook these into a focused GC procedure. Same might be true for dead letter which we don't clean up currently, although those might be important for debugging but in practice we should likely not keep them forever. wdyt?

db/sqlc/migrations/000004_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+    lease_token TEXT,
+
+    -- lease_until is the unix timestamp when the lease expires.
+    -- After expiry, the message becomes available for redelivery.
+    lease_until INTEGER,
+
+    -- Delivery tracking fields.
+    -- available_at is the unix timestamp when the message becomes available.
+    -- Used for scheduling initial delivery and retry delays after Nack.
+    available_at INTEGER NOT NULL,
```
bhandras: q: is there merit for adding a sort of delivery TTL? For durable inbox messages this is perhaps more important: what if the application is not started for a long time? Are we interested in like month old messages? Perhaps this is something we can ensure on the application level and maybe a bad idea to embed validity deadlines into the durable mailbox itself.

db/sqlc/migrations/000003_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+    -- created_at is the unix timestamp when the message was enqueued.
+    created_at INTEGER NOT NULL
+);
+
+-- Index for efficient polling of available messages.
+-- Covers: mailbox lookup, availability check, priority ordering.
+-- Note: We cannot use a partial index with strftime() since it's non-deterministic.
+-- The query handles lease expiry filtering at runtime.
+CREATE INDEX IF NOT EXISTS idx_mailbox_messages_available
+    ON mailbox_messages(mailbox_id, available_at, priority DESC);
```
bhandras: nit: if you agree with my other comments re: ordering, should we key this index by (mailbox_id, priority DESC, available_at ASC, created_at ASC) instead?

baselib/actor/outbox_publisher.go
```diff
@@ -0,0 +1,217 @@
...
+	// Default: 100ms.
+	PollInterval time.Duration
+
+	// BatchSize is the maximum number of messages to process per poll.
+	// Default: 100.
+	BatchSize int
+
+	// MaxDeliveryAttempts is the maximum delivery attempts before dead-lettering.
+	// Default: 10.
+	MaxDeliveryAttempts int
```
bhandras, line 32: Doesn't seem to be used/enforced atm.

bhandras (commented):
Added a few comments i consider higher importance. It's a huge PR so still going through it in detail which means i might add more over time as i process things but hope this is already enough as a start.

db/sqlc/migrations/000004_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+
+    -- attempts tracks how many times delivery has been attempted.
+    attempts INTEGER NOT NULL DEFAULT 0,
+
+    -- max_attempts is the maximum delivery attempts before dead-lettering.
+    max_attempts INTEGER NOT NULL DEFAULT 10,
+
+    -- created_at is the unix timestamp when the message was enqueued.
+    created_at INTEGER NOT NULL
```
bhandras: I think all of these timestamps are now int32 instead of int64 (not flagging the rest but applies everywhere).

db/sqlc/migrations/000004_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+    -- Delivery tracking fields.
+    -- available_at is the unix timestamp when the message becomes available.
+    -- Used for scheduling initial delivery and retry delays after Nack.
+    available_at INTEGER NOT NULL,
+
+    -- attempts tracks how many times delivery has been attempted.
+    attempts INTEGER NOT NULL DEFAULT 0,
+
+    -- max_attempts is the maximum delivery attempts before dead-lettering.
+    max_attempts INTEGER NOT NULL DEFAULT 10,
```
bhandras, line 62: Maybe we don't really even need the max attempts? Maybe an application side configurable (exponential) backoff is all we need?

db/sqlc/migrations/000004_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+    payload BLOB NOT NULL,
+
+    -- domain_key is an optional natural idempotency key.
+    -- For example: "round:abc123:phase:nonces" ensures the same round/phase
+    -- combination is only processed once by the receiver.
+    domain_key TEXT,
+
+    -- version is a monotonic counter for ordering within a domain.
+    -- Higher versions supersede lower versions for the same domain_key.
+    version INTEGER NOT NULL DEFAULT 0,
```
bhandras, line 141: What do we use the versions for? Seems like we set it but don't use it to enforce anything right now.

db/sqlc/migrations/000004_durable_mailbox.up.sql
```diff
@@ -0,0 +1,244 @@
...
+    -- message_type is the type name for deserialization dispatch.
+    message_type TEXT NOT NULL,
+
+    -- payload contains the TLV-encoded message data.
+    payload BLOB NOT NULL,
+
+    -- domain_key is an optional natural idempotency key.
+    -- For example: "round:abc123:phase:nonces" ensures the same round/phase
+    -- combination is only processed once by the receiver.
+    domain_key TEXT,
```
bhandras, line 137: This also doesn't seem like it's used.

baselib/actor/delivery.go
```diff
@@ -0,0 +1,257 @@
...
+	// CorrelationID links DurableAsk requests to their responses.
+	// The caller uses this to match responses to original requests.
+	// Empty for regular Ask/Tell messages.
+	CorrelationID string
+
+	// LeaseToken is the opaque token that must match for ack/nack to succeed.
+	LeaseToken string
+
+	// LeaseUntil is the deadline by which Ack/Nack must be called.
+	LeaseUntil time.Time
```
bhandras, line 55: I don't think this is correctly used yet. There's IsLeaseExpired and then there's LeaseRemaining but those are just generated to fulfill the test but not really enforced anywhere.

baselib/actor/delivery.go
```diff
@@ -0,0 +1,257 @@
...
+	Promise Promise[R]
+
+	// CallerCtx preserves the original caller's context for deadline
+	// propagation. Used when completing Ask promises.
+	CallerCtx context.Context
+
+	// CallbackActorID is set for DurableAsk to route the response.
+	// The response will be delivered to this actor's mailbox via outbox.
+	// Empty for regular Ask/Tell messages.
+	CallbackActorID string
```
bhandras, line 44: Is there a purpose of CallbackActorID in the current codebase?

baselib/actor/delivery.go
```diff
@@ -0,0 +1,257 @@
...
+	LeaseToken string
+
+	// LeaseUntil is the deadline by which Ack/Nack must be called.
+	LeaseUntil time.Time
+
+	// Attempts is the number of delivery attempts for this message.
+	Attempts int
+
+	// MaxAttempts is the maximum allowed attempts before dead-lettering.
+	MaxAttempts int
```
bhandras, line 61: Commented elsewhere too, this isn't currently enforced. (maybe also shouldn't, so better to exponential backoff).

baselib/actor/delivery.go
```diff
@@ -0,0 +1,257 @@
...
+		reason := "max attempts reached"
+		if err != nil {
+			reason = fmt.Sprintf("max attempts reached: %v", err)
+		}
+
+		if dlErr := d.store.MoveToDeadLetter(ctx, d.ID, reason); dlErr != nil {
+			return fmt.Errorf("move to dead letter: %w", dlErr)
+		}
+
+		if delErr := d.store.DeleteMessage(ctx, d.ID); delErr != nil {
```
bhandras, line 225: Shall we atomically delete the message once it's in dead letter box?

baselib/actor/durable_mailbox.go
```diff
@@ -0,0 +1,363 @@
...
+	// wake signals the receive loop to poll immediately.
+	wake chan struct{}
+
+	// actorCtx is the actor's lifecycle context.
+	actorCtx context.Context
+
+	// promiseRegistry maps message IDs to in-flight promises for Ask messages.
+	// This allows the delivery to complete the promise after processing.
+	promiseRegistry   map[string]any
+	promiseRegistryMu sync.RWMutex
```
bhandras, line 88: Is there a need to have two separate mutexes? We lock the promise registry mutex while the close mutex is locked.

baselib/actor/durable_mailbox.go
```diff
@@ -0,0 +1,363 @@
...
+					ctx, leased.ID, leased.LeaseToken, 60*time.Second,
+				)
+
+				continue
+			}
+
+			// Retrieve the promise from the registry if this is an Ask.
+			var promise Promise[R]
+			if leased.PromiseID != "" {
+				m.promiseRegistryMu.Lock()
```
bhandras, line 328: Ah ok, maybe makes sense to have the two mutexes.

baselib/actor/durable_actor.go
```diff
@@ -0,0 +1,941 @@
...
+			txCtx, delivery.ID, a.id, a.deduplicationTTL,
```
bhandras: I don't think this deduplicaton ttl is actually used in the db at all right now.

baselib/actor/durable_actor.go
```diff
@@ -0,0 +1,941 @@
...
+
+			promise.Complete(fn.Err[R](err))
+		}
+	}
+
+	return promise.Future()
+}
+
+// trySendToDLO attempts to send a message to the dead letter office.
+func (ref *durableActorRefImpl[M, R]) trySendToDLO(msg M) {
```
bhandras: Shall we receive the context?

baselib/actor/durable_actor.go
```diff
@@ -0,0 +1,941 @@
...
+		// Mailbox full or other failure.
+		return ErrMailboxFull
+	}
+
+	return nil
+}
+
+// Ask sends a message and returns a Future for the response.
+func (ref *durableActorRefImpl[M, R]) Ask(ctx context.Context, msg M) Future[R] {
+	log.TraceS(ctx, "Sending Ask to durable actor",
```
bhandras, line 931: Just a note for a follow up, we should eliminate this global logger.

litbot-9000:

baselib/actor/durable_mailbox.go
Roasbeef: Yeah uuidv7 makes sense here.

baselib/actor/durable_actor.go
Roasbeef: > we ack only if processing was successful <= this to feels more correct

Hmm, yeah I think this makes sense. One thing that pops into my mind is: potential infinite loops if for w/e reason a message/transition truly cannot be processed. Right now, most of the state machines when up doing a self transition to an error state in the case of an error, but I don't think they all do. 

So given that, perhaps it's more cautious to ack as we _did_ fully process? 

Will make an issue to revisit this.

db/sqlc/queries/mailbox.sql
Roasbeef: Makes sense re ensuring we enforce ordering guarantees here.

db/sqlc/queries/mailbox.sql
Roasbeef: Agreed we need a GC policy here, will remove this for now, then make an issue to follow up with a decision on an a GC policy here.

db/sqlc/migrations/000004_durable_mailbox.up.sql
Roasbeef: Hmm, I think this could make some sense. For example: server sends a round message, client is just flat out offline, server wants to close the round in N seconds. We can translate that into a TTL internally. 

Agree this is something we can revisit, potentially at the app level.

db/sqlc/migrations/000003_durable_mailbox.up.sql
Roasbeef: Make sense, will adopt.

db/sqlc/migrations/000004_durable_mailbox.up.sql
Roasbeef: Hmm, they _should_ be using `BIG INT` iirc, as in `tapd` land we use a replace to make sure we use the largest available integer types across sqlite/postgres.

db/sqlc/migrations/000004_durable_mailbox.up.sql
Roasbeef: Correct not used right now, mainly for future extensibility (hand wave)

db/sqlc/migrations/000004_durable_mailbox.up.sql
Roasbeef: Will trace this through

Roasbeef:
# Durability Branch Deep Review (Codex 5.3)

> **Reviewer**: Codex 5.3 (CodexDurability agent)
> **Date**: 2026-02-06
> **Branch**: `durability` (`abdad2a`)
> **Baseline**: `origin/main` (`b1b61e1`)
> **Scope**: durable mailbox, durable actor runtime, transactional outbox, DB store, message delivery semantics, integration wiring

## Method and Validation

Reviewed code paths and schema from first principles against crash-stop, retry, and concurrent-delivery failure models. Focused on:

- OS/filesystem durability behavior (SQLite WAL/fsync configuration)
- Database transaction boundaries and atomicity
- At-least-once vs exactly-once semantics
- Lease safety and stale-ack prevention
- Cross-actor message propagation via CDC outbox
- Request/response durability (`Ask` and `DurableAsk`)

Validation commands run:

- `go test ./... -count=1` (pass)
- `go test ./baselib/actor -count=1` (pass)
- `go test ./db -count=1` (pass)
- `go test ./internal/actortest -count=1` (pass)
- `go test -race ./baselib/actor -run 'TestDurableActor' -count=1` (pass)

## Executive Summary

This branch is a substantial and thoughtful durability foundation. The lease model, persistence interfaces, tests, and docs are all strong. However, there are several correctness gaps where documented guarantees (especially "exactly-once processing" and "crash-safe DurableAsk") are not currently met under realistic failure modes.

Most important: current outbox delivery and `Ask`/`DurableAsk` completion order can acknowledge success before durable commit/finalization, and can duplicate cross-actor effects when retries occur.

## Findings (Ordered by Severity)

### 1) Critical: `DurableAsk` can acknowledge success even when response outbox write failed (non-tx path)

**Property violated**: crash-safe request/response durability.

In `processWithoutTransaction`, if `writeAskResponseToOutbox` fails, code logs a warning and still calls `Ack` and `MarkProcessed`. That permanently drops the response while treating the request as complete.

**Recommended fix**: For `DurableAsk`, treat outbox-write failure as processing failure. Nack for retry, do not ack/mark-processed when the response cannot be durably enqueued.

### 2) Critical: Outbox retry path can duplicate downstream effects because message identity is not preserved end-to-end

**Property violated**: exactly-once effects across actor boundaries.

`OutboxPublisher` delivers by calling target `Tell(decoded)`. Target mailbox creates a fresh message ID for each delivery attempt. If publisher delivers successfully but fails before `CompleteOutbox`, retry will re-deliver same logical message with a new mailbox ID. Dedup keyed by mailbox ID cannot collapse this duplicate.

**Recommended fix**: Preserve and propagate stable delivery identity from outbox to inbox. Use `domain_key/version` as enforced idempotency key on receiver side.

### 3) High: `Ask` promise can complete before durable ack/transaction commit

**Property violated**: linearizable request completion / truthful success signaling.

`Delivery.Ack` completes in-memory promise before mailbox ack deletion succeeds. In tx path, commit happens after this. If commit fails, caller may already have a success while durable state rolled back.

**Recommended fix**: Defer promise completion until after durable ack/commit success.

### 4) High: Potential data race on `Delivery` state between heartbeat and ack/nack paths

**Property violated**: thread safety under concurrent lease extension and completion.

`heartbeat` goroutine calls `delivery.Extend` while main processing goroutine may call `Ack`/`Nack`. `Delivery.acked` and `LeaseUntil` are mutated/read without synchronization.

**Recommended fix**: Guard `Delivery` mutable fields with mutex/atomic discipline.

### 5) High: Poison decode/type-mismatch messages can become unreachable, not dead-lettered

**Property violated**: bounded retries with terminal observability.

Decode/type-cast errors call `NackMessage` directly. Once attempts reaches `max_attempts`, lease query stops selecting the row, but no dead-letter move occurs.

**Recommended fix**: On decode/type failure, explicitly dead-letter when attempts exhausted.

### 6) Medium: Outbox claim protocol has no in-flight lease/status transition

**Property violated**: single-claimer semantics and ordering under multiple publishers.

`ClaimOutboxBatch` increments `delivery_attempts` but keeps `status='pending'` until completion. With multiple publisher workers/processes, same pending rows can be re-claimed before completion.

**Recommended fix**: Introduce explicit in-flight state with claim lease token + expiry.

### 7) Medium: Time fields are truncated to `int32` unix seconds (2038 and precision risk)

Most persisted times are cast to `int32`. This introduces Y2038 overflow risk and sub-second precision loss.

**Recommended fix**: Use `int64` for persisted unix timestamps end-to-end.

### 8) Medium: Ask promise registry leak on enqueue failure

Ask promise is inserted into `promiseRegistry` before durable enqueue. If enqueue fails, method returns false without removing registry entry.

**Recommended fix**: Register promise only after successful enqueue, or cleanup registry on error.

### 9) Medium: `Send` failure modes collapse into `ErrMailboxFull` at caller edge

Encode/store failures return `false` from mailbox, then actor refs map that to `ErrMailboxFull` if context/termination checks do not match.

**Recommended fix**: Plumb concrete send error from mailbox to refs, preserving cause.

## Strengths

- Strong lease-token stale-ack guard model
- Good SQLite durability posture (`WAL`, `synchronous=full`, `fullfsync=true`)
- Extensive tests across unit/e2e/property dimensions

## Bottom Line

The branch is very strong groundwork and close to production-grade durability, but it is not yet safe to claim exactly-once end-to-end behavior under crash, retry, and partial-failure conditions. The fixes above are tractable and mostly localized to delivery finalization ordering and outbox/inbox identity handling.

Roasbeef:
## Durability Review Fixes Pushed

Five commits addressing findings from the Codex 5.3 deep review:

### `baselib/actor: harden Delivery with mutex and deferred promise completion`
- **Fix #4**: `sync.Mutex` guards `acked` and `LeaseUntil` against concurrent heartbeat/ack races
- **Fix #3**: Promise completion moved after durable ack; new `deferPromise` flag for tx path

### `baselib/actor: nack DurableAsk on outbox failure and defer tx promise`
- **Fix #1**: DurableAsk now nacks for retry when outbox write fails (was silently acking → dropped response)
- **Fix #3** (tx-side): `processInTransaction` defers promise until after `ExecTx` commits
- DurableAsk in non-tx path defers `MarkProcessed` until after outbox write succeeds

### `baselib/actor: harden mailbox with poison dead-lettering and outbox dedup`
- **Fix #5**: Poison messages (decode/type-cast failures) dead-lettered after `max_attempts` exhaustion
- **Fix #8**: Promise registry cleaned up on `EnqueueMessage` failure
- **Fix #2**: Outbox ID propagated via context for receiver-side deduplication; SQL query gains `ON CONFLICT (id) DO NOTHING`

### `db/sqlc: regenerate queries after mailbox ON CONFLICT change`
- Generated code from `make sqlc`

### `baselib/actor: add tests for durability review fixes`
- 14 new tests covering all 5 fixes, all passing with `-race` (7.3s total)
- Mock store updated with ON CONFLICT semantics and per-operation error injection

**Remaining items** (lower severity, separate PRs):
- Finding #6 (Medium): Outbox claim protocol lacks in-flight lease
- Finding #7 (Medium): int32 timestamp truncation
- Finding #9 (Medium): Send error propagation
- Issue #101: Formal P/Quint specification

baselib/actor/delivery.go
```diff
@@ -0,0 +1,304 @@
...
+
+		if err := result.Err(); err != nil {
+			errorText = err.Error()
+		} else {
+			// For standard Ask, only the success status is persisted,
+			// not the result value itself. See the doc comment above.
+			resultBlob = nil
+		}
+
+		err := d.store.SaveAskResult(ctx, AskResultParams{
```
bhandras: Potential correctness issue: `SaveAskResult` happens before `AckMessage` lease validation in `Delivery.Ack`. In a non-tx path, a stale lease holder can persist an Ask result, then fail `AckMessage` with lease expiry, and because `ask_results` is `ON CONFLICT DO NOTHING` the later valid worker may not be able to overwrite it. Can we reorder so persistence happens only after a successful ack ownership check (or couple this in one tx-safe operation)?

db/sqlc/queries/mailbox.sql
```diff
@@ -0,0 +1,281 @@
...
+    domain_key,
+    version,
+    created_at
+) VALUES ($1, $2, $3, $4, $5, $6, $7, $8);
+
+-- name: ClaimOutboxBatch :many
+-- Claim a batch of pending outbox messages for delivery.
+-- Updates status to 'pending' with incremented delivery_attempts.
+-- Returns messages ordered by creation time.
+UPDATE outbox_messages
```
bhandras, line 157: Still seeing the original medium-risk outbox-claim concern: this query increments `delivery_attempts` but keeps rows as `status = pending`, with no claim token/lease. With concurrent publishers, pending rows can be reclaimed repeatedly before completion, which can inflate attempts and dead-letter healthy messages early. Should we add an explicit in-flight claim state (or claim lease columns) as a follow-up?

db/sqlc/models.go
```diff
@@ -52,6 +60,63 @@ type ClientTreeTxid struct {
...
+	ID              string
+	MailboxID       string
+	MessageType     string
+	Payload         []byte
+	PromiseID       sql.NullString
+	CallbackActorID sql.NullString
+	CorrelationID   sql.NullString
+	Priority        int32
+	LeaseToken      sql.NullString
+	LeaseUntil      sql.NullInt32
```
bhandras: Timestamp representation still looks int32 in generated model/store paths (e.g. `LeaseUntil`, `AvailableAt`, `CreatedAt`, `CompletedAt`). That keeps the Y2038 boundary and narrows headroom for long-lived clients. Can we track an `int64` migration follow-up for mailbox/outbox/processed/fsm time fields end-to-end?

baselib/actor/durable_actor.go
```diff
@@ -0,0 +1,1068 @@
...
+			return ErrActorTerminated
+		}
+
+		// Check if caller's context was cancelled.
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+
+		// Mailbox full or other failure.
+		return ErrMailboxFull
```
bhandras, line 923: Error propagation note: `DurableMailbox.Send` returns only bool, so this path maps all non-context send failures to `ErrMailboxFull`. That flattens encode/store/tx failures into a capacity signal and makes retries/diagnostics harder. Would it make sense to plumb a concrete send error through mailbox/ref (while preserving existing API where needed)?

bhandras (approved):
LGTM pending last set of codex comments.

Roasbeef:
Re: `Mailbox.Send` returning `bool` instead of `error`

Agreed this is worth fixing — `DurableMailbox.Send` currently returns `bool` which loses the underlying error context (encode failure, DB error, context cancellation all collapse to `false`). This makes debugging delivery failures harder and prevents callers from distinguishing transient vs permanent failures.

Tracking this as a follow-up since it touches the `Mailbox` interface contract and all callers (`Tell`, `Ask`, `DurableAsk`, outbox publisher, tests). See the linked issue for the full scope.

PR #50, opened 2025-12-19
systest: add new systest framework build on top of existing harness, add wallet boarding tests

Roasbeef:
In this PR, we add a new `systest` framework built on top of the existing harness. It wraps the harness with an active actor system, and some other useful utilities. 

Along the way we fix some subtle bugs related to cotnexts (deciding when to make a new one vs pass on the parent context). 

We also revamp the way logging is set up for the actors, going back to the old style of relying primarily on a logger passed in as config. We then supplement that as needed with a logger passed in via a context. 

Finally we update CI to run the new systests using both postgres and sqlite.

### DIFF
--- a/chainbackends/lndclient_adapters.go
+++ b/chainbackends/lndclient_adapters.go
@@ -1,0 +1,369 @@
+package chainbackends
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btclog/v2"
+	"github.com/lightninglabs/darepo-client/build"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightningnetwork/lnd/chainntnfs"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
+)
+
+// LndClientTxBroadcaster implements TxBroadcaster using
+// lndclient.WalletKitClient.
+type LndClientTxBroadcaster struct {
+	walletKit lndclient.WalletKitClient
+}
+
+// NewLndClientTxBroadcaster creates a new broadcaster backed by lndclient.
+func NewLndClientTxBroadcaster(
+	walletKit lndclient.WalletKitClient) *LndClientTxBroadcaster {
+
+	return &LndClientTxBroadcaster{
+		walletKit: walletKit,
+	}
+}
+
+// PublishTransaction broadcasts the given transaction to the network via lnd.
+func (b *LndClientTxBroadcaster) PublishTransaction(ctx context.Context,
+	tx *wire.MsgTx, label string) error {
+
+	return b.walletKit.PublishTransaction(ctx, tx, label)
+}
+
+// Ensure LndClientTxBroadcaster implements TxBroadcaster.
+var _ TxBroadcaster = (*LndClientTxBroadcaster)(nil)
+
+// LndClientFeeEstimator implements chainfee.Estimator using lndclient.
+type LndClientFeeEstimator struct {
+	walletKit lndclient.WalletKitClient
+}
+
+// NewLndClientFeeEstimator creates a new fee estimator backed by lndclient.
+func NewLndClientFeeEstimator(
+	walletKit lndclient.WalletKitClient) *LndClientFeeEstimator {
+
+	return &LndClientFeeEstimator{
+		walletKit: walletKit,
+	}
+}
+
+// Start is a no-op for lndclient-backed estimators since the connection is
+// already established.
+func (e *LndClientFeeEstimator) Start() error {
+	return nil
+}
+
+// Stop is a no-op for lndclient-backed estimators.
+func (e *LndClientFeeEstimator) Stop() error {
+	return nil
+}
+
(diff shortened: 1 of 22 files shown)


chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,366 @@
...
+			}
+
+		case <-ctx.Done():
+		}
+	}()
+
+	return &chainntnfs.SpendEvent{
+		Spend:  spendChan,
+		Reorg:  make(chan struct{}, 1),
+		Done:   make(chan struct{}),
```
bhandras: Here i think we want this done channel defined before the goroutine and close it when the parent goroutine stops.

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,366 @@
...
+				)
+			}
+
+		case <-ctx.Done():
+		}
+	}()
+
+	return &chainntnfs.ConfirmationEvent{
+		Confirmed: confChan,
+		Cancel:    cancel,
```
bhandras, line 205: We want to set `Done: done` here.

chainsource/block_epoch_actor.go
```diff
@@ -48,20 +70,26 @@ type BlockEpochActor struct {
...
-func NewBlockEpochActor(backend ChainBackend,
-	parentCtx context.Context) *BlockEpochActor {
-
-	ctx, cancel := context.WithCancel(parentCtx)
+// configuration. The config must include Backend; use WithLogger() to inject
+// a specific logger.
+func NewBlockEpochActor(cfg BlockEpochConfig) *BlockEpochActor {
+	// Use background context for internal cancellation since the actor
+	// needs to outlive any request context. Logger is passed via config.
+	ctx, cancel := context.WithCancel(context.Background())
```
bhandras, line 78: Somewhat related (to lifecycle management/context cancellation) is that currently we cancel this context from `Stop` but we also want these actors to be `Stoppable` so `OnStop` can be called: https://github.com/lightninglabs/darepo-client/blob/62d15e6a4eddee220a92abb9128fa1d034faacd9/baselib/actor/actor.go#L270

bhandras (approved):
LGTM (pending some small issues) 🎉 

Perhaps consider addressing this potential panic too (or in a follow up): https://github.com/lightninglabs/darepo-client/issues/55

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,366 @@
...
+	satPerKw := chainfee.SatPerVByte(satPerVByte).FeePerKWeight()
+
+	return satPerKw, nil
+}
+
+// RelayFeePerKW returns the minimum fee rate required for transactions to be
+// relayed. For remote lnd, we return a reasonable default.
+func (e *LndClientFeeEstimator) RelayFeePerKW() chainfee.SatPerKWeight {
+	// Default relay fee of 1 sat/vbyte = 250 sat/kw.
+	return chainfee.SatPerKWeight(250)
```
ellemouton, line 94: should we make configurable?

chainbackends/lndclient_adapters.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// EstimateFeePerKW returns the estimated fee rate in satoshis per kilo-weight
+// unit for the given confirmation target.
+func (e *LndClientFeeEstimator) EstimateFeePerKW(
+	numBlocks uint32) (chainfee.SatPerKWeight, error) {
+
+	satPerVByte, err := e.walletKit.EstimateFeeRate(
+		context.Background(), int32(numBlocks),
+	)
```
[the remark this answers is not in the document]
ellemouton: shall we keep the context?

ellemouton (approved):

chainbackends/lndclient_adapters.go
Roasbeef: Hmm, so the current `lndclient` interface doesn't actually send on this at all. It's meant to be listened on once things "can't" re re-org'd out: 
```
	// Done is a channel that gets sent upon once the confirmation request
	// is no longer under the risk of being reorged out of the chain.
	//
	// NOTE: This channel must be buffered.
	Done chan struct{}
```

chainbackends/lndclient_adapters.go
Roasbeef: Ended up removing the `done` channel above, as it isn't actually used for a purpose. Also as mentioned above, the `Done` channel has a distinct case, will set it for now though to prevent a panic.

chainbackends/lndclient_adapters.go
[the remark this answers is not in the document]
Roasbeef: There's no context to keep here, unless we move away from using the `lnd` interfaces. I can add the 30 second timeout here though as noted

chainbackends/lndclient_adapters.go
Roasbeef: For mock purposes? FWIW this is there just to meet the `lnd` interface. If it's `lnd`, it'll use the _actual_ min relay fee it knows of, which factors in the `feefilter` message amongst other sources.

chainsource/block_epoch_actor.go
Roasbeef: Yep gonna do that in a new PR. Want to note here that this was a needed bug fix, as we want to make sure the blcok epoch actor _outlives_ the context was passed in to _make the request_.

PR #52, opened 2025-12-20
OOR client FSM (grand summary)

bhandras:
## Summary
This is the grand roll-up for the full OOR client FSM work.

The reviewable changes are split across smaller PRs for incremental review; this PR tracks the consolidated branch for end-to-end context and CI only.

## Split PRs (review targets)
- PR77: OOR client split 1 (primitives + validation)
- PR78: OOR client split 2 (core FSM + actor wrapper)
- PR79: OOR client split 3 (tx split + signing + resume)
- PR81: OOR client split 4 (retry/confirm polish + tests)

## Notes
- Base: `main`.
- This PR should not be the primary review surface; use the split PRs above.

### DIFF
--- a/lib/tx/oor/taptree.go
+++ b/lib/tx/oor/taptree.go
@@ -1,0 +1,168 @@
+package oor
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+)
+
+var (
+	// TapTreePSBTKey is the v0 convention for storing a taproot tree
+	// encoding in a PSBT input unknown field.
+	//
+	// We treat this as part of the OOR PSBT profile so client and server
+	// implementations can deterministically attach, validate, and later use
+	// the same metadata during finalization.
+	TapTreePSBTKey = []byte("taptree")
+)
+
+// EncodeTapTree encodes a set of tapscript leaves into a single byte blob.
+//
+// EncodeTapTree intentionally uses a simple leaf list representation that is
+// sufficient for v0 OOR transfers. Each leaf is encoded at depth 1 with the
+// base tapscript leaf version. The encoding uses Bitcoin varint (compact size)
+// lengths and is compatible with how many BIP-371 encodings represent tap
+// trees.
+//
+// This encoding is part of the PSBT profile for OOR transfers. If we ever need
+// to support richer trees (multiple depths), this function must become
+// versioned rather than changing behavior silently.
+func EncodeTapTree(leaves [][]byte) ([]byte, error) {
+	var buf bytes.Buffer
+
+	err := wire.WriteVarInt(&buf, 0, uint64(len(leaves)))
+	if err != nil {
+		return nil, fmt.Errorf("unable to write leaf count: %w",
+			err)
+	}
+
+	for _, leaf := range leaves {
+		err := buf.WriteByte(1)
+		if err != nil {
+			return nil, fmt.Errorf("unable to write depth: %w",
+				err)
+		}
+
+		err = buf.WriteByte(byte(txscript.BaseLeafVersion))
+		if err != nil {
+			return nil, fmt.Errorf("unable to write leaf "+
+				"version: %w", err)
+		}
+
+		err = wire.WriteVarInt(&buf, 0, uint64(len(leaf)))
+		if err != nil {
+			return nil, fmt.Errorf("unable to write leaf "+
+				"length: %w", err)
+		}
+
+		_, err = buf.Write(leaf)
+		if err != nil {
+			return nil, fmt.Errorf("unable to write leaf "+
+				"script: %w", err)
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+// DecodeTapTree decodes a tap tree encoding produced by EncodeTapTree.
+//
+// DecodeTapTree is intentionally lenient about leaf depth and version in v0:
+// it reads and ignores them. The returned value is the list of raw script
+// bytes for each leaf.
+func DecodeTapTree(data []byte) ([][]byte, error) {
+	buf := bytes.NewReader(data)
+
+	leafCount, err := wire.ReadVarInt(buf, 0)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read leaf count: %w",
+			err)
+	}
+
+	leaves := make([][]byte, 0, leafCount)
+	for i := uint64(0); i < leafCount; i++ {
+		_, err := buf.ReadByte()
+		if err != nil {
+			return nil, fmt.Errorf("unable to read depth: %w",
+				err)
+		}
+
+		_, err = buf.ReadByte()
+		if err != nil {
+			return nil, fmt.Errorf("unable to read leaf "+
+				"version: %w", err)
+		}
+
+		scriptLen, err := wire.ReadVarInt(buf, 0)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read script "+
+				"length: %w", err)
+		}
+
+		scriptBytes := make([]byte, scriptLen)
+		_, err = buf.Read(scriptBytes)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"unable to read script bytes: %w", err,
+			)
+		}
+
+		leaves = append(leaves, scriptBytes)
+	}
+
+	if buf.Len() != 0 {
+		return nil, fmt.Errorf("trailing bytes in tap tree "+
+			"encoding (%d bytes)", buf.Len())
+	}
+
+	return leaves, nil
+}
+
+// PutTapTreePSBTInput stores an encoded tap tree blob into the given PSBT input
+// unknown fields, using TapTreePSBTKey.
+func PutTapTreePSBTInput(pkt *psbt.Packet, inputIndex int,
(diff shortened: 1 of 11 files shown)


lib/tx/arktx/canonical.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// CanonicalizeOrdering sorts the transaction inputs and outputs in-place
+// according to the v0 rules.
+//
+// CanonicalizeOrdering does not insert or remove outputs. It assumes the
+// caller has already constructed an Ark tx that includes exactly one anchor
+// output of value 0. If the anchor output is missing or invalid, this returns
+// an error rather than guessing what to do.
+func CanonicalizeOrdering(tx *wire.MsgTx) error {
```
Roasbeef: Seems we can use the existing BIP 69 sort here? We use a slight variant elsewhere in lnd.

lib/tx/checkpoint/taptree.go
```diff
@@ -0,0 +1,111 @@
...
+// EncodeTapTree intentionally uses a simple leaf list representation that is
+// sufficient for v0 OOR transfers. Each leaf is encoded at depth 1 with the
+// base tapscript leaf version. The encoding uses Bitcoin varint (compact size)
+// lengths and is compatible with how many BIP-371 encodings represent tap
+// trees.
+//
+// This encoding is part of the PSBT profile for OOR transfers. If we ever need
+// to support richer trees (multiple depths), this function must become
+// versioned rather than changing behavior silently.
+func EncodeTapTree(leaves [][]byte) ([]byte, error) {
```
Roasbeef: Any reason to not use TLV here? 

You could also in theory pass in `txscript.IndexTapscriptTree` here.

Roasbeef: My round persistence PR has  TLV encoding for `waddrmgr.Tapscript`, which is used elsewhere and is a super set of just the leaves.

lib/tx/oor/taptree.go
```diff
@@ -0,0 +1,85 @@
...
+	// We treat this as part of the OOR PSBT profile so client and server
+	// implementations can deterministically attach, validate, and later use
+	// the same metadata during finalization.
+	//
+	// NOTE: PSBT unknown keys are a shared namespace. We use a short,
+	// stable key here for v0 tests and in-process wiring.
+	//
+	// A future version should consider namespacing this (for example,
+	// `ark/taptree`) to reduce collision risk with other PSBT extensions.
+	TapTreePSBTKey = []byte("taptree")
```
Roasbeef, line 19: Is it critical that we encode this within the PSBT, or is it fine to just transmit the tapscript leaf information along side?

oor/actor.go
```diff
@@ -0,0 +1,400 @@
...
+
+		for _, followUp := range followUps {
+			// Feed follow-up events into the FSM.
+			// Recursively execute any emitted outbox work.
+			// Stop when none remains.
+			nextOutbox, err := a.askEvent(ctx, fsm, followUp)
+			if err != nil {
+				return err
+			}
+
```
Roasbeef: Does this handle fault tolerance w.r.t persistence + delivery of these outbox messages the way https://github.com/lightninglabs/darepo-client/pull/48 does?

Roasbeef: Like let's say you're sending a message to the chain resolver sub-system that things failed and we need to go on chain. It gets this message but doesn't actually process it. This is now critical on restart that we start the un roll process, does this pattern have a cut out to handle such patterns?

oor/events.go
```diff
@@ -0,0 +1,189 @@
...
+// This is a foundational design choice for mobile safety: after a crash, the
+// application can restore a persisted snapshot and re-drive the outbox implied
+// by the current state.
+type Event interface {
+	eventSealed()
+}
+
+// StartTransferEvent requests starting an OOR transfer by building a submit
+// package (checkpoint PSBTs + Ark PSBT).
+type StartTransferEvent struct {
```
Roasbeef: What about the receiver side? Is your idea to have another state machine for receiving, or it'll be bundled into this state machine?

oor/events.go
```diff
@@ -0,0 +1,189 @@
...
+// IncomingTransferEvent notifies the client about an incoming OOR transfer.
+//
+// This event is intended to be delivered by some higher layer (RPC push,
+// polling, or push-notification wakeup) once the server has accepted and
+// finalized the transfer.
+//
+// The incoming transfer FSM is intentionally separate from the outgoing FSM so
+// applications can handle notifications and acknowledgements independently of
+// initiating transfers.
+type IncomingTransferEvent struct {
```
Roasbeef: Ah scratch my other comment, I see this now for incoming transfers.

oor/receive_transitions.go
```diff
@@ -0,0 +1,141 @@
...
+		// identical bytes.
+		err := oor.ValidateCanonicalArkPSBT(evt.ArkPSBT)
+		if err != nil {
+			return nil, err
+		}
+
+		// Extract recipients and surface the notification to the
+		// application layer.
+		//
+		// The outbox is intentionally ordered:
```
Roasbeef: Could move this comment down below re when the outbox fields are being set.

oor/receive_transitions.go
```diff
@@ -0,0 +1,141 @@
...
+		//   1) notify the app/UI so it can show the transfer;
+		//   2) materialize incoming VTXOs into local state; and
+		//   3) ack receipt to the server (best-effort, idempotent).
+		recipients, err := ExtractArkRecipients(evt.ArkPSBT)
+		if err != nil {
+			return nil, err
+		}
+
```
Roasbeef: Signing looks to be missing?

oor/actor.go
Roasbeef: Also depending on the fault tolerance properties of the `OutboxHandler` (eg: does it handle all the messages in a single db transaction?), we may end up with inconsistent state on disk. Still getting through things atm, so persistence isn't super clear quite yet.

vtxo/actor.go
```diff
@@ -250,26 +262,49 @@ type VTXOActorResponse struct {
...
+			VTXO:   vtxo,
+			Reason: recoveredMsg,
+		}
 
 	default:
-		return &LiveState{VTXO: vtxo, LastCheckedHeight: vtxo.CreatedHeight}
+		return &LiveState{
+			VTXO:              vtxo,
+			LastCheckedHeight: vtxo.CreatedHeight,
+		}
```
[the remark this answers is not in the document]
bhandras: Fixed by changing the default case to return `FailedState` with an explicit "unknown vtxo status" reason, so we fail closed on unexpected statuses.

lib/tx/arktx/canonical.go
bhandras: Updated canonical output ordering to BIP69 (amount then pkScript) and aligned the builders/validators and design doc; anchor output remains last.

lib/tx/checkpoint/taptree.go
bhandras: Switched the tap tree encoding to the TLV leaf format used by waddrmgr.Tapscript, so we share the same durable representation while still encoding the full leaf set.

lib/tx/oor/taptree.go
bhandras: Not strictly required, but embedding it in the PSBT keeps the transfer package self-contained and deterministic across the client/server boundary and restarts; the same bytes being signed also carry the tap tree metadata, which avoids out-of-band mismatches during finalize/restore. The tradeoff is a slightly larger PSBT and tighter coupling to the PSBT profile/unknown-key namespace; passing it alongside is lighter-weight but adds coordination/state to keep metadata and signed bytes in sync.

oor/actor.go
bhandras: Thank you for pointing this out :pray: The current actor only persists the session snapshot and re-derives the outbox from state on resume, but it does not persist the outbox itself. That means correctness depends on the `OutboxHandler` providing durable, idempotent delivery (ideally a transactional outbox: persist snapshot + enqueue outbox in one DB tx), which is exactly what the durable actor layer in #48 provides. Without that, there’s a window where state can advance but the side effect isn’t durably queued, or a side effect is delivered and the follow-up transition isn’t persisted. So the intent here is to keep the boundary compatible with #48’s transactional outbox semantics; the handler implementation is where we must enforce atomicity/dedup for critical effects.

oor/events.go
bhandras: Plan is a separate receive-side FSM. The incoming transfer path is modeled independently (see `IncomingTransferEvent` and receive states), so it’s not bundled into the outgoing FSM; this keeps notification/ack flows and failure handling separate.

oor/receive_transitions.go
bhandras: Moved the explanatory comment to sit next to the outbox emission so it’s aligned with where the fields are set.

oor/receive_transitions.go
bhandras: Receive-time signing isn’t required in the current model. The incoming path only validates the canonical Ark PSBT, surfaces the transfer, and materializes VTXOs; the VTXO descriptor is built so the recipient can sign later when spending (collab or timeout). If you want an explicit receipt signature or other policy (e.g., cryptographic ack), we can add a dedicated outbox step, but that would be a protocol change beyond the current design.

bhandras:
Closing this as this is replaced by the splits.

PR #68, opened 2026-01-08
tx: add forfeit transaction structure validation

Roasbeef:
This PR adds foundational validation logic for forfeit transactions used in the Ark batch swap protocol. The `ValidateForfeitTx` function verifies that a transaction has the correct structure: exactly two inputs (VTXO outpoint at index 0, connector outpoint at index 1) and exactly two outputs (penalty output to operator, ephemeral anchor for fee bumping).

The validation ensures structural correctness before signing, preventing malformed transactions from being processed. It checks that the VTXO and connector outpoints match expected values, the penalty output pays to the operator's script, and the anchor output uses the standard P2A script with zero value.

An optional expected amount can be provided to verify the penalty output value matches. This is useful when the caller knows the exact amount that should be paid (e.g., from protocol parameters). When set to zero, amount validation is skipped.

Comprehensive test coverage includes:
- Valid transaction structure verification  
- Input/output count validation
- Outpoint matching for VTXO and connector
- Penalty script and amount validation
- Anchor output format validation
- Round-trip test with various BTC amounts

This is the first PR in a series implementing VTXO lifecycle management with forfeit flow support.

### DIFF
--- a/lib/tx/forfeit.go
+++ b/lib/tx/forfeit.go
@@ -1,6 +1,7 @@
 package tx
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/btcsuite/btcd/btcutil"
@@ -19,6 +20,14 @@ const (
 
 	// ForfeitConnectorInputIndex is the index of the connector input.
 	ForfeitConnectorInputIndex = 1
+
+	// ForfeitPenaltyOutputIndex is the index of the penalty output that
+	// pays to the server's forfeit address.
+	ForfeitPenaltyOutputIndex = 0
+
+	// ForfeitAnchorOutputIndex is the index of the ephemeral anchor output
+	// for fee bumping via CPFP.
+	ForfeitAnchorOutputIndex = 1
 )
 
 // VTXOSpendContext describes the VTXO being spent.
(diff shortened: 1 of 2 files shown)


Roasbeef:
@claude review this 

@codex review this

lib/tx/forfeit.go
```diff
@@ -140,3 +141,92 @@ func NewVTXOCollabSignDescriptor(vtxo *VTXOSpendContext,
...
+	}
+
+	// Forfeit tx must have exactly 2 outputs (penalty + anchor).
+	if len(forfeitTx.TxOut) != 2 {
+		return fmt.Errorf("forfeit tx has %d outputs, expected 2",
+			len(forfeitTx.TxOut))
+	}
+
+	// Verify output 0 pays to server's forfeit address.
+	if !bytes.Equal(forfeitTx.TxOut[0].PkScript, params.ServerForfeitScript) {
```
ellemouton: i think worth creating constants just like the above `ForfeitConnectorInputIndex` and `ForfeitVTXOInputIndex` maybe?

ellemouton: ah yes - gemini makes same comment below 👍

ellemouton (approved):
LGTM!

maybe this is coming later on in the series but i think we'd also want to verify that the given connector is actually part of the batch tx (although... maybe the client actually doesnt care about this and it is just the operators responsibility to ensure this 🤷‍♀️ )

Roasbeef:
> maybe this is coming later on in the series but i think we'd also want to verify that the given connector is actually part of the batch tx

Yep, this is later on in the series. Wanted to break this one out as it was pretty independent.

PR #69, opened 2026-01-08
vtxo: implement FSM for VTXO lifecycle management

Roasbeef:
This PR implements a finite state machine for managing VTXO lifecycle in the Ark protocol client. The FSM tracks VTXOs through their various states from creation through expiry or forfeiture, enabling proper coordination during batch swaps.

The core states include: `LiveState` (active VTXO), `RefreshingState` (awaiting refresh), `ForfeitingState` (forfeit tx signed, awaiting confirmation), `RefreshedState` (replaced by new VTXO), `ForfeitedState` (forfeit confirmed), `ExpiringState` (approaching expiry), and `ExpiredState` (past expiry window). Each state defines valid transitions based on events like block epochs, forfeit requests, and refresh completions.

The expiry monitoring logic tracks block heights relative to VTXO creation, calculating remaining blocks until expiry. When thresholds are crossed (configurable warning/critical levels), the FSM emits refresh requests or escalates to on-chain resolution. This ensures VTXOs are proactively refreshed before the operator can unilaterally claim them.

For crash recovery, the `ForfeitingState` persists the signed forfeit transaction via `MarkForfeiting()`. On restart, `statusToState()` reconstructs the state with the persisted forfeit tx, allowing the actor to resume monitoring for on-chain confirmation without re-signing.

This PR also includes `lib/actormsg` (shared message interfaces for cross-package actor communication) and `round/vtxo_messages.go` (VTXO actor message types) which define the event types consumed by the FSM. Comprehensive test coverage verifies state transitions, expiry calculations, and forfeit flow behavior.

## Dependencies
- Depends on #68 (tx: add forfeit transaction structure validation)

### DIFF
--- a/vtxo/events.go
+++ b/vtxo/events.go
@@ -1,0 +1,152 @@
+package vtxo
+
+import (
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/baselib/actor"
+)
+
+// VTXOEvent is a sealed interface for all events that can be processed by the
+// VTXO state machine. The sealed pattern prevents external packages from
+// adding unvalidated event types. Extends actor.Message for actor system
+// compatibility.
+type VTXOEvent interface {
+	actor.Message
+	vtxoEventSealed()
+}
+
+// BlockEpochEvent is received when a new block is connected to the blockchain.
+// This triggers expiry monitoring logic to check if the VTXO needs to be
+// refreshed or escalated to the chain resolver.
+type BlockEpochEvent struct {
+	actor.BaseMessage
+
+	// Height is the block height.
+	Height int32
+
+	// Hash is the block hash.
+	Hash chainhash.Hash
+
+	// Timestamp is the block timestamp from the header.
+	Timestamp int64
+}
+
+func (e *BlockEpochEvent) vtxoEventSealed() {}
+
+// MessageType returns the message type for logging.
+func (e *BlockEpochEvent) MessageType() string { return "BlockEpochEvent" }
+
+// ForfeitRequestEvent is received from the round actor when this VTXO is being
+// forfeited as part of a batch swap. The VTXO actor should sign the forfeit
+// transaction and transition to the forfeiting state.
+//
+// The forfeit transaction structure:
+//   - Input 0: VTXO being forfeited (collaborative spend - client + operator)
+//   - Input 1: Connector output from new commitment tx (operator key spend)
+//   - Output 0: Full VTXO value to operator's forfeit address
+//   - Output 1: Anchor output (zero-value P2A for CPFP)
+type ForfeitRequestEvent struct {
+	actor.BaseMessage
+
+	// RoundID is the new round where the refreshed VTXO will be created.
+	RoundID string
+
+	// ConnectorOutpoint is the connector output from the new commitment tx
+	// that the forfeit tx must spend. This links the forfeit atomically to
+	// the new round.
+	ConnectorOutpoint wire.OutPoint
+
+	// ConnectorPkScript is the scriptPubKey of the connector output.
+	ConnectorPkScript []byte
+
+	// ConnectorAmount is the value of the connector output in satoshis.
+	ConnectorAmount int64
+
+	// ServerForfeitPkScript is the operator's taproot script where the
+	// forfeited VTXO value will be paid.
+	ServerForfeitPkScript []byte
+}
+
+func (e *ForfeitRequestEvent) vtxoEventSealed() {}
+
+// MessageType returns the message type for logging.
+func (e *ForfeitRequestEvent) MessageType() string { return "ForfeitRequestEvent" }
+
+// ForfeitSignedEvent indicates the forfeit transaction has been signed and
+// submitted to the round. This is an internal event triggered after the VTXO
+// actor signs its portion of the forfeit transaction.
+type ForfeitSignedEvent struct {
+	actor.BaseMessage
+
+	// ForfeitTxID is the txid of the forfeit transaction.
+	ForfeitTxID chainhash.Hash
+}
+
+func (e *ForfeitSignedEvent) vtxoEventSealed() {}
+
+// MessageType returns the message type for logging.
+func (e *ForfeitSignedEvent) MessageType() string { return "ForfeitSignedEvent" }
+
+// ForfeitConfirmedEvent indicates the new commitment transaction has been
+// confirmed on-chain, meaning the forfeit is final. The old VTXO is now
+// permanently forfeited and the new VTXO is live.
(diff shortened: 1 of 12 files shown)


Roasbeef:
@claude review this 

@codex review this

vtxo/interfaces.go
```diff
@@ -0,0 +1,185 @@
...
+// VTXOStateMachineCfg is a type alias for the VTXO FSM configuration.
+type VTXOStateMachineCfg = protofsm.StateMachineCfg[
+	VTXOEvent, VTXOOutMsg, *VTXOEnvironment,
+]
+
+// VTXOStatus represents the lifecycle state of a VTXO.
+type VTXOStatus int
+
+const (
+	// VTXOStatusLive indicates the VTXO is active and can be spent.
```
ellemouton, line 213: what about when it is spent (via an OOR)? should we have a state for that too?

vtxo/interfaces.go
```diff
@@ -0,0 +1,185 @@
...
+		return "failed"
+	default:
+		return "unknown"
+	}
+}
+
+// Descriptor contains all information needed to track and spend a VTXO. This
+// is the canonical representation persisted to storage and passed between
+// actors.
+type Descriptor struct {
```
ellemouton, line 265: i think this right now only defines "confirmed vtxos" (ie, vtxt leaves) - if we are describing a pre-confirmed vtxo, then that would also need to point to a set of checkpoint txs that it is spending (which would then point to parent vtxos)

vtxo/events.go
```diff
@@ -0,0 +1,152 @@
...
+	Timestamp int64
+}
+
+func (e *BlockEpochEvent) vtxoEventSealed() {}
+
+// MessageType returns the message type for logging.
+func (e *BlockEpochEvent) MessageType() string { return "BlockEpochEvent" }
+
+// ForfeitRequestEvent is received from the round actor when this VTXO is being
+// forfeited as part of a batch swap. The VTXO actor should sign the forfeit
```
ellemouton: > forfeited as part of a batch swap.

or a leave request

vtxo/events.go
```diff
@@ -0,0 +1,152 @@
...
+// transaction and transition to the forfeiting state.
+//
+// The forfeit transaction structure:
+//   - Input 0: VTXO being forfeited (collaborative spend - client + operator)
+//   - Input 1: Connector output from new commitment tx (operator key spend)
+//   - Output 0: Full VTXO value to operator's forfeit address
+//   - Output 1: Anchor output (zero-value P2A for CPFP)
+type ForfeitRequestEvent struct {
```
ellemouton: i assume there will be a wrapping struct that the round-fsm will use to indicate which vtxo is actually being forfeited? ie, at this level, we already know which one this is directed at

vtxo/events.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (e *ForfeitRequestEvent) vtxoEventSealed() {}
+
+// MessageType returns the message type for logging.
+func (e *ForfeitRequestEvent) MessageType() string { return "ForfeitRequestEvent" }
+
+// ForfeitSignedEvent indicates the forfeit transaction has been signed and
+// submitted to the round. This is an internal event triggered after the VTXO
+// actor signs its portion of the forfeit transaction.
+type ForfeitSignedEvent struct {
```
ellemouton: i feel like we should come up with a nice naming rule to make it clear what events are coming in to the FSM vs which are being emitted out of the FSM (although... this does become tricky when internal events i guess 🤔 ). 

just hard to tell at the moment what the inputs and outputs are. Perhaps something similar to the `MessageSpec` idea you had earlier for the wallet actor interface

vtxo/states.go
```diff
@@ -0,0 +1,171 @@
...
+func (s *ForfeitingState) vtxoStateSealed() {}
+
+// ForfeitedState is a terminal state indicating the VTXO has been forfeited.
+// The VTXO's value has been transferred to a new VTXO in a new round.
+type ForfeitedState struct {
+	// VTXO is the descriptor for this VTXO.
+	VTXO *Descriptor
+
+	// NewRoundID is the round where the replacement VTXO was created.
+	NewRoundID string
```
ellemouton: forfeit req doesnt necessarily mean there is a "new round" with a replacement vtxo. could have resultsed in a leave req.

vtxo/states.go
```diff
@@ -0,0 +1,171 @@
...
+func (s *ForfeitedState) vtxoStateSealed() {}
+
+// ExpiringState is a terminal state indicating the VTXO is critically close to
+// expiry and has been sent to the chain resolver for unilateral exit handling.
+type ExpiringState struct {
+	// VTXO is the descriptor for this VTXO.
+	VTXO *Descriptor
+
+	// Reason explains why the VTXO is expiring.
+	Reason string
```
ellemouton, line 138: is there more than one reason?

vtxo/transitions.go
```diff
@@ -0,0 +1,488 @@
...
+		}, nil
+
+	default:
+		return nil, fmt.Errorf("live: unexpected event: %T", event)
+	}
+}
+
+// handleBlockEpoch processes a new block notification and checks if the VTXO
+// needs to be refreshed or escalated.
+func (s *LiveState) handleBlockEpoch(
```
ellemouton, line 50: very nice

ellemouton (approved):
looks great! The only thing i think this is missing (but i think perhaps just create an issue for it) is to:
- handle the case where the VTXO gets spent by an OOR tx
- extend the descriptor so that it can be used to support pre-confirmed vtxos (ie, vtxos that are the results of ark OOR txs that are not direct VTXT leaves)

Roasbeef:
I'm gonna hold off on this series until we get https://github.com/lightninglabs/darepo-client/pull/73 in.

vtxo/interfaces.go
Roasbeef: Makes sense, will add.

vtxo/interfaces.go
Roasbeef: That's true, I'll make note of this in the PRs that start to implement the OOR flow.

vtxo/events.go
Roasbeef: Yep this should be in the next version where we start to add the interaction with the round actor.

vtxo/events.go
Roasbeef: Hmm, yeah that's a good idea. I can add a `MessageSpec` file/section just like I did for the wallet.

vtxo/states.go
Roasbeef: Just meant as debug information, can remove for now.

vtxo/states.go
Roasbeef: Updated the comment.

Roasbeef:
@claude review this

PR #70, opened 2026-01-08
vtxo: implement VTXO actor and manager with round integration

Roasbeef:
This PR implements the VTXO actor and manager components that bridge the FSM with the actor system. The actor wraps the FSM with a message-based interface, subscribing to block epochs for expiry monitoring and routing outbox messages to their destinations. The manager tracks active VTXO actors, spawning them when VTXOs are created and handling lifecycle notifications.

The VTXO actor processes events through the FSM and dispatches resulting outbox messages:
- `ForfeitSignatureSubmission` → Round actor for batch swap signature aggregation
- `RefreshRequest` → Round actor when VTXO approaches expiry
- `VTXOStatusUpdate` → Store persistence (including `MarkForfeiting()` with forfeit tx)
- `VTXOTerminatedNotification` → Manager for cleanup
- `ExpiringNotification` → Chain resolver for on-chain fallback

The manager maintains a map of outpoint → actor references, using the actor system's service registry for direct addressing. When the round actor needs to send a forfeit request, it looks up the VTXO actor by service key rather than routing through the manager.

This PR also includes necessary round package extensions:
- `RefreshVTXORequest` extended with VTXO descriptor fields (key, script, operator, expiry)
- `VTXOCreatedNotification` extended for manager integration
- `ClientMsg` refactored to use `actormsg.RoundReceivable` marker interface

Actor-level tests verify outbox message routing, store interactions, and crash recovery via `statusToState()` with persisted forfeit transactions.

## Dependencies
- Depends on #69 (vtxo: implement FSM for VTXO lifecycle management)

PR #71, opened 2026-01-08
round: implement forfeit signature collection in FSM and actor

Roasbeef:
This PR extends the round FSM and actor to support batch swap forfeit flow, where existing VTXOs are forfeited to create new VTXOs in a commitment transaction.

The FSM gains `ForfeitSignaturesCollectingState` for coordinating signature collection from participating VTXOs. When the commitment transaction contains a connector tree and forfeit mappings, the FSM transitions to this state after receiving the commitment. Each participating VTXO actor signs its forfeit transaction and sends a `ForfeitSignatureResponse`, which the FSM collects until all expected signatures are received.

The round actor adds handlers for VTXO actor messages:
- `handleRefreshVTXORequest`: Processes refresh requests from expiring VTXOs and sends acknowledgement
- `handleForfeitSignatureResponse`: Routes signatures to the appropriate round's FSM
- `processOutbox` extensions: `ForfeitRequestToVTXO` sends forfeit requests via service key lookup, `ForfeitConfirmedToVTXO` notifies VTXOs of on-chain confirmation, `VTXOCreatedNotification` is forwarded to the VTXO manager

`CommitmentTxBuilt` event now includes connector tree structure and forfeit mappings that link each VTXO to its connector leaf. The `validateConnectorTree()` function verifies the tree structure matches the commitment transaction outputs before signing.

Comprehensive tests verify the forfeit flow transitions, connector tree validation, and integration between round and VTXO actors.

## Dependencies
- Depends on #70 (vtxo: implement VTXO actor and manager with round integration)

### DIFF
--- a/round/events.go
+++ b/round/events.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
@@ -159,9 +160,43 @@ type CommitmentTxBuilt struct {
 	// sufficient for the client to verify their VTXOs and perform
 	// unilateral exit if needed.
 	VTXOTreePaths map[int]*tree.Tree
+
+	// ForfeitMappings maps each VTXO outpoint to its connector leaf info.
+	// This allows VTXO actors to find their connector output and construct
+	// the forfeit transaction. Only set when refresh requests are present.
+	ForfeitMappings map[wire.OutPoint]*ConnectorLeafInfo
+
+	// ServerForfeitPkScript is the operator's taproot script where all
+	// forfeited VTXO values are paid. The forfeit transaction output uses
+	// this script.
+	ServerForfeitPkScript []byte
 }
 
 func (e *CommitmentTxBuilt) clientEventSealed() {}
+
+// ConnectorLeafInfo contains the information needed to construct a forfeit
+// transaction for a specific VTXO. The forfeit tx spends the VTXO and its
+// connector output, paying the VTXO value to the operator's forfeit address.
+type ConnectorLeafInfo struct {
+	// LeafIndex is the position of this connector in the connector tree.
+	LeafIndex int
+
+	// ConnectorOutpoint is the outpoint of the connector output in the
+	// commitment transaction that this forfeit tx must spend.
+	ConnectorOutpoint wire.OutPoint
+
+	// ConnectorPkScript is the scriptPubKey of the connector output.
+	ConnectorPkScript []byte
+
+	// ConnectorAmount is the value of the connector output in satoshis.
+	// Connectors typically have minimal value (dust limit).
+	ConnectorAmount int64
+
+	// VTXOAmount is the value of the VTXO being forfeited. The forfeit tx's
+	// penalty output must equal this amount. This field enables validation
+	// that prevents value theft by ensuring the correct amount is forfeited.
+	VTXOAmount btcutil.Amount
+}
 
 // CommitmentTxValidated is emitted after the client successfully validates
 // the commitment transaction and VTXT path. This is a critical security
--- a/round/events.go
+++ b/round/events.go
@@ -100,8 +100,12 @@ type BoardingUTXOConfirmed struct {
 func (e *BoardingUTXOConfirmed) clientEventSealed() {}
 
 // VTXORequestsReceived is emitted when the client submits VTXO requests that
-// should be included in the next round registration.
+// should be included in the next round registration. This event can be sent
+// from both internal sources (e.g., wallet) and external actors (e.g., VTXO
+// actor requesting a new VTXO during refresh).
 type VTXORequestsReceived struct {
+	actor.BaseMessage
+
 	// Requests are the VTXO requests to include in the next join round
 	// request.
 	Requests []types.VTXORequest
@@ -109,6 +113,14 @@ type VTXORequestsReceived struct {
 
 // clientEventSealed prevents external implementations.
 func (e *VTXORequestsReceived) clientEventSealed() {}
+
+// RoundReceivable implements actormsg.RoundReceivable marker interface.
+func (e *VTXORequestsReceived) RoundReceivable() {}
+
+// MessageType returns the message type for logging.
+func (e *VTXORequestsReceived) MessageType() string {
+	return "VTXORequestsReceived"
+}
 
 // RegistrationRequested is emitted when the FSM is ready to join a round with
 // the currently confirmed set of boarding intents. The actor should treat this
--- a/round/events.go
+++ b/round/events.go
@@ -177,11 +177,6 @@ type CommitmentTxBuilt struct {
 	// This allows VTXO actors to find their connector output and construct
 	// the forfeit transaction. Only set when refresh requests are present.
 	ForfeitMappings map[wire.OutPoint]*ConnectorLeafInfo
-
-	// ServerForfeitPkScript is the operator's taproot script where all
-	// forfeited VTXO values are paid. The forfeit transaction output uses
-	// this script.
-	ServerForfeitPkScript []byte
 }
 
 func (e *CommitmentTxBuilt) clientEventSealed() {}
--- a/round/states.go
+++ b/round/states.go
@@ -173,10 +173,6 @@ type CommitmentTxValidatedState struct {
 	// refresh rounds. Empty for boarding-only rounds. Carried forward
 	// through MuSig2 signing states until forfeit collection.
 	ForfeitMappings map[wire.OutPoint]*ConnectorLeafInfo
-
-	// ServerForfeitPkScript is the operator's forfeit address for refresh
-	// rounds. Empty for boarding-only rounds.
-	ServerForfeitPkScript []byte
 }
 
 func (s *CommitmentTxValidatedState) String() string {
@@ -227,9 +223,6 @@ type ForfeitSignaturesCollectingState struct {
 	// CollectedForfeits maps VTXO outpoints to their forfeit responses.
 	// When len(CollectedForfeits) == len(ExpectedForfeits), we proceed.
 	CollectedForfeits map[wire.OutPoint]*ForfeitSignatureResponse
-
-	// ServerForfeitPkScript is the operator's forfeit address.
-	ServerForfeitPkScript []byte
 }
 
 func (s *ForfeitSignaturesCollectingState) String() string {
@@ -273,9 +266,6 @@ type NoncesSentState struct {
 	// refresh rounds. Carried forward until forfeit collection after
 	// VTXO tree signing.
 	ForfeitMappings map[wire.OutPoint]*ConnectorLeafInfo
-
-	// ServerForfeitPkScript is the operator's forfeit address.
-	ServerForfeitPkScript []byte
 }
 
 func (s *NoncesSentState) String() string {
@@ -322,9 +312,6 @@ type NoncesAggregatedState struct {
 	// refresh rounds. Carried forward until forfeit collection after
 	// VTXO tree signing.
 	ForfeitMappings map[wire.OutPoint]*ConnectorLeafInfo
-
-	// ServerForfeitPkScript is the operator's forfeit address.
-	ServerForfeitPkScript []byte
 }
 
 func (s *NoncesAggregatedState) String() string {
@@ -368,9 +355,6 @@ type PartialSigsSentState struct {
 	// refresh rounds. After VTXO tree signature validation, if non-empty,
 	// transitions to ForfeitSignaturesCollectingState.
 	ForfeitMappings map[wire.OutPoint]*ConnectorLeafInfo
-
-	// ServerForfeitPkScript is the operator's forfeit address.
-	ServerForfeitPkScript []byte
 }
 
 func (s *PartialSigsSentState) String() string {
--- a/round/states.go
+++ b/round/states.go
@@ -47,6 +47,9 @@ func (s *Idle) clientStateSealed() {}
 // on-chain outpoint for efficient lookup when confirmation events arrive. Once
 // all intents reach the required confirmations, the FSM transitions to round
 // registration.
+//
+// This state also tracks VTXOs pending refresh. When VTXOs are approaching
+// expiry, they send RefreshVTXORequest to be included in the next round.
 type PendingRoundAssembly struct {
 	// Boarding contains the collected boarding intents to include in the
 	// next round.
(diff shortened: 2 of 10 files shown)


Roasbeef:
@claude review this 

@codex review this

Roasbeef:
@claude review this

@codex review this

/gemini review

round/events.go
```diff
@@ -159,10 +160,51 @@ type CommitmentTxBuilt struct {
 	// sufficient for the client to verify their VTXOs and perform
 	// unilateral exit if needed.
 	VTXOTreePaths map[int]*tree.Tree
+
+	// ConnectorTree is the connector tree for batch swap forfeit
+	// transactions. Each connector output in this tree corresponds to a
+	// VTXO being refreshed, and the forfeit transaction spends both the old
+	// VTXO and its connector output atomically. This field is only set when
+	// the round includes refresh requests.
+	ConnectorTree *tree.Tree
```
ellemouton: i think we dont actually send the client the whole connector tree (at the moment at least). we just send them the info about the connector that they need in order to create the forfeit tx. This is cause i think the client doesnt actually care if the connector is valid as it would only affect the server negatively if the connector is not valid.

ellemouton: also, like the `VTXOTreePaths` above, we can have multiple connector outputs in the batch. But again i think the client doesnt actually need to care about this

round/events.go
```diff
@@ -159,10 +160,51 @@ type CommitmentTxBuilt struct {
...
+
+	// ForfeitMappings maps each VTXO outpoint to its connector leaf info.
+	// This allows VTXO actors to find their connector output and construct
+	// the forfeit transaction. Only set when refresh requests are present.
+	ForfeitMappings map[wire.OutPoint]*ConnectorLeafInfo
+
+	// ServerForfeitPkScript is the operator's taproot script where all
+	// forfeited VTXO values are paid. The forfeit transaction output uses
+	// this script.
+	ServerForfeitPkScript []byte
```
ellemouton: i think currently on the server side, we assume that this has been communicated via `Terms` and is not round specific. so perhaps can just be an `Environment` variable?

round/states.go
```diff
@@ -172,6 +186,63 @@ func (s *CommitmentTxValidatedState) IsTerminal() bool {
 
 func (s *CommitmentTxValidatedState) clientStateSealed() {}
 
+// ForfeitSignaturesCollectingState indicates the client is waiting for forfeit
+// signatures from VTXO actors before proceeding with MuSig2 signing. This state
```
ellemouton: `before proceeding with MuSig2 signing` -> dont think this is correct. Forfeit signing happens after the signing of the VTXO tree which is the only place where musig signing is happening. Forfeit sigs & boarding sigs are multisig

round/states.go
```diff
@@ -172,6 +186,63 @@ func (s *CommitmentTxValidatedState) IsTerminal() bool {
...
+// is entered when the round includes refresh requests (VTXOs being rolled over
+// to the new round). The FSM waits until all expected forfeit signatures are
```
ellemouton: or leave requests (but i can add that part of the comment in my pr i guess)

round/states.go
```diff
@@ -172,6 +186,63 @@ func (s *CommitmentTxValidatedState) IsTerminal() bool {
...
+// collected, then submits them to the server and transitions to nonce
+// generation.
```
ellemouton: the nonce generation part is not correct

ellemouton: ah hmm ok i see it is just how things are done on the client side. but yeah: for the client <> server interaction, vtxo nonce&sig negotiation happens before moving on to boarding sigs and forfeit sigs flow since client is only happy to sign boarding and forfeit sigs once it has determined that it's new vtxo or leave UTXO is in the batch tx

round/transitions.go
```diff
@@ -505,6 +584,141 @@ func validateBoardingInputs(commitmentTx *wire.MsgTx,
...
+// connector leaf maps correctly to its info in ForfeitMappings. This is a
+// critical security check - without it, a malicious server could omit
+// connector outputs, breaking atomic forfeit guarantees.
+//
+// The validation ensures:
+//  1. Tree structure is internally consistent.
+//  2. Tree root spends the correct commitment tx output.
+//  3. Each leaf matches its ForfeitMappings entry (amount, script, outpoint).
+//  4. All tree transactions have valid anchor outputs.
+func validateConnectorTree(
```
ellemouton: i think the client doesnt need to care about if the connector is valid or not 🤷

round/transitions.go
```diff
@@ -505,6 +584,141 @@ func validateBoardingInputs(commitmentTx *wire.MsgTx,
...
+		// Compute the leaf transaction hash and verify it matches.
+		leafTxID, err := leaf.TXID()
+		if err != nil {
+			return fmt.Errorf("failed to compute leaf %d txid: %w",
+				connInfo.LeafIndex, err)
+		}
+
+		expectedOutpoint := wire.OutPoint{
+			Hash:  leafTxID,
+			Index: 0, // Connector is always first output.
```
ellemouton: let's reference a `lib` const for this?

round/transitions.go
```diff
@@ -284,8 +306,28 @@ func (s *PendingRoundAssembly) ProcessEvent(ctx context.Context,
 
 		return &ClientStateTransition{
 			NextState: &PendingRoundAssembly{
-				Boarding: slices.Clone(s.Boarding),
-				VTXOs:    updatedVTXOIntents,
+				Boarding:        slices.Clone(s.Boarding),
+				VTXOs:           updatedVTXOIntents,
+				RefreshingVTXOs: s.RefreshingVTXOs,
+			},
```
ellemouton, line 397: for here and above, let's use `slices.Clone` to remain consistent re how we copy things over

round/transitions.go
```diff
@@ -314,6 +351,14 @@ func (s *PendingRoundAssembly) ProcessEvent(ctx context.Context,
 			totalOutput += vtxo.Amount
 		}
 
+		// Include refresh amounts in totals. For refreshes, the
+		// forfeited VTXO provides the input and the new VTXO is the
+		// output.
+		for _, req := range s.RefreshingVTXOs {
+			totalInput += btcutil.Amount(req.Amount)
+			totalOutput += btcutil.Amount(req.Amount)
+		}
```
ellemouton: hmm wondering if this slightly conflicts with the design around boarding/vtxo creation. So this is again assuming a 1:1 (ie, give up one vtxo, get a refreshed one). but it could be that we refresh one vtxo and want to get: a new vtxo and a new leave output (or a new _set_ of vtxos) etc. So i'd think we'd instead just want this to add to `totalInput` amt and then let `VTXOReqs` and `LeaveReqs` add to totalOutput.

round/transitions.go
```diff
@@ -362,9 +407,42 @@ func (s *PendingRoundAssembly) ProcessEvent(ctx context.Context,
 		boardingReqs := fn.Map(s.Boarding, buildBoardingRequest)
 		vtxoReqs := slices.Clone(s.VTXOs)
 
+		// Build refresh requests and VTXO requests for VTXOs being
+		// refreshed. Each refresh creates a new VTXO with the same key.
+		numRefresh := len(s.RefreshingVTXOs)
+		refreshReqs := make([]*RefreshRequest, 0, numRefresh)
+		for _, req := range s.RefreshingVTXOs {
+			// Build the RefreshRequest for the server's connector
+			// tree.
```
ellemouton: related to the above, i think let's remove this.

round/transitions.go
```diff
@@ -624,16 +844,87 @@ func (s *CommitmentTxReceivedState) ProcessEvent(
...
+			}
+
+			// Transition to forfeit collection state. We'll wait
+			// for all VTXO actors to respond with their signatures
+			// before proceeding to nonce generation.
+			return &ClientStateTransition{
+				NextState: &ForfeitSignaturesCollectingState{
+					RoundID:               s.RoundID,
+					CommitmentTx:          s.CommitmentTx,
+					VTXOTreePaths:         s.VTXOTreePaths,
```
ellemouton: related to comment before, we should only be signing forfeits once we have completed the VTXO tree signing & ie, have confirmed that our new vtxos/utxos are included in the batch tx

round/transitions.go
```diff
@@ -772,6 +1064,122 @@ func (s *CommitmentTxValidatedState) ProcessEvent(
...
+			return nil, fmt.Errorf("invalid forfeit tx for VTXO "+
+				"%s: %w", evt.VTXOOutpoint, err)
+		}
+
+		// Check for duplicate response.
+		_, already := s.CollectedForfeits[evt.VTXOOutpoint]
+		if already {
+			// Already have this signature, ignore duplicate.
+			return &ClientStateTransition{NextState: s}, nil
+		}
```
ellemouton, line 970: check before the validation check above?

round/transitions.go
```diff
@@ -772,6 +1064,122 @@ func (s *CommitmentTxValidatedState) ProcessEvent(
...
+		forfeitTxs := make(map[wire.OutPoint]*wire.MsgTx)
+		forfeitedVTXOs := make([]wire.OutPoint, 0, len(updatedForfeits))
+		for outpoint, resp := range updatedForfeits {
+			forfeitSigs[outpoint] = resp.Signature
+			forfeitTxs[outpoint] = resp.ForfeitTx
+			forfeitedVTXOs = append(forfeitedVTXOs, outpoint)
+		}
+
+		// Submit forfeit signatures to server and transition to
+		// CommitmentTxValidatedState to continue with MuSig2 signing.
```
ellemouton: yeah this needs to be swapped. First musig signing, then forfeit sig sending

round/transitions.go
```diff
@@ -1252,6 +1663,37 @@ func (s *InputSigSentState) ProcessEvent(
...
+		batchExpiry := evt.BlockHeight + sweepDelay
+
+		// Build outbox messages starting with standard notifications.
+		outbox := []ClientOutMsg{
+			&VTXOCreatedNotification{
+				VTXOs:          vtxos,
+				RoundID:        s.RoundID.String(),
+				CommitmentTxID: evt.TxID,
+				BatchExpiry:    batchExpiry,
+				CreatedHeight:  evt.BlockHeight,
```
ellemouton, line 1579: should we not wait for batch tx confirmation before sending `VTXOCreatedNotification`?

ellemouton: ie, in `ConfirmedState`

round/transitions.go
```diff
@@ -1252,6 +1663,37 @@ func (s *InputSigSentState) ProcessEvent(
...
+
+		// If this round included refresh requests, notify the old VTXO
+		// actors that their forfeit is now confirmed. This allows them
+		// to transition to the terminal Forfeited state.
+		for _, vtxoOutpoint := range s.ForfeitedVTXOs {
+			outbox = append(outbox, &ForfeitConfirmedToVTXO{
+				VTXOOutpoint:   vtxoOutpoint,
+				CommitmentTxID: evt.TxID,
+				BlockHeight:    evt.BlockHeight,
+			})
```
ellemouton, line 1596: same for this: think wait till `ConfirmedState`?

ellemouton (commented):
Nice - main blocker is the ordering of states.

round/transitions.go
```diff
@@ -362,9 +407,42 @@ func (s *PendingRoundAssembly) ProcessEvent(ctx context.Context,
...
-			slog.Int("vtxo_requests", len(vtxoReqs)))
+			slog.Int("vtxo_requests", len(vtxoReqs)),
+			slog.Int("refresh_requests", len(refreshReqs)))
+
+		// Build Intents after refresh VTXOs are added so downstream
+		// validation includes all VTXO outputs.
+		intent := Intents{
+			Boarding: slices.Clone(s.Boarding),
+			VTXOs:    vtxoReqs,
+		}
```
ellemouton, line 517: the intent should include the forfeit reqs.

round/events.go
Roasbeef: Can modify this instead to just be the connector information. Good point that the server is the one that's left holding the bag if they send over invalid connector information.

round/events.go
Roasbeef: Sure, will update.

round/states.go
Roasbeef: Good point, there still needs to be lock step here, as otherwise we'd sign off the forfeit before we even know we have something in the VTXO tree. Likely would've caught this when I went to catch up the client+server. Will address this now.

round/states.go
Roasbeef: Yep, thinking we'll do leave requests after the client+server catch up.

round/states.go
Roasbeef: Yep as mentioned above, this'll happen after the vtxo nonce&sig negotiation.

round/transitions.go
Roasbeef: Hmm, yeah I think you're right. I didn't have this iniitally, but then was reviewing locally and thought it was a gap. It is the case though that the server is the one that needs to ensure they don't mess up w.r.t the connector tree, as they're the ones that need to unroll it in order to claim a forfeited UTXO.

round/transitions.go
Roasbeef: SGTM. Will make a new `lib` const.

round/transitions.go
Roasbeef: SGTM.

round/transitions.go
Roasbeef: Yeah I added this last night, I think it depends on the domain model. Will revert this then add some unit tests around a value conservation constraint here re:
  * input additions:
      * boarding inputs 
      * refreshed vtxos
   * output additions:
      * vtxo reqs 
      * leave reqs

round/transitions.go
Roasbeef: Why remove this? Don't we need to tell the server of VTXOs we want to refresh in the round? As only then do they know how to make the forfeit transaction.

round/transitions.go
Roasbeef: Yep, it's the server's funds to lose if they publish the batch for us without doing this forfeit step. Will re-order the state transitions here.

round/transitions.go
Roasbeef: Sure

round/transitions.go
Roasbeef: So we trransition to the confirmed state after the actor sends `BoadingConfirmed`. I think i'll make that a more generic name as here we're doing more than boarding, so can just be `RoundConfirmed`.

Roasbeef: Nothing actually happens in `ConfirmedState` atm.

round/transitions.go
Roasbeef: See above.

round/transitions.go
Roasbeef: Thought I added this last night, will extend `Intents` to store the refresh and update the related logic.

Roasbeef:
@claude review this

@code review this

/gemini review

Roasbeef:
@claude

> The forfeit signature collection currently happens AFTER VTXO tree signing (see ellemouton's comments on lines around transitions.go:1283), but the correct protocol flow requires forfeit signatures to be collected BEFORE proceeding to VTXO tree signing. 

This is incorrect. The correct flow does forfeit signature collection for VTXOs after signing the VTXO tree. 

> ellemouton's suggestion: Remove the one-to-one assumption. Let refresh VTXOs contribute to totalInput, and let VTXOReqs and LeaveReqs contribute to totalOutput independently.

This is done. 

> Issue: Notifying actors before the commitment tx confirms could cause race conditions if the tx doesn't confirm.

We make sure that batch is confirmed before we send out messages. The confirmed state doesn't really do anything right now. 

>  Issue: The Intents struct tracks Boarding and VTXOs, but doesn't include Refreshes. This means refresh outputs aren't validated downstream.

It does now.

litbot-9000:

PR #77, opened 2026-01-12
OOR client 1/4: package primitives + validation baseline

bhandras:
## Context
Companion client stack for OOR epic: lightninglabs/darepo#89.

This split introduces protocol primitives and validators that all later OOR FSM
and durability work depends on.

## Scope
- Add canonical OOR submit/finalize package primitives.
- Add validation rules for OOR package shape and semantics.
- Add checkpoint helper script primitives used by OOR tx construction.
- Include baseline cleanup needed by this stack.

## Included
- `lib/tx/oor` primitives + validators.
- Checkpoint helper scripts.
- Unit coverage for package validation.

## Not Included
- FSM orchestration.
- Actor runtime integration.
- Snapshot/retry behavior.

## Testing
- Unit tests for submit/finalize validators and primitives.

## Stack
- Next: #78
- Then: #79
- Then: #81

### DIFF
--- a/lib/scripts/checkpoint_oor.go
+++ b/lib/scripts/checkpoint_oor.go
@@ -1,0 +1,103 @@
+package scripts
+
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/lightningnetwork/lnd/input"
+)
+
+// CheckpointPolicy defines the parameters for constructing an OOR checkpoint
+// taproot tree.
+//
+// This is intentionally interface-first and minimal: it provides enough
+// information to deterministically derive a checkpoint output pkScript, while
+// allowing the underlying closure system to evolve later.
+type CheckpointPolicy struct {
+	// OperatorKey is the public key required by the operator-controlled CSV
+	// unroll leaf.
+	OperatorKey *btcec.PublicKey
+
+	// CSVDelay is the relative timelock enforced by the
+	// operator-controlled leaf.
+	//
+	// This is a raw BIP-68 sequence value interpreted by
+	// OP_CHECKSEQUENCEVERIFY.
+	CSVDelay uint32
+}
+
+// CheckpointTapScript constructs the tapscript for an OOR checkpoint output.
+//
+// The checkpoint tree for v0 is a simple two-leaf tree:
+//
+//   - an operator-controlled CSV unroll leaf (operator key + relative
+//     timelock),
+//   - a collaborative leaf between operator and VTXO owner (provided by the
+//     caller as raw script bytes).
+//
+// "Owner" here means the owner of the VTXO being refreshed. The checkpoint
+// output itself is still operator-controlled on the CSV timeout path.
+//
+// For v0, the checkpoint output always uses the ARK NUMS internal key so there
+// is no key-path spend and all spends go through one of the script leaves.
+//
+// This function does not validate that ownerLeafScript is "a correct Ark
+// closure". That validation belongs in higher layers once the canonical closure
+// system is in place (see the closures PRs). For now, this gives OOR primitives
+// a deterministic way to bind checkpoint scripts.
+func CheckpointTapScript(policy CheckpointPolicy,
+	ownerLeafScript []byte) (*waddrmgr.Tapscript, error) {
+
+	switch {
+	case policy.OperatorKey == nil:
+		return nil, fmt.Errorf("operator key must be provided")
+
+	case len(ownerLeafScript) == 0:
+		return nil, fmt.Errorf("owner leaf script must be provided")
+	}
+
+	unrollLeaf, err := UnilateralCSVTimeoutTapLeaf(
+		policy.OperatorKey, policy.CSVDelay,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to construct unroll leaf: %w",
+			err)
+	}
+
+	ownerLeaf := txscript.NewBaseTapLeaf(ownerLeafScript)
+
+	tapscript := input.TapscriptFullTree(&ARKNUMSKey, unrollLeaf, ownerLeaf)
+
+	// Compute and set the root hash since TapscriptFullTree doesn't
+	// populate it.
+	tree := txscript.AssembleTaprootScriptTree(tapscript.Leaves...)
+	rootHash := tree.RootNode.TapHash()
+	tapscript.RootHash = rootHash[:]
+
+	return tapscript, nil
+}
+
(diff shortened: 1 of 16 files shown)


litbot-9000:

lib/tx/oor/taptree.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+var (
+	// TapTreePSBTKey is the v0 convention for storing a taproot tree
+	// encoding in a PSBT input unknown field.
+	//
+	// We treat this as part of the OOR PSBT profile so client and server
+	// implementations can deterministically attach, validate, and later use
+	// the same metadata during finalization.
+	TapTreePSBTKey = []byte("taptree")
```
ellemouton, line 19: wondering what is a good rule for knowing when to add data to PSBT kv pairs vs when to explicitly communicate it. 

For example, for the VTX tree signing, I opted to explicitly communicate co-signers rather than embedding it in the PSBT like was done in arkade.

lib/tx/oor/submit.go
```diff
@@ -0,0 +1,168 @@
...
+//     checkpoint tx output 0 (script + value), and
+//   - each Ark PSBT input includes the `taptree` metadata needed for later
+//     finalization.
+//
+// ValidateSubmitPackage does not validate:
+//
+// - client or operator signatures,
+// - whether the checkpoint txs correctly spend live VTXOs,
+// - whether checkpoint scripts match operator policy, or
+// - VTXO set state / locking.
```
ellemouton, line 43: 👍

lib/scripts/checkpoint_oor.go
```diff
@@ -0,0 +1,114 @@
...
+// This is intentionally interface-first and minimal: it provides enough
+// information to deterministically derive a checkpoint output pkScript, while
+// allowing the underlying closure system to evolve later.
+type CheckpointPolicy struct {
+	// InternalKey is the internal taproot key used for the checkpoint
+	// output.
+	//
+	// For v0 OOR transfers, we default to the unspendable ARK NUMS key to
+	// avoid creating an unintentional key path spend.
+	InternalKey *btcec.PublicKey
```
ellemouton: i think it's always the ARKNUMs point right? ie, perhaps simpler to remove the explicit setting of it?

lib/scripts/checkpoint_oor.go
```diff
@@ -0,0 +1,114 @@
...
+	// OperatorKey is the public key required by the operator-controlled CSV
+	// unroll leaf.
+	OperatorKey *btcec.PublicKey
+
+	// CSVDelay is the relative timelock enforced by the
+	// operator-controlled leaf.
+	//
+	// This is a raw BIP-68 sequence value interpreted by
+	// OP_CHECKSEQUENCEVERIFY.
+	CSVDelay uint32
```
ellemouton, line 28: would the client key not be part of this policy?

lib/scripts/checkpoint_oor.go
```diff
@@ -0,0 +1,114 @@
...
+
+// CheckpointTapScript constructs the tapscript for an OOR checkpoint output.
+//
+// The checkpoint tree for v0 is a simple two-leaf tree:
+//
+//   - an operator-controlled CSV unroll leaf (operator key + relative
+//     timelock),
+//   - an owner-controlled collaborative leaf (provided by the caller as raw
+//     script).
+//
```
ellemouton, line 42: hmm im a bit confused here. in my mind, the checkpoint is "owned" by the operator since the operator is the party who can sweep after the CSV (so that is the operator-controlled CSV leaf path), then : the other leave is a collab multisig between operator and client. ie, it is important that that is the exact script

ellemouton (commented):
nice work! great structure. 

Main blocking questions are for the very last commit wrt checkpoint txs. But perhaps those questions will be answered as I make my way through the series

lib/scripts/checkpoint_oor.go
ellemouton: ah, i see - this is the operator provided terms basically. nvm!

lib/tx/checkpoint/build.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	checkpointPkScript, err := scripts.CheckpointPkScript(
+		policy, in.OwnerLeafScript,
+	)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
bhandras: Updated in latest push: simplified this path by deriving taproot key from tapscript once and building the output script directly (PayToTaprootScript), removing the redundant recomputation path.

lib/tx/oor/taptree.go
```diff
@@ -0,0 +1,85 @@
...
+					"tap tree value is empty",
+				)
+			}
+
+			return u.Value, nil
+		}
+	}
+
+	return nil, fmt.Errorf("tap tree not found")
+}
```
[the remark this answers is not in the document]
bhandras: Addressed: GetTapTreePSBTInput now rejects duplicate tap tree entries instead of accepting first-match.

lib/tx/oor/taptree.go
bhandras: Rule we’re using: put tx-bound signing material in PSBT; keep session/protocol control data explicit in RPC fields.

lib/scripts/checkpoint_oor.go
bhandras: Good catch, agreed. Removed InternalKey from CheckpointPolicy and we now always build checkpoint with ARK NUMS key.

lib/scripts/checkpoint_oor.go
bhandras: Yes, client key participation is represented via the owner leaf/script path used to build checkpoint policy.

lib/scripts/checkpoint_oor.go
bhandras: Clarified the docs: checkpoint has operator CSV timeout leaf + collaborative operator/owner leaf, and this helper only commits the provided script bytes into the tree. Higher layers are responsible for enforcing that those bytes are the exact expected closure script.

lib/tx/checkpoint/taptree.go
```diff
@@ -0,0 +1,228 @@
...
+				LeafVersion: txscript.TapscriptLeafVersion(
+					leafVersion,
+				),
+			}
+
+			if t, ok := parsedTypes[typeTapLeafScript]; ok {
+				if t == nil {
+					leaf.Script = script
+				}
+			}
```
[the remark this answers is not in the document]
bhandras: Fixed in split-1: tapleaf decode now sets `leaf.Script` whenever the tapleaf-script type is present, so decode no longer depends on the map entry value being nil.

ellemouton (approved):
noice!

PR #79, opened 2026-01-12
OOR client 3/4: durable actor, TLV codecs, signing context

bhandras:
## Context

Third client PR in the OOR stack. Split 2 (#78) landed the outgoing/incoming FSMs and the actor wrapper. This split makes the actor **durable**: every command, FSM state, and signing artifact survives a restart so in-flight transfers resume exactly where they left off.

### Tx primitive reorganization
`lib/tx/oor` is split into focused packages (`arktx`, `checkpoint`, `psbtutil`) and a versioned submit-package encoding is added so OOR payloads have a stable binary format for persistence and replay.

### VTXO signing context
Transfer inputs now carry full VTXO descriptors (keys, scripts, expiry) instead of raw checkpoint-only data. This context is threaded through FSM states so checkpoint signing, outbox events, and crash-resume snapshots all have access to it without a separate lookup path.

### TLV durability codecs
JSON command/snapshot serialization is replaced with typed TLV records. Start, session, restore, and drive-event payloads each get their own encode/decode path with overflow-checked numeric conversions and length-scoped leaf decoding.

### Durable actor plumbing
- Restart enqueue uses epoch timestamps so ordering is clock-independent.
- `DriveEventRequest` is persisted through the command codec.
- Session identity is bound to the proposed spend; duplicate session IDs are rejected on restore.
- Outbox events emit typed protobuf `Any` envelopes with explicit error variants.

### Hardening and docs
- TLV numeric overflow checks on decode, taptree leaf boundary scoping.
- Exported-type GoDocs across the OOR and tx helper packages.

## Client durability model

Two persistence scopes with distinct ownership:

### 1) Runtime durability (generic actor delivery)
Owned by `db/actordelivery` and `baselib/actor.DurableActor`:
- Durable inbox (`mailbox_messages`)
- Processing leases + ack/nack/redelivery
- Dedup markers (`processed_messages`)
- Ask result persistence (`ask_results`)
- Actor checkpoint row (`fsm_checkpoints`)
- Dead-letter handling (`dead_letters`)

### 2) OOR session snapshot durability (checkpoint payload)
Owned by OOR durable behavior in `oor/actor.go`:
- Per-session outgoing FSM state exported into `OutgoingSnapshot`
- Session map stored in checkpoint payload
- Restart reconstructs in-memory session handles from checkpoint data

## Persistence map

| Concern | Storage | Encoding | Written at | Read at |
|---|---|---|---|---|
| Durable actor message | `mailbox_messages.payload` | TLV | enqueue/send | durable actor receive loop |
| OOR durable command envelope | message payload bytes | TLV | `Receive` dispatch | command decode |
| Start/get/resume/export/restore payloads | command payload bytes | TLV | command construction | command decode |
| OOR session checkpoint | `fsm_checkpoints.state_data` | TLV | post-handle | actor startup/restart |
| Outgoing PSBT/session artifacts | checkpoint payload | TLV + PSBT bytes | snapshot export | snapshot restore |

## Commits

1. **tx: split arktx and checkpoint primitives**
2. **tx: add PSBT and package encodings**
3. **oor: add VTXO signing context**
4. **oor: add TLV codec for durable actor command payloads**
5. **oor: validate TLV numeric conversions on decode**
6. **oor: stabilize start-session contract and session ID binding**
7. **oor: implement outbox proto envelopes**
8. **actor: make restart enqueue independent of wall clock**
9. **oor: make DriveEventRequest durable**
10. **test: improve OOR primitive coverage**
11. **tx: harden taptree TLV decoding boundaries**
12. **oor: add GoDocs for exported types**
13. **oor: document TransferInputs FSM threading rationale**

## Not included

- Retry-backoff flow and scheduling (#81).
- Crash-window resume hardening tests (#81).

## Stack

Parent: #78 · Next: #81


litbot-9000:

lib/tx/arktx/canonical.go
```diff
@@ -0,0 +1,215 @@
+package arktx
```
ellemouton, line 1: > The existing `lib/tx/oor` package keeps its public API stable via thin
wrappers and aliases, so server/client call sites (and tests) continue
to work while we migrate incrementally.

just checking (maybe answered later): but think we dont have to keep things backwards compatible rn and can just do big refactors (i know the agents like to do this bw compat thing)

oor/actor_messages.go
```diff
@@ -27,8 +27,11 @@ type StartTransferRequest struct {
 	// Policy defines the checkpoint output tap tree policy.
 	Policy scripts.CheckpointPolicy
 
-	// Inputs are the VTXO inputs to convert into checkpoint txs.
-	Inputs []oortx.CheckpointInput
+	// Inputs are the VTXOs to transfer.
+	//
+	// Each input includes enough context for the outbox handler to request
+	// wallet signatures deterministically.
```
ellemouton: 👍

oor/states.go
```diff
@@ -116,8 +135,78 @@ type AwaitingFinalizeAccepted struct {
...
+	WaitForArkConfirmation bool
+
+	// ArkConfirmDepth is the minimum required confirmation depth when
+	// waiting for Ark tx confirmation.
+	ArkConfirmDepth uint32
+}
+
+// AwaitingArkConfirmation indicates the server accepted finalize and the
+// client is optionally waiting for the Ark tx to confirm on-chain.
+type AwaitingArkConfirmation struct {
```
ellemouton: still making way through diff but at this point not entirely sure i follow why we need this - does that mean the client is expected to unroll?

ellemouton: in my mind, stronger semantics would perhaps mean "send via batch swap"

ellemouton (commented):
nice!

oor/actor_resume_test.go
```diff
@@ -524,6 +541,104 @@ func TestOORClientActorResumeAfterServerCoSigned(t *testing.T) {
...
+			PkScript: newTestTaprootPkScript(t, clientKey.PubKey()),
+			Value:    inputValue,
+		},
+	}
+
+	store := NewInMemoryOutgoingSessionStore()
+	handler := &cosignedButDroppedHandler{
+		t:            t,
+		clientSigner: clientSigner,
+	}
```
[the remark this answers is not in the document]
bhandras: Updated: dedup is now in. Resume tests share resumeTestFixture/newResumeTestFixture, so setup boilerplate is centralized while scenarios stay readable.

lib/tx/arktx/canonical.go
bhandras: Updated: transitional OOR canonical wrappers were removed and call sites now use lib/tx/arktx directly, so there is no extra compatibility layer left here.

oor/states.go
bhandras: This is mainly for durability/unilateral-exit readiness: client needs enough artifacts to recover path construction after restart.

ellemouton (commented):
will review from snapshot/resume support stuff tomorrow. 

looks like lots of comments from previous prs in the series were actually addressed here. 
I know the agents are not very good at splitting things & knowing where to apply fixups yet - but it does make it a bit hard to review

oor/states.go
bhandras: You were right to question this — that was a design mistake on our side.

We initially introduced an `AwaitingArkConfirmation` phase (`WaitForArkConfirmation` / `ArkConfirmDepth`) while shaping restart semantics, but this does not belong in the normal collaborative OOR flow. It made the state model noisier without a clear protocol requirement, so we removed it from split 3.

Current model no longer includes that phase or fields.

Ref: https://github.com/lightninglabs/darepo-client/commit/d897a4b77ded0fb9d0daeb5a900cab3df8b89ea8

lib/tx/oor/package.go
```diff
@@ -0,0 +1,261 @@
...
+const (
+	// PackageVersionV1 is the first versioned encoding of OOR
+	// submit/finalize packages.
+	PackageVersionV1 PackageVersion = 1
+)
+
+const (
+	submitPackageVersionRecordType     tlv.Type = 1
+	submitPackageArkPSBTRecordType     tlv.Type = 3
+	submitPackageCheckpointsRecordType tlv.Type = 5
```
ellemouton: just noting that we dont do this skipping of numbers on server side

oor/checkpoint_sign.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/lightningnetwork/lnd/input"
+)
+
+// SignCheckpointPSBTs attaches the client-side collaborative VTXO spend
+// signatures to each checkpoint PSBT.
+//
+// Each checkpoint PSBT is expected to spend exactly one VTXO (input index 0).
+// The TransferInput slice is expected to match the checkpoint PSBT slice
+// 1:1.
+func SignCheckpointPSBTs(signer input.Signer, inputs []TransferInput,
```
ellemouton, line 20: can the server not also make use of this rather than re-defining it?

oor/states.go
```diff
@@ -52,6 +52,10 @@ type AwaitingSubmitAccepted struct {
 
 	// CheckpointPSBTs are the checkpoint tx PSBTs for this session.
 	CheckpointPSBTs []*psbt.Packet
+
+	// TransferInputs are the vtxo descriptors and scripts needed later on
+	// to sign the checkpoint PSBTs.
+	TransferInputs []TransferInput
```
ellemouton, line 63: does the FSM itself actually need this? isnt it only used for signing (ie at the actor level)?

oor/actor_durable_message.go
```diff
@@ -0,0 +1,819 @@
...
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/baselib/actor"
+	"github.com/lightninglabs/darepo-client/lib/scripts"
+	oortx "github.com/lightninglabs/darepo-client/lib/tx/oor"
+	"github.com/lightningnetwork/lnd/tlv"
+)
+
+const (
+	oorDurableCommandTLVType tlv.Type = 42001
```
ellemouton: reason behind the large type number?

oor/transfer_input_snapshot.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+// TransferInputSnapshot is a portable encoding of the client-side signing
+// context required to finalize checkpoint PSBTs.
+//
+// The goal is to persist enough information to reconstruct the VTXO tapscript
+// deterministically, without serializing Go wallet types directly.
+type TransferInputSnapshot struct {
+	// Outpoint is the VTXO outpoint being transferred.
+	Outpoint wire.OutPoint `json:"outpoint"`
```
ellemouton: why do we need the json tags if we are using tlv?

oor/actor.go
```diff
@@ -3,11 +3,19 @@ package oor
 import (
```
ellemouton, line 3: code here could use some docs!

oor/actor.go
```diff
@@ -46,34 +63,195 @@ func NewOORClientActor(cfg ClientActorCfg) *OORClientActor {
...
+		a.durable.Stop()
+	}
+}
+
+type oorDurableBehavior struct {
+	cfg ClientActorCfg
+
+	actorID string
+
+	sessions map[SessionID]*sessionHandle
```
ellemouton, line 171: never cleaned up i think?

oor/outbox_messages_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/lib/scripts"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
```
ellemouton, line 13: godocs missing for this commit and before

ellemouton (approved):

lib/tx/oor/package.go
bhandras: Fixed: TLV record types now use sequential numbering (1, 2, 3) matching the server-side convention.

oor/checkpoint_sign.go
bhandras: Good call. The helpers (`AddTapLeafScript`, `AddTaprootScriptSpendSig`) are now in `lib/tx/psbtutil` so both client and server can use them. The server-side dedup (replacing its local copies with imports from psbtutil) is tracked in darepo#91.

oor/states.go
bhandras: You're right — the FSM transitions don't use it. It's threaded through the states so the actor-level outbox events (which need the signing context) and checkpoint snapshots (which capture it for crash-resume) can access it without a separate lookup path. Added a comment on both states explaining this rationale.

oor/actor_durable_message.go
bhandras: Fixed: renamed to `0x7003` with a doc comment explaining the high range avoids collisions with the actor framework's reserved types (e.g. `actor.RestartTLVType`). The old `42001` was arbitrary.

oor/transfer_input_snapshot.go
bhandras: Fixed: JSON tags removed. They were leftover from an earlier iteration before the TLV encoding was in place.

oor/actor.go
bhandras: Fixed: added godoc comments to `oorDurableBehavior`, `Receive`, `handleRestart`, `restoreFromCheckpoint`, `resumeRestoredSessions`, and `persistCheckpoint`.

oor/actor.go
bhandras: Good catch — removed. `actorID` was redundant with `cfg.ActorID`. All references now go through `b.cfg.ActorID`.

oor/outbox_messages_test.go
bhandras: Fixed: added godoc comments to all test functions in this file.

PR #80, opened 2026-01-13
round: bug fixes that came up during e2e testing

Roasbeef:
In this PR, we fix a series of bugs that came up during end to end testing.

### DIFF
--- a/round/actor.go
+++ b/round/actor.go
@@ -313,23 +313,6 @@ func (a *RoundClientActor) createNewRound(ctx context.Context) (*RoundFSM, error
 	return roundFSM, nil
 }
 
-// findIdleRound finds an existing round FSM that is in Idle state and can
-// accept new boarding intents. Returns nil if no idle round exists.
-func (a *RoundClientActor) findIdleRound() *RoundFSM {
-	for _, roundFSM := range a.rounds {
-		state, err := roundFSM.FSM.CurrentState()
-		if err != nil {
-			continue
-		}
-
-		if _, ok := state.(*Idle); ok {
-			return roundFSM
-		}
-	}
-
-	return nil
-}
-
 // findAssemblingRound finds a round that is currently assembling intents.
 // It prioritizes PendingRoundAssembly (which already has boarding inputs)
 // over Idle rounds. This ensures VTXOs are attached to rounds that have
--- a/round/actor.go
+++ b/round/actor.go
@@ -213,16 +213,18 @@ func (a *RoundClientActor) createRoundFSM(ctx context.Context,
 	)
 
 	return &RoundFSM{
-		FSM:     &fsm,
-		RoundID: round.RoundID,
-		TxID:    txid,
+		FSM:          &fsm,
+		RoundID:      round.RoundID,
+		TxID:         txid,
+		CommitmentTx: round.CommitmentTx,
 	}, nil
 }
 
 // registerCommitmentConfirmation registers for confirmation monitoring of a
-// commitment transaction with the chainsource actor.
+// commitment transaction with the chainsource actor. The commitmentTx is used
+// to extract the pkScript for LND's confirmation tracking.
 func (a *RoundClientActor) registerCommitmentConfirmation(ctx context.Context,
-	txid chainhash.Hash) {
+	txid chainhash.Hash, commitmentTx fn.Option[*psbt.Packet]) {
 
 	callerID := fmt.Sprintf("commitment-tx-%s", txid.String())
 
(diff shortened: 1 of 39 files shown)


Roasbeef:
@claude review this

@codex review this

round/actor.go
```diff
@@ -238,14 +240,46 @@ func (a *RoundClientActor) registerCommitmentConfirmation(ctx context.Context,
...
+	})
+
+	// Query ChainSource for current block height to use as HeightHint.
+	// LND requires HeightHint > 0 for confirmation scanning.
+	var heightHint uint32
+	heightFuture := a.cfg.ChainSource.Ask(
+		ctx, &chainsource.BestHeightRequest{},
+	)
+	heightResult := heightFuture.Await(ctx)
+	heightResp, err := heightResult.Unpack()
```
ellemouton, line 441: think i had same question on server side: 
doesnt this result in only blocks from current height being checked for confirmation? what if the tx is already confirmed? ie, wont this result in us potentially missing that?

round/actor.go
```diff
@@ -503,6 +503,76 @@ func (a *RoundClientActor) migrateRoundToActiveFSM(ctx context.Context,
...
+		CommitmentTx: fn.Some(inputSigState.CommitmentTx),
+	}
+	a.activeRounds[roundID] = roundFSM
+	a.commitmentTxIndex[txid] = roundID
+
+	// Register for commitment tx confirmation.
+	a.registerCommitmentConfirmation(ctx, txid, roundFSM.CommitmentTx)
+
+	// Create a NEW primaryFSM starting at Idle. This allows the actor to
+	// immediately accept new boarding intents for subsequent rounds.
```
ellemouton: why only after InputSigSentState? can we not do this immediately after the round has been joined? otherwise there might be quite a gap between when the current active/primary round gets going and when the client can make a request to join a new round

round/actor.go
```diff
@@ -879,12 +921,14 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
...
 				slog.String("round_id", m.RoundID.String()))
 
-			// Primary FSM checkpointed a round. Migrate to
-			// dedicated FSM.
-			err := a.migrateRoundToActiveFSM(ctx, m.RoundID)
+			// Promote the current primaryFSM to activeRounds. It's
+			// already in InputSigSentState and will handle the
+			// BoardingConfirmed event when it arrives.
+			err := a.promotePrimaryFSMToActiveRounds(ctx, m.RoundID)
 			if err != nil {
```
ellemouton: should be in previous commit

round/actor.go
```diff
@@ -844,21 +850,57 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
...
+			if heightHint == 0 {
+				heightFuture := a.cfg.ChainSource.Ask(
+					ctx, &chainsource.BestHeightRequest{},
+				)
+				heightResult := heightFuture.Await(ctx)
+				heightResp, err := heightResult.Unpack()
+				if err != nil {
+					return fmt.Errorf("get best height "+
+						"for confirmation: %w", err)
+				}
```
ellemouton, line 1146: wondering if we should employ the use of interfaces to hide/wrap this `Ask`->`Await`->`Unpack`->type cast pattern?

round/actor.go
```diff
@@ -844,21 +850,57 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
 				},
 			)
 
+			// Query ChainSource for current block height to use as
+			// HeightHint. LND requires HeightHint > 0 for
+			// confirmation scanning.
+			heightHint := m.HeightHint
```
ellemouton, line 1136: should it even be a field on the `RegisterConfirmationRequest` struct if the FSM doesnt have the ability to fill out the field itself? i think it is currently not ever set

round/actor.go
```diff
@@ -844,21 +850,57 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
...
+			// processing completes.
+			a.cfg.ChainSource.Tell(context.Background(), confReq)
 
 		case *VTXOCreatedNotification:
-			_ = m.VTXOs
+			// Forward to VTXO manager to spawn actors for the new
+			// VTXOs if configured.
+			if a.cfg.VTXOManager != nil {
+				a.cfg.VTXOManager.Tell(ctx, m)
+			}
```
ellemouton, line 1183: i think let's always set it?

ellemouton (approved):
lgtm! very nice

round/actor.go
Roasbeef: Hmm yeah you're right about that. I'll extend the itests to take down the round actor, mine N blocks, then restart to make sure it picks up the confirmation. Right now I just do a normal restart. 

I think what we actually need to be tracking here is: the height that the round aggregation started at. Then we avoid this query all together, as we just do it at the very start when we go to initiate a new round.

round/actor.go
Roasbeef: So there was a bug here w.r.t a round client being able to join subsequent rounds back to back. Before this logic would make a _new_ FSM to store in the pending map, but then leave the existing FSM in the "inputs sig sent state". Then when I attempted to trigger a new round, it would fail as the message was unexpected in that state. 

The new logic makes a new primary FSM after we cross the point of no return that starts in a blank slate, so it can then handle any additional requests while the prior round waits for confirmation. 

> otherwise there might be quite a gap between when the current active/primary round gets going and when the client can make a request to join a new round

Can you expand on this? 

One limitation rn is that you can't actually make a _new_ round while the other one is ongoing. I think a better way to handle this would be to sort of drop the primary FSM approach, and just store the FSMs on a single level keyed by round ID or temp round ID (we make before we know the true round ID). This way we can support N concurrent rounds that have been started at the same time.

round/actor.go
Roasbeef: Sure, I like that, this does indeed clutter the call sites a bit.

round/actor.go
Roasbeef: Hmm, one route here would be to set the current height when the FSM is created in the first place. This height hint is mainly for neutrino nodes and nodes running w/o a txindex. So I think we can add the current/starting height directly to the env for the FSM.

round/actor.go
Roasbeef: So `VTXOManager` actually does something in the future PR series we're working through rn. I added this so in the e2e test I'd have an event driven way to know when the round was finished (this vtxo notification sent).

round/actor.go
ellemouton: > Can you expand on this?

what i mean is: 
1) we send a JoinReq with a set of boarding/vtxo/forfeit/leave reqs and get a `Success` response but then it may take a while to reach `InputSigSentState` since the server will: 
      a) keep the registration phase open for a period of time and then 
      b) it will take time to do the nonce/sig exchange with the various clients
during this time,  the client may want to  start collecting _new_ boarding/vtxo/forfeit/leave reqs from the user and send a new JoinReq that may be part of a new concurrent round. 

With the current impl, the user would need to wait for the (a)(b) steps above to complete before it can interact with a new concurrent round, right? 

Im then saying that i think we can let them interact with concurrent rounds before the primary round reaches `InputSigSentState`. 

So: i think `primaryFSM` should only represent the fsm where we are still collecting things from the user that we want to send in our `JoinReq`. Once we've sent the JoinReq & received a Success response, then we can move the FSM over to the `active map` (which we can do now since from this point on, any messages from server for the round will include the round ID) & create a new `primary FSM` so that we can immediately continue accepting new reqs from the user

Roasbeef: > With the current impl, the user would need to wait for the (a)(b) steps above to complete before it can interact with a new concurrent round, right?

Yup, for the initial version of this diff, that was correct. 

The latest version now has a flat map and is able to support multiple concurrent rounds being created. If a user VTXO request arrives once a round is on going, then we make a new round and send over the request to the server.

Roasbeef: It's also a bit simpler as we don't need to do any rotation at all.

PR #81, opened 2026-01-13
OOR client 4/4: retry-backoff, outbox error handling, crash-resume tests

bhandras:
## Context

Final client PR in the OOR stack. Split 3 (#79) made the actor durable with TLV codecs and signing context. Split 3b (#119) added operator signature validation. This split adds **retry-backoff runtime behavior** so transient outbox failures don't drop sessions, and adds crash-resume test coverage for the co-sign window.

### Retry-backoff FSM state
`RetryBackoff` is introduced as an explicit outgoing FSM state that stores a resume snapshot, retry delay, and failure reason. Snapshot restore and resume logic can represent retry intent without losing session progress after a restart.

### Retry transitions and scheduling
Retry progression is driven through the durable actor boundary with explicit `RetryDueEvent` and `ScheduleRetryRequest` messages. `OutboxForState` emits the expected retry scheduling request for restored sessions so restart replay stays consistent with the normal runtime path.

### Outbox error handling
Retryable outbox failures are routed through retry-backoff states and durable events. Non-retryable errors terminate the session. This classification happens at the actor level so the FSM stays pure.

### Crash-resume and edge-case tests
- Actor resume test simulating a crash after server co-sign but before the client observes submit acceptance. Restores from persisted state, re-drives the outbox, and asserts stable package behavior through completion.
- Targeted unit tests for DriveEvent handling, late/terminal event behavior, and retryable vs non-retryable outbox error transitions.

## Storage model

This split extends split 3's replay behavior — it does not add a new durability backend.

| Concern | Storage/Boundary | Trigger |
|---|---|---|
| Retry scheduling intent | `ScheduleRetryRequest` outbox event | retryable outbox error |
| Crash/restart replay | durable mailbox + TLV checkpoint | actor restart |

The retry delay and resume snapshot are captured inside the `RetryBackoff` FSM state and persisted through the existing `OutgoingSnapshot` → checkpoint TLV path from split 3.

## Commits

1. **oor: add retry-backoff state for snapshot restore path**
2. **oor: wire retry-backoff transitions for outgoing snapshots**
3. **oor: make retry-backoff resumable**
4. **oor: add retry-backoff flow and outbox error handling**
5. **oor: add crash-resume test for co-sign**
6. **test: cover OOR actor/FSM edge cases**
7. **oor: reconcile split4 with split3 validation base**

## Not included

- Full transport RPC integration beyond typed envelopes.
- System-level restart/retry integration tests (unit/in-process focus here).

## Stack

Parent: #119 · Final split in the client OOR stack.

### DIFF
--- a/oor/transitions.go
+++ b/oor/transitions.go
@@ -320,6 +320,32 @@ func (s *RetryBackoff) ProcessEvent(ctx context.Context,
 	_ = env
 
 	switch evt := event.(type) {
+	case *RetryDueEvent:
+		_ = evt
+
+		if s.ResumeSnapshot == nil {
+			return nil, fmt.Errorf(
+				"resume snapshot must be provided",
+			)
+		}
+
+		nextState, err := OutgoingStateFromSnapshot(s.ResumeSnapshot)
+		if err != nil {
+			return nil, err
+		}
+
+		nextOutbox, err := OutboxForState(nextState)
+		if err != nil {
+			return nil, err
+		}
+
+		return &StateTransition{
+			NextState: nextState,
+			NewEvents: fn.Some(EmittedEvent{
+				Outbox: nextOutbox,
+			}),
+		}, nil
+
 	case *FailEvent:
 		return &StateTransition{
 			NextState: &Failed{Reason: evt.Reason},
--- a/oor/transitions.go
+++ b/oor/transitions.go
@@ -213,7 +213,7 @@ func (s *AwaitingCheckpointSignatures) ProcessEvent(ctx context.Context,
 		}
 
 		// Validate finalize package before emitting request.
-		err = oortx.ValidateFinalizePackageSigned(
+		err = oortx.ValidateFinalizePackage(
 			s.ArkPSBT, evt.FinalCheckpointPSBTs,
 		)
 		if err != nil {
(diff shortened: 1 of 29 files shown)


ellemouton (approved):

oor/transitions.go
```diff
@@ -303,6 +442,53 @@ func (s *Failed) ProcessEvent(ctx context.Context, event Event,
...
+		return nil, fmt.Errorf("internal: missing session id")
+	}
+
+	snap, err := NewOutgoingSnapshot(env.SessionID, current)
+	if err != nil {
+		return nil, err
+	}
+
+	after := evt.RetryAfter
+	if after == 0 {
```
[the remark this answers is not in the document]
bhandras: Addressed: switched to named constant (defaultRetryDelay) instead of inline literal.

litbot-9000:
@ellemouton: review reminder

oor/actor_drive_event_retry_due_test.go
```diff
@@ -0,0 +1,30 @@
+package oor
```
ellemouton: btw - maybe tell the agent not to put all testst into their own file?

also - godocs for tests 🙏

oor/retry_backoff_test.go
```diff
@@ -0,0 +1,85 @@
+package oor
```
ellemouton: same here

ellemouton (approved):

oor/actor_drive_event_retry_due_test.go
bhandras: Good call — cleaned this up.

I moved the retry-due drive-event roundtrip test into `actor_durable_message_test.go` (next to the other durable command roundtrip coverage) and added godoc comments for the test functions there.

oor/retry_backoff_test.go
bhandras: Addressed this one as well.

I moved the retry-backoff tests out of `retry_backoff_test.go` into the existing `actor_drive_event_integration_test.go` coverage and added godoc comments on the moved tests.

oor/actor_drive_event_retry_due_test.go
bhandras: done

oor/retry_backoff_test.go
bhandras: done

PR #84, opened 2026-01-14
multi: decouple vtxo requests from intents

ellemouton:
This commit fundamentally changes how VTXO requests relate to boarding intents, enabling flexible M:N mappings instead of the previous 1:N constraint.

The schema locked each VTXO request to a specific boarding intent via a foreign key constraint:

    round_vtxo_templates (
        round_id, outpoint_hash, outpoint_index, template_index,
        FOREIGN KEY (round_id, outpoint_hash, outpoint_index)
            REFERENCES round_boarding_intents
    )

This enforced a rigid 1:N relationship where one boarding intent could produce multiple VTXOs, but multiple boarding intents could NOT be combined to produce a single VTXO.

VTXO requests are now first-class citizens of the round, tied directly to the round rather than to individual intents:

    round_vtxo_requests (
        round_id, request_index,
        FOREIGN KEY (round_id) REFERENCES rounds
    )

This enables flexible M:N mappings:
- Fan-in: Multiple boarding intents can contribute UTXOs to fund a single VTXO output
- Fan-out: One boarding intent can still produce multiple VTXOs
- Any combination: The round FSM can now optimize tree construction by grouping inputs and outputs independently

The decoupling allows the round FSM to collect all inputs and all requested outputs separately, then pair them optimally during tree construction rather than being constrained by the intent-request relationship.

Fixes https://github.com/lightninglabs/darepo-client/issues/59

### DIFF
--- a/db/sqlc/migrations/000003_round_tables.up.sql
+++ b/db/sqlc/migrations/000003_round_tables.up.sql
@@ -117,17 +117,14 @@ CREATE TABLE IF NOT EXISTS round_boarding_intents (
 CREATE INDEX IF NOT EXISTS idx_round_boarding_intents_round_id
     ON round_boarding_intents(round_id);
 
--- Round VTXO templates table.
--- Stores requested VTXOs for each boarding intent.
--- One boarding intent can request multiple VTXOs (1:N relationship).
-CREATE TABLE IF NOT EXISTS round_vtxo_templates (
-    -- Composite FK to round_boarding_intents.
+-- Round VTXO requests table.
+-- Stores VTXO requests for the round.
+CREATE TABLE IF NOT EXISTS round_vtxo_requests (
+    -- round_id links to the parent round.
     round_id TEXT NOT NULL,
-    outpoint_hash BLOB NOT NULL,
-    outpoint_index INTEGER NOT NULL,
 
-    -- template_index is the order of this VTXO within the request.
-    template_index INTEGER NOT NULL,
+    -- request_index preserves request order.
+    request_index INTEGER NOT NULL,
 
     -- VTXORequest.Amount - value in satoshis.
     amount BIGINT NOT NULL,
@@ -153,10 +150,8 @@ CREATE TABLE IF NOT EXISTS round_vtxo_templates (
     -- VTXORequest.SigningKey.PubKey - 33-byte compressed public key.
     signing_pubkey BLOB NOT NULL,
 
-    PRIMARY KEY (round_id, outpoint_hash, outpoint_index, template_index),
-    FOREIGN KEY (round_id, outpoint_hash, outpoint_index)
-        REFERENCES round_boarding_intents(round_id, outpoint_hash, outpoint_index)
-        ON DELETE CASCADE
+    PRIMARY KEY (round_id, request_index),
+    FOREIGN KEY (round_id) REFERENCES rounds(round_id) ON DELETE CASCADE
 );
 
 -- Client trees table.
--- a/round/actor.go
+++ b/round/actor.go
@@ -313,23 +313,6 @@ func (a *RoundClientActor) createNewRound(ctx context.Context) (*RoundFSM, error
 	return roundFSM, nil
 }
 
-// findIdleRound finds an existing round FSM that is in Idle state and can
-// accept new boarding intents. Returns nil if no idle round exists.
-func (a *RoundClientActor) findIdleRound() *RoundFSM {
-	for _, roundFSM := range a.rounds {
-		state, err := roundFSM.FSM.CurrentState()
-		if err != nil {
-			continue
-		}
-
-		if _, ok := state.(*Idle); ok {
-			return roundFSM
-		}
-	}
-
-	return nil
-}
-
 // findAssemblingRound finds a round that is currently assembling intents.
 // It prioritizes PendingRoundAssembly (which already has boarding inputs)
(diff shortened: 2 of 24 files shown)


Roasbeef:
Will follow this up with an extensino to the server side tests to confirm multiple boarding inputs before starting a round.

db/sqlc/migrations/000003_round_tables.up.sql
```diff
@@ -149,10 +146,8 @@ CREATE TABLE IF NOT EXISTS round_vtxo_templates (
     -- VTXORequest.SigningKey.PubKey - 33-byte compressed public key.
     signing_pubkey BLOB NOT NULL,
 
-    PRIMARY KEY (round_id, outpoint_hash, outpoint_index, template_index),
-    FOREIGN KEY (round_id, outpoint_hash, outpoint_index)
-        REFERENCES round_boarding_intents(round_id, outpoint_hash, outpoint_index)
-        ON DELETE CASCADE
+    PRIMARY KEY (round_id, request_index),
+    FOREIGN KEY (round_id) REFERENCES rounds(round_id) ON DELETE CASCADE
```
Roasbeef, line 154: 👍

round/actor.go
```diff
@@ -418,6 +421,9 @@ func (a *RoundClientActor) Receive(ctx context.Context,
 	case *WalletBoardingConfirmed:
 		return a.handleWalletBoardingConfirmed(ctx, m)
 
+	case *RegisterVTXORequestsRequest:
```
Roasbeef, line 591: What do we expect to send in this new message? Presumably the wallet when it decides to board new inputs?

round/actor_messages.go
```diff
@@ -155,12 +156,46 @@ type RegisterBoardingIntentResponse struct {
...
 }
 
+// clientRespSealed marks this as a client response message.
 func (m *RegisterBoardingIntentResponse) clientRespSealed() {}
 
+// RegisterVTXORequestsRequest informs the FSM of VTXO request amounts to
+// include in the next round registration.
+type RegisterVTXORequestsRequest struct {
+	actor.BaseMessage
+
```
Roasbeef, line 148: May not be needed _now_, but in the future it may be usful to be able to also specify the outpoints here (a la coin selection).

round/states.go
```diff
@@ -42,15 +42,16 @@ func (s *Idle) IsTerminal() bool {
...
+// VTXO requests are ready, the FSM transitions to round registration.
 type PendingRoundAssembly struct {
-	// Intents maps outpoint to boarding intent. Only intents with on-chain
-	// UTXOs (i.e., ChainInfo.OutPoint set) are included in this map.
-	Intents map[wire.OutPoint]BoardingIntent
+	// Intents are the confirmed boarding intents for the next round.
+	Intents []BoardingIntent
+
+	// VTXORequests are the pending VTXO requests for the next round.
+	VTXORequests []types.VTXORequest
```
Roasbeef: 👍

Roasbeef (commented):
The changes read well. Can we split up the commit a bit for the posterity of the git history? Minimal one commit per local pacakge/file boundary.

round/actor_messages.go
ellemouton: can you expand?

This represents the user saying "i want a vtxo of size x". So at this point, there is no outpoint. 

or are you saying that we should be able to specify vtxo outpoints to forfeit in order to construct the new vtxos with the given amounts?  

If so: i think perhaps we can keep the abstraction & separation here and rather: expand to include `RegisterForfeitReqs` which specifies outpoints to spend along with `RegisterLeaveReqs` which specifies wanted on-chain UTXO scripts. 

Then we group them all together into a JoinReq that makes sense.

round/actor.go
ellemouton: i think we want another layer that decides on grouping etc. This layer will also handle automatic refreshes etc. 
Currently we have that a user can request a boarding utxo but right now that gets automatically converted into a single vtxo of the same amount. I think we should have a separate actor that handles deciding how to group things/split things/fan-in/fan-out. Yes - perhaps the wallet is the place that would manage this 👍  

Some examples of the logic this actor would manage:
- User requests a leave req with amount x: system will look at existing vtxos, gather the oldest ones until total leave amount is reached and then send the RoundActor ([]ForfeitReq, []LeaveReq). 
- Automatic tracking of VTXOs to ensure that batch swaps occur within good time. Sends RoundActor([]Forfeit, []VTXO)
- etc

Roasbeef: >  I think we should have a separate actor that handles deciding how to group things/split things/fan-in/fan-out

SGTM. Will merge this, then update the systest to cover this case, and also exercise that both sides can support concurrent on going rounds.

Roasbeef:
Will rebase this, then update the tests.

Roasbeef:
@claude review this

@codex review this

Roasbeef:
@claude

`slice.Clone` can handle `nil` entries: 
```
// Clone returns a copy of the slice.
// The elements are copied using assignment, so this is a shallow clone.
// The result may have additional unused capacity.
// The result preserves the nilness of s.
func Clone[S ~[]E, E any](s S) S {
	// Preserve nilness in case it matters.
	if s == nil {
		return nil
	}
	// Avoid s[:0:0] as it leads to unwanted liveness when cloning a
	// zero-length slice of a large array; see https://go.dev/issue/68488.
	return append(S{}, s...)
}
```

PR #86, opened 2026-01-16
MailboxRPC 2/7: contract + codegen + make rpc

bhandras:
## Summary
This PR adds initial scaffolding for an RPC-over-mailbox transport layer:

- `mailboxrpc`: a small runtime interface package for sending/awaiting RPCs and registering handlers.
- `mailboxpb`: the public protobuf contract for the mailbox edge transport.
- `protoc-gen-mailboxrpc`: a repo-local protoc plugin that generates typed RPC-over-mailbox client stubs and server registration helpers.
- `make rpc` wiring to generate mailbox RPC stubs alongside existing protos.

## Why
We want a durable, retry-safe RPC layer that can be transported over a mailbox edge with at-least-once delivery semantics and cursor-based acking. This is part of the durability work tracked in lightninglabs/darepo-client#48 and the mailbox/RPC spec tracked in lightninglabs/darepo#71.

## Notes
- Generated `*.pb.go` and `*_mailboxrpc.pb.go` files are checked in.
- Contributors can regenerate with `make rpc` (Docker-based tooling).

## Follow-ups
- Implement a concrete `mailboxrpc.RPCClient` and a router backed by the mailbox transport.
- Add a conformance test harness once the runtime implementation exists.

PR #89, opened 2026-01-16
MailboxRPC 5/7: edge-backed mailboxrpc client

bhandras:
Adds `mailboxclient.Client`, a concrete `mailboxrpc.RPCClient` implementation
backed by the mailbox edge gRPC API (`mailboxpb.MailboxService`: Send/Pull/AckUpTo).

Key behaviors:
- Encodes requests into `mailboxpb.Envelope` with `google.protobuf.Any` payloads.
- Long-polls `Pull` for responses.
- Caches pulled responses by correlation id before advancing the ack cursor
  (prevents response loss with concurrent in-flight RPCs).

Stack order:
- Base: #48
- Depends on: #88 (router), #86 (contract/codegen)

### DIFF
--- a/mailbox/client/client.go
+++ b/mailbox/client/client.go
@@ -1,0 +1,419 @@
+package mailboxclient
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"log/slog"
+	"sync"
+	"time"
+
+	mailboxpb "github.com/lightninglabs/darepo-client/mailbox/pb"
+	mailboxrpc "github.com/lightninglabs/darepo-client/mailbox/rpc"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+// Client implements mailboxrpc.RPCClient by sending and receiving mailbox
+// envelopes through a mailboxpb.MailboxServiceClient.
+type Client struct {
+	cfg Config
+
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+
+	mu sync.Mutex
+
+	cursor uint64
+
+	pending map[string][]byte
+	waiters map[string][]chan struct{}
+}
+
+// New constructs and starts a mailboxclient.Client.
+func New(cfg Config) (*Client, error) {
+	cfg = applyDefaults(cfg)
+
+	if cfg.Edge == nil {
+		return nil, fmt.Errorf("edge is required")
+	}
+	if cfg.LocalMailboxID == "" {
+		return nil, fmt.Errorf("local mailbox id is required")
+	}
+	if cfg.RemoteMailboxID == "" {
+		return nil, fmt.Errorf("remote mailbox id is required")
+	}
+	if cfg.PullMaxEnvelopes == 0 {
+		return nil, fmt.Errorf("pull max envelopes must be > 0")
+	}
+	if cfg.PullWaitTimeout <= 0 {
+		return nil, fmt.Errorf("pull wait timeout must be > 0")
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	c := &Client{
+		cfg: cfg,
+
+		cancel: cancel,
+
+		pending: make(map[string][]byte),
+		waiters: make(map[string][]chan struct{}),
+	}
+
+	c.wg.Add(1)
+	go func() {
+		defer c.wg.Done()
+		c.run(ctx)
+	}()
+
+	log.InfoS(ctx, "Mailbox client started",
+		slog.String("local_mailbox", cfg.LocalMailboxID),
+		slog.String("remote_mailbox", cfg.RemoteMailboxID))
+
+	return c, nil
+}
+
+// Stop shuts down background polling and unblocks any waiters.
+func (c *Client) Stop() {
+	log.InfoS(context.TODO(), "Stopping mailbox client")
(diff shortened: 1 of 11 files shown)


mailbox/client/client.go
```diff
@@ -0,0 +1,365 @@
...
+
+		for _, env := range resp.Envelopes {
+			c.handleEnvelope(env)
+		}
+
+		if resp.NextCursor > cursor {
+			ackOK := c.ackUpTo(ctx, resp.NextCursor)
+			if ackOK {
+				c.storeCursor(resp.NextCursor)
+			}
```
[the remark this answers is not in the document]
bhandras: Agreed.

Updated the run loop to apply the same backoff on `AckUpTo` failures as we do for transient `Pull` failures (prevents a tight loop if Pull succeeds but Ack is failing persistently).

bhandras:
@codex review pls

litbot-9000:

mailbox/client/client.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+
+		delete(c.waiters, correlationID)
+	}
+}
+
+// SendRPC sends a request payload and returns its correlation id and
+// idempotency key.
+func (c *Client) SendRPC(ctx context.Context, service string, method string,
+	req proto.Message, opts mailboxrpc.RPCOptions) (string, string, error) {
```
Roasbeef: Should either use type aliases for the output here, or return a new struct.

mailbox/client/client.go
```diff
@@ -0,0 +1,368 @@
...
+
+	c.wg.Add(1)
+	go func() {
+		defer c.wg.Done()
+		c.run(ctx)
+	}()
+
+	return c, nil
+}
+
```
Roasbeef, line 77: Generally we lack logging in this pacakge so far.

mailbox/client/client_test.go
```diff
@@ -0,0 +1,280 @@
...
+			return err
+		}
+		if pull.Status == nil || !pull.Status.Ok {
+			return fmt.Errorf("operator pull failed")
+		}
+
+		for _, env := range pull.Envelopes {
+			if err := handleOperatorEnvelope(
+				ctx,
+				edge,
```
Roasbeef: Can pack more args onto a given line here. Repeated elsewhere in teh diff.

Roasbeef:
@claude review this

mailbox/client/client.go
Roasbeef: Same the `string` for service+method can instead be like some composite struct even

mailbox/client/client.go
```diff
@@ -0,0 +1,368 @@
...
+	if !statusOK(resp.Status) {
+		return "", "", statusError("Send", resp.Status)
+	}
+
+	return correlationID, idempotencyKey, nil
+}
+
+// AwaitRPC blocks until a response for correlationID is received.
+func (c *Client) AwaitRPC(ctx context.Context, correlationID string,
```
Roasbeef, line 178: One thing I'm wondering here (need to check out the other PRs in the series is): is there not a single unified call that the client can use to get (90 + 91) _all_ the current pending messages for it? That would simplify some of the general client logic as well IIUC. 

As then we have an actor on the client side that'll continuously read messages, deliver them to the destination (who may be durable actors), then ACK that the messages were delivered. 

Still getting through the series.

Roasbeef: As then this hooks up into serverconn as a concrete implementation.

Roasbeef:
@claude review this

mailbox/client/client.go
bhandras: codex: Yes, this is exactly the direction of the rewrite.

I’m consolidating onto a single `serverconn` connector loop that continuously pulls pending messages, durably dispatches to destination actors, and only then advances `AckUpTo`.

That gives us one unified receive path for both unary-response traffic and passive server->client events, with durable actor semantics as the source of truth.

mailbox/client/client.go
bhandras: codex: Agreed on the API shape concern.

As part of the rewrite, I’ll update the unary facade signatures to use explicit request/result structures (instead of loosely-typed string tuples) so service/method/correlation semantics are encoded in types and align with the connector boundary.

mailbox/client/client.go
bhandras: codex: Agreed. I’ll carry forward structured logging across the rewrite so the new `serverconn` pull/dispatch/ack lifecycle has first-class observability (pull failures, dispatch outcomes, ack retries, cursor/ack watermarks).

mailbox/client/client_test.go
bhandras: codex: Acked. I’ll clean up the test formatting/style pass in the rewrite branch so the new tests follow the expected compact call formatting and readability conventions.

bhandras:
codex: Quick update on direction after reviewing the full series and architecture comments:

I’m consolidating onto a `serverconn` connector runtime as the unified receive/send path, backed by durable actor semantics from `#48`.

Concretely:
- `Pull`/`AckUpTo` remains transport,
- connector does durable dispatch first and ack second,
- unary RPC facade remains for caller ergonomics,
- mailbox-specific SQL response storage is being removed to avoid parallel durability paths.

I’ll follow up with rewrite commits wired around this boundary.

PR #90, opened 2026-01-16
MailboxRPC 6/7: sqlc-backed durable response storage

bhandras:
Adds an injectable `mailboxclient.Store` used by `mailboxclient.Client` to
persist:
- Pull cursor state, and
- response payloads keyed by correlation id.

This PR uses SQLC (project-standard) for persistence (`mailboxclient.SQLCStore`)
and includes a restart test proving responses survive client restarts with
cursor-based AckUpTo semantics.

Stack order:
- Base: #48
- Depends on: #89 (edge-backed client), #86 (contract/codegen)


mailbox/client/client.go
```diff
@@ -211,14 +238,23 @@ func (c *Client) run(ctx context.Context) {
...
+			continue
 		}
 
 		if resp.NextCursor > cursor {
 			ackOK := c.ackUpTo(ctx, resp.NextCursor)
 			if ackOK {
-				c.storeCursor(resp.NextCursor)
+				c.storeCursor(ctx, resp.NextCursor)
 			}
 		}
```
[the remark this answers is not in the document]
bhandras: Agreed — ordering matters here.

Reworked the client cursor/ack handling so we persist the cursor first and then drive `AckUpTo` based on the persisted cursor. We also keep an in-memory `ackTo` watermark to ensure we retry acks (with backoff) until they succeed, including across restarts.

mailboxrpcclient/bolt_store.go
```diff
@@ -0,0 +1,250 @@
...
+
+		raw := cursors.Get([]byte(mailboxID))
+		if raw == nil {
+			return nil
+		}
+
+		decoded, ok := decodeUint64(raw)
+		if !ok {
+			return nil
+		}
```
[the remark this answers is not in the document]
bhandras: Acked.

This comment refers to the older BoltDB-backed store (`mailboxrpcclient/bolt_store.go`). The branch now uses the SQLC-backed `mailboxclient.SQLCStore` and the Bolt store file is no longer present, so the specific failure mode described here does not apply.

mailbox/client/client.go
```diff
@@ -37,6 +37,9 @@ func New(cfg Config) (*Client, error) {
 	if cfg.Edge == nil {
 		return nil, fmt.Errorf("edge is required")
 	}
+	if cfg.Store == nil {
+		return nil, fmt.Errorf("store is required")
+	}
```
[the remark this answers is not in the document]
bhandras: Agreed.

Removed the redundant `cfg.Store == nil` check in `New` since `applyDefaults` already assigns an in-memory store when unset.

mailbox/client/client.go
```diff
@@ -152,11 +163,24 @@ func (c *Client) AwaitRPC(ctx context.Context, correlationID string,
...
 				DiscardUnknown: true,
 			}).Unmarshal(data, resp)
+			if err != nil {
+				return err
+			}
+
+			c.deletePending(correlationID)
+			_ = c.cfg.Store.DeleteResponse(
+				ctx, c.cfg.LocalMailboxID, correlationID,
+			)
```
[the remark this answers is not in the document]
bhandras: Good catch.

`DeleteResponse` is now handled as a best-effort cleanup step with a small bounded retry/backoff. We still return the decoded response to the caller even if cleanup fails.

bhandras:
@codex review pls

litbot-9000:

db/sqlc/migrations/000005_mailboxrpcclient_store.up.sql
```diff
@@ -0,0 +1,28 @@
+-- mailboxrpcclient durable store migration.
+--
+-- This migration creates a small persistence schema for mailboxrpcclient. It
+-- is intentionally separate from the durable actor mailbox tables: this store
```
Roasbeef, line 4: Wouldn't it be better if we just used the existing system for this? What's the rationale behind a distinct store here? 

An example of how things would've flowed in my mind:
  * `serverconn` actor uses the client mailbox to consume new mail that it may have using the RPC call that enables it to read the next relevant message. Using a unified pull call, similar to: 
  
```go
  service Mailbox {
rpc Pull (PullRequest) returns (PullResponse);
rpc Ack (AckRequest) returns (Empty);
rpc Nack (NackRequest) returns (Empty);
rpc Send (SendRequest) returns (Empty);
}


message PullRequest { string device_id; string durable; int32 max; int32 wait_ms; string filter_subject; }
message Pulled { Envelope msg; uint64 seq; string subj; }
message AckRequest { string durable; uint64 seq; }
message NackRequest { string durable; uint64 seq; string reason; }
message SendRequest { Envelope env; }
```

from https://github.com/lightninglabs/darepo/issues/4#issuecomment-3524711062 + https://github.com/lightninglabs/darepo/issues/4#issuecomment-3524711077

That durable serverconn actor then as part of a existing db connection used to consume mailbox messages spawns a goroutine (or it's another actor) to do long polling with `Pull`, that then sends an actor messages in for the message. The `serveconn` maps that message into a service key, looks that up, then will send to the actor. If the send is succesful, then outside the db connection, it'll call `Ack`. 

If we crash before `Ack`, then we'll pull the message again. If not, then it's in the mailbox of the destination, and they handle it when they can. 

Then for send they send to `serverconn` as normal, who then calls `Send` wit hthe envelop, etc.

Roasbeef: Ah going back I see the prior one does have `Pull`, etc -- I think what's missing maybe is the `servercon` wrapper and interaction.

Roasbeef: Yeah I don't think anything I sent above is incompatible. I realize there's two aspects:
  * Unary requests on the RPC server side of things. This would be for basically for swapdk talking to walletdk. This has the response+request structure that's implemented right now with the protos. 
  * Passive communication (potentially triggered by the above), which involves `serverconn` as I mention.

Roasbeef: I still would've thought that the server mux is what handles the lookup+dispatch for the unary request, but maybe in prod we'd just have a layer of indirection there.

mailbox/client/sqlc_store.go
```diff
@@ -0,0 +1,132 @@
...
+// SQLCStore is a Store implementation backed by the project's SQLC queries.
+//
+// This store is intended for production use, where the client already depends
+// on the project's SQL database stack.
+type SQLCStore struct {
+	q *sqlc.Queries
+}
+
+// NewSQLCStore constructs a Store backed by the given SQLC Queries instance.
+func NewSQLCStore(q *sqlc.Queries) *SQLCStore {
```
Roasbeef, line 22: I think this may be redundant with the durable store/queue items we added.

bhandras:
codex: Thanks for the detailed review. I re-read `darepo#4`, `darepo#71`, and the OOR durability direction (`darepo#89` + `darepo#92`), and I agree with the main concern.

I’m doing a clean rewrite of this slice with this boundary:

1. Keep the Mailbox Edge contract (`Send`/`Pull`/`AckUpTo`) as transport semantics.
2. Move durability ownership to the existing durable actor runtime (`darepo-client#48`) via a `serverconn` connector actor.
3. Remove the mailbox-specific SQL response store path (`mailbox/client/sqlc_store.go` + associated migration) to avoid dual/overlapping durability systems.

Planned runtime model:
- `serverconn` ingress loop long-polls `Pull`.
- It durably enqueues/dispatches to destination actors first.
- It calls `AckUpTo` only after durable local record/commit.
- If we crash before ack, redelivery is expected and handled by durable dedupe/idempotent actor processing.
- Egress uses durable outbox -> `Send` with stable idempotency keys.

So yes: this PR will stop trying to make a parallel mailbox-local durable store the source of truth and will instead wire the connector actor path directly.

I’ll push this as a rewrite update and keep the unary RPC facade for swapdk/walletdk usage on top of that connector/runtime.

db/sqlc/migrations/000005_mailboxrpcclient_store.up.sql
bhandras: codex: Agreed. After re-evaluating this, I’m removing the distinct mailbox SQL response store and switching to a `serverconn` connector built on the durable actor runtime.

The transport stays `Pull` + `AckUpTo`, but correctness/durability moves to the existing durable inbox/outbox path, so we don’t maintain two parallel durability mechanisms.

mailbox/client/sqlc_store.go
bhandras: codex: Agreed. I’m removing this path in the rewrite.

`mailbox/client/sqlc_store.go` and the related migration-backed response storage will be dropped; the durable actor delivery runtime will be the single durability authority.

Roasbeef:
Carrying the torch here in this PR: https://github.com/lightninglabs/darepo-client/pull/116

Will build https://github.com/lightninglabs/darepo-client/pull/91 on top of it.

PR #94, opened 2026-01-27
client: add VTXO leave (offboard) support

Roasbeef:
This PR adds client-side support for VTXO leave (offboard) requests, enabling users to exit the Ark by forfeiting their VTXO and receiving funds to an on-chain address rather than rolling over into a new VTXO.

The leave flow routes through the actor system: wallet receives the user's leave request, forwards to the round actor, which triggers the VTXO actor, which then emits a leave request back to the round actor for tracking. This maintains the same message flow patterns established for refresh while producing a different outcome.

The key insight is that leave reuses the existing forfeit mechanism. When a VTXO is being left, it enters the same RefreshRequestedState and goes through the same forfeit signing process. The difference is that instead of creating a replacement VTXO, the server includes an on-chain output in the batch transaction.

For leave-only rounds (no new VTXOs being created), the client skips the MuSig2 nonce generation phase entirely. Since there are no VTXO tree leaves requiring partial signatures, the client transitions directly from commitment transaction validation to forfeit signature collection. This optimization prevents the server from waiting for nonces that would never be used.

The implementation spans several packages: wallet handles the entry point, actormsg defines cross-actor message types, vtxo handles the VTXO-side state transitions, and round coordinates the overall flow. Each component follows the established patterns from the refresh implementation, making the leave flow a natural extension of existing functionality.

Testing is covered by the companion PR in the server repo which adds TestVTXOLeaveE2E, demonstrating the complete lifecycle from boarding through VTXO creation, leave trigger, forfeit signing, and final on-chain balance verification.

### DIFF
--- a/round/events.go
+++ b/round/events.go
@@ -342,9 +342,9 @@ func (e *RoundComplete) clientEventSealed() {}
 // expiry and needs to be refreshed in a new round. The round actor should
 // queue this VTXO for inclusion in the next batch swap.
 //
-// This request contains all information needed to build both the server's
-// RefreshRequest (for the connector tree) and a VTXORequest (for the new VTXO
-// in the VTXT). The same client key is typically reused for the new VTXO.
+// This request contains all information needed to build a forfeit request
+// (for the connector tree) and a VTXORequest (for the new VTXO in the VTXT).
+// The same client key is typically reused for the new VTXO.
 type RefreshVTXORequest struct {
 	actor.BaseMessage
 
--- a/round/events.go
+++ b/round/events.go
@@ -381,6 +381,40 @@ func (e *RefreshVTXORequest) MessageType() string {
 	return "RefreshVTXORequest"
 }
 
+// LeaveVTXORequest is sent from a VTXO actor (or wallet) when the user wants
+// to exit the Ark by forfeiting a VTXO and receiving an on-chain output. This
+// is similar to RefreshVTXORequest except the output is on-chain rather than a
+// new VTXO.
+//
+// The leave flow uses the same forfeit mechanism as refresh: the old VTXO is
+// forfeited via a connector output, and the leave output is included directly
+// in the batch transaction.
+type LeaveVTXORequest struct {
+	actor.BaseMessage
+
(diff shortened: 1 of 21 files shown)


round/events.go
```diff
@@ -381,6 +381,40 @@ func (e *RefreshVTXORequest) MessageType() string {
...
+// new VTXO.
+//
+// The leave flow uses the same forfeit mechanism as refresh: the old VTXO is
+// forfeited via a connector output, and the leave output is included directly
+// in the batch transaction.
+type LeaveVTXORequest struct {
+	actor.BaseMessage
+
+	// VTXOOutpoint identifies the VTXO to forfeit.
+	VTXOOutpoint wire.OutPoint
```
ellemouton, line 396: as mentioned on server side: i think this should not be explicit cause it limits us to one VTXO being forfeit for a leave. I think leave req should just be an amount & pk script. Then, separately, we should have Forfeit Reqs which specify which vtxos to forfeit. this would then allow us to forfeit many vtxos and create one leave req OR forfeit one large VTXO and create a set of new vtxos (refresh) and UTXOs (leaves)

round/transitions.go
```diff
@@ -980,6 +980,54 @@ func (s *CommitmentTxValidatedState) ProcessEvent(
 			slog.Int("boarding_intent_count", len(s.Intents.Boarding)),
 			slog.Int("vtxo_intent_count", len(s.Intents.VTXOs)))
 
+		// For leave-only rounds (no new VTXOs), skip nonce generation
+		// and go directly to forfeit collection. The server doesn't
+		// need nonces when there are no VTXOs to sign.
+		if len(s.Intents.VTXOs) == 0 && len(s.ForfeitMappings) > 0 {
```
ellemouton: technically (although we should set up the fee structure to disincentivise against this), it is possible to have a batch tx with only boarding inputs and leave outputs. so i think instead of using `len(s.ForfeitMappings)` here, we should instead explicitly check NumLeaveReqs

Roasbeef:
@claude review this

@codex review this

ellemouton (approved):

PR #95, opened 2026-01-27
systest: add leave request e2e test and bridge support

Roasbeef:
This PR adds end-to-end testing for VTXO leave (offboard) requests along with the necessary bridge conversion support. The test demonstrates the complete leave lifecycle: boarding funds, creating a VTXO, triggering leave to an on-chain address, and verifying the forfeited VTXO's value appears in the client's wallet.

The bridge conversion in convertToActorMsg needed updating to handle LeaveRequests. Each leave request must be converted to both a LeaveReq (containing the destination output) and a ForfeitReq (for the VTXO being forfeited). This mirrors how the server processes leaves: the forfeit mechanism handles the VTXO consumption while the leave output goes into the batch transaction.

The TestClient gains several helper methods for leave operations: TriggerVTXOLeave routes through the wallet actor to initiate the flow, GetOnChainBalance and WaitForOnChainBalance verify funds after the leave completes, and GetNewAddress generates taproot addresses for receiving leave outputs. These follow the established patterns from the refresh test helpers.

The e2e test runs in three phases. Phase one boards funds and creates a Live VTXO. Phase two triggers leave, waits for the VTXO to transition through Forfeiting, seals the round, and completes the forfeit signing. Phase three verifies the VTXO reached Forfeited status, confirms no live VTXOs remain, and validates the on-chain balance increased by approximately the VTXO amount minus fees.

One notable detail: leave-only rounds have fewer transcript messages than regular rounds because they skip the nonce/partial signature exchange. The test uses msgsPerLeaveRound=5 instead of the standard msgsPerClientRound to account for this. A short sleep handles the async message path through the actor system before registration is triggered.

Depends on the companion client PR lightninglabs/darepo-client#94 which implements the client-side leave request handling.


Roasbeef:
Created in wrong repo by mistake

PR #97, opened 2026-01-30
multi: implement BIP-322 authentication for join requests

ellemouton:
This PR builds as follows:

1) Add the Bip322 spec implementation
2) Add an application level wrapper that let's us define how we want to use the locktime fields in the to_sign tx
3) define the TLV codec for our JoinRoundReq  -> we will use that to produce the `message challenge` to sign
4) plug things in so that the client is sending over the Auth sig of the JoinRoundReq


round/auth.go
```diff
@@ -0,0 +1,341 @@
...
+	digest := computeRequestDigest(req)
+
+	// Create the message to sign.
+	msg := bip322.JoinRequestMessage{
+		Type:              "join_round",
+		Timestamp:         time.Now().Unix(),
+		BoardingOutpoints: boardingOps,
+		ForfeitOutpoints:  forfeitOps,
+		RequestDigest:     hex.EncodeToString(digest[:]),
+	}
```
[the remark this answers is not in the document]
ellemouton: decided to use timestamp-based replay protection instead of a server-provided challenge.

lib/bip322/message.go
```diff
@@ -0,0 +1,236 @@
...
+// the request digest, the client commits to all VTXO requests and leave
+// outputs without revealing them in the BIP-322 message itself, maintaining
+// privacy while ensuring authentication.
+type JoinRequestMessage struct {
+	// Type must be "join_round" for join request authentication. This
+	// allows the operator to distinguish join requests from other message
+	// types.
+	Type string `json:"type"`
+
+
```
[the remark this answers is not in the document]
ellemouton: using timestamp based approach

ellemouton:
@claude review this

litbot-9000:

lib/bip322/message.go
```diff
@@ -0,0 +1,235 @@
...
+type IntentMessageType string
+
+const (
+	// IntentMessageTypeRegister indicates a registration intent, used
+	// when registering new outputs or state with the Ark operator.
+	IntentMessageTypeRegister IntentMessageType = "register"
+
+	// IntentMessageTypeDelete indicates a deletion intent, used when
+	// removing previously registered state from the Ark operator.
+	IntentMessageTypeDelete IntentMessageType = "delete"
```
Roasbeef: Will they expire or need to be refreshed after a period of time?

lib/bip322/message.go
```diff
@@ -0,0 +1,235 @@
...
+// verification, output classifications, validity windows, and cosigner
+// public keys for offchain output authentication.
+type IntentMessage struct {
+	BaseIntentMessage
+
+	// InputTapTrees contains hex-encoded Taproot script trees for each
+	// spent input (excluding the first BIP-322 virtual input). The index
+	// in this slice corresponds to input index + 1 in the proof
+	// transaction, allowing the operator to verify covenant scripts.
+	InputTapTrees []string `json:"input_tap_trees"`
```
Roasbeef: Are the JSON tags actually needed here?

Roasbeef: What's the actual encoding format here?

lib/bip322/message.go
```diff
@@ -0,0 +1,235 @@
...
+	// This allows time-bounded deletion requests that automatically expire
+	// if not processed within a certain timeframe.
+	ExpireAt int64 `json:"expire_at"`
+}
+
+// Encode serializes the DeleteIntentMessage to a JSON string suitable for
+// signing via BIP-322. The resulting string can be passed to New() to create
+// a proof.
+func (m *DeleteIntentMessage) Encode() (string, error) {
+	encoded, err := json.Marshal(m)
```
Roasbeef: If these are sent as part of the round or pre round messages, then why do we need json encoding?

lib/bip322/message.go
```diff
@@ -0,0 +1,235 @@
...
+// request in the Ark protocol. This message type proves ownership of inputs
+// being spent (boarding UTXOs and forfeit VTXOs) while committing to the
+// complete request payload via a digest.
+//
+// The message structure prevents replay attacks through the client-provided
+// timestamp combined with the unique nature of the outpoints being spent (each
+// UTXO can only be spent once). By signing the request digest, the client
+// commits to all VTXO requests and leave outputs without revealing them in the
+// BIP-322 message itself, maintaining privacy while ensuring authentication.
+type JoinRequestMessage struct {
```
Roasbeef: Seems like it could accept this message type as a type param to make it more generic? As then we need to keep this in sync w/ the actual version.

lib/bip322/message.go
```diff
@@ -0,0 +1,235 @@
...
+	// spent in this join request. Each outpoint is encoded as a hex string
+	// in "txid:vout" format. These are virtual UTXOs from previous rounds
+	// that are being forfeited (spent) to create new outputs.
+	ForfeitOutpoints []string `json:"forfeit_outpoints"`
+
+	// RequestDigest is a SHA256 hash of the complete VTXO requests and
+	// leave outputs included in this join request. Encoded as a hex
+	// string. By committing to this digest, the client authenticates the
+	// entire request payload without exposing sensitive details in the
+	// BIP-322 message.
```
Roasbeef: Is the info sensitive if it's all actually just sent to the server?

Also I was under the impression that we _only_ need to sign the boarding inputs, which can be done independent of a round. Then we ref a hash digest of the signed intent.

lib/bip322/proof.go
```diff
@@ -0,0 +1,401 @@
...
+type FullProof psbt.Packet
+
+// Input represents a UTXO to be included in a BIP-322 proof. This structure
+// captures the essential data needed to create a valid proof: the outpoint
+// reference, sequence number for timelock support, and the witness UTXO data
+// for script verification.
+type Input struct {
+	// OutPoint identifies the previous transaction output being spent.
+	// This is the UTXO for which ownership is being proven.
+	OutPoint *wire.OutPoint
```
Roasbeef: Always there, so can be a normal value. Same below (we always use segwit in this system).

lib/bip322/proof.go
```diff
@@ -0,0 +1,401 @@
...
+//
+//	proof, _ := New(message, inputs, nil)
+//	// ... sign proof with wallet ...
+//	sig, err := proof.Signature()
+//
+// Example (custom finalization):
+//
+//	sig, err := proof.Signature(customFinalizer)
+func (p *FullProof) Signature(
+	finalize ...func(*psbt.Packet) error,
```
Roasbeef: Functional option?

lib/bip322/README.md
```diff
@@ -0,0 +1,383 @@
...
+protection can be achieved through two approaches:
+
+**Challenge-based (traditional)**:
+- Server provides a unique nonce or challenge for each request
+- Client includes the challenge in the signed message
+- Requires request/response round-trip before signing
+- Provides strong server-controlled replay prevention
+
+**Timestamp-based (this implementation)**:
+- Client includes a timestamp in the signed message
```
Roasbeef: Block height instead perhaps? Then verification is fully hermetic from Bitcoin's PoV.

lib/bip322/signature.go
```diff
@@ -0,0 +1,309 @@
...
+// The transaction structure encodes:
+//   - Witness data proving knowledge of private keys for all inputs
+//   - The message commitment via the first input's reference to to_spend
+//   - Optional outputs demonstrating spend capability or embedding data
+//   - All input sequences and locktime values for covenant enforcement
+//
+// Signatures can be encoded to base64 for transmission and storage, then
+// decoded and verified by recipients who have access to the original message
+// and UTXO data.
+type Signature wire.MsgTx
```
Roasbeef: `TxSig` or `SigTx`? A bit odd to call a message itself a signature.

lib/bip322/signature.go
```diff
@@ -0,0 +1,309 @@
...
+		)
+		if prevout == nil {
+			return ErrPrevoutNotFound
+		}
+
+		// Create a script engine with standard verification flags. This
+		// enforces consensus rules including CLEANSTACK, WITNESS, and
+		// signature validation.
+		engine, err := txscript.NewEngine(
+			prevout.PkScript,
```
Roasbeef: Can pack more on the line, we should add this example to `AGENTS.md`.

lib/types/boarding.go
```diff
@@ -73,6 +85,10 @@ type JoinRoundRequest struct {
 
 	// ForfeitReqs specifies the requests to forfeit VTXOs.
 	ForfeitReqs []*ForfeitRequest
+
+	// AuthSignature proves ownership of all boarding and forfeit inputs via
+	// BIP-322 signature. This is required for request authentication.
+	AuthSignature *BIP322Signature
```
Roasbeef: Ah ok, we'll just package it in the join message, that makes sense. 

Will the client cache these? Eg if a round is bailed, then it needs to sign again.

round/auth.go
```diff
@@ -0,0 +1,341 @@
...
+	// Add VTXO requests in order.
+	for _, vtxo := range req.VTXOReqs {
+		// Commit to pkScript.
+		data = append(data, vtxo.PkScript...)
+
+		// Commit to amount explicitly (little-endian encoding).
+		amountBytes := make([]byte, 8)
+		amount := uint64(vtxo.Amount)
+		for i := 0; i < 8; i++ {
+			amountBytes[i] = byte(amount >> (8 * i))
```
Roasbeef: Should use `binary.LittleEndian` here.

round/auth.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	// Build list of forfeit outpoints for the message.
+	forfeitOps := make([]string, 0, len(req.ForfeitReqs))
+	for _, fr := range req.ForfeitReqs {
+		forfeitOps = append(forfeitOps, fr.VTXOOutpoint.String())
+	}
+
+	// Compute digest of the request contents (VTXOs + leaves).
+	digest := computeRequestDigest(req)
```
Roasbeef: I'm a bit confused here re the mapping/framing. Why not just put _everything_ into this message digest? RIght now we have this, but then `msg` below that gets encoded into JSON. Why not skip the JSON layer entirely. 

The client can make a digest hash of all the relevant contents (eg: block hash as well for extra anti-replay). That's the main BIP 322 message being signed. It sends that to the server. This is included in the join req, so it's able to reconstruct the digest using the block hash it sent in the `StartRound` message or w/e it's called.

round/auth.go
```diff
@@ -0,0 +1,341 @@
...
+	req *types.JoinRoundRequest,
+	authInputs map[wire.OutPoint]*authInput,
+) ([]bip322.Input, []*authInput, error) {
+
+	numInputs := len(req.BoardingReqs) + len(req.ForfeitReqs)
+	inputs := make([]bip322.Input, 0, numInputs)
+	ordered := make([]*authInput, 0, numInputs)
+
+	// Add boarding UTXOs.
+	for _, br := range req.BoardingReqs {
```
Roasbeef: Can just sort here and below.

round/transitions_auth_helpers.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/lib/scripts"
+	"github.com/lightninglabs/darepo-client/lib/types"
+)
+
+// convertBoardingReqs converts round.BoardingRequest messages to
+// types.BoardingRequest for authentication.
+func convertBoardingReqs(
+	reqs []types.BoardingRequest,
```
Roasbeef: This and the ones below can just be a generic func iic. Like `map(ptr, struct)`

round/transitions.go
```diff
@@ -606,6 +606,39 @@ func (s *PendingRoundAssembly) ProcessEvent(ctx context.Context,
 			slog.Int("forfeit_requests", len(forfeitReqs)),
 			slog.Int("leave_requests", len(leaveReqs)))
 
+		// Create BIP-322 authentication signature.
+		authReq := &types.JoinRoundRequest{
```
Roasbeef: Why's this conversion needed at all?

Roasbeef (changes requested):

lib/bip322/validation_test.go
```diff
@@ -0,0 +1,231 @@
...
+	t.Parallel()
+
+	challengeScript, sig := buildHelloWorldVectorFullSig(t)
+
+	mutated := &Sig{
+		ToSign: sig.ToSign.Copy(),
+	}
+	require.NotEmpty(t, mutated.ToSign.TxIn[0].Witness)
+	require.NotEmpty(t, mutated.ToSign.TxIn[0].Witness[0])
+	mutated.ToSign.TxIn[0].Witness[0][10] ^= 0x01
```
bhandras, line 43: Nice 👍

lib/types/codec.go
```diff
@@ -0,0 +1,478 @@
...
+type JoinRoundAuth struct {
+	// Message is the canonical join message signed by the client.
+	Message []byte
+
+	// Signature is the serialized full-format BIP-322 signature
+	// payload.
+	Signature []byte
+}
+
+// JoinRoundAuthMessage serializes a JoinRoundRequest into deterministic,
```
bhandras, line 98: Nice doc!

round/join_auth.go
```diff
@@ -139,3 +170,430 @@ func computeTotalForfeitAmount(
...
+	// input (proof-of-funds). These become inputs 1..N in the
+	// to_sign transaction. Each proves ownership of a boarding
+	// UTXO or forfeited VTXO via its unilateral timeout path.
+	additionalInputs := buildJoinAuthAdditionalInputs(
+		signingInputs,
+	)
+
+	// Compute the block-window upper bound so the server can
+	// reject stale proofs.
+	validUntil := joinAuthValidUntil(env.StartHeight)
```
bhandras: This was marked by codex to be somewhat wrong, because we have two choices here: 
1) start the window from signing 
2) start the window from round fsm start height

I think both are correct and maybe codex is right in that 1) is slightly more correct.

round/join_auth_test.go
```diff
@@ -0,0 +1,543 @@
+package round
```
bhandras, line 1: Do we also want to test rejecting join auth if the signature is invalid/tampered?

bhandras (approved):
LGTM, very nice and readable! 🎉

lib/bip322/validation.go
```diff
@@ -0,0 +1,292 @@
...
+	VerificationStateInconclusive VerificationState = "inconclusive"
+
+	// VerificationStateInvalid indicates verification failed.
+	VerificationStateInvalid VerificationState = "invalid"
+)
+
+// VerificationResult contains the result of validating a BIP-322 auth package.
+type VerificationResult struct {
+	// State is the verification state.
+	State VerificationState
```
Roasbeef, line 44: 👍

lib/bip322/validation.go
```diff
@@ -0,0 +1,292 @@
...
+
+	return engine.Execute()
+}
+
+// isUpgradeableScriptError identifies script failures that BIP-322 maps to the
+// inconclusive state.
+func isUpgradeableScriptError(err error) bool {
+	return txscript.IsErrorCode(
+		err, txscript.ErrDiscourageUpgradableNOPs,
+	) ||
```
Roasbeef: Non-blocking, but had to squint a bit here to see that it's just chaining the `IsErrorCode` calls? I wonder if we can't just use `errors.Is` here?

lib/bip322/block_window.go
```diff
@@ -0,0 +1,126 @@
+package bip322
+
+import "fmt"
+
+// BlockWindow defines application-level block-height validity bounds for a
+// BIP-322 signature.
+//
+// This wrapper layer maps the bounds onto BIP-322's surfaced lock metadata:
+//   - ValidFromBlock -> to_sign nLockTime
```
Roasbeef: Nice utilization of the Bitcoin-specific portions of to_sign 👍

Roasbeef: IIUC this isn't neccerily part of the OG BIP, but with `txscript.Engine` we're able to layer on additional constraints on the signature?

lib/bip322/join_round.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/btcsuite/btcd/txscript"
+)
+
+// JoinRoundMessageChallenge derives the challenge script used for round
+// join authorization signatures from the request identifier key.
+//
+// The challenge is a key-path Taproot output (BIP-86 style, no script
+// tree) so the signer can satisfy input 0 with a single Schnorr
+// signature.
+func JoinRoundMessageChallenge(
```
Roasbeef, line 16: So this is the main challenge? I thought the server would bind to the current round somehow, ensuring that sigs can't be re-used across rounds?

round/join_auth.go
```diff
@@ -7,16 +7,47 @@ import (
...
+	"github.com/lightninglabs/darepo-client/lib/scripts"
+	"github.com/lightninglabs/darepo-client/lib/types"
+	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
 )
 
 const (
+	// joinRoundAuthWindowBlocks is the default validity window length
+	// applied to join authorization signatures.
+	joinRoundAuthWindowBlocks uint32 = 144
```
Roasbeef, line 25: Ah ok, so we don't care it it's strongly bound to a round (eg: sign round ID), we just want them to update the sigs every 24 hours.

AGENTS.md
```diff
@@ -50,6 +50,23 @@
 - Add explanatory comments between stanzas
 - Spacing between switch/select cases
 - When wrapping function calls, put closing paren on its own line with all args on new lines
+- When wrapping function **definitions**, keep the first parameter on the same line as the function name, and the closing `)` on the same line as the last parameter (no trailing comma). This differs from function calls.
+
+```go
+// WRONG: expanded style with dangling close paren
+func foo(
+	arg1 Type1,
```
Roasbeef, line 58: Did the same in a recent PR lol

If this fails, we can reach for `ast-grep` again.

lib/bip322/validation.go
```diff
@@ -0,0 +1,343 @@
...
+				inputIndex, err,
+			))
+		}
+	}
+
+	// Step 8: Surface the lock metadata exactly as BIP-322 specifies,
+	// then apply any application-level policy checks.
+	result := validResult(toSign.LockTime, toSign.TxIn[0].Sequence)
+
+	return applyBlockWindowValidationPolicy(result, validationOpts)
```
Roasbeef: So what if one of the inputs signed actually has a CSV/CLTV in the script. It's possible then that once we apply our application level sequence+locktime (our augmentation on BIP 322), we actually cause an otherwise valid witness to be valid (or the other way around). 

In that case, I wonder if we shouldn't just put this information into the message hash, which is then extract for this validation step (`applyBlockWindowValidationPolicy`). Re-using sequence+locktime in this domain is indeed cool, but I think we should double check things here to ensure there isn't some cross layer contamination.

lib/types/codec.go
```diff
@@ -0,0 +1,477 @@
...
+// versioned TLV bytes for BIP-322 signing and verification.
+//
+// The resulting bytes are the message that the client signs with BIP-322
+// to prove authorization of the join request contents.
+//
+// Top-level envelope (TLV stream):
+//
+//	+-------+-------------------------------------------+
+//	| type  | field                                     |
+//	+-------+-------------------------------------------+
```
Roasbeef, line 108: Is there a reason we _don't_ sign the round ID? By the time they're sending inputs, it's known right?

lib/bip322/validation.go
```diff
@@ -0,0 +1,343 @@
...
+	// validate against the exact commitment model defined by the spec.
+	messageHash := MessageHash(pkg.Message)
+	toSpend, err := BuildToSpend(messageHash, pkg.MessageChallenge)
+	if err != nil {
+		return invalidResult(fmt.Sprintf("build to_spend: %v", err))
+	}
+
+	// Step 3: Work on a copy of to_sign so validation never mutates caller
+	// state.
+	toSign := pkg.Sig.ToSign.Copy()
```
Roasbeef, line 177: Before this we should reject if the number of inputs exceeds some sane value. In Bitcoin land we have the policy and consensus level limits to precept against DoS, we should have something similar here.

lib/types/codec_test.go
```diff
@@ -0,0 +1,262 @@
...
+func TestJoinRoundAuthMessageOmitsIdentifierRecord(t *testing.T) {
+	t.Parallel()
+
+	req := testJoinRoundAuthRequest(t, false)
+
+	raw, err := JoinRoundAuthMessage(req)
+	require.NoError(t, err)
+
+	var (
+		version     uint64
```
Roasbeef: Sub struct with a `Decode` method?

lib/types/codec.go
```diff
@@ -0,0 +1,477 @@
...
+	records = append(records, tlv.MakePrimitiveRecord(
+		joinRoundAuthMessageLeaveRecordType, &leaveBlob,
+	))
+
+	return encodeJoinAuthTLV(records)
+}
+
+// encodeJoinAuthBoardingRequests serializes boarding request data in
+// request order.
+func encodeJoinAuthBoardingRequests(requests []*BoardingRequest) ([]byte,
```
Roasbeef, line 933: Might be missing something, but where's the decode? Decode should also enforce max limits for all the outpoints, requests, etc.

lib/types/codec.go
```diff
@@ -0,0 +1,477 @@
...
+//	|   7   | leaves    (blob list)                     |
+//	+-------+-------------------------------------------+
+//
+// Each blob list is encoded as:
+//
+//	varint(count) || for each entry: varint(len) || entry_bytes
+//
+// Boarding entry TLV:
+//
+//	1: outpoint hash  |  2: outpoint index
```
Roasbeef, line 124: Include amount as well?

round/join_auth.go
```diff
@@ -0,0 +1,599 @@
...
+	wallet ClientWallet) (keychain.KeyDescriptor, error) {
+
+	if wallet == nil {
+		return keychain.KeyDescriptor{}, fmt.Errorf(
+			"wallet signer must be provided",
+		)
+	}
+
+	keyDesc, err := wallet.DeriveNextKey(
+		ctx, joinRoundAuthIdentifierKeyFamily,
```
Roasbeef, line 65: Should we cache this to a degree? Otherwise, each time the round starts up we'll generate a new key (eg: someone else was slow, needed to restart).

round/join_auth.go
```diff
@@ -0,0 +1,599 @@
+package round
```
Roasbeef, line 1: We should add some basic logging here for when we run things in a real context (eg: systest debug, etc).

Roasbeef:
@claude review this

lib/bip322/validation.go
ellemouton: cool - added a commit at the end to address this 👍

lib/bip322/block_window.go
ellemouton: yeah the OG spec just says that the validation spits out these locktime values - but it is up to the application layer to specify what these are interpreted as

lib/types/codec.go
ellemouton: but the bip322 sig is sent when the JoinRoundReq is sent at which point we dont yet know the Round ID that we will be assigned to

lib/types/codec.go
ellemouton: reason is: the amount is already implicitly encoded via the outpoint hash & isnt currently explicitly sent via the Join Reqeust - so it would then require a blockchain lookup just for the server to verify the sig

lib/types/codec.go
ellemouton: we shoudn't need a decode - this encoding is purely to create a reproducible digest that is being signed. 

So: 1) cllient creates Join Req - creates digest & signs -> sends to server
2) server recomputes digest to validate sig

lib/bip322/join_round.go
ellemouton: bip 322 message challenge is different from the message just fyi. message challenge = the key that the client signs the first input of the to_sign tx which spends the to_spend tx. 

the message is then the digest of the content we are signing

round/join_auth.go
ellemouton: this is client side - so this is the client saying "this sig of mine is only valid for this many blocks"

lib/types/codec_test.go
ellemouton: added a Decode helper - but see my other comment re why the decode isnt needed

round/join_auth.go
ellemouton: ahhhhh good catch. updated to be "signed" height. 


forgot that the client side FSM doesnt start at "round join time" but potentially wayyyyyyy before that

ellemouton: updated - ptal

round/join_auth_test.go
ellemouton: added

lib/bip322/validation.go
ellemouton: done

round/join_auth.go
ellemouton: added

lib/bip322/validation.go
ellemouton: good point - take a look at this new commit `multi: sign join auth with intent envelope` to see if you agree with that new approach

bhandras (approved):
LGTM! 🥇

lib/types/codec.go
Roasbeef: Ah right, makes sense.

lib/types/codec.go
Roasbeef: Makes sense, server handles decode.

Roasbeef (approved):
LGTM 🫐

Seperation of concerns re intent layer addresses my locktime overload comment.

PR #98, opened 2026-01-30
unroller: add VTXO unroller actor for on-chain tree broadcast

sputn1ck:
## Summary
- Add unroller package for broadcasting pre-signed tree transactions during unilateral VTXO exits
- UnrollerActor manages unroll state and coordinates level-by-level broadcasts with CSV delays
- Uses `round.ClientVTXO` directly for tree path access (matches database layer return type)
- Includes BDD tests verifying unroll flow with mock chain source

## Test plan
- [x] Unit tests pass: `make unit pkg=unroller`
- [x] Linter passes: `make lint`
- [ ] Integration test in darepo server repo (systest)


sputn1ck:
@claude review please

unroller/state.go
```diff
@@ -0,0 +1,164 @@
...
+	// Error tracks any failure reason.
+	Error error
+
+	// RetryCount tracks broadcast retry attempts.
+	RetryCount int
+}
+
+// LevelTxids groups transactions by tree level.
+type LevelTxids struct {
+	Level int
```
bhandras, line 98: Self documenting but just for completeness could you add godocs.

unroller/state.go
```diff
@@ -0,0 +1,164 @@
...
+	Txids []chainhash.Hash
+	Nodes []*tree.Node
+}
+
+// ConfirmationInfo tracks when a transaction confirmed.
+type ConfirmationInfo struct {
+	Height    int32
```
bhandras: nit: missing godocs.

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+			if !a.isLevelConfirmed(state, state.CurrentLevel) {
+				continue
+			}
+
+			// Get confirmation height of current level.
+			currentLevelHeight := a.getLevelConfirmHeight(
+				state, state.CurrentLevel,
+			)
+
+			if evt.Height >= currentLevelHeight+csvDelay {
```
bhandras: I don't really understand why we need to wait here for each level? We should only wait after the vtxo itself is confirmed right?

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+		signedTx, err := node.ToSignedTx()
+		if err != nil {
+			a.cfg.Logger.ErrorS(ctx, "Failed to construct tx", err,
+				slog.String("txid", txid.String()))
+
+			continue
+		}
+
+		// Broadcast transaction.
+		broadReq := &chainsource.BroadcastTxRequest{
```
bhandras: This won't work as-is, because these presigned transactions are all zero fee, so we'd need to form a package spending the anchor with a fee paying child (see old ark code too).

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+		return fn.Err[UnrollerResp](
+			fmt.Errorf("fetch VTXO: %w", err),
+		)
+	}
+
+	a.cfg.Logger.InfoS(ctx, "Starting VTXO unroll",
+		slog.String("outpoint", outpointKey))
+
+	// Extract level-ordered transactions from tree.
+	levelOrder := extractLevelOrder(vtxoDesc.TreePath)
```
bhandras: Should we sanity check that we actually managed to extract the path correctly?

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+		future := a.cfg.ChainSource.Ask(ctx, broadReq)
+		result := future.Await(ctx)
+
+		if result.IsErr() {
+			a.cfg.Logger.WarnS(
+				ctx, "Broadcast failed", result.Err(),
+				slog.String("txid", txid.String()),
+				slog.Int("level", level))
+
+			// Continue with other transactions, will retry later.
```
bhandras: Copilot is right, this isn't handled in the current code.

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+			slog.String("outpoint", outpointKey))
+
+		return fn.Err[UnrollerResp](err)
+	}
+
+	a.activeUnrolls[outpointKey] = state
+
+	// Start broadcasting from level 1. Level 0 is the root/pool transaction
+	// which is already confirmed on-chain.
+	a.broadcastLevel(ctx, state, 1)
```
bhandras: IIUC the tree root is an output of the batch tx (which is the one confirmed) and one batch can have multiple tree outputs, so we should start level 0 right?

unroller/state.go
```diff
@@ -0,0 +1,164 @@
...
+
+		levelMap[level] = append(levelMap[level], txid)
+		nodeMap[level] = append(nodeMap[level], node)
+
+		if level > maxLevel {
+			maxLevel = level
+		}
+
+		// Enqueue children for next level.
+		for _, child := range node.Children {
```
bhandras: Even though we expect to just extract a single path always (IIUC) we should ensure that iteration is stable, so we could first sort child keys.

unroller/state.go
```diff
@@ -0,0 +1,164 @@
...
+	for len(queue) > 0 {
+		item := queue[0]
+		queue = queue[1:]
+
+		node, level := item.node, item.level
+
+		// Get this node's txid.
+		txid, err := node.TXID()
+		if err != nil {
+			continue
```
bhandras: I guess this can/should never fail but still let's return the error here instead.

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+
+		// Mark as broadcast.
+		state.BroadcastTxids[txid] = true
+
+		// Register for confirmation monitoring.
+		a.registerConfirmation(ctx, state, txid)
+	}
+
+	// Subscribe to block epochs for CSV tracking.
+	a.subscribeBlockEpochs(ctx, state)
```
bhandras: Could we just have a single block epoch subscription? Also we only really need to care about the csv for the leaf right?

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+	// Persist state.
+	if err := a.cfg.Store.SaveUnrollState(ctx, state); err != nil {
+		a.cfg.Logger.ErrorS(ctx, "Failed to save unroll state", err,
+			slog.String("outpoint", outpointKey))
+
+		return fn.Err[UnrollerResp](err)
+	}
+
+	a.activeUnrolls[outpointKey] = state
+
```
bhandras, line 87: Would it be safe to release the lock here? If we ask any actors in boradcastLevel we might try to re-acquire the mutex causing a deadlock.

unroller/state.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// ConfirmationInfo tracks when a transaction confirmed.
+type ConfirmationInfo struct {
+	Height    int32
+	BlockHash chainhash.Hash
+}
+
+// extractLevelOrder converts tree path to level-ordered TXIDs and nodes.
+func extractLevelOrder(treePath *tree.Tree) []LevelTxids {
```
bhandras: I'm not even sure we need all this level order thing, as we expect a single tx per level right? Could be i'm wrong here but my limited understanding says so.

unroller/unroll.go
```diff
@@ -0,0 +1,431 @@
...
+		txid := levelTxids.Txids[i]
+
+		// Convert to signed transaction.
+		signedTx, err := node.ToSignedTx()
+		if err != nil {
+			a.cfg.Logger.ErrorS(ctx, "Failed to construct tx", err,
+				slog.String("txid", txid.String()))
+
+			continue
```
bhandras: I think we should actually fail here. Could be something we test too.

unroller/unroller.feature
```diff
@@ -0,0 +1,94 @@
...
+    And confirmation subscriptions should be re-registered
+    And block epoch subscriptions should be re-registered
+
+  Scenario: Handle broadcast failures gracefully
+    Given a VTXO with a 2-level tree
+    When I request unroll for the VTXO
+    And the broadcast fails for level 1 transaction
+    Then the actor should log the failure
+    And the unroll should remain in "broadcasting" status
+    And retry should be attempted on next block
```
bhandras: This is interesting because seemingly the test passes but the code isn't correct. Maybe we need to be more defensive with the cucumber tests.

bhandras (commented):
Very nice! I think this is close but there are some assumptions which imo are wrong or worth discussing. Also some of the agent review comments are pretty high signal and should be checked as well.

litbot-9000:

PR #103, opened 2026-02-08
indexer: add cursor-aware sync sdk (draft)

bhandras:
## Summary
Add a high-level indexer sync SDK on top of the mailbox RPC client.

This introduces cursor-aware polling helpers for:
- VTXO event feed sync
- OOR recipient event sync

## What’s included
- `indexer.SyncClient` facade for restart-friendly polling
- `SyncCursorStore` abstraction
- `MemorySyncCursorStore` default implementation
- unit tests for monotonic cursor persistence and replay-safe progression

## Why
We now have script-gated indexer APIs and durable server-side event logs. The client needs a higher-level polling surface so apps can consume these feeds without reimplementing cursor bookkeeping.

## Notes
- Stacked usage: this branch is intended to pair with server PR https://github.com/lightninglabs/darepo/pull/90
- Keeping draft until we finish remaining indexer TODOs.


mailbox/client/client.go
```diff
@@ -0,0 +1,563 @@
...
+					"missing response body",
+				)
+			}
+
+			err := (proto.UnmarshalOptions{
+				DiscardUnknown: true,
+			}).Unmarshal(env.Body.Value, resp)
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
bhandras: Updated this path to wrap response-body unmarshal failures with `mailboxrpc.ErrFromMalformedResponse(...)` for consistent malformed-response handling.

arkrpc/indexer.proto
```diff
@@ -0,0 +1,368 @@
...
+    // ListVTXOEventsByScripts returns a monotonic change feed for VTXOs
+    // matching the provided scripts, gated by proof-of-control.
+    //
+    // Wallets can use this to catch up after being offline and to reconcile
+    // local state deterministically.
+    rpc ListVTXOEventsByScripts (ListVTXOEventsByScriptsRequest)
+        returns (ListVTXOEventsByScriptsResponse);
+}
+
+message RegisterReceiveScriptRequest {
```
Roasbeef, line 58: Will this get _pushed_ to it by the server?

arkrpc/indexer.proto
```diff
@@ -0,0 +1,368 @@
...
+
+    // label is informational and can be used by clients for debugging.
+    string label = 3;
+
+    oneof proof {
+        // taproot_schnorr proves control of the P2TR output key.
+        TaprootSchnorrProof taproot_schnorr = 10;
+
+        // bip322 is reserved for future BIP-322 script-satisfaction proofs.
+        BIP322Proof bip322 = 11;
```
Roasbeef, line 73: 👍

indexer/client.go
```diff
@@ -0,0 +1,224 @@
...
+// arkrpc.IndexerServiceMailboxClient that helps construct canonical receive
+// script registration proofs.
+//
+// This package intentionally does not dictate how the mailbox principal is
+// minted or stored; callers provide the canonical principal identifier string
+// used in the signed message (typically the mailbox ID, e.g. "client:<id>").
+type Client struct {
+	rpc       *arkrpc.IndexerServiceMailboxClient
+	serverID  string
+	principal string
```
Roasbeef, line 42: principal?

indexer/client.go
```diff
@@ -0,0 +1,224 @@
...
+	return &Client{
+		rpc:       arkrpc.NewIndexerServiceMailboxClient(rpc),
+		serverID:  serverID,
+		principal: principal,
+	}
+}
+
+// ReceiveScriptRegistrationMessage is the canonical JSON message signed to
+// bind a receive script to a mailbox principal.
+type ReceiveScriptRegistrationMessage struct {
```
Roasbeef: Why are we combining JSON and protos here?

indexer/client.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return sig.Serialize(), nil
+}
+
+// RegisterReceiveScriptTaproot registers a single P2TR receive script using a
+// schnorr signature proof under the output key.
+func (c *Client) RegisterReceiveScriptTaproot(ctx context.Context,
```
Roasbeef, line 473: Ah ok, this is something like an `lndclient` over the base indexer RPCs.

indexer/client.go
```diff
@@ -0,0 +1,224 @@
...
+	Nonce       string `json:"nonce"`
+}
+
+// MarshalCanonical returns the canonical JSON encoding for m.
+func (m *ReceiveScriptRegistrationMessage) MarshalCanonical() (string, error) {
+	if m.Type == "" {
+		m.Type = registrationMessageType
+	}
+
+	b, err := json.Marshal(m)
```
Roasbeef: How do we know the ordering, etc here will actually be canonical? Can't we just sign the proto?

indexer/client.go
```diff
@@ -64,6 +92,20 @@ type ReceiveScriptRegistrationMessage struct {
 	Nonce       string `json:"nonce"`
 }
 
+// ScriptScopeMessage is the canonical JSON message signed to prove control of
+// a script for a specific purpose (RPC method / feed).
+type ScriptScopeMessage struct {
```
Roasbeef: Same here re JSON inside of protos via gRPC...

indexer/client.go
```diff
@@ -40,6 +40,34 @@ const (
 	// offlineReceiveProofTTL is the lifetime used for proof-gated script
 	// queries. It should be short to limit replay windows.
 	offlineReceiveProofTTL = 10 * time.Minute
+
+	// scriptScopeMessageType is the canonical proof "type" string used for
```
Roasbeef: Also perhaps let's pick a diff name than indexer here? Something like `arkview` perhaps?

indexer/sync_client.go
```diff
@@ -0,0 +1,212 @@
...
+	vtxoCursorNamespace = "vtxo_events"
+
+	// oorCursorNamespace is the namespace used for OOR recipient event cursors.
+	oorCursorNamespace = "oor_recipient_events"
+)
+
+// SyncBackend captures the high-level indexer client methods used by SyncClient.
+//
+// *Client satisfies this interface.
+type SyncBackend interface {
```
Roasbeef, line 29: This layer seems a bit unnecessary. If it's a simple request/response (eg: `GetInfo` to the server), do we need to write anything to disk at all? In that they client will come up again, then just do `GetInfo` again right? 

The case I think we need the records on disk are for the round related messages.

Roasbeef (changes requested):

litbot-9000:

Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #106, opened 2026-02-11
lwwallet: add lightweight in-process wallet

sputn1ck:
## Summary

- Add new `lwwallet` package providing a lightweight, in-process Bitcoin wallet that replaces the LND dependency on the client side
- Implements `wallet.BoardingBackend`, `round.ClientWallet` (`input.Signer` + `input.MuSig2Signer`), and `chainsource.ChainBackend` as drop-in replacements for LND-backed implementations
- Uses LND-compatible HD keyring (`m/1017'/coinType'/family'/0/index`) and Esplora/mempool.space for chain data

### Components

| File | Purpose |
|------|---------|
| `keyring.go` | HD key derivation compatible with LND's BIP43 purpose 1017 |
| `signer.go` | `input.Signer` implementation (Schnorr, key-path, script-path) |
| `musig2.go` | `input.MuSig2Signer` with in-memory session management |
| `esplora.go` | HTTP client for Esplora/mempool.space REST API |
| `chain_backend.go` | `chainsource.ChainBackend` via Esplora polling |
| `boarding_backend.go` | `wallet.BoardingBackend` with taproot script imports |
| `coin_select.go` | Largest-first greedy coin selection for P2TR |
| `wallet.go` | Top-level wallet: addresses, UTXOs, SendCoins, Balance |
| `config.go` / `doc.go` | Configuration and package documentation |

### Test coverage

44 unit tests covering keyring derivation, coin selection, MuSig2 sessions, Esplora client mocking, chain backend registrations, and signer operations.

## Test plan

- [ ] `go test ./lwwallet/...` -- all 44 unit tests pass
- [ ] `go build ./...` -- clean compile
- [ ] `make lint` -- linter passes
- [ ] Integration: systest PR in darepo runs all E2E tests with both LND and lwwallet backends

### DIFF
--- a/lwwallet/boarding_backend.go
+++ b/lwwallet/boarding_backend.go
@@ -1,0 +1,245 @@
+package lwwallet
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"math"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btclog/v2"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/lightninglabs/darepo-client/wallet"
+	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/lightningnetwork/lnd/lnwallet/btcwallet"
+)
+
+// BoardingBackendAdapter implements wallet.BoardingBackend by wrapping
+// btcwallet.BtcWallet for key derivation and script import, while
+// using the Esplora API directly for UTXO queries. This is necessary
+// because btcwallet's internal UTXO tracking skips addresses imported
+// under non-default key scopes (like LND's m/1017' scope), so we
+// query Esplora for UTXOs at imported boarding addresses instead.
+type BoardingBackendAdapter struct {
+	btcWallet   *btcwallet.BtcWallet
+	esplora     *EsploraClient
+	chainParams *chaincfg.Params
+
+	// log is the structured logger for this boarding backend
+	// instance.
+	log btclog.Logger
+
+	// chainKeyScope is the key scope used for script imports.
+	// This matches LND's m/1017'/coinType' derivation.
+	chainKeyScope waddrmgr.KeyScope
+
+	// mu protects importedAddrs.
+	mu sync.Mutex
+
+	// importedAddrs tracks addresses imported via ImportTaprootScript.
+	// ListUnspent queries Esplora for UTXOs at each of these
+	// addresses, since btcwallet's internal UTXO tracking skips
+	// non-default scope addresses in its addRelevantTx credit
+	// marking.
+	importedAddrs map[string]btcutil.Address
+}
+
+// NewBoardingBackendAdapter creates a new boarding backend adapter
+// wrapping the given btcwallet instance.
+func NewBoardingBackendAdapter(btcw *btcwallet.BtcWallet,
+	esplora *EsploraClient, chainParams *chaincfg.Params,
+	coinType uint32, logger btclog.Logger) *BoardingBackendAdapter {
+
+	return &BoardingBackendAdapter{
+		btcWallet:   btcw,
+		esplora:     esplora,
+		chainParams: chainParams,
+		log:         logger,
+		chainKeyScope: waddrmgr.KeyScope{
+			Purpose: keychain.BIP0043Purpose,
+			Coin:    coinType,
+		},
+		importedAddrs: make(map[string]btcutil.Address),
+	}
+}
+
+// DeriveNextKey derives the next key in the specified key family. This
+// delegates to btcwallet's keyring which uses the waddrmgr for HD key
+// derivation following the m/1017'/coinType'/family'/0/index path.
+func (b *BoardingBackendAdapter) DeriveNextKey(ctx context.Context,
+	family keychain.KeyFamily) (*keychain.KeyDescriptor, error) {
+
+	keyRing := keychain.NewBtcWalletKeyRing(
+		b.btcWallet.InternalWallet(),
+		b.chainKeyScope.Coin,
+	)
+
+	desc, err := keyRing.DeriveNextKey(family)
+	if err != nil {
+		return nil, fmt.Errorf("derive next key: %w", err)
+	}
+
+	b.log.DebugS(ctx, "Derived next key",
+		slog.Int("family", int(family)),
+		slog.Int("index", int(desc.Index)))
+
+	return &desc, nil
+}
+
+// ImportTaprootScript imports a taproot script into btcwallet and
+// tracks the resulting address for Esplora-based UTXO queries. The
+// import into btcwallet registers the address for chain notifications
+// via NotifyReceived, while the local address tracking enables
+// ListUnspent to query Esplora directly.
+func (b *BoardingBackendAdapter) ImportTaprootScript(
+	ctx context.Context,
+	script *waddrmgr.Tapscript) (btcutil.Address, error) {
+
+	managedAddr, err := b.btcWallet.ImportTaprootScript(
+		b.chainKeyScope, script,
+	)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"import taproot script: %w", err,
+		)
+	}
+
+	addr := managedAddr.Address()
+
+	// Track the imported address so ListUnspent can query
+	// Esplora for UTXOs at this address. We use Esplora
+	// directly because btcwallet's addRelevantTx skips credit
+	// marking for non-default scope addresses (1017).
+	b.mu.Lock()
+	b.importedAddrs[addr.String()] = addr
+	b.mu.Unlock()
+
+	b.log.DebugS(ctx,
+		"Imported taproot script via btcwallet",
+		slog.String("address", addr.String()),
+		slog.Int("tracked_addrs", len(b.importedAddrs)))
+
+	return addr, nil
+}
+
+// ListUnspent returns UTXOs at all imported boarding addresses by
+// querying the Esplora API directly. This bypasses btcwallet's
+// internal UTXO tracking, which skips credit marking for addresses
+// in non-default key scopes (like LND's m/1017' scope). For each
+// imported address, Esplora's /address/:addr/utxo endpoint is
+// queried and results are filtered by confirmation count.
+func (b *BoardingBackendAdapter) ListUnspent(ctx context.Context,
+	minConfs, maxConfs int32) ([]*wallet.Utxo, error) {
+
+	// Treat maxConfs of 0 as "no upper bound" so callers that
+	// omit the parameter don't accidentally filter everything.
+	if maxConfs == 0 {
+		maxConfs = math.MaxInt32
+	}
+
+	// Get the current tip height for confirmation calculation.
+	tipHeight, err := b.esplora.GetTipHeight()
+	if err != nil {
+		return nil, fmt.Errorf("get tip height: %w", err)
+	}
+
+	b.mu.Lock()
+	addrs := make(map[string]btcutil.Address, len(b.importedAddrs))
+	for k, v := range b.importedAddrs {
+		addrs[k] = v
+	}
+	b.mu.Unlock()
+
+	var utxos []*wallet.Utxo
+
+	for addrStr, addr := range addrs {
+		esploraUtxos, err := b.esplora.GetAddressUtxos(addrStr)
+		if err != nil {
+			b.log.WarnS(ctx,
+				"Failed to query Esplora for address UTXOs",
+				err,
+				slog.String("address", addrStr))
+
+			continue
+		}
+
+		// Compute the pkScript for this address once, then
+		// reuse it for all UTXOs at this address.
+		pkScript, err := txscript.PayToAddrScript(addr)
+		if err != nil {
+			continue
+		}
+
+		for _, eu := range esploraUtxos {
+			// Compute confirmations from block height.
+			var confs int32
+			if eu.Status.Confirmed {
+				confs = tipHeight -
+					int32(eu.Status.BlockHeight) + 1
+			}
+
+			// Apply confirmation filters.
+			if confs < minConfs || confs > maxConfs {
+				continue
+			}
+
+			txid, err := chainhash.NewHashFromStr(eu.Txid)
+			if err != nil {
+				continue
+			}
+
+			utxos = append(utxos, &wallet.Utxo{
+				Outpoint: wire.OutPoint{
+					Hash:  *txid,
+					Index: eu.Vout,
+				},
+				PkScript:      pkScript,
+				Amount:        btcutil.Amount(eu.Value),
+				Confirmations: confs,
+			})
+		}
+	}
+
+	b.log.DebugS(ctx, "ListUnspent called",
+		slog.Int("min_confs", int(minConfs)),
+		slog.Int("max_confs", int(maxConfs)),
+		slog.Int("tracked_addrs", len(addrs)),
+		slog.Int("utxo_count", len(utxos)))
+
+	return utxos, nil
+}
+
+// GetTransaction returns the full transaction for the given txid.
+// It first attempts to fetch from btcwallet's transaction store,
+// falling back to the Esplora API for transactions not yet indexed
+// by the wallet.
+func (b *BoardingBackendAdapter) GetTransaction(ctx context.Context,
+	txid chainhash.Hash) (*wire.MsgTx, error) {
+
+	// Try btcwallet's transaction store first.
+	tx, err := b.btcWallet.FetchTx(txid)
+	if err == nil && tx != nil {
+		return tx, nil
+	}
+
+	// Fall back to Esplora for transactions not in the wallet DB.
+	b.log.DebugS(ctx,
+		"Transaction not in wallet, falling back to Esplora",
+		slog.String("txid", txid.String()))
+
+	tx, err = b.esplora.GetRawTx(txid)
+	if err != nil {
+		return nil, fmt.Errorf("get transaction: %w", err)
+	}
+
+	return tx, nil
+}
+
+// Compile-time check that BoardingBackendAdapter implements
+// wallet.BoardingBackend.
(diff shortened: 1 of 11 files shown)


sputn1ck:
@claude review pls

litbot-9000:

Roasbeef:
@claude review this

lwwallet/boarding_backend.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	// Compute and store the pkScript.
+	pkScript, err := txscript.PayToTaprootScript(outputKey)
+	if err != nil {
+		return nil, fmt.Errorf("create pkscript: %w", err)
+	}
+
+	b.mu.Lock()
+	b.monitoredScripts[addr.String()] = scriptInfo{
```
Roasbeef: I think this needs to be persisted, it needs to survive restarts to reflect in the current balance.

lwwallet/chain_backend.go
```diff
@@ -0,0 +1,597 @@
...
+
+	oldHeight := b.bestHeight.Load()
+	if newHeight <= oldHeight {
+		// No new blocks, but still check registrations in case
+		// of reorgs or newly broadcast transactions.
+		b.checkConfirmations()
+		b.checkSpends()
+		return
+	}
+
```
Roasbeef, line 432: What about re-orgs?

lwwallet/chain_backend.go
```diff
@@ -0,0 +1,597 @@
+package lwwallet
```
Roasbeef, line 1: Entire package is missing logging rn.

lwwallet/coin_select.go
```diff
@@ -0,0 +1,142 @@
...
+// with the given target amount and fee rate. Uses a largest-first strategy
+// for simplicity and predictability.
+//
+// Parameters:
+//   - available: UTXOs available for selection
+//   - targetAmount: amount to send (excluding fees)
+//   - feePerVByte: fee rate in satoshis per virtual byte
+//   - numOutputs: number of outputs in the transaction (excluding change)
+//   - dustLimit: minimum output value to avoid creating dust
+func SelectCoins(available []*walletUtxo, targetAmount btcutil.Amount,
```
Roasbeef: Just re-use btcwallet coin selection here?

Roasbeef: It's exposed as libraries.

lwwallet/keyring.go
```diff
@@ -0,0 +1,202 @@
...
+// lnd's keychain package so that key locators (family + index pairs) produce
+// identical keys regardless of whether they were derived by LND or by this
+// lightweight keyring.
+//
+// The keyring tracks the next unused index for each key family in memory.
+// On restart, callers must restore index state via RestoreState before
+// deriving new keys to avoid index reuse.
+type Keyring struct {
+	// purposeKey is the derived key at m/1017' (hardened BIP43 purpose).
+	purposeKey *hdkeychain.ExtendedKey
```
Roasbeef: Hmm, don't we need an _even smaller_ wrapper layer here? In that we can re-use all the `btcwallet` code and connect it to this backend instead? The defined backend doesn't need to live in `btcwallet`, should be able to get by with an index. 

If all of this is just in memory, what's the purpose of this wallet? It would need to do rescans everytime to get new data, and I don't see any rescan logic  in here rn.

lwwallet/musig2.go
```diff
@@ -0,0 +1,357 @@
...
+		keyring:  keyring,
+		sessions: make(map[input.MuSig2SessionID]*localMuSig2Session),
+	}
+}
+
+// MuSig2CreateSession creates a new MuSig2 signing session using the local
+// key identified by the key locator. The complete list of all public keys
+// must be provided including the local signer's key. Other signers' nonces
+// can be provided upfront if already known.
+func (m *MuSig2SessionManager) MuSig2CreateSession(
```
Roasbeef: Same here, recreating a lot of what we already have.

Roasbeef (dismissed):

lwwallet/boarding_backend.go
sputn1ck: This PR intentionally did not include any persistance to keep it light, but can add

sputn1ck:
@claude review the last 3 commits

lwwallet/boarding_backend.go
bhandras: Let's add an issue tracking these follow ups.

lwwallet/boarding_backend.go
```diff
@@ -0,0 +1,212 @@
...
+		for _, eu := range esploraUtxos {
+			// Calculate confirmations.
+			var confs int32
+			if eu.Status.Confirmed {
+				confs = tipHeight -
+					int32(eu.Status.BlockHeight) + 1
+			}
+
+			// Filter by confirmation range.
+			if confs < minConfs || confs > maxConfs {
```
bhandras, line 187: If maxConfs defaults to zero then this will filter everything that's confirmed.

lwwallet/chain_backend.go
bhandras: +1, logging should be added.

lwwallet/chain_backend.go
```diff
@@ -0,0 +1,597 @@
...
+		blockRegs:    make(map[uint64]*blockRegistration),
+		stopCh:       make(chan struct{}),
+	}
+}
+
+// Start initializes the chain backend by fetching the current chain tip
+// and starting the polling loop.
+func (b *ChainBackend) Start() error {
+	// Fetch the initial chain tip.
+	height, err := b.esplora.GetTipHeight()
```
bhandras, line 125: I think the safe way to get the tip height and hash without risking drift would be to use `GetTipHeight()` like you do here, but then use https://github.com/blockstream/esplora/blob/master/API.md#get-block-heightheight to get the hash for that height.

lwwallet/chain_backend.go
```diff
@@ -0,0 +1,597 @@
...
+	if err != nil {
+		log.WarnS(context.TODO(), "Poll tip height failed", err)
+		return
+	}
+
+	oldHeight := b.bestHeight.Load()
+	if newHeight <= oldHeight {
+		// No new blocks, but still check registrations in case
+		// of reorgs or newly broadcast transactions.
+		b.checkConfirmations()
```
bhandras, line 428: I don't think is is correct as is, as on confirmation we'll remove conf and spend registrations.

bhandras: Perhaps this could also be added to the follow up issue as the base case is covered as is.

lwwallet/chain_backend.go
```diff
@@ -0,0 +1,597 @@
...
+		select {
+		case <-reg.cancelCh:
+			continue
+
+		default:
+		}
+
+		if reg.txid == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
bhandras: This is important.

lwwallet/coin_select.go
bhandras: +1, it'd greatly reduce the diff.

lwwallet/config.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/btcsuite/btcd/chaincfg"
+)
+
+// Config holds the configuration for the lightweight wallet.
+type Config struct {
+	// Seed is the 32-byte master seed used for HD key derivation. The
+	// caller is responsible for seed generation, encryption at rest, and
+	// BIP39 mnemonic handling. The wallet only uses the raw seed bytes.
+	Seed [32]byte
```
bhandras, line 30: I don't see this used anywhere yet.

lwwallet/doc.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/btcsuite/btclog/v2"
+)
+
+// Subsystem defines the logging code for this subsystem.
+const Subsystem = "LWWL"
+
+// log is a logger that is initialized with no output filters. This means the
+// package will not perform any logging by default until the caller requests
+// it.
+var log = btclog.Disabled
```
bhandras: Oh no!!!

bhandras (commented):
Nice!! Looking pretty good, but imo we still need a few fixes before merge.

Could you pls squash fix commits to the first one, otherwise it's a bit messy to review. Also could you pls ask the agent to reply to fixed comments so its easier to build context.

lwwallet/config.go
sputn1ck: This is passed over to the btcwallet constructor

lwwallet/coin_select.go
sputn1ck: there is no more coin selection here.

lwwallet/chain_backend.go
sputn1ck: Added structured logging throughout the package (27 log statements). Lifecycle events at Info level, registrations/fulfillments/operations at Debug, errors at Warn. Logger setup moved to dedicated log.go with LWWL subsystem.

lwwallet/boarding_backend.go
sputn1ck: Created follow-up issue: https://github.com/lightninglabs/darepo-client/issues/127

For now, `wallet.Start()` re-imports persisted boarding addresses on startup so in-memory backends recover their state.

lwwallet/chain_backend.go
sputn1ck: Created follow-up issue: https://github.com/lightninglabs/darepo-client/issues/128

lwwallet/keyring.go
sputn1ck: Done — replaced the hand-rolled keyring, signer, MuSig2, and coin selection with btcwallet.BtcWallet. The package now wraps LND's btcwallet directly with an Esplora-based chain.Interface.

lwwallet/musig2.go
sputn1ck: Removed — btcwallet handles MuSig2 sessions now.

lwwallet/boarding_backend.go
sputn1ck: Fixed — `ListUnspent` now treats `maxConfs == 0` as no upper bound (`math.MaxInt32`).

lwwallet/chain_backend.go
sputn1ck: Fixed — all three call sites (`ChainBackend.Start()`, `EsploraChainService.Start()`, `EsploraChainService.GetBestBlock()`) now fetch height first, then resolve the hash via `/block-height/:height`.

lwwallet/chain_backend.go
sputn1ck: Added to follow-up issue: https://github.com/lightninglabs/darepo-client/issues/128

lwwallet/chain_backend.go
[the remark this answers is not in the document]
sputn1ck: Implemented — `checkSingleConf` dispatches to `checkConfByTxid` when txid is non-nil and `checkConfByScript` when txid is nil (scans the script's UTXOs for a confirmed transaction matching the pkScript).

lwwallet/doc.go
sputn1ck: Moved logger setup to dedicated `log.go` file (matching the pattern from round, oor, actor packages). `doc.go` now contains only the package documentation.

lwwallet/chain_backend.go
sputn1ck: Done — see reply to Roasbeef's comment above.

lwwallet/log.go
```diff
@@ -1,14 +1,20 @@
...
 
 // Subsystem defines the logging code for this subsystem.
-const Subsystem = "SVRC"
+const Subsystem = "LWWL"
 
 // log is a logger that is initialized with no output filters. This means the
 // package will not perform any logging by default until the caller requests
 // it.
-var log = btclog.Disabled
+var log btclog.Logger
```
bhandras: Can you pls avoid using global logger?

lwwallet/chain_backend.go
```diff
@@ -0,0 +1,737 @@
...
+
+	return btcutil.Amount(math.Ceil(bestRate)), nil
+}
+
+// BestBlock returns the current best block height and hash.
+func (b *ChainBackend) BestBlock(_ context.Context) (int32,
+	chainhash.Hash, error) {
+
+	height := b.bestHeight.Load()
+	hashVal := b.bestHash.Load()
```
bhandras: I think instead of these atomic values we should guard these with the backent's mutex, otherwise there's a (very) small chance of the poll loop setting one but not the other yet while someone asks for BestBlock.

lwwallet/config.go
bhandras: Ah ok, missed that.

lwwallet/wallet.go
```diff
@@ -0,0 +1,302 @@
...
+	chainSvc := NewEsploraChainService(esplora, cfg.PollInterval)
+
+	// The ChainBackend implements chainsource.ChainBackend for the
+	// actor system (confirmation/spend/block registrations).
+	chainBackend := NewChainBackend(esplora, cfg.PollInterval)
+
+	coinType := coinTypeForNet(cfg.ChainParams)
+	blockCache := blockcache.NewBlockCache(defaultBlockCacheSize)
+
+	btcw, err := btcwallet.New(btcwallet.Config{
```
bhandras, line 111: Nice!

lwwallet/wallet.go
```diff
@@ -0,0 +1,302 @@
...
+		CoinType:       coinType,
+		RecoveryWindow: cfg.RecoveryWindow,
+		LoaderOptions: []btcwallet.LoaderOption{
+			btcwallet.LoaderWithLocalWalletDB(
+				dbDir, false, 60*time.Second,
+			),
+		},
+	}, blockCache)
+	if err != nil {
+		os.RemoveAll(dbDir)
```
bhandras: I don't think we want this right? It might accidentally delete an existing wallet!

lwwallet/wallet.go
```diff
@@ -0,0 +1,302 @@
...
+// service goroutine to fully exit before removing the DB directory
+// to avoid racing with btcwallet's internal writes.
+func (w *Wallet) Stop() {
+	log.InfoS(context.Background(), "Stopping lightweight wallet",
+		slog.String("db_dir", w.dbDir))
+
+	_ = w.btcWallet.Stop()
+	w.chainSvc.WaitForShutdown()
+	_ = w.chainBackend.Stop()
+	_ = os.RemoveAll(w.dbDir)
```
bhandras: Why do we delete the db dir?

lwwallet/wallet.go
```diff
@@ -0,0 +1,302 @@
...
+
+	addr, err := w.btcWallet.NewAddress(
+		lnwallet.TaprootPubkey, false,
+		lnwallet.DefaultAccountName,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	log.DebugS(context.Background(), "Generated new P2TR address",
```
bhandras: Let's use the passed in context.

lwwallet/wallet.go
```diff
@@ -0,0 +1,302 @@
...
+// wallet-managed addresses. Confirmed balance requires at least 1
+// confirmation.
+func (w *Wallet) Balance(
+	_ context.Context) (btcutil.Amount, btcutil.Amount, error) {
+
+	// Log sync state for debugging. The sync height determines
+	// whether confirmed transactions are counted.
+	syncedTo := w.btcWallet.InternalWallet().SyncedTo()
+	chainSynced := w.btcWallet.InternalWallet().ChainSynced()
+	log.DebugS(context.Background(), "Checking wallet balance",
```
bhandras: Same, let's use function arg context.

lwwallet/wallet.go
```diff
@@ -0,0 +1,302 @@
...
+	total, err := w.btcWallet.ConfirmedBalance(
+		0, "",
+	)
+	if err != nil {
+		return 0, 0, fmt.Errorf("get total balance: %w", err)
+	}
+
+	unconfirmed := total - confirmed
+
+	log.DebugS(context.Background(), "Wallet balance result",
```
bhandras: Same

bhandras (commented):
Very nice! Last set of comments from me.

lwwallet/log.go
sputn1ck: Done — all structs now receive a `btclog.Logger` through their constructors. `Config.Logger` flows from `Wallet.New()` to all sub-components (ChainBackend, EsploraChainService, BoardingBackendAdapter, EsploraClient). Falls back to the package-level logger if nil.

lwwallet/chain_backend.go
sputn1ck: Good catch — replaced `atomic.Int32`/`atomic.Value` with plain `int32`/`chainhash.Hash` fields guarded by the existing `mu sync.Mutex`. Both values are now read and written atomically under the same lock, so `BestBlock()` can never return a mismatched height/hash pair.

lwwallet/wallet.go
sputn1ck: Agreed — moved DB directory ownership to the caller. `Config` now takes a `DBDir string` path and the wallet no longer creates or deletes directories. The systest harness creates a temp dir and cleans it up in its own `Stop()`.

lwwallet/wallet.go
sputn1ck: Removed — the wallet no longer owns or deletes the DB directory. The caller provides `Config.DBDir` and manages its lifecycle.

lwwallet/wallet.go
sputn1ck: Fixed — changed `_ context.Context` to `ctx context.Context` and pass it to the log calls in `NewAddress`, `Balance`, and all `BoardingBackendAdapter` methods.

lwwallet/wallet.go
sputn1ck: Fixed.

lwwallet/wallet.go
sputn1ck: Fixed.

lwwallet/boarding_backend.go
sputn1ck: Created: #127 (persistence for monitored scripts/boarding addresses) and #128 (re-org detection in chain backend).

lwwallet/log.go
sputn1ck: Removed `log.go` entirely — no more global mutable logger state. When `Config.Logger` is nil, the fallback is now `btclog.Disabled` (silent discard). The `Subsystem` constant moved to `config.go` for callers constructing sub-loggers.

bhandras (approved):
LGTM 🥏

PR #109, opened 2026-02-11
mailboxrpc: merge RPC options and guard empty services

bhandras:
This PR addresses two codegen robustness issues:

1) Merge variadic RPCOptions (instead of silently dropping all but opts[0]).
2) Avoid method-only imports/ident resolution for services with zero RPC methods.

Commits:
- 6bc7668 mailboxrpc: merge variadic RPC options
- 17d539f mailboxrpc: avoid method-only imports for empty services


bhandras:
Superseded by #110 (rebased on origin/main with clean history).

PR #110, opened 2026-02-11
mailboxrpc: merge RPC options and guard empty services

bhandras:
This PR addresses two codegen robustness issues:

1) Merge variadic RPCOptions (instead of silently dropping all but opts[0]).
2) Avoid method-only imports/ident resolution for services with zero RPC methods.

Commit:
- eb5d7cd mailboxrpc: merge variadic RPC options (includes empty-service guard)


litbot-9000:

PR #112, opened 2026-02-13
db/actordelivery: isolate durable-store migrations and constructor

bhandras:
## Summary

This PR is intentionally split into two commits:

1. `db: use config-based internal migration runner`
- Introduces a shared migration runner in `db/migrate`.
- Uses a config-based API for migration execution.
- Centralizes shared migration concerns, including:
  - `ErrMigrationDowngrade`
  - Postgres schema replacements (`PostgresSchemaReplacements()`).

2. `db/actordelivery: split actor-delivery store package`
- Moves durable actor-delivery persistence into `db/actordelivery`.
- Adds isolated actor-delivery SQL migrations and sqlc bindings.
- Keeps actor-delivery migration state in a dedicated migrations table
  (`actor_delivery_schema_migrations`).
- Provides construction from raw DB handles via
  `NewTxAwareDeliveryStoreFromDB(...)`.
- Renames the concrete store type to `actordelivery.Store`.

## Motivation

We want server-side code to reuse durable actor-delivery SQL persistence from
`darepo-client` directly, without copying implementation files.

That requires:
- reusable/shared migration infrastructure, and
- an isolated actor-delivery package with its own migration stream and
  constructor surface.

This PR provides that separation with a small, reviewable commit split.

## Design Notes

- Main DB migrations and actor-delivery migrations both use the shared
  `db/migrate` runner.
- Postgres schema text replacement is shared via `db/migrate`, instead of being
  implemented separately inside actor-delivery.
- Actor-delivery migrations remain isolated in
  `db/actordelivery/migrations/*`.

## Follow-up (server)

After this lands, server code can import `db/actordelivery`, run isolated
actor-delivery migrations with its preferred migration table name, and stop
vendoring/copied durable-mailbox DB code.


db/actordelivery/migrations/000001_durable_mailbox.down.sql
```diff
@@ -0,0 +1,29 @@
+-- Rollback durable mailbox migration.
```
sputn1ck, line 1: would it make sense to put the db folder in `mailbox/` folder?

db/actordelivery/migrations.go
```diff
@@ -0,0 +1,201 @@
...
+
+	stat, err := f.Stat()
+	if err != nil {
+		return nil, err
+	}
+	if stat.IsDir() {
+		return f, nil
+	}
+
+	return newReplacerFile(f, t.replaces)
```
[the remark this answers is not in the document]
bhandras: Good catch, fixed in follow-up commit:
https://github.com/lightninglabs/darepo-client/commit/5fb919c4e36bb0ddeeb360259de5f8297f681d0b

`replacerFS.Open` now closes the opened file when `Stat()` fails and also closes it when `newReplacerFile(...)` returns an error.

db/actordelivery/migrations.go
```diff
@@ -0,0 +1,201 @@
...
+	return t.parentFile.Stat()
+}
+
+func (t *replacerFile) Read(b []byte) (int, error) {
+	return t.buf.Read(b)
+}
+
+func (t *replacerFile) Close() error {
+	return nil
+}
```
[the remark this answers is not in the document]
bhandras: Fixed in the same follow-up commit:
https://github.com/lightninglabs/darepo-client/commit/5fb919c4e36bb0ddeeb360259de5f8297f681d0b

`replacerFile.Close()` now forwards to `parentFile.Close()` so the wrapped descriptor is always released.

db/actordelivery/migrations/000001_durable_mailbox.down.sql
bhandras: I kept this under `db/actordelivery` intentionally so the migration runner and constructor stay as a DB integration unit that can be reused without coupling it to the higher-level `mailbox/` package layout.

If you prefer, I can move it under `mailbox/` in a follow-up, but for this extraction PR I wanted to mirror the same isolated-store shape we are using server-side.

db/actordelivery/README.md
```diff
@@ -0,0 +1,17 @@
+# Actor Delivery SQL Store
```
sputn1ck: I wonder if the readme is warranted. Should something like this live in `doc.go` ? Is doc.go a common pattern? Maybe doc.go helps llms via lsp?

db/actordelivery/store.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/btcsuite/btclog/v2"
+	"github.com/lightninglabs/darepo-client/baselib/actor"
+	"github.com/lightninglabs/darepo-client/db"
+	"github.com/lightninglabs/darepo-client/db/sqlc"
+	"github.com/lightningnetwork/lnd/clock"
+)
+
+// NewTxAwareDeliveryStoreFromDB constructs a transaction-aware actor delivery
+// store from a raw sql.DB connection and backend type.
+func NewTxAwareDeliveryStoreFromDB(sqlDB *sql.DB, backend sqlc.BackendType,
```
sputn1ck, line 17: IMO it would make sense to have sqlc emit the go files in this isolated package.

db/actordelivery/README.md
bhandras: done

db/actordelivery/store.go
bhandras: yeah, good p, done

db/sqlc/migrations/000004_durable_mailbox.down.sql
```diff
@@ -1,29 +1,6 @@
...
-DROP INDEX IF EXISTS idx_mailbox_messages_promise;
-DROP INDEX IF EXISTS idx_mailbox_messages_lease;
-DROP INDEX IF EXISTS idx_mailbox_messages_available;
-DROP TABLE IF EXISTS mailbox_messages;
+-- Compatibility no-op.
+--
+-- Durable mailbox SQL lives in db/actordelivery/migrations/ and is
+-- tracked with its own migration table. Version 4 in the main
+-- schema stream is a no-op.
+SELECT 1;
```
sputn1ck: do we really need this?

db/actor_delivery_migrations.go
```diff
@@ -0,0 +1,59 @@
...
+	// actorDeliveryLatestMigrationVersion is the latest migration
+	// version for isolated actor-delivery migrations.
+	//
+	// NOTE: This MUST be updated when a new actor-delivery migration
+	// is added.
+	actorDeliveryLatestMigrationVersion uint = 1
+
+	// actorDeliveryMigrationsTable is the migration bookkeeping table
+	// used for isolated actor-delivery migrations.
+	actorDeliveryMigrationsTable = "actor_delivery_schema_migrations"
```
sputn1ck: i somehow feel this file here is duplicated and not necessary

sputn1ck (approved):
LGTM!

db/sqlc/migrations/000004_durable_mailbox.down.sql
bhandras: no, removed

db/actor_delivery_migrations.go
bhandras: thanks! agreed, this is now a bit nicer hopefully.

PR #116, opened 2026-02-18
mailbox+serverconn: durable response store and unified connector boundary

Roasbeef:
In this PR, we lay the groundwork for making the client's mailbox RPC
layer crash-safe by introducing a `Store` abstraction for response
persistence and expanding `serverconn` into the unified connector
boundary for all client↔server mailbox traffic.

## mailbox/client: durable response store

We add a `Store` interface to `mailbox/client` that persists the pull
cursor and response payloads keyed by correlation ID. The client's main
`run` loop is then wired to use this store: on startup we load the
persisted cursor so pulls resume from where we left off, each inbound
response envelope is written to the store before notifying in-memory
waiters, and the cursor is persisted before acking with the remote
mailbox. `AwaitRPC` now does a two-tier lookup — first the in-memory
pending map, then the store — so responses survive a restart between
`PutResponse` and the caller's `AwaitRPC`.

A `MemoryStore` ships for tests and short-lived processes that don't
need crash safety. The `Config.Store` field defaults to it when unset,
so existing callers are unaffected.

## serverconn: unified connector boundary

The `serverconn` package is expanded from its previous stub into the
single ingress/egress boundary for mailbox traffic. The core pieces:

**Ack state machine** (`types.go`): Four cursors — `PullCursor`,
`DispatchCommittedTo`, `AckTarget`, `AckCommittedTo` — enforce the
invariant that `AckUpTo` only advances after durable local dispatch.
TLV encode/decode lets the state persist via the durable actor
checkpoint mechanism.

**TLV message types** (`actor.go`): `SendClientEventRequest` wraps
outbound FSM events in `anypb.Any` for type-preserving TLV
serialization. `SendRPCRequest` marshals the full mailbox envelope for
the unary path. Both carry stable TLV type identifiers for the durable
actor codec. The actor struct gains a response registry, egress
handlers that build proper mailbox envelopes, and ingress lifecycle
management.

**Ingress loop** (`ingress.go`): A self-driven background goroutine
that long-polls the remote mailbox via `Edge.Pull`, routes
`KIND_RESPONSE` envelopes to the response registry for unary waiters,
routes `KIND_REQUEST`/`KIND_EVENT` envelopes through the dispatch table
to local durable actors, and advances the ack watermark only after
committed dispatch. Transient failures get exponential backoff w/
jitter.

**Unary facade** (`unary_facade.go`): Implements `mailboxrpc.RPCClient`
on top of the connector. `SendRPC` goes directly through `Edge.Send`
(no durable mailbox — caller retries on failure), while `AwaitRPC`
blocks on the response registry channel until the ingress loop delivers
a matching response.

## Tests

21 tests across three files cover the ack state machine invariants,
full ingress loop integration against an in-memory mailbox edge
(dispatch, ack, checkpoint restart, shutdown), and the unary facade
round-trip including concurrent inflight isolation and context
cancellation. See each commit message for a detailed description w.r.t
the incremental changes.


serverconn/actor.go
```diff
@@ -45,91 +90,407 @@ type SendClientEventRequest struct {
...
+	records := []tlv.Record{
+		tlv.MakePrimitiveRecord(protoPayloadRecordType, &payload),
+	}
+
+	stream, err := tlv.NewStream(records...)
+	if err != nil {
+		return err
+	}
+
+	if _, err := stream.DecodeWithParsedTypes(r); err != nil {
```
Roasbeef, line 199: Could just store pure protos here and bypass the TLV round trip since we send protos in the end.

mailbox/conn/response_registry.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if pending, ok := r.pending[id]; ok {
+		select {
+		case waiter.Ch <- pending.Envelope:
+			delete(r.pending, id)
+		default:
+		}
+	}
+
+	return waiter.Ch
```
Roasbeef: Should this be an `actor.Future[mailboxpb.Envelope]` instead?

mailbox/conn/response_registry.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.pruneStaleLocked(time.Now())
+
+	if waiter, ok := r.waiters[id]; ok {
+		select {
+		case waiter.Ch <- env:
```
Roasbeef: This would then complete the promise.

serverconn/runtime.go
```diff
@@ -0,0 +1,92 @@
...
+	return "serverconn-" + localMailboxID
+}
+
+// Runtime bundles the connector behavior, durable actor wrapper, and unary
+// facade into a single composition helper. The actual application wiring can
+// instantiate this now and integrate it into a higher-level daemon bootstrap
+// later.
+type Runtime struct {
+	connector *ServerConnectionActor
+	durable   *actor.DurableActor[ServerConnMsg, ServerConnResp]
```
Roasbeef: Embed instead? Then an actual actor.

serverconn/actor.go
```diff
@@ -35,101 +83,441 @@ type ServerConnResp interface {
...
+	msgIDBytes := []byte(msgID)
+
+	idempotencyKey := m.IdempotencyKey
+	if idempotencyKey == "" {
+		idempotencyKey = mailboxconn.
+			StableEventIdempotencyKey(anyBytes)
+	}
+	idempotencyBytes := []byte(idempotencyKey)
+
+	records := []tlv.Record{
```
Roasbeef: We can make a composite type wrapper here, which handles this boiler plate for us (wrap proto in TLV for now).

serverconn/actor.go
Roasbeef: Ah we require TLV serialization rn with the package. So let's do the composite generic type thing mentioned below.

serverconn/runtime.go
Roasbeef: So it can be registered directly.

Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
@claude review this

mailbox/conn/ack_state.go
```diff
@@ -0,0 +1,123 @@
...
+// AckState tracks the four cursor variables that govern safe ack progression.
+// All fields are monotonic and must not decrease during normal operation.
+//
+// The state machine enforces:
+//
+//	ack_committed_to <= dispatch_committed_to
+//
+// Cursor never advances past non-durable local work. Repeated acks are safe
+// and idempotent.
+type AckState struct {
```
Roasbeef, line 36: Moved this here so it can be used across other instances. As part of the durable actor, you store this state for server/client conn and use that to remember what was sent vs not.

Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #117, opened 2026-02-18
oor: add package artifact store and local unroll resolver

bhandras:
## Summary
This PR adds a full local OOR artifact persistence + resolution substrate and wires it into incoming/outgoing OOR runtime paths.

It now also includes the correctness/safety follow-ups requested in review:
- enum tables for persisted OOR code values,
- immutable package/binding invariants,
- actor context-tx propagation for atomic durable writes,
- resolver max-depth guard,
- typed OOR persistence surfaces (`enum` types, `time.Time`, `fn.Option`, `keychain.KeyDescriptor`),
- explicit VTXO spent/status consistency enforcement used by unspent accounting.

## Final Commit Stack
1. `db: add OOR artifact schema and sqlc surfaces`
2. `db: add OOR artifact persistence store and query APIs`
3. `db: enforce spent/status consistency for unspent VTXO accounting`
4. `oor: persist incoming and outgoing package artifacts`
5. `db: add ResolveUnrollPackages API for OOR artifact lookup`

### 1) OOR artifact schema + SQLC surfaces
Adds OOR artifact tables:
- `oor_packages`
- `oor_package_checkpoints`
- `oor_vtxo_bindings`
- `oor_recipient_cursors`
- `owned_receive_scripts`

Adds enum tables and FK wiring:
- `oor_package_directions`
- `oor_vtxo_binding_link_kinds`
- `owned_receive_script_sources`

Also adds indexes for package listing, checkpoint ordering, and session binding lookup.

### 2) DB OOR persistence API/store
Introduces `OORArtifactPersistenceStore` with APIs for:
- package upsert/retrieval/listing,
- outpoint binding upsert/retrieval,
- recipient cursor CRUD,
- owned receive-script metadata CRUD.

Materialized package responses include Ark PSBT, finalized checkpoint PSBTs, bindings, and matched binding when applicable.

### 3) OOR runtime wiring
Adds a narrow `PackagePersistence` interface at the OOR boundary and wires:
- incoming materialization -> package + `created_output` bindings,
- outgoing finalize-accepted -> package + `consumed_input` bindings.

### 4) Local unroll resolver API
Adds `ResolveUnrollPackages(ctx, outpoint)`:
- resolves package bound to target outpoint,
- walks checkpoint-input ancestry through local bindings,
- returns ancestor-first deterministic ordering,
- returns unresolved checkpoint inputs when local ancestry is incomplete,
- enforces max traversal depth as defense in depth.

### 5) VTXO spent/status consistency hardening
Adds dedicated DB consistency commit:
- `UpdateVTXOStatus(..., Spent)` now synchronizes `spent=true`,
- unspent list/count/sum queries require both `spent=false` and `status!=Spent(4)`,
- regression test added for the invariant.

This keeps OOR/round-facing spend accounting consistent when OOR marks inputs spent via status transitions.

## Review-Driven Correctness Changes Included
- Binding identity immutability: `(outpoint, kind) -> session` cannot silently rebind.
- Direction immutability: package direction is fixed per `session_id`.
- Transaction atomicity: `TransactionExecutor` joins actor tx from context (no nested tx when actor tx is present).
- Resolver bounded traversal: explicit max-depth error path.
- Receiver polling intent clarified in schema/store comments for cursors + owned scripts.

## Validation
Executed locally:
- `make lint`
- `go test ./...`

Both pass on this branch.

## Out Of Scope
Still out of scope for this PR:
- network/indexer hydration of missing ancestry,
- unroll broadcast executor/policy layer,
- broader round model redesign beyond the explicit spent/status consistency fix included here.


db/oor_artifact_store.go
```diff
@@ -0,0 +1,855 @@
...
+	OORPackageDirectionOutgoing = "outgoing"
+
+	// OORPackageLinkKindCreatedOutput identifies bindings where the local
+	// outpoint was created by the Ark transaction.
+	OORPackageLinkKindCreatedOutput = "created_output"
+
+	// OORPackageLinkKindConsumedInput identifies bindings where the local
+	// outpoint was consumed as an OOR input.
+	OORPackageLinkKindConsumedInput = "consumed_input"
+)
```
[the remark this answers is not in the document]
bhandras: Thanks for the suggestion.

I kept these constants duplicated for now because `db` importing `oor` here would introduce an import cycle in the current test graph (`oor` package tests import `db`).

I agree with the single-source concern; the cleaner long-term direction is to move the shared direction/link-kind constants into a neutral package that both `oor` and `db` can import without cycle risk.

db/oor_artifact_store.go
```diff
@@ -0,0 +1,855 @@
...
+			LastUsedAt:      rows[i].LastUsedAt,
+		})
+	}
+
+	return records, nil
+}
+
+// loadPackageBundleBySession materializes a full package bundle by session ID.
+func loadPackageBundleBySession(ctx context.Context, q OORArtifactStore,
+	sessionID []byte) (*OORPackageBundle, error) {
```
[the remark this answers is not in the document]
bhandras: Good call — applied.

Folded into `ba93b21` (`db: add OOR artifact persistence store and query APIs`):
- added `materializePackageBundle(ctx, q, pkgRow)`
- made `loadPackageBundleBySession` a thin wrapper that only fetches the row and delegates
- updated `ListPackages` to call `materializePackageBundle` directly from `ListOORPackages*` rows
- updated `GetPackageForOutpoint` to build `sqlc.OorPackage` from the join row and call `materializePackageBundle`

This removes the redundant `GetOORPackage` round-trip in both paths.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    -- Primary key keeps one checkpoint row per package index.
+    PRIMARY KEY (session_id, checkpoint_index),
+
+    -- Session foreign key keeps checkpoint rows tied to package lifecycle.
+    FOREIGN KEY (session_id) REFERENCES oor_packages(session_id)
+        ON DELETE CASCADE
+);
+
+-- oor_vtxo_bindings maps local outpoints to stored OOR package sessions.
+CREATE TABLE IF NOT EXISTS oor_vtxo_bindings (
```
Roasbeef, line 94: Local outpoint here meaning a given vtxo?

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    UNIQUE (session_id, output_index, link_kind),
+
+    -- Session foreign key keeps bindings tied to package lifecycle.
+    FOREIGN KEY (session_id) REFERENCES oor_packages(session_id)
+        ON DELETE CASCADE
+);
+
+-- oor_recipient_cursors stores the last processed recipient event for each
+-- tracked recipient script.
+CREATE TABLE IF NOT EXISTS oor_recipient_cursors (
```
Roasbeef, line 148: Who are the cursors for in this case? The sender/receiver or it's an API thing?

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,152 @@
...
+);
+
+-- owned_receive_scripts stores locally owned receive script metadata used
+-- to drive recipient polling and package attribution.
+CREATE TABLE IF NOT EXISTS owned_receive_scripts (
+    -- pk_script is the owned receive script primary key.
+    pk_script BLOB PRIMARY KEY NOT NULL,
+
+    -- client_key_family is the wallet key family for this script.
+    client_key_family BIGINT NOT NULL,
```
Roasbeef, line 171: Non-blocking, but we can do a future migration to create an independent `keys` table to have a single place where all the key fam/index information is stored.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    client_pubkey BLOB NOT NULL,
+
+    -- operator_pubkey is the operator key used in the checkpoint taptree.
+    operator_pubkey BLOB NOT NULL,
+
+    -- exit_delay is the CSV delay used in the timeout branch.
+    exit_delay BIGINT NOT NULL,
+
+    -- source labels how this script was discovered/registered.
+    source TEXT NOT NULL,
```
Roasbeef: Should this point to an enum table? What kind of values will be stored here?

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    client_key_index BIGINT NOT NULL,
+
+    -- client_pubkey is the client key used in the checkpoint taptree.
+    client_pubkey BLOB NOT NULL,
+
+    -- operator_pubkey is the operator key used in the checkpoint taptree.
+    operator_pubkey BLOB NOT NULL,
+
+    -- exit_delay is the CSV delay used in the timeout branch.
+    exit_delay BIGINT NOT NULL,
```
Roasbeef, line 183: Is this distinct from vtxos? I would've thought we would build on top of the vtxo tables for this?

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+	// SessionID is the package session identifier (Ark txid).
+	SessionID chainhash.Hash
+
+	// OutputIndex is the package output index (or enumerated input index
+	// for consumed-input bindings).
+	OutputIndex uint32
+
+	// LinkKind identifies relation to the package
+	// (created_output/consumed_input).
+	LinkKind string
```
Roasbeef: Should use a enum here.

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+	RecipientPkScript []byte
+
+	// ValueSat is populated for created-output bindings when known.
+	ValueSat *int64
+
+	// CreatedAt is the binding creation timestamp.
+	CreatedAt int64
+
+	// UpdatedAt is the last binding update timestamp.
+	UpdatedAt int64
```
Roasbeef: `time.Time`? 

here and above

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+
+	// LinkKind identifies relation to the package
+	// (created_output/consumed_input).
+	LinkKind string
+
+	// RecipientPkScript is populated for created-output bindings.
+	RecipientPkScript []byte
+
+	// ValueSat is populated for created-output bindings when known.
+	ValueSat *int64
```
Roasbeef: Could be fn.Option

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+type OORPackageBinding struct {
+	// Outpoint is the local VTXO outpoint linked to the package.
+	Outpoint wire.OutPoint
+
+	// SessionID is the package session identifier (Ark txid).
+	SessionID chainhash.Hash
+
+	// OutputIndex is the package output index (or enumerated input index
+	// for consumed-input bindings).
+	OutputIndex uint32
```
Roasbeef, line 173: Do we need this if we have outpoint? Still updating my mental model here and catching up on the oor stuff generally.

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+	MatchedOutpointBinding *OORPackageBinding
+}
+
+// OwnedReceiveScriptRecord is a local script ownership registration row.
+type OwnedReceiveScriptRecord struct {
+	// PkScript is the tracked receive script.
+	PkScript []byte
+
+	// ClientKeyFamily is the local key family.
+	ClientKeyFamily int64
```
Roasbeef: `keychain.KeyDescriptor`?

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+}
+
+// UpsertPackage writes or replaces one finalized OOR package identified by
+// session ID.
+//
+// The method persists the Ark PSBT and then rewrites the checkpoint set in
+// contiguous index order. Existing checkpoint rows for the session are removed
+// first so retries always converge to the latest canonical package.
+func (s *OORArtifactPersistenceStore) UpsertPackage(ctx context.Context,
+	direction string, sessionID chainhash.Hash, ark *psbt.Packet,
```
Roasbeef: `ark` here is the signed VTXO? 

Perhaps we should make a composite struct that includes:
  * the OG vtxo 
  * slice of checkpoint+spend
  * final vtxo

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+		return nil
+	})
+}
+
+// UpsertBinding writes or updates one local outpoint-to-session binding.
+//
+// Bindings are used as the primary lookup surface for unroll preparation.
+// created_output bindings map newly received outputs, while consumed_input
+// bindings map local outpoints spent by an outgoing transfer.
+func (s *OORArtifactPersistenceStore) UpsertBinding(ctx context.Context,
```
Roasbeef, line 409: So a binding is when we have a _new_ vtxo spending at the end of the chain?

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+	})
+}
+
+// GetPackageForOutpoint returns the fully materialized package linked to one
+// local outpoint.
+//
+// The response includes all persisted checkpoints, all known bindings for the
+// session, and the specific binding row that matched the requested outpoint.
+func (s *OORArtifactPersistenceStore) GetPackageForOutpoint(ctx context.Context,
+	outpoint wire.OutPoint) (*OORPackageBundle, error) {
```
Roasbeef, line 532: This is a vtxo outpoint?

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+}
+
+// ListPackages returns all persisted packages, optionally filtered by
+// direction.
+//
+// Each returned entry is fully materialized with checkpoints and binding rows.
+// Callers can therefore consume the result directly without additional per-row
+// lookups.
+func (s *OORArtifactPersistenceStore) ListPackages(ctx context.Context,
+	direction *string) ([]*OORPackageBundle, error) {
```
Roasbeef: Wouldn't we want to further filter by the oupoint that the package is rooted in?

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+// This is a convenience wrapper around ListPackages with the outgoing
+// direction filter preselected.
+func (s *OORArtifactPersistenceStore) ListSentPackages(
+	ctx context.Context) ([]*OORPackageBundle, error) {
+
+	direction := OORPackageDirectionOutgoing
+	return s.ListPackages(ctx, &direction)
+}
+
+// UpsertRecipientCursor writes recipient polling progress for one script.
```
Roasbeef, line 673: Hmm, won't recipients be polling the server?

db/oor_artifact_store.go
```diff
@@ -0,0 +1,914 @@
...
+
+	return &row, nil
+}
+
+// ListRecipientCursors returns all recipient cursor rows ordered by update
+// time.
+//
+// This is primarily used by orchestration code that needs to inspect or resume
+// polling across all tracked scripts.
+func (s *OORArtifactPersistenceStore) ListRecipientCursors(
```
Roasbeef, line 744: Same here we why we're concerned with receipients as the sender. Or it's that at times we send info dierctly to them?

db/oor_unroll_resolver.go
```diff
@@ -0,0 +1,266 @@
...
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/db/sqlc"
+)
+
+// OORUnrollPackages is a local package-resolution result for one outpoint.
+type OORUnrollPackages struct {
+	// TargetOutpoint is the local outpoint requested by the caller.
+	TargetOutpoint wire.OutPoint
```
Roasbeef, line 34: Which will be a VTXO outpoint right?

db/oor_unroll_resolver.go
```diff
@@ -0,0 +1,266 @@
...
+						continue
+					}
+
+					return err
+				}
+
+				nextDepth := node.depth + 1
+				existing, ok := nodes[inputPkg.SessionID]
+				if ok {
+					if nextDepth > existing.depth {
```
Roasbeef, line 150: Perhaps we should enfroce a max depth here? Just defense in depth, as the server/client _should_ apply this as policy on a higher layer.

db/sqlc/queries/oor_artifacts.sql
```diff
@@ -0,0 +1,164 @@
...
+)
+ON CONFLICT (session_id, checkpoint_index) DO UPDATE SET
+    checkpoint_psbt = EXCLUDED.checkpoint_psbt;
+
+-- name: ListOORPackageCheckpoints :many
+SELECT * FROM oor_package_checkpoints
+WHERE session_id = $1
+ORDER BY checkpoint_index ASC;
+
+-- name: UpsertOORVTXOBinding :exec
```
Roasbeef: From my review agent: 
### H-2: Binding Upsert Silently Reassigns Outpoint to Different Session

**Confirmed by**: security-auditor (PoC), differential-review, API-safety (3/6 agents)

**File**: `db/sqlc/queries/oor_artifacts.sql` (UpsertOORVTXOBinding)

The ON CONFLICT clause overwrites `session_id` when the same
`(outpoint_hash, outpoint_index, link_kind)` is upserted with a different
session. This silently rebinds an outpoint to a different package with no
detection or logging.

**Impact**: If a malicious operator replays or fabricates an incoming
package referencing an existing outpoint, the binding redirects to the
attacker's session. The resolver then returns wrong ancestry data during
unroll, potentially causing incorrect broadcast ordering during forced
exit.

**Trust model note**: Partially mitigated by this being a local client
store, but `created_output` bindings represent factual statements about
which transaction created an outpoint -- these should be immutable once
set.

**Fix**: Either error on conflict when `session_id` differs, or at
minimum log a warning:
```go
existing, err := q.GetOORVTXOBindingByOutpoint(ctx, ...)
if err == nil && !bytes.Equal(existing.SessionID, sessionID[:]) {
    return fmt.Errorf("binding conflict: outpoint already bound to session %x", existing.SessionID)
}
```

db/sqlc/queries/oor_artifacts.sql
```diff
@@ -0,0 +1,164 @@
+-- OOR artifact store queries.
+
+-- name: UpsertOORPackage :exec
```
Roasbeef: From my review agent:
### H-3: Direction Mutation on Existing Package Upsert

**Confirmed by**: code-reviewer, differential-review, API-safety (3/6 agents)

**File**: `db/sqlc/queries/oor_artifacts.sql` (UpsertOORPackage)

The SQL `ON CONFLICT (session_id) DO UPDATE SET direction =
EXCLUDED.direction` allows a package initially stored as "incoming" to be
silently flipped to "outgoing" (or vice versa). The Go-level validation
only checks that the value is one of the two allowed strings, not that it
matches the previously stored direction.

**Impact**: Incorrect direction corrupts list filtering
(`ListReceivedPackages` vs `ListSentPackages`). In adversarial
scenarios, masks the true nature of a transfer.

**Fix**: Either remove `direction` from the ON CONFLICT SET clause
(preserving original), or add a Go-level consistency check.

oor/local_persistence_handler.go
```diff
@@ -156,6 +167,17 @@ func (h *LocalPersistenceOutboxHandler) handleMaterializeIncoming(
 	}
 
 	ownedRecipients := 0
+	sessionIDHash := chainhash.Hash(msg.SessionID)
+
+	if h.PackageStore != nil {
+		err := h.PackageStore.UpsertPackage(ctx,
```
Roasbeef, line 191: Need to read the pacakge proper to verify, but: is this a durable actor? As otherwise we can upsert he package here, but then fail to do teh writes below if we crash.

Roasbeef (dismissed):
Initial review pass. Need to catch up with the oor flow in general. But one thing that I noticed here is that none of the oor tables added in this PR seem to reference any of the vtxo tables. Eg: if we're doing an OOR send outbound, and we spend the entire vtxo, that after that if we do `ListVTXOs` it either shouldn't show up, or it should show as spent. Then we shouldn't have to worry about watching it for expiry, etc (made some follow up issues re this interaction).

Roasbeef:
## Transaction Atomicity: Stores Must Participate in the Durable Actor's Per-Message Transaction

### Background

The OOR client actor is a `DurableActor` backed by a `TxAwareDeliveryStore`. The framework already wraps each message processing in a single serializable DB transaction via `TxAwareActorDeliveryStore.ExecTx` ([`db/actordelivery/store_impl.go:1334-1367`](https://github.com/lightninglabs/darepo-client/blob/oor-client-split-5-unroll-packages-draft/db/actordelivery/store_impl.go#L1334-L1367)). That method puts the `*sql.Tx` into the context via `actor.WithTx(ctx, tx)`, and the `txCtx` is passed all the way down through:

```
processInTransaction(txCtx)
  → executeBehaviorSafely(txCtx)
    → behavior.Receive(txCtx, msg)
      → handleDriveEvent(txCtx, ...)
        → askEvent(txCtx, ...)           // FSM state transition
        → persistOutgoingPackage(txCtx)  // package + binding persistence
        → persistCheckpoint(txCtx)       // durable checkpoint save
        → driveOutbox(txCtx, ...)
          → handler.Handle(txCtx, ...)   // LocalPersistenceOutboxHandler
            → PackageStore.UpsertPackage(txCtx, ...)
            → Store.SaveVTXO(txCtx, ...)
            → PackageStore.UpsertBinding(txCtx, ...)
```

The infrastructure exists (`actor/tx_context.go` provides `WithTx`, `TxFromContext`, `HasTx`, `RequireTx`), but currently **none of the `db/` package stores check for the transaction in the context**. Each store method calls `TransactionExecutor.ExecTx` which unconditionally opens a new `BeginTx`, ignoring any existing actor transaction.

### The Problem

This means every operation in the above call chain runs in its own independent transaction:

| Operation | Own txn? | Consequence on crash |
|-----------|----------|---------------------|
| FSM state transition (`askEvent`) | ✅ (actor outer txn) | — |
| `PackageStore.UpsertPackage` | ❌ own txn | Package stored, but bindings/VTXOs may not be |
| `Store.SaveVTXO` | ❌ own txn | VTXO stored, but binding may not be |
| `PackageStore.UpsertBinding` | ❌ own txn | May never execute if crash after SaveVTXO |
| `persistCheckpoint` | ❌ own txn (`b.cfg.DeliveryStore`) | Checkpoint may commit without package artifacts |

A crash at any boundary between these independent transactions leaves the store in an inconsistent state. Most critically:

1. **Outgoing path crash gap**: `persistOutgoingPackage` runs AFTER `askEvent` transitions the FSM. If the package persistence fails but the checkpoint commits, the FSM restarts in the post-transition state and `captureFinalizeStateForEvent` never fires again (it requires `AwaitingFinalizeAccepted` + `FinalizeAcceptedEvent`). Outgoing package artifacts are permanently lost.

2. **Incoming path partial state**: `handleMaterializeIncoming` calls `UpsertPackage`, then `SaveVTXO`, then `UpsertBinding` in three separate transactions. A crash between any pair leaves orphaned packages without bindings, or VTXOs without package ancestry — both degrade the unroll resolver's ability to build the correct broadcast chain for forced exit.

### The Fix: Make `TransactionExecutor.ExecTx` Respect `actor.TxFromContext`

In `db/interfaces.go`, `TransactionExecutor.ExecTx` should check for an existing actor transaction before opening a new one:

```go
func (t *TransactionExecutor[Q]) ExecTx(ctx context.Context,
    txOptions TxOptions, txBody func(Q) error) error {

    // If the context already carries a durable actor transaction,
    // participate in it instead of opening a nested one. Retries
    // are the outer transaction's responsibility.
    if tx, ok := actor.TxFromContext(ctx); ok {
        return txBody(t.createQuery(tx))
    }

    // ... existing BeginTx + retry logic unchanged ...
}
```

This is a single change point that makes the entire `db/` store layer transaction-aware. Every store that uses `TransactionExecutor` automatically participates in the actor's outer transaction when one exists — no changes needed in individual store methods or the OOR actor/handler code.

After this change, the durable actor's per-message transaction atomically covers:
- FSM state transition
- Package artifact persistence (Ark PSBT + checkpoints)
- VTXO creation
- Outpoint binding creation
- Durable checkpoint save
- Mailbox ack/dedup

Either everything commits or nothing does. The crash gap disappears entirely.

### Caveat: Read-Only vs Write Transaction Compatibility

When reusing the outer transaction, `txOptions` (read/write) should be compatible. The actor's outer transaction is always a write transaction, so all the write operations in this PR are fine. For read-only operations like `GetPackageForOutpoint` or `ListPackages` that specify `ReadTxOption()`, they'll still work within a write transaction (a write tx can serve reads). If strict separation is desired, the context check could also verify option compatibility, but for the current use case this is unnecessary.

### Scope

This applies to both the outgoing actor path (`oor/actor.go`) and the incoming handler path (`oor/local_persistence_handler.go`) — both receive `txCtx` from the durable actor framework and both call stores that should participate in the same transaction.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: Yes. “Local outpoint” here means a local VTXO outpoint (`outpoint_hash`, `outpoint_index`). I also enforced that relationship at the DB layer: OOR bindings now reference `vtxos` via FK, so bindings can only point to locally known VTXOs.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: Receiver-side cursor. `oor_recipient_cursors` tracks per-recipient-script polling progress (last processed event id) so receiver polling can resume deterministically and remain idempotent across restarts.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: Addressed this in the OOR surface by switching owned receive-script key handling to `keychain.KeyDescriptor` in the persistence model and conversion path, so the key family/index/pubkey handling is now strongly typed and consistent through this flow.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: Done. Added enum tables and FK references for persisted OOR codes:
- `oor_package_directions`
- `oor_vtxo_binding_link_kinds`
- `owned_receive_script_sources`

Also documented the stored values inline in the migration.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: It’s now layered directly on top of `vtxos`: OOR keeps relation metadata in `oor_vtxo_bindings`, and each binding must reference an existing `vtxos` row via FK.

Also removed duplicated recipient/value fields from binding storage and derive them from `vtxos` on join.

db/oor_artifact_store.go
bhandras: Done — this now uses typed enums in Go and enum tables in SQL.

db/oor_artifact_store.go
bhandras: Done. Converted these OOR persistence timestamps to `time.Time` in the store-facing structs and conversion path.

db/oor_artifact_store.go
bhandras: Done. Optional OOR fields now use `fn.Option` (including matched binding, optional recipient/value surfaces, and optional last-used time).

db/oor_artifact_store.go
bhandras: Agreed. Removed duplicated binding payload for this case — recipient script and value are no longer persisted in `oor_vtxo_bindings`; they are derived from the referenced `vtxos` row via join.

db/oor_artifact_store.go
bhandras: Done. Switched the owned receive-script client key representation to `keychain.KeyDescriptor`.

db/oor_artifact_store.go
bhandras: Clarified and tightened the materialized package view in this direction.

`ark` is the signed Ark package PSBT, and the materialized bundle now carries the full unroll-relevant set together:
- Ark package PSBT
- finalized checkpoint PSBTs
- all local bindings for the session
- matched binding for outpoint lookup responses

db/oor_artifact_store.go
bhandras: Exactly. We now model this explicitly with binding kind:
- `created_output`: local outpoint created by the package
- `consumed_input`: local outpoint consumed by an outgoing package

db/oor_artifact_store.go
bhandras: Yes — this is a local VTXO outpoint (`hash:index`).

db/oor_artifact_store.go
bhandras: Good call. Resolver ancestry walking now uses created-output linkage explicitly, so traversal stays rooted in the outpoint relation semantics rather than relying on ambiguous first-match behavior.

db/oor_artifact_store.go
bhandras: Yes — recipients poll the server. These cursor/script tables are receiver-side local state to drive that polling and resume safely.

db/oor_artifact_store.go
bhandras: Right: this is receiver-side indexing/attribution state, not sender transfer state.

db/oor_unroll_resolver.go
bhandras: Yes — unroll resolution targets and dependency edges are all VTXO outpoints.

db/oor_unroll_resolver.go
bhandras: Implemented. Added a max-depth guard (`defaultMaxUnrollDepth`) and explicit `ErrResolveUnrollMaxDepthExceeded` when traversal exceeds that bound, with test coverage for the limit path.

db/sqlc/queries/oor_artifacts.sql
bhandras: Implemented in both SQL and store logic:
- upsert no longer rewrites `session_id` on binding conflict
- store performs an explicit outpoint+kind pre-check and errors if an existing binding points to a different session

This makes `(outpoint, kind) -> session` immutable after first write.

db/sqlc/queries/oor_artifacts.sql
bhandras: Implemented in both SQL and store logic:
- package upsert no longer updates `direction` on conflict
- store pre-checks existing package direction and returns a conflict on mismatch

This makes direction immutable per `session_id`.

oor/local_persistence_handler.go
bhandras: Implemented at the transaction-executor layer.

`ExecTx` now joins an actor transaction already present in context, instead of opening a nested transaction. That makes the durable actor write sequence atomic across the persistence steps in this flow.

bhandras:
Great catch — this is now implemented at the `TransactionExecutor` layer exactly as proposed.

What changed:
- `db/interfaces.go`: `ExecTx` now checks `actor.TxFromContext(ctx)` first.
- If a context transaction exists, it executes `txBody` against that transaction (`t.createQuery(tx)`) and does **not** open a nested `BeginTx`.
- Existing retry/begin logic remains for non-actor contexts.

Why this closes the gap:
- Durable actor message processing now uses one outer actor transaction for all DB store calls made through `TransactionExecutor`.
- So FSM transition, OOR package/binding writes, VTXO updates, checkpoint writes, and delivery-store updates are committed/rolled back atomically at message scope.

Coverage:
- Added `TestTransactionExecutorUsesContextTx` in `db/interfaces_test.go`.
- Test verifies writes done via `ExecTx` inside `actor.WithTx(ctx, outerTx)` are rolled back with the outer transaction, proving participation in the actor tx.

Also on your caveat: reads remain compatible when executing inside the actor write transaction.

bhandras:
Follow-up update after an additional review pass: I implemented the remaining medium/low correctness nits discussed.

Changes applied:
- Tightened binding immutability further:
  - `UpsertOORVTXOBinding` conflict path no longer mutates `output_index` (only `updated_at`).
  - Store-side pre-check now errors if `(outpoint, kind, session)` is retried with a different `output_index`.
  - Added regression test for output-index conflict.

- Reduced resolver query amplification and removed redundant reloads:
  - Added per-input resolution/missing caches.
  - Added stale-queue depth guard to avoid reprocessing outdated queue entries.
  - Materialize package bundles directly from join rows in resolver paths (no extra package-row fetch).

- Added context cancellation check inside resolver traversal loop.

- Test/cleanup nits:
  - Removed duplicate outpoint seeding in `TestOORArtifactStoreGetPackageForOutpointPrefersCreatedBinding`.
  - Cleaned awkward comment wrapping in `lib/types/oor_persistence.go`.

Validation re-run after these changes:
- `make lint`
- `go test ./...`

Both pass on the updated branch.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,215 @@
...
+-- oor_packages stores one finalized OOR package artifact set per session.
+CREATE TABLE IF NOT EXISTS oor_packages (
+    -- session_id is the stable OOR session identifier (Ark txid bytes).
+    session_id BLOB PRIMARY KEY NOT NULL,
+
+    -- direction encodes local package direction:
+    --   0 = incoming (received by this client)
+    --   1 = outgoing (sent by this client)
+    direction INTEGER NOT NULL CHECK (
+        direction IN (0, 1)
```
ellemouton: is this part needed if it's already defined in the enum table?

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,215 @@
...
+        ON DELETE CASCADE
+);
+
+-- oor_vtxo_bindings maps local outpoints to stored OOR package sessions.
+CREATE TABLE IF NOT EXISTS oor_vtxo_bindings (
+    -- outpoint identifies the local VTXO outpoint linked to this package.
+    outpoint_hash BLOB NOT NULL,
+
+    -- outpoint_index is the output index of the local outpoint.
+    outpoint_index INTEGER NOT NULL CHECK (outpoint_index >= 0),
```
ellemouton, line 99: wonder if our vtxo table shouldnt just have an int ID too for easy reference.. cause this makes it hard to quickly lookup all spending txs for a given vtxo

db/sqlc/migrations/000004_oor_artifact_store.up.sql
```diff
@@ -0,0 +1,215 @@
...
+
+    -- Direction enum foreign key.
+    FOREIGN KEY (direction) REFERENCES oor_package_directions(direction)
+);
+
+-- oor_package_checkpoints stores the ordered finalized checkpoint package
+-- for each OOR session.
+CREATE TABLE IF NOT EXISTS oor_package_checkpoints (
+    -- session_id references the owning OOR package row.
+    session_id BLOB NOT NULL,
```
ellemouton, line 74: i think nicer to have an integer session ID that we then use for foreign keys

oor/actor.go
```diff
@@ -325,11 +341,26 @@ func (b *oorDurableBehavior) handleDriveEvent(ctx context.Context,
...
+	}
+
 	outbox, err := b.askEvent(ctx, handle.FSM, req.Event)
 	if err != nil {
 		return fn.Err[ActorResp](err)
 	}
 
+	if finalizeState != nil {
+		err := b.persistOutgoingPackage(ctx, req.SessionID,
+			finalizeState)
```
ellemouton: nit: formatting in a couple places

db/oor_unroll_resolver_test.go
```diff
@@ -0,0 +1,237 @@
...
+		Hash:  chainhash.Hash{0x44, 0xaa},
+		Index: 0,
+	}
+
+	parentSession, parentArk, parentCheckpoints, parentOutpoint,
+		parentScript, parentValue, _ := buildTestOORPackageWithInput(
+		t, 0x44, externalInput,
+	)
+
+	err := store.UpsertPackage(ctx, OORPackageDirectionIncoming,
```
ellemouton: formattin

db/oor_unroll_resolver_test.go
```diff
@@ -0,0 +1,237 @@
...
+			OORPackageLinkKindCreatedOutput)
+		require.NoError(t, err)
+
+		prevInput = outpoint
+		targetOutpoint = outpoint
+	}
+
+	_, err := store.ResolveUnrollPackages(ctx, targetOutpoint)
+	require.Error(t, err)
+	require.ErrorIs(t, err, ErrResolveUnrollMaxDepthExceeded)
```
ellemouton: can just ErrorIs

db/oor_unroll_resolver.go
```diff
@@ -0,0 +1,341 @@
...
+
+type unrollPackageNode struct {
+	pkg   *OORPackageBundle
+	depth int
+}
+
+// ResolveUnrollPackages resolves locally known OOR package artifacts needed to
+// unroll the target outpoint.
+//
+// The resolver follows checkpoint-input ancestry through locally persisted
```
ellemouton, line 53: do we somewhere check for import cycles btw? (of packages)

db/sqlc/queries/oor_artifacts.sql
ellemouton: my agent says:

```
 The Go code does a read-then-check inside the transaction (lines 326-340), but the SQL ON CONFLICT at oor_artifacts.sql:9-11 silently overwrites direction on conflict because the SET clause doesn't
  exclude it. A concurrent writer or direct SQL migration could flip direction without the Go guard.

  Fix: Exclude direction from the SQL upsert's SET clause so it's write-once at the DB level:

  -- oor_artifacts.sql lines 9-11, change to:
  ON CONFLICT (session_id) DO UPDATE SET
      ark_psbt = EXCLUDED.ark_psbt,
      updated_at = EXCLUDED.updated_at
  WHERE oor_packages.direction = EXCLUDED.direction;

  This makes the upsert a no-op (0 rows affected) if directions differ. The Go code can then check the affected rows and return the conflict error. After changing the SQL, run make sqlc to regenerate
```

db/sqlc/queries/oor_artifacts.sql
ellemouton:
```
UpsertOORVTXOBinding ON CONFLICT only updates updated_at, but it doesn't guard against a different session_id or output_index being written for the same (outpoint_hash, outpoint_index,
  link_kind) triple. The Go guard at oor_artifact_store.go:443-457 covers this, but the SQL layer is silent.

  Fix: Add a WHERE clause to the upsert so it's a no-op on session/index mismatch:

  -- oor_artifacts.sql lines 51-52, change to:
  ON CONFLICT (outpoint_hash, outpoint_index, link_kind) DO UPDATE SET
      updated_at = EXCLUDED.updated_at
  WHERE oor_vtxo_bindings.session_id = EXCLUDED.session_id
    AND oor_vtxo_bindings.output_index = EXCLUDED.output_index;
```

db/oor_artifact_store.go
```diff
@@ -0,0 +1,1301 @@
...
+				ctx, directionCode,
+			)
+		}
+		if err != nil {
+			return err
+		}
+
+		out := make([]*OORPackageBundle, 0, len(rows))
+		for i := range rows {
+			pkg, err := materializePackageBundle(
```
ellemouton, line 630: we might want to batch collect in future if the number of rows are ever expected to become really large

ellemouton (approved):

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: Good call. I removed the redundant enum CHECK clauses and now rely on enum-table foreign keys as the single source of truth for `direction` and `link_kind`.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: Agree this could simplify query ergonomics for graph-style lookups. I deferred it to follow-up issue #129 so we can handle the wider schema migration/backfill in one scoped change.

db/sqlc/migrations/000004_oor_artifact_store.up.sql
bhandras: Agree this is a cleaner long-term FK model. Deferred to issue #129 because introducing a surrogate package/session key needs coordinated schema and API migration.

oor/actor.go
bhandras: Applied. I reformatted the multiline calls in this path to use trailing commas and put the closing `)` on its own line.

db/oor_unroll_resolver_test.go
bhandras: Applied. Updated resolver test call formatting to the same multiline style (closing `)` on its own line).

db/oor_unroll_resolver_test.go
bhandras: Applied. Simplified this assertion to `require.ErrorIs(...)`.

db/oor_unroll_resolver.go
bhandras: Current guard is `go list ./...` / `go test ./...`, which fails on package import cycles. I didn't add a dedicated extra cycle-check target in this PR.

db/sqlc/queries/oor_artifacts.sql
bhandras: Applied. I added the SQL-level guard (`WHERE oor_packages.direction = EXCLUDED.direction`), switched this query to `:execrows`, regenerated sqlc, and now handle zero-row guarded no-op/conflict explicitly in store logic.

db/sqlc/queries/oor_artifacts.sql
bhandras: Applied. I added the SQL `WHERE` guard on `session_id` + `output_index`, switched to `:execrows`, regenerated sqlc, and now handle zero-row guarded no-op/conflict explicitly in store logic.

db/oor_artifact_store.go
bhandras: Agree. I tracked batching/pagination follow-up in issue #129: https://github.com/lightninglabs/darepo-client/issues/129

PR #118, opened 2026-02-18
oor: add unified OOR service for outgoing + incoming orchestration

bhandras:
## Summary

This split introduces a high-level `OORService` API on top of existing OOR
primitives so callers no longer manually orchestrate:

- outgoing actor commands,
- incoming receive-session lifecycle,
- outbox follow-up recursion,
- recipient-cursor progression.

The service provides one cohesive surface for:

1. outgoing transfer lifecycle,
2. incoming transfer sync (once + background worker),
3. unroll package resolution delegation.

Base: `oor-client-split-5-unroll-packages-draft`.

## Architecture Alignment

This PR now supports the same actor-system wiring pattern used by VTXO/round:

- outgoing OOR can be resolved via actor-system service key,
- service sends `Ask` to `ActorRef` instead of hard-wiring direct calls,
- direct local actor construction remains available as a compatibility fallback.

Concretely:

- added `ActorServiceKey(actorID)` in `oor/service_key.go`,
- added `ServiceConfig.ActorSystem`, `ServiceConfig.OutgoingServiceKey`, and
  `ServiceConfig.OutgoingRef`,
- `NewOORService` resolves outgoing in this order:
  1. `OutgoingRef` (explicit injection),
  2. actor-system service-key lookup,
  3. local `NewOORClientActor` fallback (requires `DeliveryStore`).

This keeps integration flexible while matching the system-wide actor plumbing
model when actor system wiring is available.

## Problem This Split Solves

By split 5, OOR primitives were correct, but app/harness integration still
required low-level choreography in multiple places:

- direct actor command handling for outgoing,
- ad-hoc receive-session driving for incoming,
- manual cursor discipline,
- custom worker-loop lifecycle handling.

This split centralizes that orchestration in one tested runtime service.

### 1) Service contracts and adapters

- `oor/service_types.go`
  - new `OORService` interface
  - typed request/status/state view structs
  - `IncomingEventSource`, `IncomingCursorStore`, `UnrollPackageResolver`
  - actor-system outgoing wiring config fields
- `oor/service_db_adapters.go`
  - `DBIncomingCursorStore` adapter from
    `db.OORArtifactPersistenceStore` to `IncomingCursorStore`
- `oor/service_key.go`
  - actor-system service-key helper for outgoing OOR actor lookup

### 2) Service runtime

- `oor/service.go`
  - `NewOORService(cfg ServiceConfig)` constructor
  - outgoing facade:
    - `StartOutgoing`
    - `GetOutgoingState`
  - incoming orchestration:
    - `SyncIncomingOnce`
    - `StartIncomingSync`
    - `StopIncomingSync`
    - `GetIncomingSyncStatus`
  - unroll facade:
    - `ResolveUnrollPackages`
  - lifecycle:
    - `Stop`
  - outgoing actor selection supports actor-system lookup and local fallback

### 3) Service-level tests

- `oor/service_test.go`
  - outgoing happy path through service API
  - outgoing flow through actor-system service-key wiring
  - constructor failure when actor-system lookup has no registered actor
  - incoming sync success and cursor advancement
  - ack failure does **not** advance cursor
  - resolver passthrough
  - background worker lifecycle start/stop

### Outgoing path

`OORService.StartOutgoing` -> configured outgoing actor endpoint:

- explicit `OutgoingRef`, or
- actor-system `ServiceKey.Ref(...)`, or
- local durable `OORClientActor` fallback.

Service callers use `GetOutgoingState` for typed state view.

### Incoming path

`SyncIncomingOnce` (or worker cycle):

1. list owned receive scripts,
2. load per-script cursor,
3. fetch events `afterEventID`,
4. for each event:
   - validate ordering + payload shape,
   - drive receive FSM (`IncomingTransferEvent`),
   - execute outbox through local persistence/transport boundary,
   - require terminal `ReceiveCompleted`,
   - **only then** write cursor to `event_id`.

Flow:

`script -> cursor -> events -> validate -> materialize -> ack -> cursor++`

Cursor safety invariant:

- cursor is not advanced on partial failure,
- retries are at-least-once and converge via idempotent handlers.

## Scope Notes

- This split does **not** add new server RPC/stream endpoints.
- Incoming source remains pluggable (`IncomingEventSource`), so a future
  long-poll or stream source can be injected without changing service API.

## Testing

Executed locally:

- `go test ./...`
- `make lint`

All green.

### DIFF
--- a/oor/service.go
+++ b/oor/service.go
@@ -1,0 +1,718 @@
+package oor
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/baselib/actor"
+	"github.com/lightninglabs/darepo-client/db"
+)
+
+// oorService is the default OORService implementation.
+type oorService struct {
+	outgoingRef   actor.ActorRef[ActorMsg, ActorResp]
+	outgoingActor *OORClientActor
+
+	incomingSource IncomingEventSource
+	cursorStore    IncomingCursorStore
+	unrollResolver UnrollPackageResolver
+
+	incomingOutboxHandler OutboxHandler
+
+	incomingPageSize     int32
+	incomingPollInterval time.Duration
+	incomingPollJitter   time.Duration
+	jitterRand           *rand.Rand
+	jitterRandMu         sync.Mutex
+
+	workerMu     sync.Mutex
+	workerCancel context.CancelFunc
+	workerDone   chan struct{}
+
+	incomingRunMu sync.Mutex
+
+	statusMu sync.RWMutex
+	status   IncomingSyncStatus
+}
+
+// NewOORService constructs a high-level OOR service for outgoing and incoming
+// flow orchestration.
+//
+// The constructor wires local persistence handling in front of the provided
+// transport/signing handler so both outgoing and incoming paths share one
+// consistent outbox boundary.
+func NewOORService(cfg ServiceConfig) (OORService, error) {
+	if cfg.VTXOStore == nil {
+		return nil, fmt.Errorf("vtxo store must be provided")
+	}
+
+	if cfg.OperatorKey == nil {
+		return nil, fmt.Errorf("operator key must be provided")
+	}
+
+	if cfg.ResolveIncomingClientKey == nil {
+		return nil, fmt.Errorf("incoming client key resolver must be " +
+			"provided")
+	}
+
+	if cfg.ResolveIncomingMetadata == nil {
+		return nil, fmt.Errorf("incoming metadata resolver must be " +
+			"provided")
+	}
+
+	actorID := cfg.ActorID
+	if actorID == "" {
+		actorID = DefaultActorServiceKeyName
+	}
+
+	pageSize := cfg.IncomingPageSize
+	if pageSize <= 0 {
+		pageSize = DefaultIncomingPageSize
+	}
+
+	pollInterval := cfg.IncomingPollInterval
+	if pollInterval <= 0 {
+		pollInterval = DefaultIncomingPollInterval
+	}
+
+	localHandler := &LocalPersistenceOutboxHandler{
+		Next:                     cfg.TransportOutboxHandler,
+		Store:                    cfg.VTXOStore,
+		PackageStore:             cfg.PackageStore,
+		OperatorKey:              cfg.OperatorKey,
+		ExitDelay:                cfg.ExitDelay,
+		ResolveIncomingClientKey: cfg.ResolveIncomingClientKey,
+		ResolveIncomingMetadata:  cfg.ResolveIncomingMetadata,
+	}
+
+	outgoingRef, outgoingActor, err := buildOutgoingActorClient(
+		cfg, actorID, localHandler,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &oorService{
+		outgoingRef:           outgoingRef,
+		outgoingActor:         outgoingActor,
+		incomingSource:        cfg.IncomingSource,
+		cursorStore:           cfg.IncomingCursorStore,
+		unrollResolver:        cfg.UnrollResolver,
+		incomingOutboxHandler: localHandler,
+		incomingPageSize:      pageSize,
+		incomingPollInterval:  pollInterval,
+		incomingPollJitter:    cfg.IncomingPollJitter,
+		// #nosec G404 -- non-crypto jitter for polling.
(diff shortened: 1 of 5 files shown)


oor/service.go
```diff
@@ -0,0 +1,638 @@
...
+	}
+
+	s.workerMu.Lock()
+	defer s.workerMu.Unlock()
+
+	if s.workerCancel != nil {
+		return nil
+	}
+
+	loopCtx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
bhandras: Good catch, fixed.

`StartIncomingSync` now derives the worker context from the provided `ctx` (`context.WithCancel(ctx)`) instead of `context.Background()`, with a nil-context guard that falls back to `context.Background()`.

This ties worker lifetime to caller cancellation while preserving explicit `StopIncomingSync` shutdown behavior.

oor/service.go
```diff
@@ -0,0 +1,638 @@
...
+		return delay
+	}
+
+	jitterMax := big.NewInt(int64(s.incomingPollJitter) + 1)
+	noise, err := cryptorand.Int(cryptorand.Reader, jitterMax)
+	if err != nil {
+		return delay
+	}
+
+	jitter := time.Duration(noise.Int64())
```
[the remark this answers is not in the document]
bhandras: Applied.

The jitter path now uses a non-crypto PRNG (`math/rand`) stored on the service struct (`jitterRand`) instead of `crypto/rand`.

This keeps the polling delay jitter fast and avoids blocking entropy reads for non-security randomness.

oor/service.go
```diff
@@ -0,0 +1,644 @@
...
+
+// StartOutgoing starts one outgoing transfer via the underlying durable actor.
+func (s *oorService) StartOutgoing(ctx context.Context,
+	req StartOutgoingRequest) (SessionID, error) {
+
+	if s == nil || s.actor == nil {
+		return SessionID{}, fmt.Errorf("service must be provided")
+	}
+
+	result := s.actor.Receive(ctx, &StartTransferRequest{
```
Roasbeef: Why direct recv on actor vs adding it to the system and using a service key to get a normal ref?

oor/service.go
```diff
@@ -0,0 +1,644 @@
...
+	if !ok || resp == nil {
+		return SessionID{}, fmt.Errorf("unexpected response type: %T",
+			result.UnwrapOr(nil))
+	}
+
+	return resp.SessionID, nil
+}
+
+// GetOutgoingState returns a caller-facing state summary for one session.
+func (s *oorService) GetOutgoingState(ctx context.Context,
```
Roasbeef, line 142: So then what ends up driving this? Thought we had a like coordinator actor that would talk to the vtxo actors when things need to happen based on the wallet or server?

oor/service.go
```diff
@@ -0,0 +1,644 @@
...
+func (s *oorService) processIncomingEvent(ctx context.Context,
+	event *IncomingRecipientEvent) error {
+
+	session, err := NewReceiveSession(ctx, event.ArkPSBT, event.SessionID)
+	if err != nil {
+		return err
+	}
+	defer session.FSM.Stop()
+
+	outbox, err := askFSMEvent(ctx, session.FSM, &IncomingTransferEvent{
```
Roasbeef, line 446: Same here. Does this need to be a durable actor? Then can makes sure stuff like send/recv OOR is properly restarted.

oor/service.go
bhandras: I took a closer look at the `serverconn`/mailbox plumbing direction in PR #116, and I think your point stands: if we route incoming through that connector path, the incoming orchestration I added here is likely unnecessary.

Given that direction, this PR should probably be treated as temporary scaffolding (or closed) rather than the long-term shape.

oor/service.go
bhandras: Agreed. After looking at PR #116 more closely, the intended model is to drive incoming server events through the unified connector/mailbox path into actors, not through a service-local polling loop.

So the incoming orchestration in this PR is likely redundant with that target architecture.

oor/service.go
bhandras: Yes — after reviewing PR #116, I agree the long-term durable restart story should come from mailbox/serverconn ingress + actor dispatch, not this service-level incoming loop.

Given that, the incoming receive path in this PR is probably not worth carrying forward as-is. We can close this and reintroduce only the pieces that still fit once the connector plumbing lands.

bhandras:
Closing this as the direction is already set in https://github.com/lightninglabs/darepo-client/pull/116 and therefore this won't be necessary or not in this current form. We can always revisit once we're in the plumbing stage.

PR #120, opened 2026-02-19
chainresolver: add chain resolver coordinator and FSM

sputn1ck:
## Summary

- Add `chainresolver` package with coordinator actor, per-VTXO resolver FSM, CPFP child builder, and package relayer
- Add chain resolver state persistence (DB schema, migrations, queries, store) for durable resolver tracking across restarts
- Coordinator routes chainsource events (spend, conf, block epoch) to per-VTXO resolvers and manages outbox-driven side effects (broadcast, watch registration)

## Test plan

- [x] Unit tests for coordinator message routing (`coordinator_test.go`)
- [x] Unit tests for resolver FSM state transitions (`resolver_transitions_test.go`)
- [x] Unit tests for CPFP child builder (`cpfp_test.go`)
- [x] Unit tests for unroll logic (`unroll_test.go`)
- [ ] E2E system tests in root repo (separate PR)


chainresolver/coordinator.go
```diff
@@ -0,0 +1,897 @@
...
+func (c *Coordinator) spawnResolver(ctx context.Context,
+	outpoint wire.OutPoint, initialState ResolverState,
+	resolverCtx *ResolverContext) {
+
+	env := NewResolverEnvironment(
+		fmt.Sprintf("resolver.%s", outpoint.String()),
+		resolverCtx,
+	)
+
+	c.resolvers[outpoint] = &resolverInstance{
```
Roasbeef, line 404: As mentioned offline, what I thing could work here is first an aggregration, then a split later. So eg if we have 5 vtxos in the same tree we need to go on chain for, we create an actor which groups then together until the greatest common ancestor in the tree. Then at that point it splits out into smaller groups, until finally it's a actor that handles the baseline path down to the vtxo manifestation.

chainresolver/coordinator.go
```diff
@@ -0,0 +1,897 @@
...
+	c.cfg.Logger.InfoS(ctx, "Chain resolver coordinator stopped")
+}
+
+// Receive processes incoming messages and routes them to the appropriate
+// resolver FSM.
+func (c *Coordinator) Receive(ctx context.Context,
+	msg ChainResolverMsg) fn.Result[ChainResolverResp] {
+
+	switch m := msg.(type) {
+	case *ExpiringVTXORequest:
```
Roasbeef, line 121: So we should modify the message sent from the `vtxo` actor to be this instead? See the expiry detection logic there.

Roasbeef: Also does this need to be a durable actor? Or we expect that the vtxo actor will re-send the go-to-chain message on restart?

PR #121, opened 2026-02-19
mailbox: fix AwaitRPC lost-wakeup flake

bhandras:
## Summary
This fixes a flaky timeout in `TestClient_ConcurrentInFlightDoesNotDrop` by
eliminating a lost-wakeup race in `AwaitRPC`.

## Root Cause
`AwaitRPC` previously did these steps separately:
1. Check `pending` for the response.
2. Register a waiter channel.

If a response arrived between steps 1 and 2, `handleEnvelope` would cache
the response and notify the **current** waiter set — which did not yet
include the new waiter. The newly registered waiter would then block until
context timeout.

## Fix
Make the response check and waiter registration atomic under one lock via
`popPendingOrAddWaiter`:
- if pending response exists: return it immediately and skip waiter registration
- otherwise: register waiter before releasing the lock

This removes the race window where notifications can be missed.

## CI Failure
Observed CI failure (PR #79, unit-cover):
- https://github.com/lightninglabs/darepo-client/actions/runs/22197814323/job/64202618296?pr=79

Failure signature:
- `TestClient_ConcurrentInFlightDoesNotDrop`
- `context deadline exceeded`

## Validation
- Reproduced pre-fix flake locally with high repetition.
- Post-fix stress runs passed:
  - `go test ./mailbox/client -run TestClient_ConcurrentInFlightDoesNotDrop -count=20000 -timeout=60m`
  - `go test -covermode=atomic -coverpkg=github.com/lightninglabs/darepo-client/... -tags='dev nolog' ./mailbox/client -run TestClient_ConcurrentInFlightDoesNotDrop -count=1000 -timeout=20m`

### DIFF
--- a/mailbox/client/client.go
+++ b/mailbox/client/client.go
@@ -179,14 +179,15 @@ func (c *Client) AwaitRPC(ctx context.Context, correlationID string,
 	resp proto.Message) error {
 
 	for {
-		data, ok := c.popPending(correlationID)
-		if ok {
-			return (proto.UnmarshalOptions{
+		data, ch, hasPending := c.popPendingOrAddWaiter(correlationID)
+		if hasPending {
+			unmarshal := proto.UnmarshalOptions{
 				DiscardUnknown: true,
-			}).Unmarshal(data, resp)
+			}
+
+			return unmarshal.Unmarshal(data, resp)
 		}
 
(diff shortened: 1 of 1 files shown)


Roasbeef:
Folded into the latest series.

PR #123, opened 2026-02-22
darepod: add client daemon scaffold with actor system

Roasbeef:
In this PR, we add the initial `darepod` client daemon -- the main entry
point that boots up, connects to lnd and the ark server, initializes the
actor system, and exposes a gRPC API for external tooling (CLI, GUI, etc).

The daemon follows the same lifecycle pattern used by lnd and tapd: cobra
for CLI parsing, viper for config/env overrides, lndclient for the lnd
connection (blocking until chain-synced and wallet-unlocked), and a signal
interceptor for graceful shutdown. The startup flow is sequential and
numbered for readability -- connect lnd, dial the ark server, init the
actor system, start the chain source, then fire up the gRPC listener.

## Indexer RPC Schema and Client

The first four commits (carried over from `indexer-draft-oor-lite`) add the
arkrpc proto definitions for the indexer service and a mailbox-backed Go
client in `indexer/`. The indexer client takes an `RPCClient` interface
(satisfied by the serverconn `UnaryFacade`) and wraps the generated
`IndexerServiceMailboxClient` stubs. We also add taproot script-scope auth
helpers for proof-of-control signing, and a cursor-based sync client with
full test coverage.

## Daemon Scaffold

In `cmd/darepod/main.go`, we wire up cobra with flags for `--network`,
`--lnd.host`, `--lnd.tlspath`, `--lnd.macaroonpath`, `--server.host`,
`--server.insecure`, `--rpc.listenaddr`, and the mailbox ID pair
(`--server.localmailboxid`, `--server.remotemailboxid`). Config structs
live in `darepod/config.go` with sensible defaults and a `Validate()`
method.

`darepod/server.go` is the main orchestrator. The `RunUntilShutdown` method
walks through the numbered startup steps:

1. Connect to lnd via `lndclient.NewLndServices`
2. Dial the ark server's mailbox edge (with TLS modes: custom cert, system
   pool, or insecure for regtest)
3. Initialize `actor.NewActorSystem()` with deferred graceful shutdown
4. Create the chain backend via `chainbackends.NewLNDBackendFromLndClient`
   and register a `ChainSourceActor` under `chainsource.ChainSourceKey`
5. Placeholder for the serverconn `Runtime` (blocked on `DeliveryStore`,
   a.k.a the DB layer)
6. Start the gRPC server with `DaemonService` registered on both local
   gRPC and a mailbox `ServeMux` for dual-transport access
7. Block until shutdown

## DaemonService RPC

We add a `daemonrpc/daemon.proto` with a single `GetInfo` RPC that returns
version, commit hash, network, lnd pubkey, lnd alias, and current block
height (fetched via `ChainKit.GetBestBlock`). The `RPCServer` implements
both the gRPC server interface and the `DaemonServiceMailboxServer`
interface, so the same handler serves both transports. The mailbox stubs
are generated by `protoc-gen-mailboxrpc` as part of `make rpc`.

## What's Left

The serverconn `Runtime` needs a persistent `DeliveryStore` before it can
be wired -- that's the DB layer work. Once that lands, the mailbox `ServeMux`
plugs into `ConnectorConfig.Dispatchers` and the indexer client gets
initialized via `indexer.New(runtime.Unary(), ...)`. The remaining actors
(round, OOR, VTXO, wallet) similarly depend on the store.

changes.


Roasbeef:
@claude review this

Roasbeef:
Addressed the actionable items from the review in 85c49a6:

1. **Env var key replacer** — Added `v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))` so `DAREPOD_LND_HOST` works for `lnd.host`. Good catch.

2. **Zero time.Time wrapping** — Guarded the `expiresAt→uint64` cast so zero times produce `0` (server default) instead of wrapping.

3. **Missing comments** — Added GoDoc on `TaprootScriptScope` fields, function comments on `randomNonceHex` and `newTaprootScope`.

4. **Dead code TODOs** — Added TODO comments tying `newMailboxEdge` and the gRPC TLS config fields to the serverconn runtime wiring step.

5. **TLS MinVersion** — Already fixed in 033ad08 (set `tls.VersionTLS12` on both config paths).

Re item 3 (unused `RPCConfig.TLSCertPath`/`TLSKeyPath`): keeping the fields with a TODO — they'll be wired once auto-generated TLS material is in place.

Re item 10 (gRPC error log after graceful stop): `grpcServer.Serve` returns nil on `GracefulStop`, so the error branch only fires on actual failures. The cancelled ctx in the log is fine since btclog doesn't inspect context state.

Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
Addressed all actionable items from both Claude reviews, the Codex review, and Gemini comments:

**From Claude review #2 (latest):**
1. **TLV proof expiry mismatch for zero time** — Moved `expiresAtUnixS` zero-guard before `encodeRegistrationProofTLV` so both the signed TLV message and the proto field use the same value.
2. **Nil dereference if `env.Rpc` is nil** — Added `env.Rpc == nil` guard in `handleInboundRPC`.
3. **Makefile clean redundancy** — Removed `$(RM) -r ./bin/darepod` line.
4. **Validate() mailbox IDs** — Added `LocalMailboxID` and `RemoteMailboxID` checks to `Config.Validate()`.

**From Codex review:**
5. **Tilde expansion bug** — Fixed `expandTilde` to strip the leading `/` after `~` so `filepath.Join` produces the correct path.

**Lint fixes:**
- Rewrapped `cmd/darepod/main.go` Long string to 80 cols
- Added `nolint:ll` for unavoidable long const in `indexer/client.go`  
- Rewrapped comments and interface signatures in `indexer/sync_client.go`

All fixups squashed into their target commits, branch force-pushed.

Roasbeef:
@claude review the latest push addressing your review comments: (1) MuSig2 pubkey serialization now uses schnorr.SerializePubKey for RC2+ and SerializeCompressed for v040, (2) added resp.Pubkey length guard in fetchOperatorTerms, (3) added TODO for future dispatcher methods, (4) improved compile-time interface check comment with NOTE about import cycle.

sputn1ck (approved):
LGTM! Started with some glueing between server and this and it seems to work, but would be very helpful to have a good master/base branch to start from

PR #125, opened 2026-02-25
db: make TransactionExecutor join actor's outer transaction

Roasbeef:
In this PR, we fix a gap in the durable actor framework's atomicity
guarantee by teaching `TransactionExecutor.ExecTx` to respect the outer
database transaction that `TxAwareActorDeliveryStore` injects into the
context via `actor.WithTx`.

Previously, every call to `ExecTx` unconditionally created a new database
transaction via `BeginTx`, even when the context already carried one from
the actor framework. This meant that inner stores (e.g., `PackageStore`,
`VTXOStore`) each ran in their own isolated transactions rather than
participating in the actor's per-message transaction. A crash between any
two of these inner transactions would leave the store in an inconsistent
state: a package persisted but its bindings lost, or an FSM state advanced
but the outgoing artifacts never written.

The fix is a small early-return at the top of `ExecTx`: we check
`actor.TxFromContext(ctx)` and, when present, pass the existing `*sql.Tx`
directly to `createQuery`, bypassing the retry loop entirely (the outer
transaction owner handles retries and commit/rollback). When no outer
transaction is present, existing behavior is preserved unchanged.

## Tests

Four new tests in `db/tx_executor_ctx_test.go` verify the behavior:

`TestExecTx_JoinOuterActorTx` confirms that writes through `ExecTx` land
in the outer transaction and aren't visible globally until the outer tx
commits. `TestExecTx_StandaloneNoTx` is the regression test for the
existing standalone path. `TestExecTx_OuterTxRollbackAtomicity` exercises
the key invariant: two sequential `ExecTx` calls joining the same outer
transaction are both discarded atomically on rollback. Finally,
`TestExecTx_OuterTxErrorPropagation` verifies that errors from `txBody`
propagate cleanly without touching the outer transaction's lifecycle.

changes.

### DIFF
--- a/db/interfaces.go
+++ b/db/interfaces.go
@@ -237,9 +237,12 @@ func NewTransactionExecutor[Querier any](db BatchedQuerier,
 func (t *TransactionExecutor[Q]) ExecTx(ctx context.Context,
 	txOptions TxOptions, txBody func(Q) error) error {
 
-	// If a durable actor transaction is already attached to the context,
-	// participate in that transaction instead of starting a nested one.
-	// Retry semantics are handled by the outer transaction executor.
+	// If the context already carries a database transaction from the
+	// durable actor framework, join it instead of creating a new one.
+	// This ensures that all store operations within a single actor
+	// message are executed atomically in the same transaction. The
+	// provided txOptions are ignored in this case; isolation level
+	// and read-only semantics are governed by the outer transaction.
 	if tx, ok := actor.TxFromContext(ctx); ok {
 		return txBody(t.createQuery(tx))
 	}
--- a/db/tx_executor_ctx_test.go
+++ b/db/tx_executor_ctx_test.go
@@ -1,0 +1,265 @@
+package db
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


Roasbeef:
@claude review this

Roasbeef:
@claude Rebased onto origin/main and addressed all review feedback. Here's what changed:

- **txOptions comment (Gemini)**: Added note clarifying that `txOptions` are ignored when joining an outer transaction — isolation and read-only semantics are governed by the outer tx.
- **Missing defer (Gemini)**: Added `defer outerTx.Rollback()` in `TestExecTxOuterTxRollbackAtomicity` for consistent cleanup.
- **Postgres placeholders (Codex)**: Added `//go:build !test_postgres` constraint since the test uses raw SQLite-specific DDL/DML (`?` placeholders, `INTEGER PRIMARY KEY`).
- **Test naming**: Removed underscores from all 4 test function names (e.g. `TestExecTx_JoinOuterActorTx` → `TestExecTxJoinOuterActorTx`).
- **Linter**: Replaced 8x `context.Background()` with `t.Context()` to fix `usetesting` lint violations.

All tests pass, linter is clean. Please review again.

PR #126, opened 2026-02-25
darepod: add SDK-style test lifecycle and bufconn support

sputn1ck:
## Summary

- Add `RunWithContext` method for test/SDK lifecycle management via context cancellation (instead of requiring a process-global `signal.Interceptor`)
- Add `RPCAddr()` to expose the resolved gRPC listen address for test harnesses using OS-assigned ports
- Add optional `Listener` field to `RPCConfig` for in-memory transports (bufconn) and SDK-style embedding
- Add darepod binary to `.gitignore`

Companion server PR: lightninglabs/darepo#114

## Test plan

- [x] `make build` passes on server with these changes
- [x] `make arktest-verbose` passes (TestDaemonGetInfo via bufconn)
- [x] `make systest` passes (18/18 tests)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ DS_Store
 
 # Go workspace sum file - regenerated locally as needed.
 go.work.sum
+
+bin/darepod
--- a/daemonrpc/daemon.pb.go
+++ b/daemonrpc/daemon.pb.go
@@ -76,7 +76,11 @@ type GetInfoResponse struct {
 	// operator is active and pulling envelopes.
 	ServerConnected bool `protobuf:"varint,6,opt,name=server_connected,json=serverConnected,proto3" json:"server_connected,omitempty"`
 	// lnd_alias is the alias of the connected lnd node.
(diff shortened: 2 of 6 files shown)


bhandras:
Closing this draft as superseded by the current `darepod` embedding / SDK stack.

The core ideas from this PR have now landed in smaller, focused slices:

- `RunWithContext`, `RPCAddr()`, and `DaemonReady()` are already on `main`.
- injected `RPCConfig.Listener` support and the focused listener-path tests landed in `#289`.
- the richer daemon status surface via `GetInfo.ServerInfo` landed in `#290`.
- the consumer-facing Go SDK on top of that runtime boundary landed in `#291`.

So the original thesis of this PR is still the right direction, but the implementation has been absorbed into the newer stack in a cleaner shape:
- keep `darepod` as the canonical runtime
- make embedding a transport/lifecycle decision via injected listeners / bufconn
- build `sdk/ark` on top of the daemon boundary rather than shipping one large draft bundle

Because of that, there is no remaining unique work in `#126` that still needs to be merged as-is. Thanks again for pushing this direction early — the current stack builds directly on it.

PR #132, opened 2026-03-02
round: unify pool additions via single IntentPackage event

ellemouton:
## Summary

- **Move actor-only messages** (`RefreshVTXORequest`, `LeaveVTXORequest`,
  `buildVTXORequestFromRefresh`) out of `events.go` into `actor.go`, clarifying
  the boundary between FSM events and actor messages
- **Replace redundant types** — remove `ForfeitIntent` (carried duplicate
  amount), `round.ForfeitRequest`, and `round.LeaveRequest`; use
  `types.ForfeitRequest` and `types.LeaveRequest` directly, eliminating
  conversion loops in join-round authorization
- **Consolidate 4 FSM events into 1** — replace `BoardingUTXOConfirmed`,
  `VTXORequestsReceived`, `ResumeBoardingIntents`, and the existing
  `IntentPackage` with a single `IntentPackage` carrying all four pools
  (boarding, forfeits, VTXOs, leaves). Input validation moves to the actor
  layer where it belongs
- **Update README** to reflect the unified event model

Net result: ~350 fewer lines, one FSM event entry point instead of four,
and forfeit amounts resolved from VTXOStore at registration time rather
than carried through the event chain.

## Test plan

- [x] `make build` passes at each commit
- [x] `make unit pkg=round` passes at each commit
- [x] `make lint` passes

### DIFF
--- a/round/actor.go
+++ b/round/actor.go
@@ -780,6 +780,36 @@ func (a *RoundClientActor) handleWalletBoardingConfirmed(ctx context.Context,
 		slog.Int("amount", int(walletIntent.ChainInfo.Amount)),
 		slog.Int("conf_height", int(walletIntent.ChainInfo.ConfHeight)))
 
+	// Validate chain data that the FSM previously checked.
+	confTx := walletIntent.ChainInfo.ConfTx
+	if confTx == nil {
+		return fn.Err[actormsg.RoundActorResp](fmt.Errorf(
+			"boarding confirmation missing tx"))
+	}
+	if int(walletIntent.Outpoint.Index) >= len(confTx.TxOut) {
+		return fn.Err[actormsg.RoundActorResp](fmt.Errorf(
+			"invalid outpoint index %d for tx %s",
+			walletIntent.Outpoint.Index,
+			walletIntent.Outpoint.Hash))
+	}
+
+	// Build the boarding request from the wallet intent. Chain-level
+	// information (ConfHeight, ConfHash, ConfTx, TxProof, Amount) is
+	// carried through the embedded wallet.BoardingIntent.ChainInfo
+	// and remains available to downstream consumers (e.g.,
+	// buildJoinRoundAuthRequest uses ChainInfo.Amount for BIP-322
+	// proof construction).
+	boardingReq := types.BoardingRequest{
+		Outpoint:    &walletIntent.Outpoint,
+		ClientKey:   walletIntent.Address.KeyDesc.PubKey,
+		OperatorKey: walletIntent.Address.OperatorKey,
+		ExitDelay:   walletIntent.Address.ExitDelay,
+	}
+	intent := BoardingIntent{
+		BoardingIntent: *walletIntent,
+		Request:        boardingReq,
+	}
+
 	// Find an existing assembling round (Idle or PendingRoundAssembly) or
 	// create a new one. This allows multiple boarding confirmations to
 	// accumulate in the same round.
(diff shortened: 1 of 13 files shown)


ellemouton:
@claude review this

ellemouton:
@claude - were all your comments addressed in the latest push?

round/events.go
```diff
@@ -338,13 +346,46 @@ type RoundComplete struct{}
...
+// as a single event. The FSM unpacks the package and appends each item to its
+// respective pool. All items in a package are accepted together — the upstream
+// caller decides what must travel as a unit.
+//
+// Examples:
+//   - Refresh: {Forfeits: [1], VTXOs: [1]}
+//   - Leave:   {Forfeits: [1], Leaves: [1]}
+//   - Consolidate N-to-1: {Forfeits: [N], VTXOs: [1]}
+//   - Split 1-to-N: {Forfeits: [1], VTXOs: [N]}
+type IntentPackage struct {
```
Roasbeef, line 292: 👍

round/interfaces.go
```diff
@@ -228,6 +240,11 @@ type Intents struct {
 	// outputs. Each leave forfeits a VTXO and creates an on-chain output
 	// in the batch transaction instead of a new VTXO.
 	Leaves []*LeaveRequest
+
+	// Forfeits contains the VTXOs being forfeited as inputs. Decoupled
```
Roasbeef: Do we need _both_ this _and_ `IntentPackage`? Seems like we can consolidate even further.

round/actor.go
```diff
@@ -793,21 +818,11 @@ func (a *RoundClientActor) handleWalletBoardingConfirmed(ctx context.Context,
 		}
 	}
 
-	// Create the FSM event from the wallet's confirmed intent. Wallet only
-	// notifies after min confs, so we set confirmations to 1. Include the
-	// Address and TxProof for building the Request.
-	confirmEvt := &BoardingUTXOConfirmed{
-		Outpoint:      walletIntent.Outpoint,
-		Address:       walletIntent.Address,
-		BlockHeight:   walletIntent.ChainInfo.ConfHeight,
```
Roasbeef, line 707: We no longer need this chain level information? Eg: the round actor could wait for more confirmations to pass before permitting the input to be boarded in a round.

round/transitions.go
```diff
@@ -295,122 +187,53 @@ func (s *PendingRoundAssembly) ProcessEvent(ctx context.Context,
...
 	case *IntentPackage:
-		// An atomic bundle of intents. Unpack into our pools.
+		// An atomic bundle of intents. Unpack into our pools,
+		// deduplicating boarding intents by outpoint.
 		if evt.isEmpty() {
 			return selfLoop(s), nil
 		}
 
+		// Deduplicate boarding intents by outpoint.
+		updatedBoarding := slices.Clone(s.Boarding)
```
Roasbeef, line 204: Insert into an `fn.Set` and let that handle it?

Roasbeef: Same here below re dup check.

Roasbeef (dismissed):
I dig the simplification. I think what we're missing here is: actor level tests for the new mapping+verification.

round/interfaces.go
ellemouton: Done — `IntentPackage` now embeds `Intents`, so the four pool fields are defined once and shared between event transport and state storage.

_— by claude_

round/actor.go
ellemouton: Good point — the chain info is still fully available via the embedded `wallet.BoardingIntent.ChainInfo` (ConfHeight, ConfHash, ConfTx, TxProof, Amount), so downstream consumers like `buildJoinRoundAuthRequest` still access it. But you're right that we lost the ability for the *round actor* to enforce its own confirmation policy independently of the wallet — e.g., requiring more confs for high-value inputs. Currently the wallet gates on `MinConfirmations` and the round actor trusts that. Is that something you'd want added back here, or is wallet-level gating sufficient?

_— by claude_

round/transitions.go
ellemouton: Done — both boarding and forfeit dedup now use `fn.Set[wire.OutPoint]` for O(1) lookups instead of `slices.ContainsFunc`.

_— by claude_

ellemouton (commented):
Added `TestActorIntentMapping` covering the actor's translation of raw inputs into `IntentPackage` events:

- Boarding chain info preserved through type embedding
- Actor rejects nil ConfTx and invalid outpoint indices
- Duplicate boarding outpoints deduplicated by `fn.Set`
- Refresh maps correctly to forfeit input + VTXO output
- Leave maps correctly to forfeit input + leave output
- Mixed rounds accumulate all four pools
- Duplicate forfeit outpoints deduplicated

_— by claude_

sputn1ck (approved):
LGTM!

PR #136, opened 2026-03-03
actor: strip sender tx in DurableMailbox.Send

Roasbeef:
## Summary

- Add `WithoutTx()` to the actor package that strips database
  transactions from context at actor boundaries
- Use it in `DurableMailbox.Send()` before `EnqueueMessage()` to
  prevent the sender's transaction from leaking into the receiver's
  mailbox store

## Problem

When a `DurableActor` processes a message inside
`TxAwareActorDeliveryStore.ExecTx`, the context carries the actor's
`*sql.Tx` via `actor.WithTx`. If the behavior calls another actor via
`Ask` (e.g., `server.Receive(ctx, msg)`), the context flows through:

```
DurableMailbox.Send(ctx) → Store.EnqueueMessage(ctx) → TransactionExecutor.ExecTx(ctx)
```

The `TxFromContext` check in `ExecTx` joins the **sender's**
transaction, which causes two issues:

1. **Cross-DB actors**: the message is written to the sender's DB,
   invisible to the receiver's processing goroutine (hang/deadlock)
2. **Same-DB Ask calls**: the sender blocks waiting for a response
   while the receiver cannot see the uncommitted message (deadlock)

## Why this is safe

The outbox atomicity guarantee is **unaffected**:

- The actor's own store operations (FSM state writes, outbox table
  writes) still join the actor's transaction via `TxFromContext`
- Only cross-actor mailbox enqueues are isolated
- This matches the production gRPC boundary where Go context values
  are naturally stripped when crossing process boundaries
- The `OutboxPublisher` already uses a fresh context (`WithOutboxID`
  only, no tx) when delivering messages

## Test plan

- [x] All OOR e2e tests pass (previously hung indefinitely)
- [x] All OOR actor unit tests pass
- [x] All `ExecTx` transaction joining tests pass

### DIFF
--- a/baselib/actor/durable_mailbox.go
+++ b/baselib/actor/durable_mailbox.go
@@ -107,8 +107,13 @@ func NewDurableMailbox[M TLVMessage, R any](
 // envelope is accepted, the provided context is cancelled, or the actor's
 // context is cancelled.
 //
-// If the context contains a transaction (via WithTx), the message is written
-// within that transaction, enabling atomic outbox writes.
+// The sender's database transaction is intentionally stripped before
+// enqueueing. Mailbox sends cross actor boundaries: the receiver's
+// delivery store must manage its own transaction lifecycle. Inheriting
+// the sender's transaction would either target the wrong database
+// (cross-DB actors) or deadlock for Ask-style calls where the sender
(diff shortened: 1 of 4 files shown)


Roasbeef:
@claude review this

PR #155, opened 2026-03-07
darepod+darepocli: add wallet RPC surface, agent-first CLI, and MCP server

Roasbeef:
In this PR, we extend the `darepod` daemon with a full wallet lifecycle RPC
surface and introduce `darepocli`, a new CLI binary designed from the ground up
for both human operators and AI agent consumption. The daemon now supports two
wallet backends (`lnd` and `lwwallet`), and the CLI exposes every RPC through
three parallel surfaces: cobra subcommands for humans, raw proto-JSON input for
agents (via `--json`), and a Model Context Protocol stdio server for tool-use
LLMs (via `darepocli mcp serve`).

changes.

## Proto and Daemon RPC Surface

We add nine new RPCs to `daemon.proto`: `GenSeed`, `InitWallet`, `UnlockWallet`
for the wallet lifecycle; `GetBalance`, `NewAddress` for basic wallet ops;
`ListVTXOs` with status/amount filtering for VTXO inventory; and `SendVTXO`,
`SendOOR`, `RefreshVTXOs` for transfers. The `GetInfoResponse` is extended with
`wallet_type` and `wallet_ready` fields so clients can detect the daemon's
startup state before issuing wallet-dependent calls.

On the daemon side, `RunUntilShutdown` is restructured to support dual backends
selected via `wallet.type` config. In `lwwallet` mode, the gRPC listener starts
_before_ the wallet is ready so the daemon can accept `GenSeed`/`InitWallet`/
`UnlockWallet` calls. A `walletReady` channel gates everything else. If
`DAREPOD_WALLET_PASSWORD` or `--wallet.password_file` is configured, the daemon
auto-unlocks at startup for fully unattended operation.

## Seed Manager and Key Material Handling

The seed manager (`seed_manager.go`) handles aezeed mnemonic generation,
seed encryption via scrypt + NaCl secretbox (XSalsa20-Poly1305), and
loading seeds from env vars or encrypted files on disk. After review, we
added `zeroBytes` defer calls in `EncryptSeed`, `DecryptSeed`, and the
plaintext recovery path to clear scrypt-derived keys, NaCl secret keys, and
decrypted seed bytes from memory once they're no longer needed.

## Wallet State Machine Hardening

The wallet state field (`WalletState`) is changed from a plain `int` to
`atomic.Int32`, enabling lock-free concurrent access from gRPC handler
goroutines and the startup goroutine. `InitWallet` uses `CompareAndSwap` to
atomically claim the `None → Locked` transition, preventing TOCTOU races when
two clients call `InitWallet` simultaneously. The `walletReady` channel close
is wrapped in `sync.Once` to guard against a double-close panic.

The identity pubkey derivation is switched from `DeriveNextKey` (which
increments the index on every call) to `DeriveKey` with a fixed
`KeyLocator{Family: 6, Index: 0}`, so the node identity is stable across
restarts. A new `DeriveKey` method is added to `lwwallet.Wallet` to support
this.

## Agent-First CLI (`darepocli`)

The CLI follows agent-first design principles: JSON-only output on stdout,
structured errors on stderr, and a `parseRequest[T]` generic helper that
dispatches between the `--json` agent path (proto-JSON unmarshaled directly
into the request) and the bespoke flags human path. Commands are organized as
git-style nested subcommands: `wallet create|unlock|balance|newaddress`,
`vtxos list|refresh`, `send inround|oor`, plus `getinfo`.

Password input follows a strict priority chain: stdin pipe >
`DAREPOD_WALLET_PASSWORD` env > `--wallet_password_file` flag > interactive
prompt. Passwords are never accepted as bare CLI arguments. Mutating commands
support `--dry-run` for safe validation before execution.

For context window discipline, `vtxos list` supports `--fields` (field mask
filtering, so agents can request only the columns they need) and `--ndjson`
(newline-delimited JSON, one object per VTXO, suitable for streaming into
`jq` or `wc -l`).

## Schema Introspection and MCP Server

`darepocli schema` provides runtime API introspection so agents can self-serve
method signatures, parameter types, and enum values without burning tokens on
static docs. The schema registry in `schema_registry.go` is the single source
of truth for both the `schema` command and the MCP tool definitions.

`darepocli mcp serve` starts a Model Context Protocol stdio server using the
official Go MCP SDK (`github.com/modelcontextprotocol/go-sdk`). Each daemon RPC
is exposed as a typed MCP tool with auto-generated input schemas from Go struct
tags. All 10 RPCs are registered as tools.

A Go AST-based verification script (`scripts/verify-schema-registry/main.go`)
parses the source to ensure schema registry entries, MCP tool registrations, and
cobra leaf commands stay in sync. It's wired up as `make schema-check` for CI.

## Config and Safety

`expandTilde` now returns `(string, error)` instead of silently swallowing
`os.UserHomeDir` failures. `NetworkDir()` and `LogDir()` propagate the error,
and all call sites handle it.

A mainnet safety guard in `Config.Validate()` requires `--allow-mainnet` (or
`allow-mainnet=true` in config) to run on mainnet, preventing accidental mainnet
operation during development.

The default gRPC transport in `darepocli` is changed from insecure to TLS using
the system certificate pool (TLS 1.2 minimum). `--no-tls` is still available
for local dev.

The incorrect `SendVTXO` implementation (which dispatched a self-refresh instead
of a directed send) is replaced with `codes.Unimplemented`, clearly documenting
that in-round directed sends require round protocol support for recipient
outputs.


Roasbeef:
Merged piece wise.

PR #168, opened 2026-03-10
vtxo+wallet: implement VTXO coin selection with atomic locking

### LINKED ISSUE #150: darepod: implement full VTXO coin selection + locking for OOR sends
## Summary

Implement proper VTXO coin selection with locking in the wallet actor
for OOR (out-of-round) sends. Currently the RPC layer uses a
`SelectAndLockVTXOsRequest` message to the wallet actor, but the wallet
actor doesn't implement actual locking semantics (marking VTXOs as
in-use to prevent double-spending across concurrent OOR sends).

## Current State

- `wallet/messages.go` defines `SelectAndLockVTXOsRequest`,
  `SelectAndLockVTXOsResponse`, `UnlockVTXOsRequest`,
  `UnlockVTXOsResponse`, and `SelectedVTXO` types
- `darepod/rpc_server.go` SendOOR handler uses these messages via
  the wallet actor Ask pattern
- The wallet actor needs to actually implement the lock/unlock handlers

## Design

- Wallet actor maintains an in-memory set of locked outpoints
- `SelectAndLockVTXOsRequest` handler: query vtxo store for live VTXOs,
  run coin selection (largest-first or similar), lock selected outpoints,
  return `SelectedVTXO` list
- `UnlockVTXOsRequest` handler: remove outpoints from locked set
- Locked VTXOs are excluded from future coin selection queries
- Locks are transient (in-memory) — cleared on daemon restart
- Consider TTL-based lock expiry for crash recovery

## Files

- `wallet/actor.go` — add lock state and message handlers
- `wallet/messages.go` — types already defined

ellemouton:
## VTXO coin selection + locking for transfers

Closes https://github.com/lightninglabs/darepo-client/issues/150

Implements full VTXO coin selection with in-memory locking to prevent
double-spends across concurrent OOR sends and in-round directed
transfers. The wallet actor selects VTXOs using a largest-first
strategy and delegates lock management to the VTXO manager via the
actor system's service key pattern.

### Architecture

Locking lives in the **VTXO manager** rather than the wallet because:

- The manager owns VTXO lifecycle (spawn, terminate, status tracking)
- Auto-unlock on termination is a natural fit
- Actor serialization (`Receive`) eliminates explicit synchronization
- The wallet avoids the `wallet → vtxo → round → wallet` import cycle
  by discovering the manager through the well-known
  `VTXOManagerServiceKey`

### Select & Lock Flow

Selection and locking happen in a **single atomic manager message**
(`SelectAndLockVTXOsRequest`) — no race window between list and lock.

```
┌────────┐  SelectAndLockVTXOsRequest(target)
│ Caller │─────────────────────────────────────►┌────────────────┐
│(RPC/   │                                      │  Wallet Actor  │
│ Round) │                                      └───────┬────────┘
└────────┘                                              │
            ① Ask: SelectAndLockVTXOsRequest            │
               (service key lookup)                     ▼
                                          ┌─────────────────────────┐
                                          │      VTXO Manager       │
                                          │  (single Receive call)  │
                                          │                         │
                                          │  ListLiveVTXOs()        │
                                          │  filter locked outpoints│
                                          │  selectCoinsLargestFirst│
                                          │  lockedOutpoints.Add()  │
                                          └─────────────────────────┘
            ② SelectAndLockVTXOsResponse                │
               { Selected, TotalSelected } ◄────────────┘
                        │
                        ▼
            ③ Wallet adapts to SelectedVTXO list
               and returns to caller
```

### Unlock Flow (failure / cancellation)

```
┌────────┐  UnlockVTXOsRequest(outpoints)
│ Caller │─────────────────────────────────────►┌────────────────┐
└────────┘                                      │  Wallet Actor  │
                                                └───────┬────────┘
            Forwards to VTXO manager via service key    │
                                                        ▼
                                                ┌───────────────┐
            lockedOutpoints.Remove(op) ─────────│ VTXO Manager  │
                                                └───────────────┘
            VTXOs available for next selection
```

### Auto-Unlock on Termination

```
  VTXO actor terminates (spent / forfeited / expired)
              │
              ▼
  VTXOTerminatedMsg ────────────────────────────┌───────────────┐
                                                │ VTXO Manager  │
  handleVTXOTerminated:                         │               │
    delete from actors map                      │  auto-unlock  │
    lockedOutpoints.Remove(op) ◄────────────────└───────────────┘
```

### Double-Spend Prevention

Locked VTXOs are invisible to subsequent selections. Two concurrent
callers requesting 50k each will never receive the same VTXO:

```
Caller A: SelectAndLock(50k) → gets vtxo1(60k), locks it
Caller B: SelectAndLock(50k) → vtxo1 filtered out, gets vtxo2(55k)
```

All access is serialized through the manager's `Receive` — no explicit
mutex needed.

### Design Notes

- **Locks are transient** (in-memory `fn.Set`): cleared on daemon
  restart. After restart there are no in-flight transfers, so nothing
  needs to stay locked.
- **Largest-first selection** minimizes input count. The pure function
  sorts a copy to avoid mutating the caller's slice.
- **`ManagerMsg` / `ManagerResp` are type aliases** (not embedding
  interfaces) so Go generics treat them as identical to
  `actormsg.VTXOManagerMsg` / `actormsg.VTXOManagerResp` — required for
  `RegisterWithSystem` to accept the service key's type parameters.
- **TTL-based lock expiry** (mentioned in issue) is deferred: auto-unlock
  on termination + transient locks cover the practical cases.

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -37,6 +37,7 @@ import (
 	"github.com/lightninglabs/darepo-client/rpc/roundpb"
 	"github.com/lightninglabs/darepo-client/serverconn"
 	"github.com/lightninglabs/darepo-client/timeout"
+	"github.com/lightninglabs/darepo-client/vtxo"
 	"github.com/lightninglabs/darepo-client/wallet"
 	"github.com/lightninglabs/lndclient"
 	lndbuild "github.com/lightningnetwork/lnd/build"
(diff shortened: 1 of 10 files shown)


vtxo/manager.go
```diff
@@ -219,6 +245,174 @@ func (m *Manager) handleVTXOTerminated(ctx context.Context,
 	return fn.Ok[ManagerResp](&VTXOTerminatedResp{})
 }
 
+// handleListAvailable returns all live VTXOs that are not currently locked.
+// This queries the store for live VTXOs and filters out any that appear in
+// the locked outpoints set.
+func (m *Manager) handleListAvailable(
```
Roasbeef, line 251: Should add tests for these new methods. 

Also do the vtxos themselves need to know that they're locked?

wallet/wallet.go
```diff
@@ -713,6 +719,115 @@ func (a *Ark) handleLeaveVTXOs(ctx context.Context,
...
+		return fn.Err[WalletResp](
+			fmt.Errorf("no actor system for VTXO selection"),
+		)
+	}
+
+	// Ask the VTXO manager to select and lock in a single actor message.
+	mgrKey := actormsg.VTXOManagerServiceKey()
+	mgrRef := mgrKey.Ref(a.actorSystem)
+
+	future := mgrRef.Ask(ctx, &actormsg.SelectAndLockVTXOsRequest{
```
Roasbeef, line 738: We could prob make a helper function (type param'd w/ the message type) for the 10 or so lines below this (actor call boiler plate).

wallet/wallet.go
```diff
@@ -713,6 +719,115 @@ func (a *Ark) handleLeaveVTXOs(ctx context.Context,
...
+	if a.actorSystem == nil {
+		return fn.Err[WalletResp](
+			fmt.Errorf("no actor system for VTXO unlock"),
+		)
+	}
+
+	mgrKey := actormsg.VTXOManagerServiceKey()
+	mgrRef := mgrKey.Ref(a.actorSystem)
+
+	future := mgrRef.Ask(
```
Roasbeef, line 802: Same here re the boiler plate: req, await, call, type cast.

darepod/server.go
```diff
@@ -1523,9 +1524,110 @@ func (s *Server) initRoundActor(ctx context.Context,
...
 }
 
+// initVTXOManager creates, registers, and starts the VTXO manager actor.
+// The manager spawns individual VTXO actors for each live VTXO and
+// maintains an in-memory lock set for coin selection. It is registered
+// under the well-known VTXOManagerServiceKey so the wallet actor can
+// find it via service key lookup for lock/unlock operations.
+func (s *Server) initVTXOManager(ctx context.Context,
+	clientWallet vtxo.VTXOWallet,
+	chainSourceRef actor.ActorRef[
```
Roasbeef, line 1554: Non-blocking, but can add some type aliases to the respective packages to cut down on the type decl length here.

internal/coinselect/coinselect_integration_test.go
```diff
@@ -0,0 +1,332 @@
+package coinselect
```
Roasbeef, line 1: Any reason to put put these in `systest`? I guess it doesn't necessarily need access to any of the backends?

wallet/wallet.go
```diff
@@ -713,6 +719,115 @@ func (a *Ark) handleLeaveVTXOs(ctx context.Context,
...
+	mgrRef := mgrKey.Ref(a.actorSystem)
+
+	future := mgrRef.Ask(ctx, &actormsg.SelectAndLockVTXOsRequest{
+		TargetAmount: int64(req.TargetAmount),
+	})
+	result := future.Await(ctx)
+
+	resp, err := result.Unpack()
+	if err != nil {
+		return fn.Err[WalletResp](
```
Roasbeef, line 745: These errors paths should unlock the locked vtxos. Also we may want to consider a sort of auto unlock timeout here as well. 

Also I think the `vtxo` actors themelves need to also know about this locking, as right now the path for refresh is wallet -> round <-> vtxo. Otherwise we could try an oor, but then it races with a refresh attempt.

vtxo/manager.go
```diff
@@ -219,6 +245,174 @@ func (m *Manager) handleVTXOTerminated(ctx context.Context,
...
+	})
+}
+
+// handleLockVTXOs adds the given outpoints to the locked set.
+func (m *Manager) handleLockVTXOs(ctx context.Context,
+	req *actormsg.LockVTXOsRequest) fn.Result[ManagerResp] {
+
+	var locked int
+	for _, op := range req.Outpoints {
+		if !m.lockedOutpoints.Contains(op) {
```
Roasbeef, line 319: Only add them if they're actually vtxo outpoints? The manager has a map of all the actual outpoints.

Roasbeef (changes requested):

PR #169, opened 2026-03-10
multi: implement in-round SendVTXO directed sends

ellemouton:
## Summary

Implements a first-class in-round send API, completing the `SendVTXO`
RPC that was previously stubbed as `codes.Unimplemented`.

Closes https://github.com/lightninglabs/darepo-client/issues/156 (takes
a different approach — wallet-orchestrated flow rather than round-driven).

Depends on #168.

## Approach

**Wallet orchestrates the send flow**: `SendVTXO RPC → wallet (coin
select + lock) → wallet sends complete intent to round actor`. The round
actor receives a fully assembled package rather than having to coordinate
coin selection itself.

### Key design decisions

1. **TriggerSendForfeitEvent vs TriggerRefreshEvent**: A new VTXO actor
   event avoids the `ForfeitRequest → RefreshVTXORequest` round-trip
   that refresh uses. The round actor already has the IntentPackage from
   the wallet; VTXO actors just need to be in `RefreshRequestedState`
   for forfeit signing.

2. **Recipient key format**: v1 supports `pubkey` (32-byte x-only →
   full VTXO descriptor) and `pk_script` (pre-computed tapscript)
   destinations. `address` destinations need an Ark-specific address
   format (future work).

3. **Change handling**: `changeAmount = totalSelected - totalAmount`. If
   > 0, a change VTXORequest is added via the existing `buildVTXORequest`
   (self-VTXO with fresh key derivation).

### Data flow

```
RPC SendVTXO
  → resolve recipients to VTXO pkScripts
  → wallet.SendVTXOsRequest (Ask)
    → VTXO manager: SelectAndLockVTXOs
    → compute change
    → round actor: TriggerVTXOSendMsg (Tell)
      → build IntentPackage (forfeits + recipient VTXOs + change)
      → feed to round FSM
      → trigger each VTXO actor: TriggerSendForfeitEvent
        → transition to RefreshRequestedState (no ForfeitRequest outbox)
        → ready for forfeit signing when round provides connectors
```

## Files changed

| File | What |
|------|------|
| `lib/actormsg/interfaces.go` | `SendRecipient`, `TriggerVTXOSendMsg` |
| `wallet/messages.go` | `SendVTXOsRequest`, `SendVTXOsResponse` |
| `wallet/wallet.go` | `handleSendVTXOs` handler + `Receive()` case |
| `round/vtxo_messages.go` | `TriggerSendForfeitEvent` |
| `vtxo/events.go` | Type alias for `TriggerSendForfeitEvent` |
| `vtxo/transitions.go` | `handleTriggerSendForfeit` in `LiveState` |
| `round/actor.go` | `handleTriggerVTXOSend` + `buildSendVTXORequest` |
| `darepod/rpc_server.go` | Complete `SendVTXO` handler |

## Test plan

- [x] `make build` compiles
- [x] `make lint` passes
- [x] VTXO transition unit tests (`TestLiveStateTriggerSendForfeit`,
  `TestLiveStateTriggerSendForfeitThenForfeitRequest`)
- [x] Round actor unit tests (`TestHandleTriggerVTXOSend` — error
  without actor system, send with recipients + change, send without
  change)
- [x] Integration tests (`TestSendVTXOsIntegration` — full
  wallet→manager→round flow with real actor system: send with change,
  exact amount, multiple recipients, insufficient balance, lock
  exclusion)

### DIFF
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -8,14 +8,18 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/darepo-client/build"
 	"github.com/lightninglabs/darepo-client/daemonrpc"
+	"github.com/lightninglabs/darepo-client/lib/actormsg"
 	"github.com/lightninglabs/darepo-client/lib/scripts"
+	"github.com/lightninglabs/darepo-client/lib/tree"
 	oortx "github.com/lightninglabs/darepo-client/lib/tx/oor"
+	"github.com/lightninglabs/darepo-client/lib/types"
 	"github.com/lightninglabs/darepo-client/lwwallet"
 	"github.com/lightninglabs/darepo-client/oor"
 	"github.com/lightninglabs/darepo-client/vtxo"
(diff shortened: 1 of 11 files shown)


lib/actormsg/interfaces.go
```diff
@@ -195,6 +195,46 @@ func (m *TriggerVTXORefreshMsg) MessageType() string {
...
+type TriggerVTXOSendMsg struct {
+	actor.BaseMessage
+
+	// ForfeitOutpoints identifies the VTXOs to forfeit as inputs
+	// (from coin selection).
+	ForfeitOutpoints []wire.OutPoint
+
+	// Recipients describes the recipient outputs to create in the new
+	// round.
+	Recipients []SendRecipient
```
Roasbeef, line 223: Will change be implicit? So cases where we have an amount left over.

round/actor.go
```diff
@@ -1938,3 +1941,145 @@ func (a *RoundClientActor) handleTriggerVTXOLeave(ctx context.Context,
...
+	a.log.InfoS(ctx, "Queued in-round send",
+		slog.Int("forfeits", len(forfeits)),
+		slog.Int("recipients", len(cmd.Recipients)),
+		slog.Int64("change", cmd.ChangeAmount))
+
+	// Trigger each forfeited VTXO actor so it transitions to
+	// RefreshRequestedState for forfeit signing. We use
+	// TriggerSendForfeitEvent which emits only a status update
+	// (no ForfeitRequest outbox) since we already have the intent.
+	for _, outpoint := range cmd.ForfeitOutpoints {
```
Roasbeef, line 2039: Perhaps more natural to process these as normal outbox messages?

wallet/wallet.go
```diff
@@ -828,6 +831,105 @@ func (a *Ark) handleUnlockVTXOs(ctx context.Context,
...
+			fmt.Errorf("no actor system for VTXO send"),
+		)
+	}
+
+	// Ask the VTXO manager to select and lock VTXOs covering the total
+	// send amount.
+	mgrKey := actormsg.VTXOManagerServiceKey()
+	mgrRef := mgrKey.Ref(a.actorSystem)
+
+	future := mgrRef.Ask(ctx, &actormsg.SelectAndLockVTXOsRequest{
```
Roasbeef, line 852: Nice, builds on top of the recent work.

darepod/rpc_server.go
```diff
@@ -569,23 +573,195 @@ func (r *RPCServer) SendVTXO(ctx context.Context,
...
+	}
+
+	locked, ok := selectResp.(*wallet.SelectAndLockVTXOsResponse)
+	if !ok {
+		return nil, status.Errorf(codes.Internal,
+			"unexpected response type: %T", selectResp)
+	}
+
+	// Immediately unlock the selected VTXOs.
+	if unlockErr := r.unlockVTXOs(
```
Roasbeef, line 753: Defer unlock? So gets the error case abvove.

internal/coinselect/coinselect_integration_test.go
```diff
@@ -330,3 +331,290 @@ func TestSelectWithNoActorSystem(t *testing.T) {
...
+	)
+
+	return walletActor, roundStub
+}
+
+// TestSendVTXOsIntegration exercises the full wallet→manager→round flow:
+// the wallet selects and locks VTXOs, computes change, and forwards a
+// TriggerVTXOSendMsg to the round actor with the correct forfeit
+// outpoints, recipients, and change amount.
+func TestSendVTXOsIntegration(t *testing.T) {
```
Roasbeef, line 436: More of a candidate for an addition to teh `systest` vs the pure coin selection (as here we can assert the new round occurs, and the receiver picks it up via the indexer or direct info send to it).

darepod/rpc_server.go
```diff
@@ -569,23 +573,195 @@ func (r *RPCServer) SendVTXO(ctx context.Context,
...
+		}
+
+		return desc.PkScript, nil
+
+	case *daemonrpc.Output_PkScript:
+		if len(d.PkScript) == 0 {
+			return nil, fmt.Errorf("pk_script is empty")
+		}
+
+		return d.PkScript, nil
```
Roasbeef, line 699: Does this need to (eventually) be wrapped in the AST system for safety re making sure all the exit paths are upheld?

round/actor.go
Roasbeef: Hmm also what if one of these fails (but not all). In the in-process realm, it's unlikely unelss there's some other race going on prob.

darepod/rpc_server.go
```diff
@@ -569,23 +573,195 @@ func (r *RPCServer) SendVTXO(ctx context.Context,
...
+			"wallet actor not initialized")
+	}
+
+	wRef := r.server.walletRef.UnsafeFromSome()
+
+	// Ask the wallet to select, lock, and forward the send to the
+	// round actor.
+	sendReq := &wallet.SendVTXOsRequest{
+		Recipients:  recipients,
+		TotalAmount: btcutil.Amount(totalAmount),
```
Roasbeef, line 624: Dust validation? 

Though could say the future `testmempoolaccept` layer can help take care of that? Or I gues this would be a server param re the dust limit?

Roasbeef (changes requested):

ellemouton:
closing in favour of https://github.com/lightninglabs/darepo-client/pull/176

PR #172, opened 2026-03-11
vtxo+round+wallet: purify VTXO FSM and move intent composition to wallet

ellemouton:
Refactor groundwork for VTXO coin selection (#150).

This PR simplifies the boundary between `vtxo`, `round`, and `wallet`
so upcoming work on FSM-based locking (#168) and directed in-round
sends can build on a cleaner ownership model.

At a high level:

- The VTXO FSM now models lifecycle only, not business intent.
- The wallet now owns intent composition.
- The round actor now validates and registers pre-composed intents
  instead of assembling them itself.

No protocol change is intended here. This is a structural refactor that
removes overlapping responsibilities, dead message paths, and context
lifetime bugs.

## Why

Before this change, intent ownership was split across multiple layers:

- The wallet triggered refresh/leave indirectly.
- VTXO actors translated those triggers into round-specific messages.
- The round actor still had to reconstruct intent packages.

That made the architecture harder to extend for coin selection and
locking, because the system did not have a single clear owner for
"which VTXOs are being consumed, and for what purpose".

This PR makes that ownership explicit:

- `wallet` decides what intent package to register.
- `round` drives protocol execution for that package.
- `vtxo` tracks availability and lifecycle while remaining agnostic to
  whether a cooperative consume is a refresh, leave, or later another
  flow.

### 1. VTXO FSM now models lifecycle, not business intent

Refresh and leave were previously represented as separate concepts in
the VTXO FSM even though, from the VTXO actor's perspective, both are
just cooperative forfeits.

This PR simplifies that model:

- `RefreshRequestedState` -> `PendingForfeitState`
- `ExpiringState` -> `UnilateralExitState`
- `TriggerRefreshEvent` and `TriggerLeaveEvent` collapse into
  `PendingForfeitEvent`
- `RefreshAcknowledgedEvent` is removed as dead/no-op state machinery

This leaves the VTXO FSM focused on lifecycle transitions:

- live
- pending cooperative consume
- forfeiting
- forfeited
- unilateral exit
- failed

### 2. Round-bound VTXO signals now route through the manager

VTXO actors no longer hold a direct round actor reference.

Instead:

- VTXO actors emit relay messages to the VTXO manager
- the manager forwards round-bound payloads
- round-to-VTXO notifications still use actor lookup by service key

This keeps VTXO actors simpler and makes outbound coordination flow
through one place.

### 3. Wallet now owns intent composition

The biggest architectural change is that the wallet now builds the full
intent package for refresh and leave flows.

That means:

- the wallet loads VTXO descriptors itself
- the wallet composes forfeits plus replacement VTXOs or leave outputs
- the round actor receives a pre-composed package via
  `RegisterIntentRequest` / `RegisterIntentMsg`
- the round actor validates and registers that package with the FSM

The round actor no longer needs separate wallet-trigger paths for
refresh/leave intent assembly.

### 4. Forfeit amounts are carried locally in intents

Forfeit amounts are now carried in the local intent data used during
registration.

This removes fragile registration-time store lookups and makes amount
validation more direct when building the round request.

The amount remains local bookkeeping, not a wire-level protocol change.

### 5. Persistence work no longer depends on request-context lifetime

Several local persistence operations now use `context.WithoutCancel`
when the work must survive beyond the actor request that triggered it.

This applies to operations like:

- round checkpoint persistence
- confirmed VTXO persistence
- local registration-time work that should not be abandoned because the
  triggering request context expired

### Ownership after this PR

```text
wallet
  -> load VTXO data
  -> compose IntentPackage
  -> register with round

round
  -> validate/register package
  -> drive round FSM
  -> notify VTXO actors when cooperative consume becomes concrete

vtxo
  -> track lifecycle / availability
  -> sign forfeits when requested
  -> escalate to unilateral exit if expiry becomes critical
```

### VTXO lifecycle before vs after

Before, the FSM encoded refresh/leave intent explicitly.
After, it models only the lifecycle of cooperative consumption.

```text
Before:
Live
  -> RefreshRequested
  -> Forfeiting
  -> Forfeited

Live
  -> Expiring
  -> Failed

After:
Live
  -> PendingForfeit
  -> Forfeiting
  -> Forfeited

Live
  -> UnilateralExit
  -> Failed
```

## Result

After this PR:

- VTXO actors no longer care why they are being cooperatively consumed
- the wallet is the single place where round intents are composed
- the round actor has a narrower, cleaner responsibility
- the codebase is in a better position for VTXO locking and coin
  selection work

## Test plan

- [x] `make unit pkg=vtxo`
- [x] `make unit pkg=round`
- [x] `make unit pkg=wallet`
- [x] `make build`
- [x] `make lint`


ellemouton:
@claude review this

ellemouton:
@claude review again

round/actor.go
```diff
@@ -1861,24 +1876,42 @@ func (a *RoundClientActor) handleTriggerVTXORefresh(ctx context.Context,
...
+
+			continue
+		}
+
 		serviceKey := actormsg.VTXOActorServiceKey(outpoint)
-		err := serviceKey.Ref(a.cfg.ActorSystem).Tell(
-			ctx, &TriggerRefreshEvent{
-				ForceRefresh: cmd.ForceRefresh,
-			},
+		err = serviceKey.Ref(a.cfg.ActorSystem).Tell(
```
Roasbeef: So the round actor speaks directly to the vtxo actors, but the vtxo actor relays through the round actor? 

Why not make this symmetric? On either end of the relationship. One other thing to think about here is: which of these actors will be made durable? Right now we have plas to make the round actor durable, so it can survive restarts when receiving messages from the wallet/server.

round/actor.go
```diff
@@ -114,6 +114,30 @@ func (e *LeaveVTXORequest) MessageType() string {
...
+// it with the FSM, then notifies affected VTXO actors.
+//
+// This replaces the older pattern where the round actor loaded VTXO
+// descriptors and composed intents itself (via TriggerVTXORefreshMsg /
+// TriggerVTXOLeaveMsg).
+type RegisterIntentRequest struct {
+	actor.BaseMessage
+
+	// Package is the fully composed round intent bundle.
+	Package *IntentPackage
```
Roasbeef, line 92: Hmm, but now both the round actor _and_ the wallet build up these intent packages?

round/actor.go
```diff
@@ -859,12 +840,6 @@ func (a *RoundClientActor) Receive(ctx context.Context,
 	case *ForfeitSignatureResponse:
 		return a.handleForfeitSignatureResponse(ctx, m)
 
-	case *actormsg.TriggerVTXORefreshMsg:
```
Roasbeef, line 808: 👍 for removing these, was working in this area earlier in the week and was a bit confused re why we had to very similar messages sent into the actor.

Roasbeef (approved):
Solid refactor 

Found a few things locally and regenerated the docs, will tack on as extra commits.

round/actor.go
Roasbeef: Nvm resolved at the end.

Roasbeef:
## Review Fixes

Pushed three commits addressing findings from code review:

### `083e68e0` — round: use findAssemblingRound for intent registration and add VTXO dedup

`handleRegisterIntent` and `handleRefreshVTXORequest` were using `findPendingRound()` which matches by temp-key status. This includes rounds in `RegistrationSentState` — feeding an `IntentPackage` to that state causes a silent self-loop, discarding the intent without error. Switched both to `findAssemblingRound()` which correctly filters by FSM state (Idle or PendingRoundAssembly).

Also added PkScript-based deduplication for VTXO requests in `PendingRoundAssembly.ProcessEvent`. The forfeit pool already deduplicates by outpoint, but duplicate VTXO output requests (from wallet + auto-expiry racing) would inflate `totalOutput` and cause the balance check to fail.

Tests: `intent_creates_new_round_when_existing_in_registration_sent`, `duplicate_vtxo_pkscript_deduplicated`.

### `71ec7c69` — round: validate forfeit amounts against VTXOStore instead of trusting embedded

`computeTotalForfeitAmount` was trusting the embedded `Amount` field on `ForfeitRequest` and skipping the VTXOStore lookup when it was non-zero. Changed to always use VTXOStore as the canonical source of truth when a store is available. The embedded amount is now only a fallback for test environments (nil store).

Updated the stale `Forfeits` field comment in `Intents` and the `sortedForfeitRequests` comment.

Test: `TestComputeTotalForfeitAmountStoreOverridesEmbedded` — verifies store amounts (100+200=300) override inflated embedded amounts (999+888).

### `a459522d` — docs: update per-package docs for VTXO FSM refactor

Updated `round/CLAUDE.md`, `vtxo/CLAUDE.md`, `wallet/CLAUDE.md`, and `ARCHITECTURE.md` with new state names (`PendingForfeit`, `UnilateralExit`), message types (`PendingForfeitEvent`, `RegisterIntentMsg`), and the wallet's intent composition ownership.

### Note on proto stubs

`make rpc` produces the same output already on the branch — the Docker container uses protoc v3.21.12. The version difference vs main (v5.28.0) is because main was generated with a local install rather than the Docker container. No action needed.

Issue #124, opened 2026-02-25
oor: integrate OOR receive path with VTXO actor lifecycle


Roasbeef:
## Summary

The OOR receive path currently writes VTXO descriptors directly to the database without notifying the VTXO manager or spawning VTXO actors. This means OOR-received VTXOs are inert rows in the DB — no actor manages their lifecycle, no one monitors their expiry, and they can't participate in refresh rounds or forfeit flows.

Additionally, the `Descriptor` tracks `TreeDepth` (position in the commitment tree) but not the **OOR chain depth** (number of OOR hops since the last on-chain commitment). This chain depth is critical for unilateral exit cost analysis and go-on-chain decisions.

## Current Behavior

When an incoming OOR transfer arrives:

1. `ReceiveNotified` emits `MaterializeIncomingVTXOsRequest`
2. `LocalPersistenceOutboxHandler.handleMaterializeIncoming` (`oor/local_persistence_handler.go:141`):
   - Resolves client key + metadata via callbacks
   - Calls `BuildIncomingVTXODescriptor()` to build a `vtxo.Descriptor`
   - Writes directly to DB via `h.Store.SaveVTXO(ctx, desc)` (line 217)
3. FSM transitions to `ReceiveAwaitingAck` → acks server → `ReceiveCompleted`

**No message is sent to the VTXO manager.** Compare with the round path where `VTXOCreatedNotification` is emitted and forwarded to the VTXO manager (`round/actor.go:1181-1190`) to spawn VTXO actors.

### Consequences

- **No VTXO actor spawned**: Nobody monitors expiry via `BlockEpochEvent`, nobody handles `ForfeitRequestEvent` for refresh rounds
- **No refresh participation**: If the client wants to refresh the OOR VTXO in a future round, the round actor can't send forfeit requests to a non-existent VTXO actor
- **No spend monitoring**: No `RegisterSpendNtfn` for detecting prior-owner unilateral exits (related: issue #63)
- **No chain depth tracking**: `TreeDepth` exists on the descriptor but represents position in the VTXT, not the number of OOR hops — these are different concepts

### 1. Notify VTXO Manager on OOR Receive

After `SaveVTXO`, the OOR receive path should send a notification to the VTXO manager (same pattern as `VTXOCreatedNotification` in the round actor). This spawns a VTXO actor for the new OOR VTXO so it can:

- Monitor expiry
- Handle forfeit requests when the VTXO is used in a future round
- Register spend notifications for double-spend monitoring

[...]

PR #173, opened 2026-03-11
oor+vtxo: activate VTXO actors for OOR receives, add ChainDepth plumbing

ellemouton:
Partial progress on #124 (Parts 1 and 2)

## Summary

This PR addresses the OOR receive lifecycle gap from #124 and lays the
data-model groundwork for chain-depth-aware exit policy.

It does two things:

1. It fixes OOR receive so incoming VTXOs do not stop at "persisted row in
   the DB". After an incoming OOR VTXO is materialized, the OOR flow now
   notifies `VTXOManager`, which spawns a live VTXO actor for that
   descriptor.
2. It adds first-class `ChainDepth` tracking to VTXOs. `ChainDepth` is
   distinct from `TreeDepth`: it represents the number of OOR checkpoint
   hops between a VTXO and the most recent on-chain commitment.

### Part 1: OOR receive now activates VTXO actors

Before this change, incoming OOR VTXOs were written to the `vtxos` table but
never forwarded to `VTXOManager`, so no VTXO actor was spawned.

This PR changes that by:
- wiring the OOR actor to a real `VTXOManager`
- notifying the manager after incoming OOR VTXOs are durably materialized
- having the manager spawn actors for already-persisted descriptors
- wiring daemon startup so both round-created and OOR-created VTXOs flow into
  the same manager

Result:
- OOR-received VTXOs now participate in the normal actor lifecycle
- expiry monitoring, forfeit handling, and related VTXO-actor-driven behavior
  are no longer bypassed for received OOR VTXOs

### Part 2: persisted `ChainDepth` plumbing

This PR also introduces `ChainDepth` on VTXOs and threads it through:
- `vtxo.Descriptor`
- OOR incoming metadata / descriptor construction
- DB schema and persistence
- daemon RPC VTXO responses
- indexer VTXO protobufs
- tests and systest coverage

Round-created VTXOs explicitly default to `ChainDepth = 0`.

## What This PR Does Not Cover

This PR does **not** complete the full production metadata wiring needed to
source authoritative non-zero incoming OOR `ChainDepth`.

That gap already existed before this branch:
- the live OOR actor setup in `darepod/server.go` still does not wire the full
  incoming metadata resolver path
- because of that, the live daemon still lacks the production source that
  should eventually provide incoming OOR lineage metadata, including chain
  depth

So while this PR makes the model, storage, and APIs ready for
chain-depth-aware behavior, it does **not** claim to fully complete Part 2
end-to-end in production.

This PR also does **not** implement Part 3 from #124:
- no fee-based unilateral-exit cost model
- no chain-depth-aware go-on-chain policy changes
- no change to the expiry policy beyond making the correct field available for
  future work

## Why This Split Still Makes Sense

Even without the final production metadata resolver wiring, this PR is still
valuable on its own because it:
- fixes the original actor-lifecycle bug for OOR receives
- eliminates the need for a future schema migration or proto break when
  chain-depth-aware policy lands
- keeps the code honest by preserving `ChainDepth` wherever authoritative
  metadata is supplied, without inventing values from `TreeDepth`

## Testing

Covered by:
- unit tests for OOR manager notification wiring
- unit tests for round manager wiring
- unit tests for `ChainDepth` propagation and round-trip persistence
- systest proving OOR materialization results in a live VTXO actor, and that
  injected `ChainDepth` survives materialization/persistence

Commands run:
- `make unit pkg=oor`, `make unit pkg=vtxo`, `make unit pkg=db`, `make unit pkg=darepod`
- `go test -tags systest ./systest -run TestOORIncomingMaterializationSpawnsVTXOActor -count=1`
- `make lint`

## Follow-up Work

Remaining follow-up after this PR:
- wire the production OOR incoming metadata resolver path in
  `darepod/server.go`
- source authoritative incoming OOR `ChainDepth` from the receive metadata
  path
- implement Part 3 policy logic using `ChainDepth` instead of overloading
  `TreeDepth`

### DIFF
--- a/oor/actor.go
+++ b/oor/actor.go
@@ -12,6 +12,7 @@ import (
 	"github.com/lightninglabs/darepo-client/baselib/actor"
 	"github.com/lightninglabs/darepo-client/build"
 	"github.com/lightninglabs/darepo-client/serverconn"
+	"github.com/lightninglabs/darepo-client/vtxo"
 	fn "github.com/lightningnetwork/lnd/fn/v2"
 )
 
@@ -64,6 +65,10 @@ type ClientActorCfg struct {
 	// ActorID is the durable mailbox id used for this actor instance.
 	// Re-using the same ActorID across restarts enables checkpoint restore.
 	ActorID string
+
+	// VTXOManager receives notifications after incoming VTXOs are durably
+	// materialized so it can spawn VTXO actors for monitoring.
+	VTXOManager actor.TellOnlyRef[vtxo.ManagerMsg]
 }
 
 // OORClientActor wraps the outgoing-transfer client FSM in a durable actor
@@ -948,6 +953,12 @@ func (b *oorDurableBehavior) driveOutbox(ctx context.Context,
 		}
 
 		for _, followUp := range followUps {
+			// When incoming VTXOs are materialized, forward
+			// them to the VTXO manager so it can spawn
+			// monitoring actors. This mirrors the rounds
+			// actor pattern for VTXOCreatedNotification.
+			b.notifyMaterializedVTXOs(ctx, followUp)
+
 			finalizeState, err := b.captureFinalizeStateForEvent(
 				fsm, followUp,
 			)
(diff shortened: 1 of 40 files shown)


oor/actor.go
```diff
@@ -142,6 +147,8 @@ func newOORActorCodec() *actor.MessageCodec {
 // messages. If startup prerequisites fail, the returned actor stores the error
 // and surfaces it on Receive.
 func NewOORClientActor(cfg ClientActorCfg) *OORClientActor {
+	wireVTXOManagerNotifier(&cfg)
+
```
ellemouton: fixing

ellemouton:
@claude review this

Roasbeef (approved):
LGTM 🌽

PR #174, opened 2026-03-11
multi: extract darepocli commands, add darepod RunWithContext

sputn1ck:
## Summary

- Extract darepocli commands from `package main` into importable
  `darepoclicommands` sub-package with exported `NewRootCmd()`,
  enabling embedding in the darepotest REPL.
- Add `RunWithContext(ctx)` to `darepod.Server` as a context-based
  alternative to `RunUntilShutdown(interceptor)`, allowing test
  harnesses and the REPL to manage multiple concurrent darepod
  instances without the process-global signal handler.
- Add `/darepocli` binary to `.gitignore`.

## Test plan
- [ ] `go build ./cmd/darepocli/` — existing binary still compiles
- [ ] `go build ./darepod/` — package compiles with new API
- [ ] Existing darepocli commands work identically (thin wrapper)
- [ ] `make unit pkg=internal/repl` — REPL tests pass
- [ ] Manual: `darepotest repl` starts darepod via `RunWithContext`


darepod/rpc_server.go
```diff
@@ -534,11 +536,23 @@ func (r *RPCServer) RefreshVTXOs(ctx context.Context,
 	}, nil
 }
 
+// boardJoinTimeout is how long the Board RPC waits for the operator
+// to acknowledge the JoinRoundRequest via the mailbox before giving
+// up. This must be long enough for the durable egress path and one
+// mailbox round-trip.
+const boardJoinTimeout = 30 * time.Second
```
Roasbeef: I think this should instead be a non-blocking call. Then we'll add a CLI `rounds` command that allows a user to check on pending/requested rounds.

darepod/rpc_server.go
```diff
@@ -552,34 +566,241 @@ func (r *RPCServer) Board(ctx context.Context,
...
+	assemblyCtx, assemblyCancel := context.WithTimeout(
+		ctx, boardJoinTimeout,
+	)
+	defer assemblyCancel()
+
+	assemblyTicker := time.NewTicker(boardPollInterval)
+	defer assemblyTicker.Stop()
+
+	var reachedAssembly bool
+	for !reachedAssembly {
```
Roasbeef: We can simplify this a lot, instead we can just go through the wallet entirely here. Then have it send the message to the round actor as needed re the VTXO requests.

darepod/rpc_server.go
```diff
@@ -552,34 +566,241 @@ func (r *RPCServer) Board(ctx context.Context,
...
+	// trigger-batch will include this client.
+	joinCtx, cancel := context.WithTimeout(ctx, boardJoinTimeout)
+	defer cancel()
+
+	ticker := time.NewTicker(boardPollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-joinCtx.Done():
```
Roasbeef: Same here, we can remove this, then make a `rounds` CLI command that lets the caller list/query the active rounds (and the state of the round), and also watch to be notified when a round concludes or advances.

darepod/server.go
```diff
@@ -1156,6 +1156,17 @@ func (s *Server) registerRoundEventRoutes(
 		)
 	}
 
+	// JoinAck: server accepted the client's join request.
+	addRoundRoute(
+		roundpb.MethodJoinAck,
```
Roasbeef, line 1165: 👍

Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #175, opened 2026-03-12
vtxo+wallet: FSM-based admission gating for VTXO spend and forfeit

ellemouton:
## Summary

Replaces #168 with a fundamentally different approach: instead of an in-memory lock set outside the FSM, **VTXO lifecycle state _is_ the lock**. The manager acts as a single admission gate for all VTXO operations — OOR spend, cooperative forfeit, and completion — using purpose-specific FSM states (`SpendingState`, `PendingForfeitState`) that survive restarts.

**Depends on:** #172 (`vtxo-fsm-refactor`)
**Replaces:** #168 (`vtxo-coin-select`)

## Why This Approach Is Better

PR #168 used an in-memory `lockedVTXOs map[wire.OutPoint]struct{}` that:

1. **Lost locks on restart** — a crash mid-OOR could double-spend a VTXO
2. **Duplicated state** — the lock set existed alongside the FSM, creating two sources of truth
3. **Required manual cleanup** — callers had to remember to unlock on every error path

This PR eliminates all three problems:

```
┌─────────────────────────────────────────────────────────┐
│                    VTXO Manager                         │
│              (single admission gate)                    │
│                                                         │
│  SelectAndReserveSpend    ReserveForfeit                │
│         │                       │                       │
│         ▼                       ▼                       │
│  ┌─────────────┐        ┌────────────────┐             │
│  │ SpendingState│        │PendingForfeit  │             │
│  │  (persisted) │        │State(persisted)│             │
│  └──────┬──────┘        └───────┬────────┘             │
│         │                       │                       │
│    ┌────┴────┐             ┌────┴────┐                  │
│    ▼         ▼             ▼         ▼                  │
│ Complete  Release      Forfeit   Release                │
│  (Spent)  (Live)     Confirmed   (Live)                 │
│                       (Forfeited)                       │
└─────────────────────────────────────────────────────────┘
```

### Admission Flow — OOR Spend

```
Wallet                    Manager                   VTXO Actor FSM
  │                         │                            │
  │ SelectAndReserveSpend   │                            │
  │ (amount, count)         │                            │
  │────────────────────────►│                            │
  │                         │  SpendReserveEvent         │
  │                         │───────────────────────────►│
  │                         │  LiveState → SpendingState │
  │                         │◄───────────────────────────│
  │  SelectedVTXO[]         │                            │
  │◄────────────────────────│                            │
  │                         │                            │
  │   ... OOR protocol ...  │                            │
  │                         │                            │
[...]

PR #176, opened 2026-03-12
multi: implement in-round directed VTXO sends

### LINKED ISSUE #156: darepod: implement in-round directed sends via SendVTXO RPC
In #155, we stubbed out `SendVTXO` with `codes.Unimplemented` because the
previous implementation was incorrect: it dispatched a `RefreshVTXOsRequest`
to the wallet actor, which only supports self-refresh (sending VTXOs back to
the caller's own address to extend expiry). A directed in-round send needs
first-class support in the round actor to include recipient outputs in the
shared transaction.

## Goal

We want a first-class in-round send API in the round actor. Today the round
actor only handles self-refresh (extend expiry by sending back to self). A
directed send where the caller specifies external recipient addresses and
amounts needs to be a native operation in the round participation flow, not
bolted on top of the refresh path.

## What needs to happen

1. **Round actor: add directed send as a first-class operation.** The round
   actor should accept a send request with `(address, amount)` recipient pairs
   and submit them as output requests to the coordinator during round
   participation. This is distinct from refresh, where the only output goes
   back to the sender. The coin selection, input signing, and round
   participation logic should treat directed sends as a peer to refresh, not
   a special case of it.

2. **New message type for the wallet/round actor.** Introduce a dedicated
   `SendInRoundRequest` (or similar) that carries the validated recipient
   list, rather than overloading `RefreshVTXOsRequest` with a recipients
   field. Keeping the two operations separate makes the actor's state machine
   cleaner.

3. **Wire `SendVTXO` RPC to the new actor message.** The RPC handler in
   `rpc_server.go` already validates `Recipients` (address + amount), so the
   plumbing from gRPC to the actor is straightforward once the actor supports
   the operation.

4. **Dry-run support.** The `--dry-run` flag is already wired in the CLI
   (`darepocli send inround`), so the actor should respect a dry-run field
   and return the coin selection result without actually participating in a
   round.

## Context

See `darepod/rpc_server.go` `SendVTXO` handler (currently returns
`codes.Unimplemented`) and the TODO comment there explaining the gap.

`SendOOR` (out-of-round transfers) already works via the OOR service key
pattern and can be used in the meantime for transfers that don't need to
wait for a round.

ellemouton:
## Summary

Replaces #169 
Fixes #156.

This PR implements first-class in-round directed sends through `SendVTXO` on
top of the admission model introduced in #175.

Instead of reusing the out-of-round spend path, directed send is treated as the
cooperative flow it actually is:

- the wallet asks the VTXO manager to atomically select and reserve inputs for
  cooperative use
- the selected VTXOs enter `PendingForfeit`
- the wallet composes the full round intent package, including recipient and
  change outputs
- the round actor registers that intent and proceeds with normal cooperative
  round handling

That makes this the PR 3 replacement for #169, but with the post-PR-1/PR-2
architecture rather than the older send-specific trigger flow.

## Why This Replaces #169

#169 was built before the current wallet/round boundary and before actor-owned
VTXO admission existed. The main problem with reviving it directly was that it
mixed directed send with the old spend-locking path.

This PR replaces that design with the same single-source-of-truth model used by
PR 2:

- OOR send: `Live -> Spending -> Spent`
- In-round directed send: `Live -> PendingForfeit -> Forfeiting -> Forfeited`

So directed send now reuses cooperative admission rather than creating a second
locking path.

## Flow

```text
SendVTXO RPC
  -> validate recipients and total
  -> wallet.SendVTXOsRequest
  -> manager.SelectAndReserveForfeitRequest
  -> selected inputs move Live -> PendingForfeit
  -> wallet builds IntentPackage:
       forfeits + recipient outputs + change output
  -> wallet sends RegisterIntentMsg to round
  -> if round rejects:
       release forfeit reservation and surface any release failure
  -> if round accepts:
       normal cooperative round lifecycle continues
```

## Main Changes

- Add `SelectAndReserveForfeitRequest` so directed send can do atomic
  cooperative coin selection and reservation in one step.
- Add dedicated wallet send messages instead of overloading refresh.
- Implement `handleSendVTXOs` with strict cleanup on every post-reservation
  failure path.
- Implement `SendVTXO` in the daemon RPC server, including recipient
  resolution for taproot addresses and x-only pubkeys.
- Return send result metadata including selected input count and explicit
  change amount.
- Persist directed-send recipient VTXOs using the resolved recipient
  `ClientKey`, not the sender signing key.
- Add unit, integration, and systest coverage for the new flow.

## Dependency

This PR depends on #175 (`vtxo-spend-state`).

It is opened against `vtxo-spend-state` because PR 3 relies on PR 2's actor-
owned admission model and manager-side cooperative select-and-reserve API.

## Test Plan

- `go test ./darepod ./wallet ./vtxo ./round ./oor`
- `go test -tags=systest ./systest -run TestSendVTXOEndToEnd -count=1`

### DIFF
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
@@ -622,9 +623,10 @@ func (r *RPCServer) Board(ctx context.Context,
 	}, nil
 }
 
-// SendVTXO initiates an in-round transfer by submitting a refresh
-// request with specific recipient outputs to the round coordinator.
-// The transfer completes asynchronously when the next round commits.
+// SendVTXO initiates an in-round directed transfer by forfeiting
+// existing VTXOs and creating new recipient VTXOs in the same round.
+// Coin selection, reservation, and round registration are handled
+// atomically by the wallet actor.
 func (r *RPCServer) SendVTXO(ctx context.Context,
 	req *daemonrpc.SendVTXORequest) (
 	*daemonrpc.SendVTXOResponse, error) {
@@ -638,43 +640,103 @@ func (r *RPCServer) SendVTXO(ctx context.Context,
 			"at least one recipient is required")
 	}
 
-	// Validate recipients and compute total amount.
+	// Resolve each recipient's pkScript and client pubkey from
+	// the proto Output destination.
+	recipients := make(
+		[]wallet.SendRecipient, 0, len(req.Recipients),
+	)
 	var totalAmount int64
+
 	for i, out := range req.Recipients {
 		if out.GetDestination() == nil {
 			return nil, status.Errorf(
 				codes.InvalidArgument,
 				"recipient %d: destination is "+
-					"required", i)
+					"required", i,
+			)
 		}
 
 		if out.AmountSat <= 0 {
 			return nil, status.Errorf(
 				codes.InvalidArgument,
 				"recipient %d: amount must be "+
-					"positive", i)
+					"positive", i,
+			)
 		}
+
+		pkScript, clientKey, err := r.resolveRecipientOutput(
+			out,
+		)
+		if err != nil {
+			return nil, status.Errorf(
+				codes.InvalidArgument,
+				"recipient %d: %v", i, err,
+			)
+		}
+
+		recipients = append(recipients, wallet.SendRecipient{
+			PkScript:  pkScript,
+			Amount:    btcutil.Amount(out.AmountSat),
+			ClientKey: clientKey,
+		})
 
 		totalAmount += out.AmountSat
 	}
 
-	// For dry_run, validate inputs and return a preview.
-	if req.DryRun {
-		return &daemonrpc.SendVTXOResponse{
-			Status:         "preview",
-			TotalAmountSat: totalAmount,
-		}, nil
+	// Fetch operator terms for fee, dust limit, exit delay, and
+	// operator key.
+	terms, err := r.server.fetchOperatorTerms(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal,
+			"unable to fetch operator terms: %v", err)
 	}
 
-	// TODO(roasbeef): In-round directed sends are not yet
-	// implemented. The wallet actor's RefreshVTXOsRequest only
-	// supports self-refresh (sending back to self), not directed
-	// transfers to external recipients. Once the round protocol
-	// supports recipient outputs, this handler should build a
-	// proper send request with the validated recipients.
-	return nil, status.Errorf(codes.Unimplemented,
-		"in-round directed sends are not yet implemented; "+
-			"use SendOOR for out-of-round transfers")
+	if !r.server.walletRef.IsSome() {
+		return nil, status.Errorf(codes.Internal,
+			"wallet actor not initialized")
+	}
+
+	wRef := r.server.walletRef.UnsafeFromSome()
+
+	sendReq := &wallet.SendVTXOsRequest{
+		Recipients:    recipients,
+		OperatorFee:   terms.MinOperatorFee,
+		DustLimit:     terms.DustLimit,
+		OperatorKey:   terms.PubKey,
+		VTXOExitDelay: terms.VTXOExitDelay,
+		DryRun:        req.DryRun,
+	}
+
+	future := wRef.Ask(ctx, sendReq)
+	result := future.Await(ctx)
+
+	resp, err := result.Unpack()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal,
+			"send failed: %v", err)
+	}
+
+	sendResp, ok := resp.(*wallet.SendVTXOsResponse)
+	if !ok {
+		return nil, status.Errorf(codes.Internal,
+			"unexpected response type: %T", resp)
+	}
+
+	log.InfoS(ctx, "SendVTXO completed",
+		slog.String("status", sendResp.Status),
+		slog.Int("selected_count",
+			sendResp.SelectedCount),
+		slog.Int64("total_selected",
+			int64(sendResp.TotalSelected)),
+		slog.Int64("change",
+			int64(sendResp.ChangeAmount)))
+
+	return &daemonrpc.SendVTXOResponse{
+		Status:          sendResp.Status,
+		TotalAmountSat:  totalAmount,
+		ChangeAmountSat: int64(sendResp.ChangeAmount),
+		SelectedCount:   int32(sendResp.SelectedCount),
+	}, nil
 }
 
 // SendOOR initiates an out-of-round transfer directly between the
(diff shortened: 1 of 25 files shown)


ellemouton:
@claude review this

lib/actormsg/vtxo_admission.go
```diff
@@ -167,3 +167,51 @@ type ReleaseForfeitResponse struct {
 
 // VTXOManagerResp implements the VTXOManagerResp marker interface.
 func (r *ReleaseForfeitResponse) VTXOManagerResp() {}
+
+// =============================================================================
+// Atomic cooperative select-and-reserve: wallet → Manager → VTXO actors
+// =============================================================================
+//
+// SelectAndReserveForfeitRequest combines coin selection with cooperative
```
Roasbeef, line 175: double comment here?

wallet/messages.go
```diff
@@ -515,3 +515,88 @@ func (m *LeaveVTXOsResponse) MessageType() string {
...
+// SendVTXOsRequest asks the wallet to execute an in-round directed
+// send. The wallet atomically selects and reserves VTXOs for
+// cooperative consumption, builds the IntentPackage (forfeits +
+// recipient VTXOs + change), and registers it with the round actor.
+type SendVTXOsRequest struct {
+	actor.BaseMessage
+
+	// Recipients is the list of send destinations with resolved
+	// pkScripts and amounts.
+	Recipients []SendRecipient
```
Roasbeef, line 616: Does this support change? So we'll add ourselves to the recipient if the input amount is greater than the output

wallet/messages.go
```diff
@@ -515,3 +515,88 @@ func (m *LeaveVTXOsResponse) MessageType() string {
...
+
+// walletMsgSealed implements the sealed WalletMsg interface.
+func (m *SendVTXOsRequest) walletMsgSealed() {}
+
+// SendVTXOsResponse contains the result of a directed send request.
+type SendVTXOsResponse struct {
+	actor.BaseMessage
+
+	// Status is "submitted" for real sends or "preview" for dry-run.
+	Status string
```
Roasbeef, line 652: Should this return a round ID or something else that can be used to query the state of the ongoing send?

systest/send_vtxo_test.go
```diff
@@ -0,0 +1,785 @@
...
+
+	return &mailboxpb.SendResponse{
+		Status: &mailboxpb.Status{Ok: true},
+	}, nil
+}
+
+// Pull returns queued envelopes for a mailbox starting at the requested cursor.
+// When the mailbox is empty it waits up to the requested timeout so the
+// daemon's ingress loop behaves like it would against a real long-poll edge.
+func (s *fakeMailboxServer) Pull(ctx context.Context,
```
Roasbeef, line 127: There should be an in-memory mailbox, I think, in the mailbox package.

Roasbeef: If not, then we can also just move this to another file instead of having it within the SendVTXOTest.go file.

darepod/rpc_server.go
```diff
@@ -870,6 +932,98 @@ func (r *RPCServer) unlockVTXOs(ctx context.Context,
...
+		addr, err := btcutil.NewAddressTaproot(
+			d.Pubkey, r.server.chainParams,
+		)
+		if err != nil {
+			return nil, nil, fmt.Errorf(
+				"derive taproot address: %w", err,
+			)
+		}
+
+		pkScript, err := txscript.PayToAddrScript(addr)
```
Roasbeef, line 971: Should we actually be making an actual script here? In that we only really want a pubkey from them, as that'll be used as the pubkey in the newly created VTXO.

wallet/wallet.go
```diff
@@ -1282,6 +1287,273 @@ func (a *Ark) handleCompleteSpendVTXOs(ctx context.Context,
...
+		if keyErr != nil {
+			return nil, fmt.Errorf(
+				"derive signing key for recipient "+
+					"%d: %w", i, keyErr,
+			)
+		}
+
+		vtxoRequests = append(vtxoRequests, types.VTXORequest{
+			Amount:      r.Amount,
+			PkScript:    r.PkScript,
```
Roasbeef: Is this actually used? Given we haven't finished the AST feature yet (not merged in).

darepod/rpc_server.go
```diff
@@ -638,43 +640,103 @@ func (r *RPCServer) SendVTXO(ctx context.Context,
...
+	}
+
+	wRef := r.server.walletRef.UnsafeFromSome()
+
+	sendReq := &wallet.SendVTXOsRequest{
+		Recipients:    recipients,
+		OperatorFee:   terms.MinOperatorFee,
+		DustLimit:     terms.DustLimit,
+		OperatorKey:   terms.PubKey,
+		VTXOExitDelay: terms.VTXOExitDelay,
```
Roasbeef, line 706: Since I had the scope of this PR, we should also have the client validate that the exit delay is actually saying otherwise if it's zero, then that means that there's no actual safety for the user.

Roasbeef (approved):
LGTM 🫐


Solid PR, no direct blocking comments, see some of the inline comments.

Roasbeef:
One other thing that we should consider here is the receiver side. For example, for OOR, we actually have the indexer and then the receiver can hit the indexer to basically know when something is sent to it and to get all the checkpoints, etc. Here we don't actually have anything like that. How is the receiver supposed to know that they have a new Inron VTXO?

I think we should make a follow-up here to put the proper plumbing to make sure that the receiver can also know whether there's a new and run VTXO because, for example, the server would also do a similar push notification as it did for O or R to the receiver so they can actually know that they have a new VTXO to get all the information and it's imported and recognized as its own.

wallet/messages.go
ellemouton: yes - but the change is handled internally in the wallet

ellemouton:
thanks for review 🙏 

working on addressing things locally. realised i think we first need to fix this bug: https://github.com/lightninglabs/darepo-client/pull/210

Roasbeef:
Rebased on top of master after the recent changes.

ellemouton:
## Follow-up items (separate PR)

The following changes are planned for a follow-up PR on top of this one:

### Hardening
- Overflow protection: validate individual recipient amounts against 21M BTC cap, overflow-safe addition
- Deferred cleanup: protect VTXO reservation with `defer` + `committed` flag so panics don't leak `PendingForfeitState`
- `context.WithoutCancel` for cleanup operations so forfeit release survives client disconnection
- Cap recipients at 256 in RPC validation
- Explicit nil destination check before type switch

### Ownership model (replace `IsOwner` with data-driven ownership)
- Add `OwnedScriptChecker` interface to round FSM's `ClientEnvironment`
- Replace `IsOwner` flag with `OwnedScriptChecker.IsOwnedScript(pkScript)` — ownership determined by checking the `owned_receive_scripts` DB store
- Remove `IsOwner` from `VTXOIntent` and `VTXORequest`
- Add `OwnedScriptRegistrar` interface on round actor — registers pkScripts for boarding, refresh, and change VTXOs at intent time
- Wire `ownedScriptCheckerAdapter` and `ownedScriptRegistrarAdapter` into daemon using `OORArtifactPersistenceStore`
- Self-send works correctly (both recipient + change VTXOs persisted)

### Server-side VTXO event publishing
- `VTXOEventPublisher` interface on server rounds actor
- After round confirmation, publish `VTXO_CREATED` indexer events for each tree leaf
- Wire adapter using `indexer.Operator.PublishVTXOEvent`

### Recipient VTXO materialization (bob sees his VTXO)
- Client-side `IncomingVTXOEvent` handler — currently no envelope route exists for `MethodIncomingVTXO`
- Handler needs: indexer query for full VTXO details, owner key lookup from `owned_receive_scripts`, persist + notify VTXO manager
- Server-side publish works (confirmed via testing), client-side handler is the remaining work

### Other
- Rename `NewOORReceiveScript` → `NewReceiveScript` (proto rename, needs `make rpc`)
- Send tracking (`send_id` concept for correlating sends with round lifecycle)
- Server itest: `TestDirectedSendSelfSend` (alice → alice, both VTXOs persisted)

Companion server itest PR: lightninglabs/darepo#205

PR #178, opened 2026-03-13
multi: add Dockerfile and TxProof construction for boarding UTXOs

Roasbeef:
In this PR, we add a Dockerfile for the darepod/darepocli binaries and
wire up client-side TxProof (SPV merkle inclusion proof) construction for
boarding UTXOs.

## Dockerfile

We add a multi-stage alpine build that produces `darepod` and `darepocli`
binaries with `CGO_ENABLED=0`. A Docker skill file documents how to build
and run the container as part of the e2e testing stack (see the companion
server PR for the full docker-compose setup).

## TxProof Construction

When a boarding UTXO confirms on-chain, the wallet now fetches the full
block via a new `GetBlock` method on `BoardingBackend`, builds a
`TxMerkleProof` proving the transaction's inclusion in the block's merkle
tree, and serializes the proof as `RawTxProof` bytes. The round actor
passes these bytes through to the server in the `tx_proof` field of the
`BoardingRequest` proto.

We add `GetBlock` to both the lnd-backed and esplora-backed boarding
backends. The lnd backend uses `ChainKit.GetBlock`, while the esplora
backend fetches the raw block hex via the REST API.

The TLV serialization codec for `TxProof` is duplicated into
`lib/types/proof_codec.go` (from `db/proof_codec.go`) to break an import
cycle: the round package needs to serialize proofs but can't import `db`.

This pairs with the server-side validation in lightninglabs/darepo#157.


Roasbeef:
@claude review this

PR #179, opened 2026-03-13
oor: real owner leaf signing and post-commit deferral

sputn1ck:
## Summary

- Add `OwnerCheckSigLeaf` helper that builds a real `<schnorr_pubkey> OP_CHECKSIG` leaf for checkpoint taproot outputs, replacing the OP_TRUE placeholder.
- Implement `SignArkPSBT` which signs Ark transaction inputs using the client key on the owner leaf path. Update `SigningOutboxHandler` to call it.
- Add `PostCommitHandler` interface to the durable actor framework, enabling behaviors to defer side-effects (like Telling sibling actors) until after the DB transaction commits, preventing SQLite write-lock deadlocks.
- Defer transport events (submit, finalize, ack) and DB-writing local events (`MarkInputsSpentRequest`) to `FlushPostCommit` in the OOR client actor. Feed optimistic follow-up events into the FSM within the transaction for atomic state transitions.
- Use background context for `Edge.Send` in serverconn to avoid inheriting closed transaction contexts.
- Add pubkey destination support for OOR sends, deriving VTXO-compatible P2TR outputs from recipient schnorr keys.
- Switch `ToProto` from custom TLV-in-Any envelopes to concrete `oorpb` proto types.

## Test plan

- [x] `go test ./oor/...` passes
- [x] `go test ./baselib/actor/...` passes
- [x] All outbox handler tests updated to use real `SignArkPSBT`
- [x] Outbox message serialization tests updated for oorpb types
- [ ] Integration test via `make arktest case=TestScenarios/oor_send` (requires server-side PR)

### DIFF
--- a/baselib/actor/durable_actor.go
+++ b/baselib/actor/durable_actor.go
@@ -412,8 +412,24 @@ func (a *DurableActor[M, R]) processInTransaction(
 		return
 	}
 
-	// Transaction committed -- now it is safe to complete the
-	// in-memory promise so the caller observes the result.
+	// Transaction committed — flush any post-commit work the
+	// behavior buffered during Receive. This is needed when the
+	// behavior must Tell another durable actor backed by the same
+	// SQLite database: doing so inside the transaction would
+	// deadlock on the write lock.
+	if pch, ok := a.behavior.(PostCommitHandler); ok {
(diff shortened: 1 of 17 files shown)


Roasbeef:
Worked into https://github.com/lightninglabs/darepo-client/pull/180

litbot-9000:

PR #180, opened 2026-03-14
multi: fix OOR checkpoint collab leaf, real signing, and same-DB tx atomicity

### LINKED ISSUE #137: oor: migrate OOR client actor from synchronous OutboxHandler to async OutboxPublisher
## Summary

Migrate the OOR client actor's outbox handling from synchronous inline execution (`driveOutbox()`) to the async `OutboxPublisher` CDC pattern. This eliminates cross-actor transaction leaks, makes FSM state + outbox writes atomic, and provides crash-safe delivery.

## Motivation

Today `driveOutbox()` in `oor/actor.go` synchronously calls `handler.Handle(ctx, sessionID, msg)` inside the actor's database transaction. When the handler performs cross-actor calls (e.g., `server.Receive()` → `ref.Ask()` → `DurableMailbox.Send()` → `EnqueueMessage()`), the receiver's `TransactionExecutor.ExecTx()` finds the sender's `*sql.Tx` in the context and either:
- **Writes to the wrong database** (client and server have separate DBs), or
- **Deadlocks** (same DB — Ask blocks waiting for response while holding the tx)

`DurableMailbox.Send()` now strips the tx via `WithoutTx()` as a defense-in-depth fix (see #136), but the proper architectural solution is the OutboxPublisher pattern: side effects run outside the actor's transaction entirely.

## Target Architecture

Instead of executing side effects inline, `driveOutbox()` writes outbox events to the outbox table within the same transaction as the FSM state change. The `OutboxPublisher` (already implemented in `baselib/actor/outbox_publisher.go`) polls the outbox table and delivers messages to handler actors with a fresh context.

<details>
<summary>Current Flow (Synchronous)</summary>

```
Actor.Receive(msg)
  ↓
TxAwareDeliveryStore.ExecTx(ctx) — opens DB transaction, attaches to ctx
  ↓
oorDurableBehavior.Receive(txCtx, msg)
  ↓
handleStartTransfer / handleDriveEvent
  ↓
askEvent(txCtx, fsm, event) — drives FSM, returns []OutboxEvent
  ↓
driveOutbox(txCtx, sessionID, fsm, outbox)
  ↓
handler.Handle(txCtx, sessionID, outboxEvent) — BLOCKS on RPC/signing
  ↓
follow-up events → askEvent → more outbox → recursive driveOutbox
  ↓
persistCheckpoint(txCtx) — writes checkpoint in same tx
  ↓
ExecTx commits — all state + checkpoint atomic
```

</details>

<details>
<summary>Target Flow (Async via OutboxPublisher)</summary>

```
Actor.Receive(msg)
  ↓
TxAwareDeliveryStore.ExecTx(ctx) — opens DB transaction
  ↓
oorDurableBehavior.Receive(txCtx, msg)
  ↓
handleStartTransfer / handleDriveEvent
  ↓
askEvent(txCtx, fsm, event) — drives FSM, returns []OutboxEvent
  ↓
enqueueOutboxEvents(txCtx, sessionID, outbox)
  ↓
DeliveryStore.EnqueueOutbox(txCtx, params) — writes to outbox table
  ↓
persistCheckpoint(txCtx) — writes checkpoint in same tx
  ↓
ExecTx commits — FSM state + outbox entries + checkpoint are atomic
  ↓
(async) OutboxPublisher polls outbox table with fresh context
  ↓
OutboxPublisher decodes message, looks up target actor
  ↓
target actor.Tell(freshCtx, decodedMsg) — handler actor receives
  ↓
handler actor executes side effect (RPC, signing)
  ↓
handler actor calls client OOR actor with follow-up event via Tell/Ask
```

</details>

## Benefits

1. **Atomic state + outbox**: FSM state and outbox entries commit or roll back together. No window where state advances but outbox is lost.
2. **No cross-actor transaction leaks**: The publisher uses its own context, so the receiver's `EnqueueMessage` never sees the sender's `*sql.Tx`.
3. **Crash-safe delivery**: If the process crashes after the commit, the publisher replays pending outbox entries on restart.
4. **Idempotent receiver-side dedup**: The publisher injects the outbox row ID via `WithOutboxID(ctx, id)` so duplicate deliveries are silent no-ops via the `ON CONFLICT (id) DO NOTHING` clause.

## What Is the Outbox Handler

<details>
<summary>Outbox handler overview</summary>

The **outbox handler** is the I/O boundary of the OOR transfer FSM. The FSM is intentionally pure: it consumes `Event` inputs and emits `OutboxEvent` outputs. Each `OutboxEvent` represents a side-effect request that must be executed outside the FSM:

- **RequestArkSignatures** — ask the wallet/signing layer to sign the Ark PSBT
- **SendSubmitPackageRequest** — send the submit package to the server via RPC
- **RequestCheckpointSignatures** — sign co-signed checkpoint PSBTs
- **SendFinalizePackageRequest** — send finalized checkpoints to server via RPC
- **MarkInputsSpentRequest** — update local VTXO store (persistence)
- **MaterializeIncomingVTXOsRequest** — persist incoming VTXOs
- **SendIncomingAckRequest** — ack incoming transfer to server
- **ScheduleRetryRequest** — schedule a delayed retry timer

The `OutboxHandler` interface is defined in `oor/actor.go`:

```go
type OutboxHandler interface {
    Handle(ctx context.Context, sessionID SessionID,
        outbox OutboxEvent) ([]Event, error)
}
```

Handlers chain via the `Next` field. For example, `LocalPersistenceOutboxHandler` (in `oor/local_persistence_handler.go`) handles persistence events and delegates transport/signing to `Next`.

**Why it matters**: Keeping side effects behind this interface makes the FSM deterministic and testable. State transitions produce outbox events; the handler executes them and returns follow-up events that drive the FSM forward.

**How `driveOutbox()` works today** (in `oor/actor.go:667-721`): For each outbox event, it synchronously calls `handler.Handle()`, collects follow-up events, feeds each follow-up into the FSM via `askEvent()`, persists checkpoints, and recursively processes any newly emitted outbox events. The entire chain — FSM transition, handler execution, checkpoint persistence — runs inside the actor's database transaction because the context carries the `*sql.Tx` from `TxAwareActorDeliveryStore.ExecTx`.

</details>

## Milestones

<details>
<summary>Milestone 1: TLV Codec Registration for OutboxEvent Types</summary>

The OutboxPublisher uses a `MessageCodec` to decode outbox payloads. Each `OutboxEvent` type that will flow through the publisher needs to be registered with the codec. Many outbox event types already embed `actor.BaseMessage` and implement `TLVMessage` (or can be made to). For types that carry complex payloads (PSBTs, outpoints), we use the existing `ToProto()` method plus a `FromProto()` decoder.

Steps:

1. In `oor/outbox_messages.go`, ensure each OutboxEvent type that needs async delivery has:
   - A stable `MessageType() string` method (can reuse `outboxType()`)
   - TLV encode/decode via the existing proto envelope pattern

2. In `oor/actor.go`, register all outbox event types with the `MessageCodec` used by the OutboxPublisher

3. Verify round-trip encode/decode with a unit test for each type

</details>

<details>
<summary>Milestone 2: Rewrite driveOutbox() to Enqueue Instead of Execute</summary>

Replace the synchronous `driveOutbox()` with `enqueueOutboxEvents()` that writes outbox events to the delivery store's outbox table.

Key design choice: `outboxTargetActorID(msg)` determines which handler actor receives each outbox event type:

- Transport events (Submit, Finalize, IncomingAck) → transport handler actor
- Signing events (RequestArk, RequestCheckpoint) → signing handler actor
- Persistence events (MarkInputsSpent, MaterializeIncoming) → persistence handler actor
- Timer events (ScheduleRetry) → timer/scheduler actor

Update all call sites of `driveOutbox()`:
- `handleStartTransfer()` line ~310
- `handleDriveEvent()` line ~370
- `handleRestart()` / `resumeRestoredSessions()`

</details>

<details>
<summary>Milestone 3: Wire OutboxPublisher into OORClientActor</summary>

Modify `NewOORClientActor` to create and start an `OutboxPublisher`. The `System` (`SystemContext`) provides `Lookup(ServiceKey)` so the publisher can find the target handler actor by its service key / actor ID.

Add `OutboxPublisher` stop to `OORClientActor.Stop()`.

</details>

<details>
<summary>Milestone 4: Create Handler Actors for Side Effects</summary>

Create `oor/outbox_handler_actor.go` with a lightweight actor that wraps the existing `OutboxHandler` interface. This preserves the existing handler chain pattern (LocalPersistenceOutboxHandler → TransportHandler) but runs it outside the client actor's transaction.

The follow-up events are delivered back to the OOR client actor as `DriveEventRequest` messages via `Tell()`. This means the FSM advances asynchronously — the handler actor does not wait for the FSM response.

</details>

<details>
<summary>Milestone 5: Handle the Async FSM Continuation</summary>

With async delivery, follow-ups arrive as separate `DriveEventRequest` messages in the client actor's mailbox. The existing `handleDriveEvent()` already handles this pattern. The only change needed is that `handleDriveEvent()` now calls `enqueueOutboxEvents()` instead of `driveOutbox()`, continuing the async chain.

This creates a natural message-driven loop — each step is independently crash-safe.

</details>

<details>
<summary>Milestone 6: Session ID Propagation</summary>

Use `sessionID.String()` as the `DomainKey` in `OutboxParams`. The handler actor extracts it from the outbox message metadata. This avoids modifying every OutboxEvent type.

</details>

<details>
<summary>Milestone 7: Update E2E Tests</summary>

The E2E tests in `darepo/oor/e2e_client_server_test.go` need to create handler actors that wrap the existing test adaptors, wire the OutboxPublisher into the test's OORClientActor, and wait for the async FSM progression to complete.

</details>

<details>
<summary>Milestone 8: Restart/Resume Compatibility</summary>

`resumeRestoredSessions()` enqueues outbox events to the outbox table instead of executing them synchronously. The OutboxPublisher picks them up and delivers to handler actors. This is actually simpler than the current flow.

</details>

## Outbox Event → Follow-up Event Mapping

| Outbox Event | Follow-up Event |
|---|---|
| `RequestArkSignatures` | `ArkSignedEvent` |
| `SendSubmitPackageRequest` | `SubmitAcceptedEvent` |
| `RequestCheckpointSignatures` | `CheckpointsSignedEvent` |
| `SendFinalizePackageRequest` | `FinalizeAcceptedEvent` |
| `MarkInputsSpentRequest` | `InputsMarkedSpentEvent` |
| `MaterializeIncomingVTXOsRequest` | `IncomingHandledEvent` |
| `SendIncomingAckRequest` | `IncomingAckSentEvent` |
| `ScheduleRetryRequest` | `RetryDueEvent` (after delay) |

## Key Files

- `baselib/actor/outbox_publisher.go` — Existing OutboxPublisher (already implemented and tested)
- `baselib/actor/delivery_store.go` — `DeliveryStore` interface with outbox methods
- `oor/actor.go` — `OORClientActor`, `oorDurableBehavior`, `driveOutbox()`
- `oor/outbox_messages.go` — Concrete `OutboxEvent` types with TLV encoding
- `oor/local_persistence_handler.go` — Persistence OutboxHandler chain

## Related

- #136 — `DurableMailbox.Send()` strips sender tx via `WithoutTx()` (defense-in-depth fix)

Full ExecPlan available at `PLAN-outbox-publisher-migration.md` in the repo root.

### LINKED ISSUE #141: actor: re-examine whether WithoutTx in DurableMailbox.Send is needed for same-DB actors
## Context

PR #136 (`actor: strip sender tx in DurableMailbox.Send`) added `WithoutTx()` to unconditionally strip the sender's database transaction in `DurableMailbox.Send` before calling `EnqueueMessage`. The stated reasons were:

1. **Cross-DB actors**: the message would be written to the sender's DB, invisible to the receiver
2. **Same-DB Ask calls**: the sender blocks waiting for a response while the receiver cannot see the uncommitted message (deadlock)

However, the original test failures that motivated this fix may have been caused by a **test setup bug on the server side**, not by a fundamental architectural flaw. Server-side PR [darepo#113](https://github.com/lightninglabs/darepo/pull/113) (merged around the same time) reveals that the OOR e2e tests were using the **same database** for both client and server delivery stores — which doesn't match production where client and server have separate databases.

## Evidence the original failures may have been a test bug

In `darepo#113`, the `oor/e2e_client_server_test.go` was fixed to use **separate databases** for client and server:

```diff
+	// The client uses its own database for the delivery store since
+	// in production the client and server have separate databases.
+	clientSQLStore := db.NewTestDB(t)
 	clientDeliveryStore, err := db.NewActorDeliveryStoreFromDB(
-		sqlStore, clock.NewDefaultClock(), btclog.Disabled,
+		clientSQLStore, clock.NewDefaultClock(), btclog.Disabled,
 	)
```

Additionally, `darepo#113` removed the `deliveryStoreShim` that was intentionally preventing `TxAwareDeliveryStore` on the server side. The shim existed because "server OOR currently relies on non-transactional processing because the rest of the OOR persistence path is not yet transaction-context-aware." Once `TransactionExecutor.ExecTx` was made context-aware, the shim was no longer needed.

### 1. Is `WithoutTx` necessary for same-DB durable actors?

When both the sender and receiver are durable actors **on the same database** (which is the common case for client-side actors like round + wallet), `WithoutTx` prevents a potentially valuable optimization: **atomic cross-actor Tell within the sender's transaction**.

Without `WithoutTx`, a `Tell` from Actor A to Actor B (same DB) inside A's `ExecTx` would write the message to B's inbox **in the same transaction** as A's state change. This means:
- If the tx commits: both A's state change AND the message to B are visible atomically
- If the tx rolls back: neither A's state change nor the message exist
- No need for the outbox pattern at all for same-DB Tell

### 2. The deadlock concern for Ask — is it real?

The PR description claims same-DB Ask deadlocks because "the sender blocks waiting for a response while the receiver cannot see the uncommitted message." But:
- For `DurableAsk`, the sender doesn't block in-process — the response comes back via the outbox/publisher asynchronously
- For in-memory `Ask`, the sender blocks, but the receiver reads via `LeaseNextMessage` which runs its own query — **SQLite WAL mode allows concurrent readers to see committed data while writers hold a transaction**

We should verify whether SQLite WAL mode actually prevents this deadlock scenario, or whether it's a real concern.

### 3. Cross-DB actors are the only hard case

The cross-DB case (reason #1) is genuinely unsolvable with shared transactions — you can't share a `*sql.Tx` across databases. But this case is already handled at the architecture level: client-to-server communication goes over gRPC, which naturally strips context values.

For **intra-process, same-DB** actors (the client daemon's round actor talking to its wallet actor), `WithoutTx` may be throwing away atomicity that we could use to simplify the system (potentially eliminating the need for the `OutboxPublisher` in some paths).

## Proposed investigation

1. Write a test with two same-DB durable actors where Actor A does `Tell` to Actor B inside its `ExecTx`, **without** `WithoutTx`, and verify the message is visible to B after A's tx commits
2. Write a test for the Ask deadlock scenario to determine if it actually deadlocks with SQLite WAL mode
3. If same-DB Tell works without `WithoutTx`, consider making the stripping conditional:
   - Strip when crossing DB boundaries (detected via some marker)
   - Preserve when same-DB, enabling atomic cross-actor messaging
4. Evaluate whether the `OutboxPublisher` can be eliminated for same-DB `DurableAsk` if the response can be written directly to the callback actor's inbox within the responder's transaction

## Related

- Client PR #136: `actor: strip sender tx in DurableMailbox.Send`
- Client PR #125: `db: make TransactionExecutor join actor's outer transaction`
- Server PR darepo#113: `db: make TransactionExecutor join actor's outer transaction` + shim removal + test fix

Roasbeef:
This PR folds in and corrects several foundational pieces for the OOR
transfer flow, fixing a protocol-level security issue in the checkpoint
collab leaf construction, adding real Ark PSBT signing, and resolving
the same-DB durable actor deadlock at its root.

## Checkpoint collab leaf: 2-of-2, not single-sig

The checkpoint output's collaborative spend path was using a single-sig
`<owner_key> OP_CHECKSIG` leaf (via `OwnerCheckSigLeaf`). Per the
protocol spec (Technical Explainer: OOR Ark and Checkpoint Transactions),
this path **must** be a 2-of-2 multisig between owner and operator:

```
<owner_key> OP_CHECKSIGVERIFY
<operator_key> OP_CHECKSIG
```

Using single-sig would allow the client to create alternative Ark txs
spending the checkpoint output without operator knowledge, breaking the
double-spend protection that checkpoints provide. The fix replaces
`OwnerCheckSigLeaf` with the existing `MultiSigCollabTapLeaf` helper
which already implements the correct 2-of-2 script.

This also required changing the FSM's submit validation from
`ValidateSubmitPackageSigned` (full script VM execution) to
`ValidateSubmitPackage` (structural only), since at submit time only the
client's half of the 2-of-2 signature is present. The operator adds
their half during the submit phase.

## Real Ark PSBT signing

The `SignArkPSBT` function is added to sign Ark tx inputs using the
client key on the checkpoint output's collab leaf. The signing outbox
handler now calls `SignArkPSBT` instead of passing the Ark PSBT through
unsigned.

Fixes from code review:
- Pass full `KeyDescriptor` (including `KeyLocator`) so hardware
  signers can locate the private key
- Search `TaprootLeafScript` entries by matching the expected leaf
  script hash instead of assuming index 0

## Same-DB tx atomicity (WithoutTx removal)

`DurableMailbox.Send()` was calling `WithoutTx(ctx)` to strip the
sender's database transaction before enqueueing messages. On same-DB
actors (all client-side actors share one SQLite DB), this forced
`EnqueueMessage` to open a **new** write transaction while the sender's
write tx was still active. SQLite WAL only allows one writer, causing a
deadlock.

The `WithoutTx` was added as a blanket safety measure for cross-DB
actors, but the root cause was a test setup bug where client and server
shared one DB (fixed in darepo#113). With `TransactionExecutor.ExecTx`
already context-aware (joins the outer tx when one exists), removing
`WithoutTx` gives us atomic cross-actor Tell for free: the enqueue is
committed in the same transaction as the sender's state change.

## Other changes

- `serverconn`: use `context.WithoutCancel` for `Edge.Send` gRPC calls
  to prevent parent cancellation from aborting outbound sends, with a
  `defaultSendEventTimeout` constant
- `darepod`: add `Output_Pubkey` destination for OOR sends, deriving a
  VTXO-compatible P2TR output from a recipient's schnorr key
- `oorpb`: switch `ToProto` from `anypb.Any` envelopes to concrete proto
  types matching the server-side dispatcher

Several commits are adapted from #179, with the critical checkpoint leaf
fix and review feedback addressed.

Fixes #137
Fixes #141

### DIFF
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/txscript"
@@ -695,15 +696,15 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 	}
 
 	// Resolve the recipient's pkScript from the destination
-	// oneof.
-	pkScript, err := r.resolveOutputPkScript(
-		req.Recipient,
-	)
+	// oneof. Pubkey destinations need operator terms to derive
+	// VTXO-compatible taproot outputs, so we pass a lazy fetcher.
+	pkScript, err := r.resolveOutputPkScript(ctx, req.Recipient)
 	if err != nil {
 		return nil, err
 	}
 
-	// For dry_run, validate inputs and return a preview.
+	// For dry_run, validate inputs and return a preview without
+	// contacting the operator.
 	if req.DryRun {
 		return &daemonrpc.SendOORResponse{
 			Status: "preview",
@@ -868,8 +869,9 @@ func (r *RPCServer) unlockVTXOs(ctx context.Context,
 
 // resolveOutputPkScript derives a pkScript from the Output's
 // destination oneof. It supports address, raw pubkey, and raw
-// pkScript destinations.
-func (r *RPCServer) resolveOutputPkScript(
+// pkScript destinations. For pubkey destinations, operator terms
+// are fetched to derive a VTXO-compatible taproot output.
+func (r *RPCServer) resolveOutputPkScript(ctx context.Context,
 	out *daemonrpc.Output) ([]byte, error) {
 
 	switch d := out.Destination.(type) {
@@ -894,6 +896,51 @@ func (r *RPCServer) resolveOutputPkScript(
 		}
 
 		return pkScript, nil
+
+	case *daemonrpc.Output_Pubkey:
+		if len(d.Pubkey) == 0 {
+			return nil, status.Errorf(
+				codes.InvalidArgument,
+				"pubkey is empty",
+			)
+		}
+
+		recipientKey, err := schnorr.ParsePubKey(d.Pubkey)
+		if err != nil {
+			return nil, status.Errorf(
+				codes.InvalidArgument,
+				"invalid pubkey: %v", err,
+			)
+		}
+
+		// For OOR sends, a pubkey destination creates a
+		// VTXO-compatible taproot output with the operator's
+		// collab key and exit delay (not a simple BIP-86
+		// key-path-only output). This ensures the recipient
+		// gets a standard Ark VTXO they can spend
+		// collaboratively or exit unilaterally.
+		terms, err := r.server.fetchOperatorTerms(ctx)
+		if err != nil {
+			return nil, status.Errorf(
+				codes.Internal,
+				"fetch operator terms for pubkey "+
+					"destination: %v", err,
+			)
+		}
+
+		tapKey, err := scripts.VTXOTapKey(
+			recipientKey, terms.PubKey,
+			terms.VTXOExitDelay,
+		)
+		if err != nil {
+			return nil, status.Errorf(
+				codes.Internal,
+				"unable to derive VTXO tap key: %v",
+				err,
+			)
+		}
+
+		return txscript.PayToTaprootScript(tapKey)
 
 	case *daemonrpc.Output_PkScript:
 		if len(d.PkScript) == 0 {
--- a/darepod/wallet_ops.go
+++ b/darepod/wallet_ops.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/darepo-client/lib/scripts"
 	"github.com/lightninglabs/darepo-client/oor"
 	"github.com/lightninglabs/darepo-client/vtxo"
 	"github.com/lightningnetwork/lnd/aezeed"
@@ -121,9 +122,23 @@ func BuildTransferInputs(ctx context.Context,
 			)
 		}
 
+		// The checkpoint output collab path is a 2-of-2 multisig
+		// between the VTXO owner and the operator, matching the
+		// VTXO's own collaborative spend path. This ensures both
+		// parties must sign the Ark tx that spends the checkpoint.
+		collabLeaf, err := scripts.MultiSigCollabTapLeaf(
+			desc.ClientKey.PubKey, desc.OperatorKey,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"build collab leaf for %s: %w",
+				op, err,
+			)
+		}
+
 		inputs = append(inputs, oor.TransferInput{
 			VTXO:            desc,
-			OwnerLeafScript: desc.PkScript,
+			OwnerLeafScript: collabLeaf.Script,
 		})
 	}
 
--- a/baselib/AGENTS.md
+++ b/baselib/AGENTS.md
@@ -31,6 +31,7 @@ build on.
 
 - Every actor message must embed `BaseMessage` to satisfy the sealed `Message` interface.
 - Actors registered with system are auto-stopped on `system.Shutdown()`.
+- `DurableMailbox.Send` preserves the sender's DB transaction in the context. Same-DB actors share the tx for atomic enqueue via `ExecTx` joining.
 - Protofsm processes events iteratively; internal events loop until exhausted before returning.
 - `EmittedEvent` separates internal events (routed back to FSM) from outbox events (dispatched externally).
 
--- a/baselib/CLAUDE.md
+++ b/baselib/CLAUDE.md
@@ -31,6 +31,7 @@ build on.
 
 - Every actor message must embed `BaseMessage` to satisfy the sealed `Message` interface.
 - Actors registered with system are auto-stopped on `system.Shutdown()`.
+- `DurableMailbox.Send` preserves the sender's DB transaction in the context. Same-DB actors share the tx for atomic enqueue via `ExecTx` joining.
 - Protofsm processes events iteratively; internal events loop until exhausted before returning.
 - `EmittedEvent` separates internal events (routed back to FSM) from outbox events (dispatched externally).
 
--- a/baselib/actor/durable_mailbox.go
+++ b/baselib/actor/durable_mailbox.go
@@ -107,13 +107,12 @@ func NewDurableMailbox[M TLVMessage, R any](
 // envelope is accepted, the provided context is cancelled, or the actor's
 // context is cancelled.
 //
-// The sender's database transaction is intentionally stripped before
-// enqueueing. Mailbox sends cross actor boundaries: the receiver's
-// delivery store must manage its own transaction lifecycle. Inheriting
-// the sender's transaction would either target the wrong database
-// (cross-DB actors) or deadlock for Ask-style calls where the sender
-// blocks on the response but the receiver cannot see the uncommitted
-// message.
+// The sender's database transaction (if any) is preserved in the context
+// passed to EnqueueMessage. When both actors share the same SQLite
+// database, ExecTx in the delivery store joins the outer transaction,
+// making the enqueue atomic with the sender's state change. This
+// eliminates the window where a crash could commit the sender's state
+// but lose the enqueued message.
 func (m *DurableMailbox[M, R]) Send(ctx context.Context, env envelope[M, R]) bool {
 	m.closeMu.RLock()
 	defer m.closeMu.RUnlock()
@@ -185,11 +184,11 @@ func (m *DurableMailbox[M, R]) Send(ctx context.Context, env envelope[M, R]) boo
 		MaxAttempts:     m.cfg.MaxAttempts,
 	}
 
-	// Strip the sender's transaction so the receiver's delivery store
-	// creates its own transaction for the enqueue operation.
-	enqueueCtx := WithoutTx(ctx)
-
-	if err := m.cfg.Store.EnqueueMessage(enqueueCtx, params); err != nil {
+	// Allow the sender's transaction (if any) to flow through so
+	// same-DB actors share the tx and the enqueue is atomic with
+	// the sender's state change. ExecTx in the delivery store
+	// joins the outer tx when one exists in the context.
+	if err := m.cfg.Store.EnqueueMessage(ctx, params); err != nil {
 		// Clean up the promise registry entry to prevent unbounded
 		// stale entries from accumulating on repeated enqueue failures.
 		if promiseID != "" {
--- a/baselib/actor/durable_mailbox_test.go
+++ b/baselib/actor/durable_mailbox_test.go
@@ -1090,11 +1090,11 @@ func (s *txCapturingStore) EnqueueMessage(ctx context.Context,
 	return s.mockDeliveryStore.EnqueueMessage(ctx, params)
 }
 
-// TestDurableMailboxSendStripsSenderTx verifies that Send strips the sender's
-// database transaction from the context before calling EnqueueMessage. This
-// prevents the receiver's delivery store from inheriting the sender's
-// transaction, which would cause cross-DB visibility issues or Ask deadlocks.
-func TestDurableMailboxSendStripsSenderTx(t *testing.T) {
+// TestDurableMailboxSendPreservesSenderTx verifies that Send preserves the
+// sender's database transaction in the context passed to EnqueueMessage.
+// This allows same-DB actors to share the transaction so the enqueue is
+// atomic with the sender's state change.
+func TestDurableMailboxSendPreservesSenderTx(t *testing.T) {
 	t.Parallel()
 
 	store := newMockDeliveryStore()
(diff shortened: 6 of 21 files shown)


Roasbeef:
@claude review this 

@codex review this

darepod/wallet_ops.go
```diff
@@ -121,9 +122,23 @@ func BuildTransferInputs(ctx context.Context,
...
+			return nil, fmt.Errorf(
+				"build collab leaf for %s: %w",
+				op, err,
+			)
+		}
+
 		inputs = append(inputs, oor.TransferInput{
 			VTXO:            desc,
-			OwnerLeafScript: desc.PkScript,
+			OwnerLeafScript: collabLeaf.Script,
```
bhandras, line 141: But isn't the desc.PkScript already the script we want? Or this is wrong because it could have other closures too right? I think claude slightly misunderstands the issue.

bhandras: Ok i get the fix now, but the commit message feels wrong. How about something like this:
```
  darepod: use collab leaf script for checkpoint inputs

  BuildTransferInputs was populating TransferInput.OwnerLeafScript with
  the VTXO output pkScript. That field is not an output script; it is the
  raw tapscript leaf that gets committed into the checkpoint output tree.

  For standard VTXOs, the correct leaf is the 2-of-2 collaborative path
  between the client and operator. Using the outer pkScript here produces
  the wrong checkpoint tree and breaks the later Ark signing flow, which
  expects to find that collaborative leaf on each checkpoint spend path.
```

darepod/rpc_server.go
```diff
@@ -895,6 +900,36 @@ func (r *RPCServer) resolveOutputPkScript(
...
+
+		recipientKey, err := schnorr.ParsePubKey(d.Pubkey)
+		if err != nil {
+			return nil, status.Errorf(
+				codes.InvalidArgument,
+				"invalid pubkey: %v", err,
+			)
+		}
+
+		tapKey, err := scripts.VTXOTapKey(
```
bhandras, line 931: codex: The new `pubkey` destination behavior no longer matches the public RPC contract. `daemon.proto` still says `Output.pubkey` is used to derive a BIP-86 taproot output, but this code now derives an Ark/VTXO output with `scripts.VTXOTapKey(...)`, which depends on operator policy and has different spend/custody semantics. External clients that follow the proto comment will derive/display the wrong destination type. This should either keep the old BIP-86 behavior or update the RPC contract/docs to make the Ark-specific semantics explicit.

darepod/rpc_server.go
```diff
@@ -694,10 +696,18 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 			"amount must be positive")
 	}
 
+	// Fetch operator terms early so pubkey destinations can derive
+	// VTXO-compatible pkScripts.
+	terms, err := r.server.fetchOperatorTerms(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal,
+			"unable to fetch operator terms: %v", err)
```
bhandras: codex: This moves operator-term fetching ahead of both the `dry_run` early return and non-pubkey destination handling. That changes behavior for address and raw `pk_script` requests: previously those paths could still return a local preview without contacting the operator, but now they fail whenever `fetchOperatorTerms` is unavailable. Was that regression intended? If pubkey destinations need terms, I think the fetch should stay conditional on that case rather than becoming mandatory for all `SendOOR` requests.

darepod/rpc_server.go
Roasbeef: Yeah the BIP 86 comment is wrong, since rn it's just the pubkey to be used in the vtxo script. Will modify.

PR #181, opened 2026-03-16
oor: production-identical OOR receive path with async materialization

Roasbeef:
## Summary

This PR implements the full production OOR incoming transfer receive
path with crash-safe async materialization. The changes enable clients
to receive OOR transfers through production connectors (serverconn +
mailbox) without SQLite write-lock starvation or ingress deadlocks.

## Detailed Fix Catalogue

See the [OOR fixes analysis gist](https://gist.github.com/Roasbeef/360a155ad12bcd8565cbe88c1d845bf3)
for the full 20-fix catalogue with production applicability matrix.

### Three-Phase Async Receive (oor/actor.go, oor/receive_*)
- Phase 1: Route persists lightweight ResolveIncomingTransferRequest (TLV durable)
- Phase 2: Actor resolves via IncomingFetcher outside DB tx (no lock contention)
- Phase 3: Actor drives receive FSM → async MaterializeIncomingVTXOs → callback

### Owner-Pubkey TLV Proof (indexer/client.go, indexer/signer_adapters.go)
- VTXO tapscripts use NUMS internal key — client can't sign output-key proof
- New TLV field carries owner pubkey; server reconstructs + verifies

### Async Materialization (oor/local_persistence_handler.go)
- Detects actor DB tx via hasActorDBTx(ctx)
- Schedules indexer queries in background goroutine
- CallbackRef delivers result as fresh DriveEventRequest

### Production Daemon Wiring (darepod/server.go)
- EnsureDefaultOORReceiveKey centralized for daemon + systest
- IncomingFetcher, CallbackRef, operator terms all wired
- SendClientEventRequest auto-derives ServiceMethod

## Test Plan

- [x] `go test ./oor -count=1` — all OOR unit tests
- [x] `go test ./darepod ./indexer ./oor -count=1` — integration
- [x] `go test ./serverconn -count=1` — connector tests
- [x] Full systest suite (417s, zero failures)

### DIFF
--- a/arkrpc/indexer.proto
+++ b/arkrpc/indexer.proto
@@ -152,6 +152,16 @@ message OORRecipientEvent {
     bytes session_id = 3;
     uint32 output_index = 4;
     uint64 value = 5;
+
+    // ark_psbt is the serialized Ark PSBT for this OOR session. The
+    // client needs this to materialize the received VTXO via the
+    // IncomingTransferEvent in the OOR FSM.
+    bytes ark_psbt = 6;
+
+    // checkpoint_psbts are the finalized checkpoint PSBTs for the
+    // session. These provide lineage and unroll proof data for the
+    // materialized VTXO.
+    repeated bytes checkpoint_psbts = 7;
 }
 
 // IncomingOOREvent is delivered as a mailbox EVENT envelope. It is
--- a/arkrpc/indexer.proto
+++ b/arkrpc/indexer.proto
@@ -288,6 +288,10 @@ message VTXO {
     // and the most recent on-chain commitment. Round-created VTXOs have
     // chain_depth 0.
     uint32 chain_depth = 15;
+
+    // tree_path_tlv is the TLV-encoded extracted tree.Tree path required for
+    // unilateral exit of this VTXO's underlying commitment lineage.
+    bytes tree_path_tlv = 16;
 }
 
 message ListVTXOsByScriptsRequest {
--- a/db/interfaces.go
+++ b/db/interfaces.go
@@ -5,6 +5,8 @@ import (
 	"database/sql"
 	"math"
 	prand "math/rand"
+	"path/filepath"
+	"runtime"
 	"time"
 
 	"github.com/btcsuite/btclog/v2"
@@ -247,6 +249,8 @@ func (t *TransactionExecutor[Q]) ExecTx(ctx context.Context,
 		return txBody(t.createQuery(tx))
 	}
 
+	callerFile, callerLine, callerFunc := txRetryCaller()
+
 	waitBeforeRetry := func(attemptNumber int) {
 		retryDelay := t.opts.randRetryDelay(attemptNumber)
 
@@ -256,6 +260,9 @@ func (t *TransactionExecutor[Q]) ExecTx(ctx context.Context,
 				"deadlock error",
 			"attempt_number", attemptNumber,
 			"delay", retryDelay,
+			"caller_file", callerFile,
+			"caller_line", callerLine,
+			"caller_func", callerFunc,
 		)
 
 		// Before we try again, we'll wait with a random backoff based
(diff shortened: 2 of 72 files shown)


arkrpc/indexer.proto
```diff
@@ -278,6 +288,10 @@ message VTXO {
     // and the most recent on-chain commitment. Round-created VTXOs have
     // chain_depth 0.
     uint32 chain_depth = 15;
+
+    // tree_path_tlv is the TLV-encoded extracted tree.Tree path required for
+    // unilateral exit of this VTXO's underlying commitment lineage.
+    bytes tree_path_tlv = 16;
```
Roasbeef, line 294: Add an an actual proto set?

db/interfaces.go
```diff
@@ -323,6 +330,43 @@ func (t *TransactionExecutor[Q]) ExecTx(ctx context.Context,
 	return ErrRetriesExceeded
 }
 
+// txRetryCaller returns the first caller above the db transaction helpers so
+// retry logs can identify which store method is contending on the database.
+func txRetryCaller() (string, int, string) {
```
Roasbeef, line 335: Debug, can be removed.

db/vtxo_store.go
```diff
@@ -59,6 +65,50 @@ func (s *VTXOPersistenceStore) SaveVTXO(
...
+		RoundID:            desc.RoundID,
+		StartHeight:        0,
+		ConfirmationHeight: confirmationHeight,
+		CommitmentTxid:     desc.CommitmentTxID[:],
+		Status:             "confirmed",
+		CreationTime:       nowUnix,
+		LastUpdateTime:     nowUnix,
+	}
+
+	return q.InsertRound(ctx, params)
```
Roasbeef, line 109: Can be an upsert instead?

Roasbeef:
## Additional Changes Since Initial PR

Several follow-up fixes landed after the initial PR was opened:

### Transport-Native Durable Unary Queries
The OOR incoming resolve path no longer uses goroutine-based callbacks or `IncomingFetcher` closures. Instead, the actor emits transport-native durable `serverconn` messages:

- `SendListOORRecipientEventsByScriptRequest` — resolves the lightweight incoming hint into the full Ark PSBT
- `SendListVTXOsByScriptsRequest` — resolves authoritative VTXO lineage metadata

Both are TLV-serializable, crash-safe, and processed by `serverconn` after commit via a new `DurableUnaryBuilder` that constructs proof-gated request bodies. Responses arrive as durable events through the `EventRouter` fallback when no in-memory waiter exists.

### `ResponseRegistry` Rich Delivery Result
`DeliverResponse` now returns a tri-state (`waiter`/`buffered`/`dropped`) so the ingress can fall back to durable route dispatch for responses that arrive with no live waiter (crash replay case).

### `NewOORReceiveScript` RPC + `darepocli oor receive`
Fresh OOR receive-target allocation is now exposed through the public `DaemonService` RPC and CLI. Returns `pk_script_hex`, pubkey, key locator metadata, and label.

### Remaining Follow-Ups
- Generalize `DurableUnaryBuilder` beyond the two OOR indexer queries
- Add `send oor --pk_script`/`--pubkey` CLI flags
- Evaluate MCP exposure for `oor.receive`

Full fix catalogue: https://gist.github.com/Roasbeef/360a155ad12bcd8565cbe88c1d845bf3

Roasbeef:
## CLI + MCP OOR Surface

Latest commit adds the public OOR surface follow-up:

- **`darepocli send oor`** now accepts `--to` (address), `--pubkey` (32-byte x-only hex), or `--pk_script` (raw hex) with mutual exclusion validation
- **MCP** gains `oor_receive` tool (allocates fresh receive script) and richer `send_oor` args
- Shared `oor_destination.go` parsing module with 32-byte x-only pubkey checks
- Regression test coverage for all destination parsing paths
- CLI guide updated

All tests green (CLI unit tests + full systest suite 417s).

Roasbeef:
Folded into smaller PR series. Now all merged.

PR #191, opened 2026-03-17
multi: add tests for OOR receive, durable messages, and VTXO store

Roasbeef:
In this PR, we add test coverage for the OOR receive path introduced in the
preceding PRs in this stack.

The OOR actor tests cover the \`ResolveIncomingTransferRequest\` TLV codec
round-trip and verify that an actor restart with a pending resolve correctly
resumes the receive FSM. The durable message tests exercise incoming session
snapshot encode/decode and checkpoint version upgrade paths.

\`actor_drive_event_integration_test.go\` is the end-to-end drive event test
that walks the full lifecycle: submit → finalize → incoming transfer →
materialize. This confirms the three-phase async receive flow works
correctly when driven through the actor's public \`DriveEventRequest\`
interface.

On the DB side, \`vtxo_store_test.go\` verifies that \`ensureRoundExists\`
creates a minimal round row for imported incoming VTXOs without overwriting
richer state on rounds that already exist. \`round/from_proto_test.go\`
covers the \`SigningKey\` field decode in \`VTXORequest\` conversion.

Builds on #190.

### DIFF
--- a/db/vtxo_store_test.go
+++ b/db/vtxo_store_test.go
@@ -167,6 +167,68 @@ func TestVTXOPersistenceStoreSaveAndGet(t *testing.T) {
 	)
 }
 
+// TestVTXOPersistenceStoreSaveVTXOCreatesMissingRound verifies that SaveVTXO
+// creates a minimal local round row for imported OOR VTXOs whose source round
+// is otherwise unknown to the client.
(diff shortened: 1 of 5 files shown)


Roasbeef:
@claude review this

PR #197, opened 2026-03-17
Part 1: client low-risk follow-ups for integration test migration

bhandras:
## Summary
This is Part 1 of the integration-test migration follow-ups on the client side.

It includes two low-risk changes:
- preserve notifier event delivery after request-context cancellation in the LND chain backend
- add non-Docker lint targets (`lint-local`, `lint-source-local`, `lint-changed-local`) with a local `custom-gcl` fallback path

## Commits
- `e5dbdd2` chainbackends: preserve notifier delivery after request context
- `e75143c` build: add non-docker lint targets

## Validation
- `make lint-changed-local base=origin/main workers=2`
- targeted chainbackend tests were previously run when this stack was prepared

## Notes
- This PR is intentionally scoped to low-risk client changes only.
- Server-side integration-test expansion is tracked in a separate Part 1 PR in `darepo`.

### DIFF
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -205,11 +205,18 @@ func (b *LNDBackend) RegisterConf(ctx context.Context,
 			err)
 	}
 
+	// The caller context only scopes registration setup. Keep the delivery
+	// forwarder alive until the registration itself is cancelled so
+	// confirmations are not dropped when the originating actor request
+	// context is released.
+	notifyCtx, cancel := context.WithCancel(context.Background())
+
 	// Create a channel to convert lnd's TxConfirmation to our type.
 	confChan := make(chan *chainsource.TxConfirmation, 1)
 
 	go func() {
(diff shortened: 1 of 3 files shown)


chainbackends/lnd.go
```diff
@@ -258,12 +268,16 @@ func (b *LNDBackend) RegisterSpend(ctx context.Context,
 		return nil, fmt.Errorf("failed to register spend: %w", err)
 	}
 
+	// Keep spend delivery alive independently of the actor request context.
+	notifyCtx, cancel := context.WithCancel(context.Background())
```
Roasbeef, line 272: 👍 

We've made a similar fix across both repos. Basically, sometimes that we pass the context down too far. Probably makes sense to make some sort of like scale or do a pass around context propagation to see if this was gonna pop up anywhere else.

Roasbeef (approved):
LGTM 🦾

PR #201, opened 2026-03-18
build: add native custom-gcl installer

bhandras:
## Summary

This splits the native local-linter bootstrap into a dedicated tooling PR.

## Changes

- add `scripts/install-custom-gcl.sh` to build a host-native `custom-gcl`
- move the local bootstrap logic out of the `Makefile` into dedicated scripts
- update `scripts/local-custom-gcl.sh` to prefer PATH, reuse an existing local
  binary, try a native build, and only then fall back to the `lll`
  approximation
- add `make install-custom-gcl`
- document the native install flow in `AGENTS.md`, `CLAUDE.md`, and the
  corresponding `tools/` docs

## Why

The existing non-Docker lint path works, but it falls back to stock
`golangci-lint` and disables the repo-local `ll` plugin. This keeps the local
workflow fast while letting `make lint-local` and `make lint-changed-local`
use the real plugin after one install step.

## Validation

- `shellcheck scripts/install-custom-gcl.sh scripts/local-custom-gcl.sh`
- `./scripts/install-custom-gcl.sh ./tools/custom-gcl-native-test`
- `./tools/custom-gcl-native-test version`
- `make local-custom-gcl`
- `make lint-changed-local base=HEAD~1 workers=1`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,6 +8,8 @@ This file is a **map**, not a manual. Follow links for details.
 |---------|---------|
 | `make build` | Compile the project |
 | `make lint` | Run linter (must pass before committing) |
+| `make lint-changed-local` | Run native local linter on changes vs base |
+| `make install-custom-gcl` | Build native `custom-gcl` for this host |
 | `make fmt` | Format all Go source files |
 | `make unit pkg=<pkg> case=<test>` | Run unit tests |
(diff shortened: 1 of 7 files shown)


Roasbeef (approved):
LGTM 🍈

PR #202, opened 2026-03-18
internal/actortest: harden outbox e2e waits

bhandras:
## Summary

This follow-up hardens the outbox end-to-end timing in `internal/actortest`.

- increase the timeout budget for waiting on the source actor to durably
  process forwarded messages in the outbox e2e tests
- use `eventuallyWithOutboxPublish(...)` for outbox delivery assertions so the
  tests actively trigger publish attempts instead of relying only on the
  publisher ticker under `-race`
- keep the change test-only and scoped to the flaky outbox integration cases

## Why

PR #195 hit a `unit-race` timeout in
`TestOutboxPublisher_DeliversToTarget` before the source actor completed the
forward-processing step. These tests exercise real SQLite persistence and can
become scheduler-sensitive in CI, so the previous timing budget was sometimes
too tight.

## Testing

- `go test -tags='dev nolog' -run 'TestOutboxPublisher_DeliversToTarget|TestOutboxPublisher_MultiHopForwarding|TestProperty_OutboxEventuallyDelivers' -count=1 ./internal/actortest`
- `go test -tags='dev nolog' -race -run 'TestOutboxPublisher_DeliversToTarget|TestOutboxPublisher_MultiHopForwarding|TestProperty_OutboxEventuallyDelivers' -count=5 ./internal/actortest`
- `go test -tags='dev nolog' -race ./internal/actortest`

### DIFF
--- a/internal/actortest/e2e_test.go
+++ b/internal/actortest/e2e_test.go
@@ -172,6 +172,18 @@ func eventuallyWithOutboxPublish(
 // delivery assertions in this test suite.
 const durableAskResponseTimeout = 10 * time.Second
 
+// outboxForwardProcessingTimeout is the timeout budget for waiting on the
+// source actor to durably process a forward request before any outbox delivery
+// assertions. This is intentionally looser than basic actor assertions because
+// these end-to-end tests run with real SQLite persistence and under `-race` in
+// CI.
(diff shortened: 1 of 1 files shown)


Roasbeef (approved):
LGTM 🫐

PR #203, opened 2026-03-18
serverconn: route header-only unary error responses

bhandras:
## Summary
- allow routed KIND_RESPONSE envelopes with nil body when gRPC error headers are present
- keep rejecting malformed envelopes with nil body and no encoded error
- add ingress coverage for routed header-only error responses reaching EventRouter adapters

## Why
OOR receive metadata queries can return mailbox-routed unary failures encoded in headers only. Dropping those envelopes prevented the receive FSM from getting deterministic fail events.

## Testing
- go test ./serverconn -run "TestIngress_ResponseDispatch(HeaderOnlyError|WithoutWaiter)" -count=1

### DIFF
--- a/serverconn/event_router.go
+++ b/serverconn/event_router.go
@@ -176,22 +176,32 @@ func AddEnvelopeRoute[M actor.Message, R any](r *EventRouter,
 	dispatcher := func(ctx context.Context,
 		env *mailboxpb.Envelope) error {
 
-		if env == nil || env.Body == nil {
-			return fmt.Errorf("nil envelope or body for %s/%s",
+		if env == nil {
+			return fmt.Errorf("nil envelope for %s/%s",
 				cfg.Service, cfg.Method)
 		}
 
-		event := cfg.NewEvent()
-		if event == nil {
-			return fmt.Errorf("nil event prototype for %s/%s",
-				cfg.Service, cfg.Method)
-		}
+		var event proto.Message
+		if env.Body != nil {
+			event = cfg.NewEvent()
+			if event == nil {
+				return fmt.Errorf(
+					"nil event prototype for %s/%s",
+					cfg.Service, cfg.Method,
+				)
+			}
 
-		if err := (proto.UnmarshalOptions{
-			DiscardUnknown: true,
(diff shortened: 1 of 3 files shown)


serverconn/event_router.go
```diff
@@ -176,24 +176,32 @@ func AddEnvelopeRoute[M actor.Message, R any](r *EventRouter,
...
+				)
+			}
+
+			if err := (proto.UnmarshalOptions{
+				DiscardUnknown: true,
+			}).Unmarshal(env.Body.Value, event); err != nil {
+				return fmt.Errorf("unmarshal %s/%s event: %w",
+					cfg.Service, cfg.Method, err)
+			}
+		} else if mailboxrpc.DecodeErrorHeaders(env.Headers) == nil {
```
Roasbeef: Third case of the body being nil with no errors?

bhandras: codex: Yes, that is the malformed third case I meant to handle here. The updated dispatcher now makes that explicit: if `env.Body == nil` and there is no encoded gRPC error in the mailbox headers, we return an error immediately and fail closed. The only nil-body case we allow is a header-only unary failure, where the server encoded the gRPC status in the headers. I also added a regression test for the malformed nil-body/no-error path so that case stays covered.

lndbackend/client_wallet.go
```diff
@@ -231,6 +218,26 @@ func (c *ClientWallet) MuSig2CreateSession(
 	return sessionInfo, nil
 }
 
+// serializeMuSig2SignerPubKeys matches lnd's RPC expectations for each
+// MuSig2 draft version. v0.4.0 uses x-only keys while v1.0.0rc2 uses
+// compressed public keys.
```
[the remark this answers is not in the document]
bhandras: Keeping the original wording here. I double-checked both the helper and the corresponding test, and they agree on the draft-version mapping that lnd expects for signer key serialization: `v0.4.0` uses x-only signer keys, while `v1.0.0rc2` uses compressed signer keys. So this comment was not swapped after all.

Roasbeef (approved):
LGTM 🌂

PR #204, opened 2026-03-18
darepod: add integration harness prerequisites

bhandras:
## Summary\n- add configurable daemon log writer output selection\n- fix MuSig2 signer key serialization for v1.0.0rc2 in the lnd backend\n- decouple indexer proof server ID from mailbox routing IDs\n- expose the bound RPC listen address for in-process harness discovery\n\n## Why\nThese commits are prerequisites for reliable real-daemon integration tests and should land separately from the routed header-only error envelope fix.\n\n## Testing\n- go test ./darepod/...\n- go test ./lndbackend/...\n

### DIFF
--- a/darepod/config.go
+++ b/darepod/config.go
@@ -2,6 +2,7 @@ package darepod
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"time"
@@ -84,6 +85,10 @@ type Config struct {
 	// the list (without a '=') sets the default for unlisted subsystems.
 	// Valid levels: trace, debug, info, warn, error, critical, off.
 	DebugLevel string `mapstructure:"debuglevel"`
+
+	// LogWriter is the sink for daemon log output. When nil, darepod
+	// writes logs to stdout.
+	LogWriter io.Writer
 
 	// Lnd configures the connection to the backing lnd node.
 	Lnd *LndConfig `mapstructure:"lnd"`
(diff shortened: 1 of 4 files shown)


darepod/config.go
```diff
@@ -85,6 +86,10 @@ type Config struct {
 	// Valid levels: trace, debug, info, warn, error, critical, off.
 	DebugLevel string `mapstructure:"debuglevel"`
 
+	// LogWriter is the sink for daemon log output. When nil, darepod
+	// writes logs to stdout.
+	LogWriter io.Writer
```
Roasbeef, line 91: Nice, yeah commented about this in the other PR.

lndbackend/client_wallet.go
```diff
@@ -231,6 +218,26 @@ func (c *ClientWallet) MuSig2CreateSession(
...
 }
 
+// serializeMuSig2SignerPubKeys matches lnd's RPC expectations for each
+// MuSig2 draft version. v0.4.0 uses x-only keys while v1.0.0rc2 uses
+// compressed public keys.
+func serializeMuSig2SignerPubKeys(version input.MuSig2Version,
+	allSignerPubkeys []*btcec.PublicKey) [][]byte {
+
+	signerBytes := make([][]byte, len(allSignerPubkeys))
+	for i, pk := range allSignerPubkeys {
```
Roasbeef, line 228: If we had this backwards,, we're somewhat surprising if the test passed otherwise?

Roasbeef (approved):
LGTM 🫐

PR #205, opened 2026-03-19
actortest: force outbox publish in delivery poll loops

Roasbeef:
Three outbox-delivery e2e tests (`TestOutboxPublisher_DeliversToTarget`,
`TestOutboxPublisher_MultiHopForwarding`,
`TestProperty_OutboxEventuallyDelivers`) relied solely on the
`OutboxPublisher`'s internal ticker to deliver pending messages during
their `eventually` polling loops. Under heavy CI scheduler pressure or
the race detector, the ticker can be delayed long enough to trip the
timeout and cause a flake.

This commit switches the delivery-wait assertions to use
`eventuallyWithOutboxPublish`, which calls `PublishPending()` on every
poll iteration. This is the same pattern the `DurableAsk` tests already
use, making delivery deterministic regardless of goroutine scheduling
jitter.

Verified with `go test -race -count=50` (150 total runs across the
three tests) — zero failures, zero data races.

### DIFF
--- a/internal/actortest/e2e_test.go
+++ b/internal/actortest/e2e_test.go
@@ -393,9 +393,14 @@ func TestOutboxPublisher_DeliversToTarget(t *testing.T) {
 	})
 
 	// Wait for target to receive the increment via OutboxPublisher.
-	eventually(t, 5*time.Second, func() bool {
-		return targetBehavior.Count() == 25
-	})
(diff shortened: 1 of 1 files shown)


Roasbeef:
Closing — this fix was already landed on main via an earlier merge. The branch is now empty after rebasing.

PR #206, opened 2026-03-19
wallet+round: mark consumed boarding intents adopted at checkpoint

bhandras:
## Summary

This PR fixes a multi-round boarding accounting bug where consumed boarding
intents could remain visible as `confirmed`, causing a later board request to
overcount spendable boarding balance.

The fix now applies the adoption transition through the wallet actor at round
checkpoint time, so boarding intent lifecycle ownership remains in the wallet
layer while still updating promptly once the checkpoint is durable.

## Changes

- Add a wallet actor message pair for checkpoint-time adoption updates.
- Add `MarkBoardingIntentsAdopted` to the wallet store interface.
- Implement `MarkBoardingIntentsAdopted` in the DB store as a single write
  transaction with deduplication and expected-state validation.
- Keep round persistence focused on round-owned tables (remove wallet-table
  status mutation from round store commit logic).
- Have the round actor issue the wallet adoption request during checkpoint
  handling.
- Add coverage in `db`, `wallet`, and `round` tests for transition behavior,
  invariants, and checkpoint propagation.

## Validation

- `go test ./... -count=1`
- `./tools/custom-gcl run --timeout=10m ./wallet ./round ./db`

### DIFF
--- a/db/boarding_wallet.go
+++ b/db/boarding_wallet.go
@@ -33,6 +33,8 @@ type IntentHeightFilter = sqlc.ListBoardingIntentsByStatusAndMinHeightParams
 
 // BoardingStore is the interface that groups all boarding-related database
 // queries. This is a subset of sqlc.Querier focused on boarding operations.
+//
+//nolint:interfacebloat
 type BoardingStore interface {
 	InsertBoardingAddress(ctx context.Context, arg NewAddrParams) error
 
@@ -61,6 +63,10 @@ type BoardingStore interface {
 	ListBoardingIntentsByStatusAndMinHeight(
 		ctx context.Context, arg IntentHeightFilter,
 	) ([]BoardingIntentRow, error)
+
+	UpdateBoardingIntentStatus(
+		ctx context.Context, arg sqlc.UpdateBoardingIntentStatusParams,
+	) error
 }
 
 // BatchedBoardingStore combines BoardingStore with transaction support via the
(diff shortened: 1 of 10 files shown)


ellemouton (commented):
I wonder if this is the correct approach given our actor architecture - does this not mix up layers/concerns a bit? 

Afaict, this is the first point where the round store now also writes to wallet related tables. Imo, i think perhaps only the wallet actor should do that? ie, would it not be possible to instead have an outbox event notifying the wallet actor that the boarding intent is now usesd & then it can update its in-memory state as well as persisted state accordingly?

I know that breaks atomicity a bit - but i think with durability & serial processing, this may be ok?

My instinct is just to keep actors as the "owners" of their tables so that they can cache & do in-memory stuff without worrying about actors writing to their stores. 

Thoughts?

bhandras:
> I wonder if this is the correct approach given our actor architecture - does this not mix up layers/concerns a bit?
> 
> Afaict, this is the first point where the round store now also writes to wallet related tables. Imo, i think perhaps only the wallet actor should do that? ie, would it not be possible to instead have an outbox event notifying the wallet actor that the boarding intent is now usesd & then it can update its in-memory state as well as persisted state accordingly?
> 
> I know that breaks atomicity a bit - but i think with durability & serial processing, this may be ok?
> 
> My instinct is just to keep actors as the "owners" of their tables so that they can cache & do in-memory stuff without worrying about actors writing to their stores.
> 
> Thoughts?

Yeah totally agree, ptal at the reworked version.

ellemouton (approved):
nice! thanks for the update 🙏

PR #212, opened 2026-03-20
arkscript: move the client to policy-first scripts

sputn1ck:
## Summary

This PR moves the client stack to a policy-first model, where VTXO ownership and spend semantics travel as `PolicyTemplate` blobs plus explicit spend-path metadata and are compiled into taproot artifacts consistently at the edges. It replaces the old raw script tuple flow with `arkscript`-backed policy data, wiring that model through shared library helpers, protocol types, state machines, persistence, wallet and VTXO runtime, and daemon-facing integration surfaces.

The stack also includes the follow-up fixes that were needed to make the migrated flow correct and reviewable after rebasing onto fresh `main`: preserving stored concrete pubkeys when rehydrating round state from DB rows, fixing local-owner healing for zero-valued key locators, teaching directed sends to accept standard `policy_template` recipients, and keeping the final transition and generated-code paths lint-clean. The daemon integration step intentionally preserves the mailbox-edge factory and round-registration harness hooks so downstream integration tests do not lose useful test surface just because the client moved to policy-backed descriptors.

## Commit Stack (18 commits, bottom-up)

1. **build: use printf for colored Makefile output** - make the top-level `Makefile` status banners shell-safe so CI and `/bin/sh` runs do not trip over the colored output.
2. **lib/types: add policy-backed shared types** - introduce the shared policy-bearing types and core boarding-side data model.
3. **lib/tree: build VTXOs from policy templates** - teach the tree builders and materializers to consume policy-backed descriptors.
4. **lib/tx/checkpoint: compile policy-backed artifacts** - move checkpoint transaction construction onto semantic policy inputs and richer checkpoint artifacts.
5. **lib/tx: compile policy-backed forfeit spends** - switch forfeit tx construction to `arkscript` and introduce explicit sequence and locktime context for policy-driven spend paths.
6. **lib/tx/oor: compile policy-backed submit artifacts** - move OOR submit builders and tapleaf helpers onto policy-backed spend information.
7. **lib/scripts: remove legacy helper package** - drop the old script-helper package after its remaining tx-layer consumers have moved away.
8. **lib/tx/oor: keep tap trees on checkpoint PSBTs** - make checkpoint PSBT outputs the canonical home for tap-tree metadata and remove the late finalize rebinding helpers.
9. **lib/arkscript: clarify validation comments** - isolate the small validation comment cleanup from the functional migration steps.
10. **rpc: update round and OOR policy wire types** - update the round and OOR transport schemas plus generated protocol code.
11. **daemonrpc: add policy-backed lookup RPCs** - add daemon and indexer lookup RPC surfaces together with regenerated stubs and guards.
12. **indexer/darepod: consume lookup RPCs** - wire the new lookup RPCs into the indexer client and daemon-side lookup handlers.
13. **baselib/actor: add durable actor wait helpers** - split reusable durable-actor support into a small prep commit.
14. **oor: migrate actor and signing flow to policy** - port the OOR actor, signing flow, and persistence helpers to policy-backed inputs.
15. **round: migrate round registration to policy** - move round registration, join auth, codec, and transition handling to the new model, including the forfeit-only submission cleanup and the final FSM refactors that keep the migrated flow lint-clean.
16. **wallet/vtxo: propagate policy-backed descriptors** - thread policy-aware descriptors through wallet and VTXO business logic, including the incoming-handler fixups needed after rebasing onto fresh `main`.
17. **db: persist policy blobs and regenerate sqlc** - align storage, migrations, generated `sqlc` output, and merged-schema artifacts with the policy model, including preservation of stored concrete owner and operator pubkeys when round state is rehydrated from DB rows and healing of zero-locator local owners.
18. **darepod: wire policy through daemon and CLI** - finish the integration through darepod, wallet ops, chain backend plumbing, CLI, and public-path tests, while deliberately preserving the mailbox-edge factory and round-registration harness hooks used by downstream integration harnesses and adding focused vHTLC custom-input coverage for the policy-first path.

## Test Plan

- [x] `make lint-local`
- [x] `make sqlc-check`
- [x] `go test ./vtxo ./round ./db ./darepod ./oor ./wallet -count=1`
- [x] `go test ./darepod ./lib/arkscript ./oor -count=1`
- [x] `go test ./darepod -run 'TestServerNewMailboxEdgeUsesFactory|TestServerTriggerRoundRegistration|TestResolveRecipientOutput(Pubkey|Address|PolicyTemplateStandard|PolicyTemplateCustomRejected|NonTaprootRejected|InvalidPubkey)$' -count=1`
- [x] `go test ./db -run 'TestVTXOStore(SaveVTXOsHealsZeroLocatorOwner|GetVTXOPreservesStoredPubKeyParity)$' -count=1`


db/boarding_wallet.go
```diff
@@ -426,94 +420,6 @@ func (b *BoardingWalletStore) GetIntent(ctx context.Context,
 	return result, err
 }
 
-// MarkBoardingIntentsAdopted marks the provided boarding intent outpoints as
-// adopted in a single transaction. This operation validates expected state and
-// fails fast if any outpoint is missing or already in an unexpected terminal
-// state.
-func (b *BoardingWalletStore) MarkBoardingIntentsAdopted(ctx context.Context,
```
Roasbeef, line 433: Deleted?

lib/arkscript/node.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// Compile the inner expression first.
+	innerScript, err := c.Inner.Script()
+	if err != nil {
+		return nil, fmt.Errorf("csv: failed to compile inner: %w", err)
+	}
+
+	// Build the full script: <inner> <lock> CSV DROP
+	builder := txscript.NewScriptBuilder()
```
Roasbeef, line 94: Can use `ScriptTemplate`: https://pkg.go.dev/github.com/btcsuite/btcd/txscript#ScriptTemplate

lib/arkscript/psbt.go
```diff
@@ -0,0 +1,276 @@
...
+		buf[0] = 0xff
+		binary.LittleEndian.PutUint64(buf[1:9], val)
+		_, err := w.Write(buf[:9])
+
+		return err
+	}
+}
+
+// readCompactSize reads a compact size integer from the reader.
+func readCompactSize(r io.ByteReader) (uint64, error) {
```
Roasbeef, line 363: Can use https://pkg.go.dev/github.com/btcsuite/btcd/wire#ReadVarInt

lib/arkscript/nums.go
```diff
@@ -0,0 +1,41 @@
+package arkscript
```
bhandras, line 1: I think it'd be slightly cleaner if we move things that we keep in a first commit, but also right now the `nums_test.go` file is permanently deleted.

lib/arkscript/node.go
```diff
@@ -0,0 +1,182 @@
+package arkscript
```
bhandras, line 1: Not necessarily something that needs action but imo introducing the node and policy template first could help with understanding the story.

darepod/server.go
```diff
@@ -370,19 +370,6 @@ func (s *Server) run(ctx context.Context,
 	// 3. Initialize the actor system.
 	// -------------------------------------------------------
 	s.actorSystem = actor.NewActorSystem()
-	// Register cleanup from least dependent to most dependent so that the
-	// defer LIFO order tears down components from most dependent to least
-	// dependent: actor system -> chain backend -> DB.
-	defer func() {
```
bhandras, line 508: This was intentional shutdown order fix. See eg. logs in this run: https://github.com/lightninglabs/darepo/actions/runs/23349606907/job/67924463825?pr=187
```
[WRN] ACTR: Failed to lease message from mailbox err="sql: database is closed" mailbox_id=oor-client
```

Added here: https://github.com/lightninglabs/darepo-client/pull/209

darepod/wallet_ops.go
```diff
@@ -144,3 +148,74 @@ func BuildTransferInputs(ctx context.Context,
...
+			return nil, fmt.Errorf("parse outpoint %q: %w",
+				ci.Outpoint, err)
+		}
+
+		var desc *vtxo.Descriptor
+
+		// If the caller provided amount and pkscript, build
+		// the descriptor directly (for VTXOs not in the local
+		// store, e.g., received via OOR).
+		if ci.AmountSat > 0 && len(ci.PkScript) > 0 {
```
bhandras, line 177: Do we need to ensure we adhere to the policy template? IIUC its not checked.

darepod/rpc_server.go
```diff
@@ -736,63 +736,83 @@ func (r *RPCServer) SendOOR(ctx context.Context,
...
 
-	selectReq := &wallet.SelectAndLockVTXOsRequest{
-		TargetAmount: targetAmt,
-	}
-	selectFuture := wRef.Ask(ctx, selectReq)
-	selectResult := selectFuture.Await(ctx)
+	if len(req.CustomInputs) > 0 {
+		// Custom inputs provided — bypass wallet selection
+		// and build TransferInputs from the specified VTXOs.
+		selectedInputs, err = BuildCustomTransferInputs(
```
bhandras, line 978: I think we should still lock vtxos otherwise we could end up with two concurrent spends.

darepod/wallet_ops.go
sputn1ck: removed in followup commits

db/boarding_wallet.go
sputn1ck: Yes — the entire call chain is gone. The old flow was: round actor sends MarkBoardingIntentsAdoptedRequest to wallet actor, which calls store.MarkBoardingIntentsAdopted(). The new flow sets BoardingStatusAdopted directly in the FSM transition (round/transitions.go:1184) before checkpointing the round, so the status is persisted atomically with the round state. This removed the race window between checkpoint and the separate adopted-marking message. The DB method, wallet handler, and message types all became dead code.

darepod/server.go
sputn1ck: Good catch — restored in c00fa27. Separate LIFO defers again so the DB outlives all actors: runtime -> OOR actor -> actor system -> lwwallet -> chain backend -> DB -> server conn.

darepod/rpc_server.go
sputn1ck: Added a doc comment in c00fa27. Custom inputs refer to non-standard VTXOs (e.g. vHTLC outputs from swaps) outside the wallet managed set, so the wallet actor does not track them for locking. For the swap use case, the swap daemon coordinates input lifecycle externally. A lock-by-outpoint mechanism for wallet-overlapping custom inputs can be added when needed.

darepod/wallet_ops.go
sputn1ck: As sputn1ck noted, addressed in the follow-up swap commits. Policy template validation happens server-side via ValidatePolicy() at submit time.

lib/arkscript/validate.go
```diff
@@ -0,0 +1,152 @@
...
+	key *btcec.PublicKey) bool {
+
+	if len(script) == 0 || key == nil {
+		return false
+	}
+
+	target := schnorr.SerializePubKey(key)
+
+	return bytes.Contains(script, target)
+}
```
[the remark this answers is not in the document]
sputn1ck: Fixed the doc comment in c00fa27 — now accurately describes the bytes.Contains heuristic. False positives are acceptable because this is a lightweight admission gate; full script VM execution at finalize time proves correctness.

indexer/client.go
```diff
@@ -218,6 +222,71 @@ func encodeProofTLV(msgType, serverID, principal, purpose string,
...
+		return nil, fmt.Errorf("build TLV stream: %w", err)
+	}
+
+	var buf bytes.Buffer
+	if err := tlvStream.Encode(&buf); err != nil {
+		return nil, fmt.Errorf("encode TLV proof: %w", err)
+	}
+
+	return buf.Bytes(), nil
+}
```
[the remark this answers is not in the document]
sputn1ck: Added clarifying comment in c00fa27. The two schemas are intentionally different: encodeProofTLV binds to a specific pk_script, while encodeScriptScopeProofTLV binds to a signerPubKey to authorize queries for all scripts derived from that key. Server distinguishes them via the script_scope message type discriminator.

darepod/server.go
```diff
@@ -466,7 +454,7 @@ func (s *Server) run(ctx context.Context,
 		&rpcMailboxAdapter{RPCServer: s.rpcServer},
 	)
 
-	lis, err := net.Listen("tcp", s.cfg.RPC.ListenAddr)
+lis, err := net.Listen("tcp", s.cfg.RPC.ListenAddr)
```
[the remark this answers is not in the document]
sputn1ck: gofmt -d reports no issues on the current state.

lib/arkscript/nums.go
sputn1ck: NUMS key is now covered by golden vector tests in golden_test.go which verify byte-identical output key computation. The commit stack was reorganized — commit 1 introduces the full lib/arkscript foundation (including nums.go) before any migration.

lib/arkscript/node.go
sputn1ck: Agreed — the rebased stack now does this. Commit 1 (arkscript: add semantic policy foundation) introduces all node types, policy template, and validation before any consumer migration.

darepod/wallet_ops.go
sputn1ck: Added client-side policy validation in 56da2d0. Both BuildCustomTransferInputs (input path) and validateOutputPolicyTemplate (recipient path) now run ValidatePolicy() to check Ark invariants (collab leaf with operator key, CSV-gated exit leaves) before initiating the OOR transfer. Previously this was only enforced server-side at submit time.

darepod/rpc_server.go
Roasbeef: So we have VTXO locking already on the client side. We can thread it through here as well.

lib/arkscript/anchor.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/btcsuite/btcd/wire"
+)
+
+var (
+	// AnchorPkScript is a standard P2A script. This is a keyless output.
+	// This corresponds to address: bc1pfeessrawgf.
+	// The script path for this output requires no witness elements,
+	// meaning that it can be spent by anyone who provides the script and
+	// control block.
+	AnchorPkScript = []byte{
```
Roasbeef, line 14: There's a `txscript` constant for this.

litbot-9000:
@bhandras: review reminder
@roasbeef: review reminder

vtxo/manager.go
```diff
@@ -810,22 +810,25 @@ func clientVTXOToDescriptor(cv *round.ClientVTXO,
...
-		return fn.Err[*Descriptor](
-			fmt.Errorf("build tapscript: %w", err),
-		)
+	// Reconstruct the standard tapscript only when the round
+	// output uses the default Ark policy shape. Custom policies
+	// carry their semantic template and explicit spend paths
+	// instead of a derived standard tapscript.
+	var tapscript *waddrmgr.Tapscript
+	if len(cv.PkScript) > 0 {
+		desc := &Descriptor{PolicyTemplate: cv.PkScript}
```
bhandras: codex: This looks like `cv.PkScript` is being assigned into `Descriptor.PolicyTemplate`. `StandardTapScript()` decodes the semantic policy from `PolicyTemplate`, so this path never reconstructs the tapscript for standard round-created VTXOs and leaves `TapScript == nil` until a later DB reload heals it. Should this be `cv.PolicyTemplate` instead?

daemonrpc/daemon.proto
```diff
@@ -320,12 +370,19 @@ message Output {
         // delay, so the recipient gets a standard Ark VTXO.
         bytes pubkey = 3;
 
-        // pk_script is the raw output script for the recipient.
-        bytes pk_script = 4;
+        // policy_template is the serialized arkscript policy for the
```
bhandras, line 373: codex: `Output.policy_template` is documented here as a first-class recipient destination, but `SendVTXO` still resolves recipients through `resolveRecipientOutput()`, which only accepts pubkeys and taproot addresses and rejects this variant. That leaves the shared `Output` API advertising a destination type that works for `SendOOR` but not for directed in-round sends. Could we either teach `SendVTXO` this path too, or narrow the proto surface so the contract stays accurate?

round/transitions.go
```diff
@@ -1667,19 +1742,95 @@ func (s *PartialSigsSentState) transitionToForfeitCollection(
...
 	intents Intents, trees map[SignerKey]*tree.Tree,
 	roundID RoundID) ([]*ClientVTXO, error) {
 
 	vtxos := make([]*ClientVTXO, 0)
 	for _, req := range intents.VTXOs {
+		if req.OwnerKey.PubKey == nil {
+			continue
+		}
+
+		if req.OwnerKey.KeyLocator == (keychain.KeyLocator{}) {
```
bhandras: codex: `KeyLocator{}` is also a valid local owner when the wallet derived family/index `0/0`. This guard treats that real locator as "foreign" and skips materializing the confirmed VTXO, so the first key in a family can disappear from local balance tracking. Could we carry an explicit ownership bit here instead of using the zero locator as the sentinel?

vtxo/manager.go
bhandras: Fixed this by rebuilding the standard tapscript from the stored semantic
policy template instead of the concrete `pk_script` bytes.

I also added a regression test on the VTXO manager path so a round-created
standard VTXO materializes with its tapscript immediately rather than only
after a later DB reload.

daemonrpc/daemon.proto
bhandras: Good catch. I taught `SendVTXO`'s recipient resolution path to accept
standard `policy_template` destinations too.

That path now decodes the standard Ark VTXO template, derives the concrete
`pk_script`, and extracts the owner key needed for the collaborative spend.
I also tightened the proto comment to make it explicit that APIs needing a
collaborative owner key require the standard VTXO shape.

round/transitions.go
bhandras: Fixed by removing the zero-valued `KeyLocator{}` sentinel from this path.

Local ownership is now carried explicitly by the presence of an owner
descriptor, so a real locally derived `0/0` key is no longer treated as
foreign. I added a regression test that exercises the zero-locator owner
case to keep it from disappearing again.

bhandras (commented):
Draft replies for addressed comments

lib/types/boarding.go
```diff
@@ -108,35 +108,54 @@ type ForfeitRequest struct {
...
-	// preserved locally when the client owns the resulting VTXO.
-	OwnerKey keychain.KeyDescriptor
+	// ClientKey is the public key of the client used in the construction
+	// of the collaborative spend path of the VTXO.
+	ClientKey *btcec.PublicKey
 
+	// OwnerKey is the local key descriptor for the VTXO owner when this
+	// client controls the resulting output. This is local-only
+	// metadata and is not serialized onto the join-round wire.
+	OwnerKey keychain.KeyDescriptor
```
Roasbeef, line 150: Style nit: missing extra new line.

lib/tx/checkpoint/build.go
```diff
@@ -29,6 +29,11 @@ type Input struct {
 	// the closure system is canonical, higher layers should construct this
 	// leaf using closure helpers and pass the resulting script bytes here.
 	OwnerLeafScript []byte
+
+	// OwnerLeafPolicy is the semantic owner-leaf policy encoding that
+	// corresponds to OwnerLeafScript. When present, higher layers can
+	// reconstruct the owner leaf without decompiling raw script.
+	OwnerLeafPolicy []byte
```
Roasbeef, line 36: Non-blocking, but a type alias would be nice here to differentiate exactly what this is (the bytes encoded ast policy).

lib/tx/oor/taptree.go
```diff
@@ -38,72 +13,3 @@ func EncodeTapTree(leaves [][]byte) ([]byte, error) {
 func DecodeTapTree(data []byte) ([][]byte, error) {
 	return checkpoint.DecodeTapTree(data)
 }
-
-// PutTapTreePSBTInput stores an encoded tap tree blob into the given PSBT input
-// unknown fields, using TapTreePSBTKey.
-func PutTapTreePSBTInput(pkt *psbt.Packet, inputIndex int,
```
Roasbeef, line 44: Nice simplification here!

lib/tx/checkpoint/build.go
```diff
@@ -156,6 +156,7 @@ func BuildPSBT(policy arkscript.CheckpointPolicy, in Input) (*Result, error) {
 	}
 
 	pkt.Inputs[0].WitnessUtxo = in.SpentVTXO.Output
+	pkt.Outputs[0].TaprootTapTree = encodedTapTree
```
Roasbeef, line 159: 👍

rpc/roundpb/service.go
```diff
@@ -8,38 +8,39 @@ const (
 	// MethodJoinAck is the push event method name for
 	// ClientSuccessResp. The server sends this to acknowledge a
 	// client's JoinRoundRequest was accepted.
-	MethodJoinAck = "JoinAck"
+	MethodJoinAck = "ClientSuccessResp"
```
Roasbeef, line 11: This was wrong all along?

lib/types/codec.go
```diff
@@ -391,11 +386,9 @@ func decodeJoinAuthBoardingRequests(raw []byte) ([]*BoardingRequest, error) {
...
 func decodeJoinAuthBoardingRequest(raw []byte) (*BoardingRequest, error) {
 	var (
-		hash        []byte
-		index       uint64
-		clientKey   []byte
-		operatorKey []byte
-		exitDelay   uint64
+		hash   []byte
+		index  uint64
+		policy []byte
```
Roasbeef, line 391: 👍

Roasbeef (approved):
LGTM 🫛

db/boarding_wallet.go
bhandras: codex: Yep — this one is fully gone now. The old `MarkBoardingIntentsAdopted` path disappeared with the policy-first registration flow, so there is no remaining caller for it.

lib/arkscript/node.go
bhandras: codex: I took a closer look at this before folding anything into `#254`. `txscript.ScriptTemplate` turns byte params into pushdata, which is fine for building standalone script fragments from scratch, but not for the `arkscript` AST composition paths here. In `CSV.Script()` and `Condition.Script()` we need to splice already-compiled script fragments inline, and the template version changes the resulting script by inserting pushes. I verified that experimentally and it broke the golden-vector / AST composition tests, so I left the `ScriptBuilder` form in place here intentionally.

lib/arkscript/psbt.go
bhandras: codex: You were right here — my earlier “superseded” reply was too hand-wavy. I picked this up in follow-up PR `#254` by switching the Ark tap tree PSBT encoding over to `wire.ReadVarInt`/`wire.WriteVarInt`.

darepod/rpc_server.go
bhandras: codex: Agreed. I left the merged path as the smaller change, but I think the right follow-up is to thread wallet locking through here as well when a custom input overlaps wallet-managed VTXOs.

lib/arkscript/anchor.go
bhandras: codex: Fixed in follow-up PR `#254` by switching this over to the `txscript` constant instead of the raw opcode byte.

lib/types/boarding.go
bhandras: codex: Fixed in follow-up PR `#254` by adding the missing blank line.

lib/tx/checkpoint/build.go
bhandras: codex: Fixed in follow-up PR `#254` by introducing an `EncodedLeafPolicy` alias here so the semantic policy bytes are named more explicitly.

lib/tx/oor/taptree.go
bhandras: codex: Thanks — this part of the stack definitely got nicer once the semantic policy became the source of truth.

lib/tx/checkpoint/build.go
bhandras: codex: Thanks!

rpc/roundpb/service.go
bhandras: codex: Yeah, pretty much. It looks like this had drifted before the policy-first port, and `#212` was the first place we actually lined the service shape back up with the real wire behavior.

lib/types/codec.go
bhandras: codex: Thanks!

PR #214, opened 2026-03-20
oor: retry mark-inputs-spent on resume

bhandras:
## Summary

This is a focused client-side fix for the real retry classification bug we
surfaced while investigating the flaky `TestOORClientResumeAfterCoSign`
failure in `lightninglabs/darepo#184`.

When the OOR FSM reaches `AwaitingLocalVTXOUpdate`, the local persistence
handler routes `MarkInputsSpentRequest` through `CompleteSpend`. If that path
fails during shutdown or restart overlap, we want to preserve retry metadata
and resume the session later. Before this change, the handler returned a plain
error, so the actor checkpointed the session as `Failed` even though the
failure was transient.

## What changed

- wrap `CompleteSpend` failures in `NewRetryableOutboxError`
- keep the fallback direct-store path unchanged
- add a handler-level regression asserting mark-inputs-spent failures are
  tagged retryable
- add a durable restart regression proving the session stays in
  `AwaitingLocalVTXOUpdate`, preserves retry metadata across restart, and
  completes after explicit resume
- make the actor log message say it is mapping handler failures to outbox
  error events rather than claiming every failure is retryable

## Testing

- `go test ./oor -count=1`
- `make lint-changed-local base=origin/main`
- `python3 scripts/commit_message.py lint --range origin/main..HEAD`

### DIFF
--- a/oor/actor_resume_test.go
+++ b/oor/actor_resume_test.go
@@ -2,6 +2,7 @@ package oor
 
 import (
 	"context"
+	"fmt"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


oor/local_persistence_handler.go
```diff
@@ -168,8 +168,13 @@ func (h *LocalPersistenceOutboxHandler) handleMarkInputsSpent(
...
 		err := h.CompleteSpend(ctx, msg.Outpoints)
 		if err != nil {
-			return nil, fmt.Errorf("complete spend via "+
-				"manager: %w", err)
+			wrappedErr := fmt.Errorf(
+				"complete spend via manager: %w", err,
+			)
+
+			return nil, NewRetryableOutboxError(
+				wrappedErr, 0,
```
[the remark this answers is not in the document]
ellemouton: makes sense

oor/local_persistence_handler_test.go
```diff
@@ -601,6 +601,9 @@ func TestLocalPersistenceHandlerMarkInputsSpentCompleterError(t *testing.T) {
 	)
 	require.Error(t, err)
 	require.ErrorContains(t, err, "actor unavailable")
+	var retryErr *RetryableOutboxError
+	require.ErrorAs(t, err, &retryErr)
+	require.Zero(t, retryErr.RetryAfter)
```
ellemouton: 👍

oor/actor_resume_test.go
```diff
@@ -1339,3 +1340,175 @@ func TestOORClientActorDurableRestartAutoResume(t *testing.T) {
 		}
 	}, 5*time.Second, 50*time.Millisecond)
 }
+
+// TestOORClientActorDurableRestartRetriesMarkInputsSpent asserts that a
+// retryable local-spend completion failure survives restart without moving the
+// session into Failed.
+func TestOORClientActorDurableRestartRetriesMarkInputsSpent(t *testing.T) {
```
ellemouton, line 1347: could be cool just to add a bit of comments in this test 🙏

ellemouton (approved):
nice catch, thanks! 🙏

oor/actor_resume_test.go
bhandras: codex: Added a few targeted comments in the test around the three non-obvious phases: the first transient `CompleteSpend` failure, the fact that retry metadata is exported before the first actor stops, and the distinction between restart scheduling retry vs. explicit resume replaying the local spend completion.

oor/local_persistence_handler.go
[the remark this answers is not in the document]
bhandras: codex: Folded this in. I switched the call site to pass `defaultRetryDelay` directly so the backoff policy is obvious at the handler boundary, and updated the handler test to assert that explicit retry delay instead of the old zero-value placeholder.

PR #210, opened 2026-03-19
multi: split VTXO descriptors from tree signers

ellemouton:
## Summary

Previously the client reused the MuSig2 tree-signing key as the VTXO
owner key in saved descriptors. When operator terms rotated and a new
signing session was needed, the stored VTXO referenced a stale key
that could not spend the output locked to the actual owner key,
making the VTXO unspendable.

This violated the Ark spec (ARK-01, ARK-02) which **mandates** key
separation:

> Signing keys **MUST be different from VTXO ownership keys.**
> They **MUST use different derivation paths** to ensure separation.
> — ARK-01 "Key Separation", ARK-02 "VTXT Signing Keys"

The spec defines three distinct key roles:

| Spec key | Implementation | Role |
|----------|---------------|------|
| `P_v` (VTXO ownership key) | `OwnerKey` (family 44) | Committed to VTXO tapscript. Long-lived. |
| `P_o` (operator key) | `OperatorKey` | Collaborative spend cosigner. |
| `P_s` (signing key) | `SigningKey` (`KeyFamilyMultiSig`) | MuSig2 tree signing. Ephemeral per round. |

### Commit 1: `multi: separate VTXO owners from tree cosigners`

Core bug fix — stops storing the signing key as the VTXO owner key.

- Derive a dedicated owner key (family 44, `VTXOOwnerKeyFamily`)
  committed to the VTXO taproot output script.
- Change `VTXORequest.ClientKey` from `*btcec.PublicKey` to
  `keychain.KeyDescriptor` so the key locator survives persistence.
- Add `OwnsClientKey` flag so the client only persists VTXOs it
  owns, allowing co-signing for foreign recipients.
- Rename `buildClientVTXOs` → `buildOwnedClientVTXOs` with
  ownership filter.
- Extend DB schema with `client_key_family`, `client_key_index`,
  `owns_client_key` columns.

### Commit 2: `multi: restructure VTXO types and derive signing keys in FSM`

Structural cleanup and type safety improvements.

**Tree layer:**
- Rename `LeafDescriptor.CoSignerKey` → `SigningKey`.
- Add `OwnerKey`, `OperatorKey`, `ExitDelay`, `SigningKey` to
  `VTXODescriptor`.

**Rename `ClientKey` → `OwnerKey`** across all types:
- `VTXORequest`, `ClientVTXO`, `vtxo.Descriptor`,
  `wallet.VTXODescriptor`.
- `OwnsClientKey` → `IsOwner`.

[...]

PR #219, opened 2026-03-24
multi: implement in-round directed VTXO sends

ellemouton:
## Summary

Replaces #176.

Implements in-round directed sends through `SendVTXO`, built on the
key-separation foundation from #210.

Instead of the old approach that mixed directed send with the OOR spend
path, directed send uses the cooperative admission model:

- The wallet asks the VTXO manager to atomically select and reserve
  inputs for cooperative use (`SelectAndReserveForfeitRequest`)
- Selected VTXOs enter `PendingForfeitState`
- The wallet builds an `IntentPackage` with forfeits + recipient
  VTXOs + change VTXO
- The round actor registers the intent and immediately triggers
  registration with the server

### Key differences from #176

- **PkScript derived correctly**: recipient VTXO output scripts are
  derived from the VTXO descriptor (`ownerKey + operatorKey + exitDelay`),
  not plain BIP-86 P2TR. This is done in the wallet where operator terms
  are available.
- **Consistent OwnerKey semantics**: `SendRecipient` carries the raw
  x-only `OwnerKey`. Only pubkey destinations are supported (address
  and pk_script are rejected since they don't provide the untweaked key
  needed for VTXO construction).
- **Ownership filtering via #210**: recipient VTXOs set `IsOwner=false`
  so `buildOwnedClientVTXOs` skips them — the sender doesn't persist or
  manage recipient VTXOs.
- **No signing keys in wallet**: signing keys are left empty in
  `VTXORequest` — the round FSM derives them per #210.
- **Tracking deferred**: round key tracking, `TriggerRegistration` RPC,
  and temp key preservation from #176 are deferred to a follow-up issue.
  The send flow works without them.

## Commits

1. `actormsg: add SelectAndReserveForfeitRequest types`
2. `vtxo: add atomic forfeit selection to manager`
3. `wallet: add SendVTXOsRequest message types`
4. `wallet: implement handleSendVTXOs for directed send`
5. `daemonrpc: add change_amount_sat and selected_count to SendVTXOResponse`
6. `darepod: implement SendVTXO RPC for directed sends`
7. `round: auto-trigger registration after directed send intent`

## Testing

- `go test ./vtxo/ ./wallet/ ./darepod/` — all pass
- Server-side itest: `TestDirectedSendIntegration` (in companion PR)
  exercises the full flow: alice boards, sends 30k to bob, round
  completes, alice has correct change VTXO

## Companion PR

Server-side itest: lightninglabs/darepo#TBD

### DIFF
--- a/daemonrpc/daemon.pb.go
+++ b/daemonrpc/daemon.pb.go
@@ -1427,8 +1427,14 @@ type SendVTXOResponse struct {
 	// total_amount_sat is the total amount being sent (sum of
 	// recipients).
 	TotalAmountSat int64 `protobuf:"varint,3,opt,name=total_amount_sat,json=totalAmountSat,proto3" json:"total_amount_sat,omitempty"`
-	unknownFields  protoimpl.UnknownFields
-	sizeCache      protoimpl.SizeCache
+	// change_amount_sat is the change returned to the sender. Zero if
+	// the selected VTXOs exactly covered the total.
+	ChangeAmountSat int64 `protobuf:"varint,4,opt,name=change_amount_sat,json=changeAmountSat,proto3" json:"change_amount_sat,omitempty"`
+	// selected_count is the number of VTXOs selected as inputs for
+	// this send.
+	SelectedCount int32 `protobuf:"varint,5,opt,name=selected_count,json=selectedCount,proto3" json:"selected_count,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *SendVTXOResponse) Reset() {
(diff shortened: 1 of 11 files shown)


Roasbeef:
## Updated: Why LND passes but btcwallet/lwwallet fail

### Primary blocker: `FinalizePsbt` empty account name

All three wallet adapters pass `""` to `btcwallet.FinalizePsbt()`:
- `lwwallet/wallet.go:193`: `w.BtcWallet.FinalizePsbt(packet, "")`
- `darepod/server.go:3460`: `w.Wallet.BtcWallet.FinalizePsbt(packet, "")`

**LND works** because its gRPC `WalletKit.FinalizePsbt` internally maps `""` → `"default"` account. The direct btcwallet Go API treats `""` as a literal lookup and returns "account name '' not found".

**Fix**: Pass `"default"` (`lnwallet.DefaultAccountName`) instead of `""`.

### Secondary: Wallet UTXO sync timing

The tests call `h.FundClientWallet()` which sends coins + generates 6 blocks. LND sees them instantly (same bitcoind, sync). btcwallet (neutrino) and lwwallet (Esplora poller) have async sync — the UTXOs may not be visible when CPFP broadcast fires.

### Tertiary: Boarding address race (btcwallet only)

btcwallet's deferred wallet init means the boarding address may not be persisted before the test queries it.

Full investigation in `debug-progress.md` on the branch.

PR #225, opened 2026-03-26
multi: harden directed send and add receiver notification

### LINKED ISSUE #223: Add receiver-side notification for in-round (InRon) VTXO sends
## Context

From PR #176 [comment](https://github.com/lightninglabs/darepo-client/pull/176#issuecomment-4086784845):

For OOR sends, we have the indexer and the receiver can hit it to know when something is sent to them and to get all the checkpoints, etc. For **in-round sends**, we don't have anything analogous — the receiver currently has no way to know that they have a new InRon VTXO.

## Problem

When an in-round send completes, the **receiver** is not notified. They have no mechanism to:

1. Discover that a new InRon VTXO was created for them.
2. Fetch the relevant information (proofs, checkpoint data, etc.).
3. Import and recognize the VTXO as their own.

## Proposed Solution

Add plumbing so the server sends a push notification to the receiver for in-round sends, similar to how it already does for OOR sends. This should allow the receiver to:

- Be notified that a new InRon VTXO exists.
- Retrieve all necessary information about the VTXO.
- Import and recognize the VTXO in their local wallet state.

## References

- PR #176 (SendVTXO in-round send implementation)

ellemouton:
## Summary

Follow-up to #176. Addresses security hardening, ownership model
improvements, and receiver-side VTXO materialization for in-round
directed sends.

Closes #223.

### 1. Hardening (`74fc7c7`)

- Validate recipient amounts against 21M BTC cap with overflow-safe addition
- Replace `releaseAndFail` with `defer` + `committed` pattern so panics don't leak `PendingForfeitState` VTXOs
- Use `context.WithoutCancel` for deferred forfeit release (survives client disconnection)
- Cap recipients at 256 in RPC validation

### 2. Replace `IsOwner` with `OwnedScriptChecker` (`e4f5aa9`)

- Add `OwnedScriptChecker` interface to round FSM's `ClientEnvironment`
- VTXO ownership is now data-driven: checked against `owned_receive_scripts` DB store
- Remove `IsOwner` from `VTXOIntent` and `VTXORequest`
- Add `OwnedScriptRegistrar` interface on round actor — registers pkScripts for boarding, refresh, and change VTXOs at intent time
- Wire adapters into daemon using `OORArtifactPersistenceStore`
- Self-send works correctly (both recipient + change VTXOs persisted)

### 3. Extend `IncomingVTXOEvent` proto (`1d2ead4`)

- Add `pk_script`, `value_sat`, `round_id`, `batch_expiry_height`, `relative_expiry` fields
- For `VTXO_CREATED` events from confirmed rounds, carries enough data for the client to materialize without a follow-up indexer query

### 4. Receiver-side VTXO materialization (`074ce93`)

- New `IncomingVTXOHandler` actor processes `IncomingVTXOEvent` push notifications
- When server publishes a `VTXO_CREATED` event for a round leaf matching a registered receive script:
  1. Looks up pkScript in `owned_receive_scripts`
  2. Derives tapscript from owner key + operator key
  3. Builds `vtxo.Descriptor` from event metadata
  4. Persists via `VTXOPersistenceStore`
  5. Notifies VTXO manager via `VTXOsMaterializedNotification`
- Envelope route registered for `(arkrpc.ArkService, IncomingVTXO)`
- Same receive-script mechanism as OOR — recipient calls `NewOORReceiveScript` once and receives VTXOs from both OOR and in-round sends

## Testing

- All existing tests pass with race detector
- Server-side itest (companion PR): `TestDirectedSendIntegration` verifies bob receives his VTXO via `ListVTXOs`, `TestDirectedSendSelfSend` verifies both VTXOs

## Companion PR

Server-side itest + VTXOEventPublisher: lightninglabs/darepo#205


darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,181 @@
...
+	if h.cfg.VTXOStore != nil {
+		saveErr := h.cfg.VTXOStore.SaveVTXO(
+			lookupCtx, desc,
+		)
+		if saveErr != nil {
+			h.log.Warnf("Failed to save incoming "+
+				"VTXO %s: %v", outpoint, saveErr)
+
+			return fn.Ok[IncomingVTXOResp](nil)
+		}
```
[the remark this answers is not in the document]
ellemouton: ~cc: This is intentional best-effort design. The IncomingVTXOHandler processes server push notifications — returning an error here would only surface to the actor framework, not to any user-facing RPC. The handler will receive the event again on next connection/restart. A warning log is sufficient for operator visibility.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,181 @@
...
+	copy(outpoint.Hash[:], op.Txid)
+	outpoint.Index = op.Vout
+
+	h.log.Infof("Received IncomingVTXOEvent: outpoint=%s "+
+		"value=%d round=%s", outpoint, evt.ValueSat,
+		evt.RoundId)
+
+	if h.cfg.ScriptStore == nil {
+		return fn.Ok[IncomingVTXOResp](nil)
+	}
```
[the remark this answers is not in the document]
ellemouton: ~cc: This nil check is for optionality during startup — the handler can be registered before all stores are wired. It's a defensive guard, not a normal code path. Adding a warning here would be noisy since it fires for every event during early init.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// Look up the pkScript in owned receive scripts.
+	lookupCtx := context.WithoutCancel(ctx)
+	rec, err := h.cfg.ScriptStore.LookupOwnedReceiveScript(
+		lookupCtx, pkScript,
+	)
+	if err != nil {
+		// Not our script — ignore.
+		return fn.Ok[IncomingVTXOResp](nil)
```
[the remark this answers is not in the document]
ellemouton: ~cc: Agreed this is a design choice. In practice, LookupOwnedReceiveScript returns a 'not found' error for scripts we don't own — which is the common case since the server pushes events for ALL round leaves, not just ours. A follow-up could distinguish 'not found' from unexpected DB errors, but the current behavior is correct for the happy path.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	if rec.ClientKey.PubKey == nil {
+		return fn.Ok[IncomingVTXOResp](nil)
+	}
```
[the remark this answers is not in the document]
ellemouton: ~cc: Good callout. A nil PubKey here would indicate a corrupted DB record. Adding a warning log for this case is reasonable — will note as a follow-up improvement.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,181 @@
...
+
+	tapscript, err := scripts.VTXOTapScript(
+		rec.ClientKey.PubKey, operatorKey, exitDelay,
+	)
+	if err != nil {
+		h.log.Warnf("Failed to derive tapscript for "+
+			"incoming VTXO %s: %v", outpoint, err)
+
+		return fn.Ok[IncomingVTXOResp](nil)
+	}
```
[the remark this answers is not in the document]
ellemouton: ~cc: This is already logging a warning with the outpoint and error details. The handler is best-effort — the VTXO event will be re-delivered on reconnect. Returning an error to the actor framework wouldn't add recovery capability here.

wallet/wallet.go
```diff
@@ -1407,37 +1415,34 @@ func (a *Ark) handleSendVTXOs(ctx context.Context,
...
+			return
 		}
 
-		return fn.Err[WalletResp](primary)
-	}
+		releaseCtx := context.WithoutCancel(ctx)
+		_ = a.releaseManagerForfeitStrict(
+			releaseCtx, reservedOutpoints,
+		)
+	}()
```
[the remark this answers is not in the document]
ellemouton: ~cc: Agreed — this pattern ensures VTXOs are always released even on context cancellation or panic.

ellemouton:
@claude review this

ellemouton:
~cc: Addressed the review feedback:

**Fixed in this push:**
1. **Logging style** — converted all `Infof`/`Warnf` calls in `incoming_vtxo_handler.go` to structured `InfoS`/`WarnS` with `slog` KV pairs
2. **ValueSat overflow guard** (#6) — added `uint64(math.MaxInt64)` and `MaxSatoshi` check before casting to `btcutil.Amount`
3. **leaves[0] invariant** (#4) — changed `len == 0` guard to `len != 1` assertion with descriptive error
4. **Dry-run release error logging** (#3) — deferred closure now logs `WarnS` on release failure instead of silently discarding
5. **commitTxID comment** (#7) — added comment explaining why outpoint hash == commitment tx ID
6. **KeyLocator zero-value comment** (#5) — documented why Family=0/Index=0 is safe to skip (VTXOOwnerKeyFamily starts at 44)
7. **nil PubKey warning** (#2 partial) — added `WarnS` log for corrupted DB records with nil client pubkey

**Acknowledged as follow-ups:**
- LookupOwnedReceiveScript distinguishing "not found" from DB errors (#2)
- `OORArtifactPersistenceStore` duplication comment (minor nit)
- Dead nil-store guard in `owned_script_checker.go` (minor nit)

wallet/wallet.go
```diff
@@ -1407,37 +1416,41 @@ func (a *Ark) handleSendVTXOs(ctx context.Context,
...
+	// Ensure reserved VTXOs are released if we don't reach the
+	// successful registration at the end. Use a background
+	// context so cleanup survives client disconnection.
+	committed := false
+	defer func() {
+		if committed {
+			return
+		}
+
+		releaseCtx := context.WithoutCancel(ctx)
```
bhandras, line 1425: 👍

darepod/owned_script_checker.go
```diff
@@ -0,0 +1,69 @@
...
+
+// ownedScriptCheckerAdapter implements round.OwnedScriptChecker by
+// looking up pkScripts in the owned_receive_scripts persistence store.
+type ownedScriptCheckerAdapter struct {
+	store *db.OORArtifactPersistenceStore
+}
+
+var _ round.OwnedScriptChecker = (*ownedScriptCheckerAdapter)(nil)
+
+func (a *ownedScriptCheckerAdapter) IsOwnedScript(ctx context.Context,
```
bhandras, line 28: nit: missing godoc.

darepod/owned_script_checker.go
```diff
@@ -0,0 +1,69 @@
...
+	if a.store == nil {
+		return false
+	}
+
+	// Use a context that survives cancellation so the DB lookup
+	// completes even if the caller's context is being torn down
+	// (e.g., during round confirmation in a shutting-down FSM).
+	lookupCtx := context.WithoutCancel(ctx)
+
+	_, err := a.store.LookupOwnedReceiveScript(lookupCtx, pkScript)
```
bhandras, line 40: I don't think this is correct as is, we should instead surface the error and a bool flag if it returns the receive script?

darepod/owned_script_checker.go
```diff
@@ -0,0 +1,69 @@
...
+// persisting pkScripts in the owned_receive_scripts table.
+type ownedScriptRegistrarAdapter struct {
+	store       *db.OORArtifactPersistenceStore
+	operatorKey *btcec.PublicKey
+	exitDelay   uint32
+}
+
+var _ round.OwnedScriptRegistrar = (*ownedScriptRegistrarAdapter)(nil)
+
+func (a *ownedScriptRegistrarAdapter) RegisterOwnedScript(
```
bhandras, line 67: nit: missing godoc

darepod/server.go
```diff
@@ -1899,21 +1899,44 @@ func (s *Server) initRoundActor(ctx context.Context,
...
+	// Build the owned-script checker from the OOR artifact store.
+	// This allows the round FSM to determine which VTXOs belong
+	// to the local wallet by looking up registered receive scripts.
+	var scriptChecker round.OwnedScriptChecker
+	var scriptRegistrar round.OwnedScriptRegistrar
+	if s.db != nil {
+		oorStore := db.NewStore(
+			s.db.DB, s.db.Queries, s.db.Backend(),
+			s.log,
+		).NewOORArtifactStore(clock.NewDefaultClock())
```
bhandras: nit: perhaps we could store the clock in the server and inject it on creation of the server?

round/actor.go
```diff
@@ -1961,6 +1985,35 @@ func (a *RoundClientActor) handleRegisterIntent(ctx context.Context,
...
+
+			if vtxo.OwnerKey.KeyLocator == (keychain.KeyLocator{}) { //nolint:ll
+				continue
+			}
+
+			regErr := a.cfg.OwnedScriptRegistrar.RegisterOwnedScript( //nolint:ll
+				ctx, vtxo.PkScript, vtxo.OwnerKey,
+			)
+			if regErr != nil {
+				a.log.WarnS(ctx,
```
bhandras: Why only warn here?

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+	IncomingVTXOMsg, IncomingVTXOResp] {
+
+	return actor.NewServiceKey[IncomingVTXOMsg, IncomingVTXOResp](
+		incomingVTXOServiceKeyName,
+	)
+}
+
+// IncomingVTXOHandlerConfig holds the handler's dependencies.
+type IncomingVTXOHandlerConfig struct {
+	Log         fn.Option[btclog.Logger]
```
bhandras: nit: missing godocs.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+	evt := msg.Event
+	if evt == nil ||
+		evt.Type != arkrpc.VTXOEventType_VTXO_EVENT_TYPE_CREATED {
+
+		return fn.Ok[IncomingVTXOResp](nil)
+	}
+
+	op := evt.GetOutpoint()
+	if op == nil || len(op.Txid) != 32 {
+		return fn.Ok[IncomingVTXOResp](nil)
```
bhandras: Should we print a warning log here? This should not normally happen (same for empty pkscript).

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Receive processes IncomingVTXOEvent messages.
+func (h *IncomingVTXOHandler) Receive(ctx context.Context,
+	msg IncomingVTXOMsg) fn.Result[IncomingVTXOResp] {
+
+	evt := msg.Event
+	if evt == nil ||
+		evt.Type != arkrpc.VTXOEventType_VTXO_EVENT_TYPE_CREATED {
+
+		return fn.Ok[IncomingVTXOResp](nil)
```
bhandras: IIUC this always need to be `VTXOEventType_VTXO_EVENT_TYPE_CREATED` right? Maybe we could add some debug logging in case something else is coming through.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+		slog.String("outpoint", outpoint.String()),
+		slog.Uint64("value_sat", evt.ValueSat),
+		slog.String("round_id", evt.RoundId))
+
+	if h.cfg.ScriptStore == nil {
+		return fn.Ok[IncomingVTXOResp](nil)
+	}
+
+	// Look up the pkScript in owned receive scripts.
+	lookupCtx := context.WithoutCancel(ctx)
```
bhandras: Do we strictly need withoutcancel here?

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+	rec, err := h.cfg.ScriptStore.LookupOwnedReceiveScript(
+		lookupCtx, pkScript,
+	)
+	if err != nil {
```
bhandras, line 24: Perhaps we could specifically only skip not-found errors (or even just log) otherwise surface the error.

darepod/incoming_vtxo_handler.go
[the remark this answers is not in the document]
bhandras: I'd say this more signals a database error or schema inconsistency which we should surface.

darepod/rpc_server.go
```diff
@@ -656,14 +664,26 @@ func (r *RPCServer) SendVTXO(ctx context.Context,
 			)
 		}
 
-		if out.AmountSat <= 0 {
+		const maxAmountSat int64 = 21_000_000 * 1e8
```
bhandras: nit: `btcutil.MaxSatoshi`

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+
+	desc := &vtxo.Descriptor{
+		Outpoint:       outpoint,
+		Amount:         btcutil.Amount(evt.ValueSat),
+		PkScript:       pkScript,
+		OwnerKey:       rec.ClientKey,
+		OperatorKey:    operatorKey,
+		TapScript:      tapscript,
+		RoundID:        evt.RoundId,
+		CommitmentTxID: commitTxID,
```
bhandras: codex: `outpoint.Hash` here is the leaf txid, not the round commitment txid. `GetNonAnchorOutpoint` derives the outpoint from the leaf transaction, while the rest of the client uses `CommitmentTxID` for commitment-lineage metadata. Storing the leaf txid under `CommitmentTxID` will leave these descriptors with inconsistent ancestry information.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+		BatchExpiry:    evt.BatchExpiryHeight,
```
bhandras: codex: This assumes `evt.BatchExpiryHeight` is already an absolute height, but the companion server PR currently populates it from `Terms.SweepDelay` when publishing the event. The rest of the client treats `BatchExpiry` as an absolute block height, so a received VTXO will look immediately critical/expired on any chain height greater than the sweep delay. Can we either send `confirmation_height + sweep_delay` from the server or recompute it here?

darepod/incoming_vtxo_handler.go
bhandras, line 24: codex: Returning `Ok` on store/materialization failures makes this mailbox event effectively lossy. `IncomingVTXOEvent` is delivered via `Tell`, so once we ack it there is no retry path here; a transient DB issue during lookup/save would silently drop the receive. Could we distinguish `not found` from real store errors and surface the latter instead of swallowing them?

litbot-9000:

darepod/rpc_server.go
```diff
@@ -635,11 +635,19 @@ func (r *RPCServer) SendVTXO(ctx context.Context,
 		return nil, err
 	}
 
+	const maxRecipients = 256
```
Roasbeef, line 640: Should make an issue to tune this param later.

wallet/wallet.go
```diff
@@ -1361,10 +1361,19 @@ func (a *Ark) handleSendVTXOs(ctx context.Context,
 			))
 		}
 
-		if r.Amount <= 0 {
+		const maxAmount btcutil.Amount = 21_000_000 * 1e8
```
Roasbeef: Can use `btcutil.MaxSatoshi` here again.

round/transitions.go
```diff
@@ -1660,34 +1663,29 @@ func buildOwnedClientVTXOs(intents Intents, trees map[SignerKey]*tree.Tree,
...
-				OperatorKey: req.OperatorKey,
-				TreePath:    clientTree,
-				RoundID:     fn.Some(roundID),
-			})
+		if len(leaves) != 1 {
+			return nil, fmt.Errorf("expected exactly "+
+				"1 leaf for signing key, got %d",
+				len(leaves))
 		}
+		leaf := leaves[0]
```
Roasbeef, line 1682: We now assume a single leaf here?

arkrpc/indexer.proto
```diff
@@ -472,11 +472,30 @@ message ListVTXOEventsByScriptsResponse {
...
+// the VTXO without a follow-up indexer query.
 message IncomingVTXOEvent {
     uint64 event_id = 1;
     VTXOEventType type = 2;
     OutPoint outpoint = 3;
     VTXOStatus status = 4;
+
+    // pk_script is the VTXO output script. Present for CREATED events
+    // so the client can match against registered receive scripts.
+    bytes pk_script = 5;
```
Roasbeef, line 502: Perhaps we should use an enum/oneof here to add more structure to the proto? So we can more easily distinguiish if this is a new VTXO from an in round send, or an oor.

darepod/incoming_vtxo_handler.go
```diff
@@ -0,0 +1,208 @@
...
+	Log         fn.Option[btclog.Logger]
+	ScriptStore *db.OORArtifactPersistenceStore
+	VTXOStore   *db.VTXOPersistenceStore
+	VTXOManager actor.TellOnlyRef[vtxo.ManagerMsg]
+}
+
+// IncomingVTXOHandler materializes VTXOs from IncomingVTXOEvent
+// notifications pushed by the server's indexer after round
+// confirmation.
+type IncomingVTXOHandler struct {
```
Roasbeef: Let's move this to a new package? Similar to the other actor sub-ssytems, then we can write some dierct unit tests against it.

Roasbeef: Can go in the `vtxo` package perhaps

darepod/server.go
```diff
@@ -1163,10 +1167,47 @@ func (s *Server) buildEventRoutes() *serverconn.EventRouter {
...
+	s.registerIncomingVTXOEventRoute(router)
 
 	return router
 }
 
+// registerIncomingVTXOEventRoute registers the IncomingVTXO push event
+// route. When the server publishes a VTXO_CREATED event for a round
+// leaf matching a registered receive script, this route dispatches it
+// to the incoming VTXO handler actor for materialization.
+func (s *Server) registerIncomingVTXOEventRoute(
```
Roasbeef, line 1467: 👍

darepod/server.go
```diff
@@ -1163,10 +1167,47 @@ func (s *Server) buildEventRoutes() *serverconn.EventRouter {
...
+
+	vtxoKey := IncomingVTXOServiceKey()
+
+	serverconn.AddRoute(router, serverconn.EventRouteConfig[
+		IncomingVTXOMsg, IncomingVTXOResp,
+	]{
+		Service: arkServiceName,
+		Method:  MethodIncomingVTXO,
+		NewEvent: func() proto.Message {
+			return &arkrpc.IncomingVTXOEvent{}
```
Roasbeef, line 1478: Ah so we added this, but it wasn't actuall used for the oor flow?

Roasbeef:
## Review comment resolutions

Pushed a force-update with all fixups squashed into their target commits. Here's what changed:

### Commit 1: `multi: harden directed send validation and cleanup`
- Replaced raw `21_000_000 * 1e8` with `btcutil.MaxSatoshi` in both `rpc_server.go` and `wallet.go` (@Roasbeef, @bhandras)
- Added `TODO(#241)` for maxRecipients cap tuning (@Roasbeef) — issue created at #241

### Commit 2: `multi: replace IsOwner flag with OwnedScriptChecker`
- Added godocs to `ownedScriptCheckerAdapter.IsOwnedScript` and `ownedScriptRegistrarAdapter.RegisterOwnedScript` (@bhandras)
- Changed `IsOwnedScript` return type from `bool` to `fn.Result[bool]` — distinguishes not-found from real store errors (@bhandras)
- Updated `ownedScriptCheckerAdapter` to return `fn.Err` on real DB errors, `fn.Ok(false)` on `sql.ErrNoRows`
- Escalated `RegisterOwnedScript` failure from warn log to error return in `round/actor.go` (@bhandras)
- Stored `clock.Clock` in `Server` struct, replaced all 8 `clock.NewDefaultClock()` calls with `s.clk` (@bhandras)
- Expanded single-leaf-per-signer comment in `transitions.go` to explain the invariant (@Roasbeef)

### Commit 3: `arkrpc: add pkScript, value, and round metadata to IncomingVTXOEvent`
- Added `VTXOOrigin` enum (`IN_ROUND` / `OOR`) to distinguish VTXO creation paths (@Roasbeef)
- Added `commitment_txid` field for round commitment tx (distinct from leaf txid in outpoint) (@bhandras/codex)
- Clarified `batch_expiry_height` proto comment: server MUST send absolute height (`confirmation_height + sweep_delay`) (@bhandras/codex)

### Commit 4: `darepod: add IncomingVTXOEvent handler for round VTXO receipt`
- **Moved handler to `vtxo` package** with `OwnedScriptLookup` and `VTXOSaver` interfaces for testability (@Roasbeef)
- Added `ownedScriptLookupAdapter` in darepod to bridge `db.OORArtifactPersistenceStore` → `vtxo.OwnedScriptLookup`
- Added 4 unit tests: owned script, unowned script, non-CREATED event, nil event
- Added field-level godocs to `IncomingVTXOHandlerConfig` (@bhandras)
- Added debug logging for non-CREATED event types, warning for nil outpoint/pkScript (@bhandras)
- Removed unnecessary `context.WithoutCancel` (@bhandras)
- Distinguished `sql.ErrNoRows` (not-found) from real store errors — returns `fn.Err` on real failures (@bhandras/codex)
- Changed `SaveVTXO` failure from warn+Ok to `fn.Err` return (@bhandras)
- Used `commitment_txid` from event instead of leaf txid for `CommitmentTxID` field (@bhandras/codex)

### Not changed (by design)
- `server.go:1190` route registration: confirmed this is intentionally a new route for in-round sends, separate from OOR event routes (@Roasbeef)

Roasbeef (approved):
LGTM 🛩️

PR #227, opened 2026-03-27
arkscript: add semantic policy foundation for tapscript composition

Roasbeef:
## Summary

Introduce the `lib/arkscript` package — a typed AST for tapscript policy
composition that replaces the raw-script-bytes model in `lib/scripts`.

This is the **foundation layer** for the policy-first architecture. It provides
the core types and compilation pipeline without migrating any existing consumers.
Subsequent PRs will propagate `lib/arkscript` through the VTXO, wallet, DB, and
daemon layers.

### What changed and why

The old model stored VTXO outputs as derived artifacts from a fixed tuple
`(ownerKey, operatorKey, exitDelay)`. Every layer independently reconstructed
the taproot tree, creating drift risk and making non-standard scripts (like
6-leaf vHTLCs) impossible without ad-hoc escape hatches.

The new model stores the **semantic policy** as the source of truth:

```
PolicyTemplate { Leaves: []LeafTemplate }
    ↓  .Compile()
CompiledPolicy { InternalKey, Leaves, RootHash, MerkleProofs }
    ↓  .OutputKey()
pkScript (derived, never stored as primary)
```

### Package structure (commit-by-commit)

1. **AST node types and validation** — `Node` interface (sealed), three leaf
   types (`Multisig`, `CSV`, `Condition`), NUMS internal key, validation
   routines
2. **Taproot tree assembly and policy templates** — `CompiledPolicy`,
   `SpendInfo`, `PolicyLeaf` with `LeafRole` classification, `PolicyTemplate`
   with binary encode/decode, golden vector tests
3. **Spend path and composition** — `SpendPath` (wraps `SpendInfo` with
   condition witnesses), `ComposeWithSiblingRoot` for combining Ark policies
   with external roots (e.g. Taproot Assets)
4. **Standard VTXO policy** — `VTXOPolicy` (owner + operator + CSV),
   `DeriveSequence`/`DeriveLockTime` for tx-context derivation with
   property-based tests
5. **VHTLC and settlement** — `VHTLCPolicy` (6-leaf hash-time-locked
   contracts), settlement pair resolution for unilateral exit paths
6. **PSBT helpers and checkpoints** — Policy-aware PSBT encoding/decoding,
   witness construction, OOR checkpoint builder
7. **Reviewer guide** — Comprehensive documentation covering motivation, type
   system, security model, and concrete vHTLC walkthrough

### The Node AST

Three node types, sealed interface (only `lib/arkscript` types can implement):

| Type | Script encoding | Purpose |
|------|----------------|---------|
| `Multisig{Keys}` | `<k0> CHECKSIGVERIFY ... <kN-1> CHECKSIG` | N-of-N signature check |
| `CSV{Lock, Inner}` | `<inner> <lock> OP_CSV OP_DROP` | Relative timelock gate |
| `Condition{Predicate, Inner}` | `<predicate> <inner>` | Opaque script prefix (hashlock, CLTV, etc.) |

### Security invariants

- **Sealed interface** — Only the three types above can implement `Node`.
  External packages cannot introduce unexpected script shapes.
- **NUMS internal key** — All policies use a nothing-up-my-sleeve internal
  key (no key-path spend possible).
- **Canonical ordering** — Leaves are sorted lexicographically by script bytes
  for deterministic tree construction.
- **Operator validation** — `ValidateArkPolicy` walks the AST to enforce that
  every policy has a collab leaf (operator cosigns) and a CSV-gated exit leaf.

### Key design decisions

- **Predicate is opaque bytes, not a typed enum.** This lets us support future
  predicate types (timelocks, hashlocks, DLCs) without changing the AST.
  The cost is that predicate parsing (e.g. `ExtractAbsoluteLockTime`) must
  use the script tokenizer.
- **LeafRole on PolicyLeaf.** Leaves carry a `collab`/`exit` role tag for
  settlement pair matching. Role is assigned at construction time by the
  policy builder (VTXO, VHTLC), not derived from script analysis.
- **Template lives alongside CompiledPolicy.** Both `VTXOPolicy` and
  `VHTLCPolicy` carry a `Template *PolicyTemplate` so callers can access
  semantic operations (settlement pairs, validation) without recompilation.

## Test plan

- [x] All existing golden vector tests pass
- [x] Node compilation tests (Multisig, CSV, Condition combinations)
- [x] Tree construction and canonical ordering tests
- [x] Policy template encode/decode round-trip tests
- [x] VHTLC 6-leaf construction, role counts, named accessors
- [x] Settlement pair derivation for VTXO and VHTLC policies
- [x] Compose with sibling root (Taproot Assets integration path)
- [x] PSBT encoding/decoding tests
- [x] `DeriveSequence`/`DeriveLockTime` property-based tests with rapid
- [x] Validation tests (missing collab, missing exit, etc.)

See `docs/policy_arkscript_review_guide.md` for a detailed walkthrough of
the type system and security model.


lib/arkscript/node_test.go
```diff
@@ -0,0 +1,310 @@
...
+	require.NoError(t, err)
+
+	// Disassemble and verify structure.
+	dis, err := txscript.DisasmString(script)
+	require.NoError(t, err)
+	t.Logf("Multisig (checksig) script: %s", dis)
+
+	// Should be: <k1> CHECKSIGVERIFY <k2> CHECKSIG
+	// Script length: (1 + 32) + 1 + (1 + 32) + 1 = 68 bytes
+	require.Len(t, script, 68)
```
bhandras: I think we could actually assemble the script manually and check equality that way.

lib/arkscript/node_test.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	script, err := node.Script()
+	require.NoError(t, err)
+
+	dis, err := txscript.DisasmString(script)
+	require.NoError(t, err)
+	t.Logf("CSV script: %s", dis)
+
+	// Should be: <key> CHECKSIG <delay> CSV DROP
+	require.Contains(t, dis, "OP_CHECKSIG")
```
bhandras: Same here, instead of containment check let's assemble the expected script and check equality.

lib/arkscript/node_test.go
```diff
@@ -0,0 +1,310 @@
...
+	locktime := uint32(500000)
+
+	script, err := AbsoluteLockTimeCondition(locktime)
+	require.NoError(t, err)
+
+	dis, err := txscript.DisasmString(script)
+	require.NoError(t, err)
+	t.Logf("absolute locktime condition: %s", dis)
+
+	require.Contains(t, dis, "OP_CHECKLOCKTIMEVERIFY")
```
bhandras: Same comment as above.

lib/arkscript/node_test.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	script, err := node.Script()
+	require.NoError(t, err)
+
+	dis, err := txscript.DisasmString(script)
+	require.NoError(t, err)
+	t.Logf("Condition script: %s", dis)
+
+	// Should be: HASH160 <hash> EQUALVERIFY <key> CHECKSIG
+	require.Contains(t, dis, "OP_HASH160")
```
bhandras: Same. Actually note for codex/claude: pls fix all tests that check node scripts against real assembled expectations, and avoid using heuristics such as "Contains" which don't really test expectations fully.

lib/arkscript/nums.go
```diff
@@ -0,0 +1,41 @@
...
+	// ARKNUMSKey is a NUMS key (nothing up my sleeves number) that has
+	// no known private key. This was generated using the following script:
+	// https://github.com/lightninglabs/lightning-node-connect/tree/
+	// master/mailbox/numsgen, with the seed phrase "Ark Protocol NUMS".
+	ARKNUMSKey = mustParsePubKey(ARKNUMSHex)
+)
+
+// mustParsePubKey parses a hex encoded public key string into a public key
+// and panics if parsing fails.
+func mustParsePubKey(pubStr string) btcec.PublicKey {
```
bhandras, line 29: Let's also add `scripts/nums_test.go` here.

lib/arkscript/validate.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// ValidatePolicy checks that a set of AST nodes satisfies the minimum Ark
+// policy invariants inferred from the signer structure:
+//
+//  1. At least one operator-containing leaf.
+//  2. At least one non-operator leaf.
+//  3. Every non-operator leaf is CSV-gated.
+//  4. Smallest inferred exit delay >= MinExitDelay.
+func ValidatePolicy(nodes []Node, opts PolicyValidationOpts) error {
```
bhandras, line 30: I think this is a great start, but perhaps we could be more strict about policy validation? Containing a leaf doesn't necessarily mean that it satisfies vtxo requirements.

lib/arkscript/validate_test.go
```diff
@@ -0,0 +1,276 @@
...
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/lightninglabs/darepo-client/internal/testutils"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/stretchr/testify/require"
+)
+
+func TestContainsKey(t *testing.T) {
```
bhandras, line 13: nit: let's add godocs to these tests.

lib/arkscript/golden_test.go
```diff
@@ -0,0 +1,388 @@
...
+	ExitDelay        uint32
+
+	// Expected outputs (hex encoded).
+	OutputKeyHex      string
+	CollabScriptHex   string
+	TimeoutScriptHex  string
+	CollabControlHex  string
+	TimeoutControlHex string
+	RootHashHex       string
+	InternalKeyHex    string
```
bhandras, line 32: Instead we could check against the nums key in the tests.

lib/arkscript/golden_test.go
```diff
@@ -0,0 +1,388 @@
...
+			"4335306b07c3c169438461b5d4749884ec65" +
+			"768dc7e0689a1fa760492e409e1a71d25122" +
+			"742802d845c745eb466536e5a379",
+		TimeoutControlHex: "c1372f225b3caee8213096de3229e" +
+			"e4335306b07c3c169438461b5d4749884ec" +
+			"65c9bcb65c1015db6b44e5c2977606cf7a5" +
+			"d8f13d226531c6d01133d0810bed17b",
+		RootHashHex: "9556f4e03cfb047183ca430d888510c8" +
+			"1f85d2900f27e6edfbfbe3d257a18449",
+		InternalKeyHex: "02372f225b3caee8213096de3229ee43" +
```
bhandras, line 65: So these are all nums, let's just reuse that.

lib/arkscript/golden_test.go
```diff
@@ -0,0 +1,336 @@
...
+		)
+		collabControlBytes, err := collabControl.ToBytes()
+		require.NoError(t, err)
+
+		timeoutProof := tree.LeafMerkleProofs[1]
+		timeoutControl := timeoutProof.ToControlBlock(
+			tapscript.ControlBlock.InternalKey,
+		)
+		timeoutControlBytes, err := timeoutControl.ToBytes()
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
bhandras: Seems legit.

lib/arkscript/tree_test.go
```diff
@@ -0,0 +1,401 @@
...
+	require.Len(t, policy.merkleProofs[0], 0)
+
+	// Root hash should be the leaf hash.
+	expectedHash := leaves[0].Leaf.TapHash()
+	require.Equal(t, expectedHash[:], policy.RootHash)
+}
+
+// TestBuildTreeTwoLeaves tests tree construction with two leaves and verifies
+// it matches btcd's AssembleTaprootScriptTree.
+func TestBuildTreeTwoLeaves(t *testing.T) {
```
bhandras, line 115: Nice!

lib/arkscript/vhtlc.go
```diff
@@ -0,0 +1,429 @@
...
+// VHTLCPolicy represents a compiled vHTLC taproot policy.
+type VHTLCPolicy struct {
+	// Template is the semantic policy template for this vHTLC.
+	Template *PolicyTemplate
+
+	*CompiledPolicy
+
+	// Individual closures for easy access.
+	ClaimClosure                           Node
+	RefundClosure                          Node
```
bhandras, line 62: Can we add godocs pls.

lib/arkscript/checkpoint.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/lightningnetwork/lnd/input"
+)
+
+// CheckpointPolicy defines the parameters for constructing an OOR checkpoint
+// taproot tree.
+//
+// This is intentionally interface-first and minimal: it provides enough
+// information to deterministically derive a checkpoint output pkScript, while
+// allowing the underlying closure system to evolve later.
+type CheckpointPolicy struct {
```
bhandras, line 18: nit: Checkpoint policy could be lifted out to its own commit.

bhandras (commented):
Solid! There are a few finds by codex we need to address, but overall looks super nice!

lib/arkscript/settlement.go
```diff
@@ -0,0 +1,351 @@
...
+				matchIndex = forfeit.index
+				break
+			}
+		}
+
+		if matchIndex < 0 {
+			continue
+		}
+
+		authPath, err := spendPathForLeaf(
```
bhandras, line 157: codex: I think there’s a canonical-vs-template index bug here.

`PolicyTemplate.Compile()` sorts leaves before building the `CompiledPolicy`, but `SettlementPairsForParticipant()` records candidate indices while iterating `p.Leaves` in template order and then reuses those indices against `compiled` in `spendPathForLeaf(...)`.

I was able to reproduce a custom 4-leaf policy where, once template order differs from canonical script order, the returned `AuthLeafIndex` / `ForfeitLeafIndex` no longer line up with the derived witness scripts and control blocks.

Since `CompiledPolicy.SpendInfo()` is indexed in canonical order, I think this needs a semantic->canonical index mapping before deriving the spend paths and returning the leaf indices.

A nice regression test shape here might be:
- build two auth/forfeit settlement pairs whose scripts sort into a different order than the original `PolicyTemplate.Leaves`
- call `SettlementPairsForParticipant()`
- assert that each returned index points at the same witness script/control block that comes back from the paired `AuthPath` / `ForfeitPath`

That would catch exactly this template-order vs canonical-order mismatch.

lib/arkscript/vtxo.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, fmt.Errorf("vtxo: collab leaf is not multisig")
+	}
+
+	exitNode, ok := template.Leaves[1].Node.(*CSV)
+	if !ok {
+		return nil, fmt.Errorf("vtxo: exit leaf is not csv")
+	}
+
+	// Build the tree using the NUMS key as internal key.
+	policy, err := template.Compile()
```
bhandras, line 66: codex: I think there’s a role-classification hole here for standard VTXOs.

`PolicyTemplate.Compile()` builds `PolicyLeaf{Leaf: ...}` without setting `Role`, so the zero value (`LeafRoleCollab`) gets assigned to every compiled leaf. `NewVTXOPolicy()` then uses `template.Compile()` directly, which means the exit leaf is currently misclassified as `collab` as well.

I verified that with a quick repro on this branch: `NewVTXOPolicy(...).Leaves[0].Role == collab`, but `Leaves[1].Role == collab` too.

My instinct would be either:
- populate the roles explicitly in `NewVTXOPolicy()` after canonical script matching, or
- introduce an `unknown` / unset state so we don’t silently classify everything as `collab`.

lib/arkscript/spend_path.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	controlBlock, err := readVarBytes(r, "spend path control block")
+	if err != nil {
+		return nil, err
+	}
+
+	requiredSequence, err := wire.ReadVarInt(r, 0)
```
bhandras, line 159: codex: I think the decode path needs an upper-bound check before narrowing this varint back to `uint32`.

Right now malformed encodings can silently wrap values instead of being rejected. I was able to feed `DecodeSpendPath()` a payload with `requiredSequence = 2^32` and `requiredLockTime = 2^32 + 5`, and it decoded successfully as `seq=0` / `lock=5`. I hit the same behavior with `DecodeNode()` for a CSV lock as well.

Since these encodings can come back from persisted artifacts later in the stack, I think it’d be safer to mirror the explicit bounds check that `parseAbsoluteLockTimePredicate()` already does and reject anything above `math.MaxUint32`.

lib/arkscript/validate.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if opts.MinExitDelay > 0 && smallestCSVLock < opts.MinExitDelay {
+		return fmt.Errorf("exit delay %d is below minimum %d",
+			smallestCSVLock, opts.MinExitDelay)
+	}
+
+	return nil
+}
+
+// ContainsKey returns true if the AST contains a multisig node that
```
bhandras: codex: Small helper-semantics nit here: the comment says this checks whether the AST contains a multisig node that references the key, but the implementation is a recursive walk over typed AST nodes, not a raw script-level search.

That’s actually nicer than I originally thought, but I still think the wording could be a touch more explicit that this only looks through `Multisig`, `CSV`, and `Condition` structure and intentionally ignores opaque predicate bytes.

A tiny doc tweak here would make it easier for future readers to understand what “contains key” really means in this package.

lib/arkscript/standard_vtxo.go
```diff
@@ -0,0 +1,205 @@
...
+
+	if operatorKey == nil {
+		return nil, fmt.Errorf("collab leaf missing operator key")
+	}
+
+	if !containsKeyBytes(collab, schnorr.SerializePubKey(ownerKey)) {
+		return nil, fmt.Errorf("collab leaf missing owner key")
+	}
+
+	return &StandardVTXOParams{
```
bhandras, line 188: codex: I think `DecodeStandardVTXOParams()` should probably reject `exit.Lock == 0` as well.

`StandardVTXOTemplate()` explicitly enforces a non-zero exit delay, but the decoder currently accepts a 2-leaf `{collab, CSV(0, owner)}` template as a valid “standard” VTXO and `IsStandardVTXOTemplate()` returns true for it.

I verified that with a small repro on this branch. That makes the semantic builder/decoder pair disagree on what the standard policy shape actually is. If the intent is that “standard” means the shape produced by `StandardVTXOTemplate()`, I think the decode path should enforce the same non-zero delay invariant too.

lib/arkscript/compose.go
```diff
@@ -0,0 +1,140 @@
...
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+)
+
+// ComposedPolicy represents a policy that combines an Ark policy root with
```
bhandras, line 12: codex: One architectural question here: the docs and comments consistently describe these as Ark policies with an unspendable NUMS internal key, and `ComposeWithSiblingRoot()` even says the internal key must be unspendable.

But `BuildTree(...)` and `ComposeWithSiblingRoot(...)` are public and neither enforces that invariant. So today it’s possible to build/compose a `CompiledPolicy` with an arbitrary spendable internal key even though the surrounding API/docs read as if key-path spends are impossible by construction.

If the package is meant to stay Ark-specific, I wonder if this invariant should be encoded in the API (or at least asserted) rather than just documented. Otherwise I’d probably soften the docs/comments to make it clear this is a caller obligation rather than something the library guarantees.

lib/arkscript/tree.go
```diff
@@ -0,0 +1,258 @@
...
+
+// BuildTree constructs a canonical balanced binary taproot tree from the
+// ordered leaf list. This implements the algorithm from the RFC:
+//   - If n == 1: the root is the single leaf hash.
+//   - If n > 1: split left = leaves[0:n/2], right = leaves[n/2:n],
+//     compute L = BuildTree(left), R = BuildTree(right),
+//     root = TapBranchHash(min(L,R), max(L,R)).
+//
+// The function also computes merkle proofs for each leaf.
+func BuildTree(leaves []PolicyLeaf, internalKey *btcec.PublicKey) (
```
bhandras, line 211: codex: I think `BuildTree(...)` should reject a nil `internalKey` up front.

Right now it happily returns a `CompiledPolicy` with `InternalKey == nil`, and then callers hit a panic later when they do something normal like `OutputKey()`. I verified that with a tiny repro on this branch: `BuildTree(..., nil)` returns `err=nil`, but `policy.OutputKey()` panics with a nil deref.

Since `CompiledPolicy` is meant to be the fully compiled/usable form, this feels like a case where the constructor should enforce the invariant rather than leaving a latent panic behind for later callers.

lib/arkscript/tree.go
```diff
@@ -0,0 +1,258 @@
...
+	// Initialize merkle proofs (one per leaf, initially empty).
+	merkleProofs := make([][]chainhash.Hash, len(leaves))
+	for i := range merkleProofs {
+		merkleProofs[i] = make([]chainhash.Hash, 0)
+	}
+
+	// Build the tree recursively, collecting proofs along the way.
+	rootHash := buildTreeRecursive(leafHashes, 0, merkleProofs)
+
+	return &CompiledPolicy{
```
bhandras, line 263: codex: `BuildTree(...)` also seems to retain the caller’s `leaves` slice by reference instead of snapshotting it.

I was able to reproduce this shape locally:
- build a `CompiledPolicy` from a one-leaf slice
- mutate `leaves[0].Leaf.Script` after `BuildTree(...)` returns
- `policy.RootHash` stays the same, but `policy.SpendInfo(0).WitnessScript` changes to the mutated script

That leaves the “compiled” policy internally inconsistent: the stored leaf scripts can drift away from the precomputed leaf hashes and merkle proofs. I think `BuildTree(...)` should defensively copy the input leaves before storing them on `CompiledPolicy`.

lib/arkscript/vhtlc.go
```diff
@@ -0,0 +1,429 @@
...
+//     Checksig(receiver)
+//  5. UnilateralRefund (exit): CSV(delay) + Multisig([sender, receiver])
+//  6. UnilateralRefundWithoutReceiver (exit): CSV(delay) +
+//     Checksig(sender)
+func NewVHTLCPolicy(opts VHTLCOpts) (*VHTLCPolicy, error) {
+	if err := opts.validate(); err != nil {
+		return nil, err
+	}
+
+	claimPredicate, err := sha256Condition(opts.PreimageHash)
```
bhandras: codex: `ClaimPath(preimage)` might want to validate the preimage length up front.

The vHTLC success predicate built by `PaymentHash160Condition(...)` starts with `OP_SIZE 32 EQUALVERIFY`, so a non-32-byte preimage can never satisfy the leaf anyway. Right now `ClaimPath()` will happily package any `[]byte` into the witness and leave the failure until script execution time.

Since this helper is already policy-specific, it seems like a nicer API to reject impossible preimages early here.

docs/policy_arkscript_review_guide.md
```diff
@@ -0,0 +1,516 @@
...
+
+```
+ownerKey  + operatorKey  + exitDelay
+    ↓
+VTXOTapScript(ownerKey, operatorKey, exitDelay)
+    ↓
+pkScript (raw 34-byte P2TR output)
+```
+
+Every layer (tree builder, OOR runtime, DB, RPC, checkpoint signer) received
```
bhandras, line 21: codex: I think parts of this reviewer guide still reflect the older broader migration PR rather than this focused foundation slice.

A few examples:
- the “Integration” section points reviewers at files like `oor/checkpoint_sign.go`, server `oor/policy_helpers.go`, and `db/rounds_codec.go`
- the “Migration Boundaries” section says this PR deletes `lib/scripts/` and changes DB/wire formats

None of that is actually in `#227`, so I think trimming the guide back to what this PR really introduces would make it a much stronger review aid.

docs/policy_arkscript_review_guide.md
```diff
@@ -0,0 +1,516 @@
...
+All encodings use Bitcoin-standard varint/varBytes. Version bytes enable
+forward-compatible evolution.
+
+---
+
+## 7. Key Files to Review
+
+### Foundation (read first)
+
+| File | What to look for |
```
bhandras, line 443: codex: I think the documented `SpendPath` binary format is stale.

The guide currently shows something like:
- `[witness_script][control_block][condition_count]...`

But the code in `spend_path.go` actually encodes:
- `[version][condition_count][conditions...][witness_script][control_block][requiredSequence][requiredLockTime]`

Since this guide is supposed to help reviewers reason about the foundation pieces, I think it’s worth updating this section so it matches the actual wire shape introduced here.

lib/arkscript/psbt.go
```diff
@@ -0,0 +1,370 @@
...
+				"psbt: failed to write script for leaf %d: %w",
+				i, err,
+			)
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+// DecodeTapTree deserializes a PSBT tap tree encoding back into leaf data.
```
bhandras, line 125: codex: I think the binary decoders could use explicit upper bounds on these count varints before allocating slices.

Right now `DecodeTapTree()` trusts `leafCount`, `DecodePolicyTemplate()` trusts `leafCount`, `DecodeSpendPath()` trusts `conditionCount`, and `DecodeNode()` trusts `keyCount`, and all of them feed those values straight into slice allocation / loops.

Since these blobs can come back from persisted artifacts or remote inputs later in the stack, a malformed value here can force a huge allocation before we ever get to the per-item size checks. The per-field byte limits are good, but I think we still want count limits as well, or at least a sanity bound tied to expected Ark policy sizes.

lib/arkscript/validate.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// ExtractCSVDelay returns the CSV lock value if the AST contains a
+// CSV node, or 0 if it does not. If multiple CSV nodes are nested,
+// the outermost lock is returned.
+func ExtractCSVDelay(node Node) uint32 {
+	return extractCSVDelay(node)
+}
+
+// ScriptContainsKey performs a byte-level scan of a compiled script
```
bhandras: codex: I think this helper/comment pair still overclaims a bit.

The implementation is just `bytes.Contains(script, schnorr.SerializePubKey(key))`, so it can return true when the key bytes are only present as pushed data and never actually fed into `CHECKSIG`. A script like `<pubkey> OP_DROP` already returns true here.

If this is intentionally just a lightweight prefilter, I’d tighten the comment to say so explicitly. Otherwise it probably wants tokenizer / push-boundary checking rather than a raw substring match.

lib/arkscript/tree.go
```diff
@@ -0,0 +1,258 @@
...
+// PolicyLeaf represents a compiled tapscript leaf in canonical ordering.
+type PolicyLeaf struct {
+	// Leaf is the compiled tapscript leaf (script bytes + leaf version).
+	Leaf txscript.TapLeaf
+
+	// Role classifies this leaf as collaborative or unilateral exit.
+	Role LeafRole
+}
+
+// CompareTo returns -1 if this leaf sorts before other, 1 if after, and 0 if
```
bhandras, line 18: codex: One subtle correctness concern here: `PolicyLeaf` includes both script bytes and leaf version, but `CompareTo` / `SortLeaves` only compare raw script bytes.

That makes the advertised canonical order input-order-dependent if two leaves share the same script bytes but differ in leaf version. I repro’d this with two leaves built from the same script and versions `0xc0` vs `0xc2`: `SortLeaves` preserved caller order, so `BuildTree` returned different canonical indices depending on input order.

If this API is meant to be generic, I think the comparator needs to include leaf version (or just sort by tapleaf hash). If base-leaf-only is the real invariant, it may be worth enforcing or documenting that instead.

lib/arkscript/psbt.go
```diff
@@ -0,0 +1,370 @@
...
+	err := wire.WriteVarBytes(&buf, 0, preimage)
+	if err != nil {
+		return nil, fmt.Errorf("psbt: failed to write "+
+			"preimage: %w", err)
+	}
+
+	return buf.Bytes(), nil
+}
+
+// DecodeConditionWitness deserializes a hashlock preimage from PSBT storage.
```
bhandras, line 231: codex: Should this reject trailing bytes too?

Most of the new binary decoders in this package check that the reader is fully consumed, but `DecodeConditionWitness()` currently accepts `<varbytes preimage> + garbage` as valid and returns just the preimage.

Since this metadata is part of the durable PSBT artifact story too, I’d lean toward rejecting extra bytes here for consistency and canonicalness.

lib/arkscript/policy_template.go
```diff
@@ -0,0 +1,538 @@
...
+
+		if err := writeVarBytes(&buf, leafBytes); err != nil {
+			return nil, err
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+// ParticipantKeys returns the unique public keys referenced by the policy.
```
bhandras, line 209: codex: Small API-contract question: because this collects into a map and then ranges over it, the returned key order is nondeterministic across runs.

If callers are meant to treat `ParticipantKeys()` as an unordered set, it may be worth saying that in the godoc. Otherwise I think it should sort before returning so the result is stable for tests / RPC surfaces / later serialization helpers.

lib/arkscript/policy_template.go
```diff
@@ -0,0 +1,538 @@
...
+
+		return &Multisig{
+			Keys: keys,
+		}, nil
+
+	case nodeKindCSV:
+		return decodeLockedNode(r)
+
+	case nodeKindCondition:
+		predicate, err := readVarBytes(r, "condition predicate")
```
bhandras, line 453: codex: Small decode-consistency question here: `Condition.Script()` rejects empty predicates, but `DecodeNode()` currently accepts an empty predicate varbytes field and returns a `Condition` that can never compile successfully.

Since `EncodeNode()` can’t produce this form in the first place, I’d expect the decode path to reject it too so the binary format stays a representation of valid semantic nodes rather than “valid syntax, invalid semantics.”

docs/policy_arkscript_review_guide.md
```diff
@@ -0,0 +1,516 @@
...
+| `Hash160Condition`         | `HASH160 <h> EQUALVERIFY`                        |
+| `AbsoluteLockTimeCondition`| `<n> CHECKLOCKTIMEVERIFY DROP`                   |
+
+### Composition
+
+These three building blocks compose arbitrarily. A vHTLC claim leaf is:
+
+```go
+Condition{
+    Predicate: PaymentHash160Condition(hash),
```
bhandras: codex: I think this example is out of sync with the actual vHTLC implementation.

In `vhtlc.go`, both the cooperative claim leaf and the unilateral claim leaf are built from `sha256Condition(opts.PreimageHash)`, but the guide uses `PaymentHash160Condition(hash)` here.

Since this doc is meant to orient reviewers, I’d update the example to match the concrete policy builder we’re actually reviewing.

lib/arkscript/vhtlc.go
```diff
@@ -0,0 +1,429 @@
...
+	*SpendInfo, error) {
+
+	return p.SpendInfo(
+		p.unilateralRefundWithoutReceiverLeafIndex,
+	)
+}
+
+// ClaimPath returns a SpendPath for claiming via the hashlock leaf.
+// The preimage is included as a condition witness element.
+func (p *VHTLCPolicy) ClaimPath(
```
bhandras, line 241: codex: I think this helper should probably validate the *content* of the preimage too, not just package it.

Right now `ClaimPath(preimage)` will accept any byte slice and build a spend path, even if `SHA256(preimage)` doesn’t match the `PreimageHash` the policy was constructed with. That means callers can get a syntactically well-formed claim witness for a preimage that can never satisfy the leaf.

Since this API is policy-specific and already knows which claim path it is building, I’d expect it to reject mismatching preimages early rather than deferring the failure to script execution time.

lib/arkscript/psbt.go
```diff
@@ -0,0 +1,370 @@
...
+	Depth uint8
+
+	// LeafVersion is the BIP-341 leaf version (typically 0xc0).
+	LeafVersion uint8
+
+	// Script is the tapscript leaf script bytes.
+	Script []byte
+}
+
+// EncodeTapTree serializes a compiled policy's leaves into the PSBT tap tree
```
bhandras: codex: I think this comment/API overstates compatibility with the standard PSBT tap-tree field.

BIP-371’s `PSBT_OUT_TAP_TREE` encoding is a depth-first sequence of `(depth, leaf version, script)` tuples; it does **not** start with a leaf-count varint. This helper currently adds a leading count and assumes canonical leaf order on decode, so it won’t round-trip a standard `TaprootTapTree` blob as-is.

If the intent is to define an Ark-specific tap-tree encoding, I’d rename/document it that way. If the intent is to populate the standard PSBT output field later, I think the format here needs to match BIP-371 exactly.

lib/arkscript/tree.go
```diff
@@ -0,0 +1,258 @@
...
+	for i := 1; i < len(leaves); i++ {
+		for j := i; j > 0 && leaves[j].CompareTo(
+			&leaves[j-1],
+		) < 0; j-- {
+			leaves[j], leaves[j-1] = leaves[j-1], leaves[j]
+		}
+	}
+}
+
+// BuildTree constructs a canonical balanced binary taproot tree from the
```
bhandras, line 200: codex: Small API-contract concern: this doc says `BuildTree` constructs a *canonical* tree, but the function doesn’t sort or verify the input leaves first.

Right now canonicalness depends entirely on the caller passing leaves that are already in canonical order. That’s true for `PolicyTemplate.Compile()` and `NewVHTLCPolicy()`, but it isn’t enforced at this public API boundary.

I’d either sort internally, or make the contract more explicit in the name/docs (for example, “build from already-sorted leaves”) so external callers don’t accidentally get a non-canonical tree from the canonical builder.

lib/arkscript/vhtlc.go
```diff
@@ -0,0 +1,429 @@
...
+		)],
+		unilateralRefundWithoutReceiverLeafIndex: si[string(
+			unilateralRefundWithoutReceiverScript,
+		)],
+		orderedNodes: ordered,
+	}, nil
+}
+
+// ClaimSpendInfo returns the spend information for the Claim path.
+func (p *VHTLCPolicy) ClaimSpendInfo() (*SpendInfo, error) {
```
bhandras, line 194: codex: I think there’s a real tx-context bug across these VHTLC spend-info / path helpers.

All of these accessors call `p.SpendInfo(...)` directly, which returns the default `RequiredSequence = 0xffffffff` / `RequiredLockTime = 0`. But for several VHTLC leaves the AST clearly implies non-default tx context:
- `UnilateralClaim` should require `UnilateralClaimDelay`
- `UnilateralRefund` should require `UnilateralRefundDelay`
- `UnilateralRefundWithoutReceiver` should require `UnilateralRefundWithoutReceiverDelay`
- `RefundWithoutReceiver` should require `RequiredLockTime = RefundLocktime` and a non-final sequence

I verified this with a quick repro on this branch: `UnilateralClaimSpendInfo()` and `RefundWithoutReceiverSpendInfo()` both currently return the default `(0xffffffff, 0)`, and the corresponding `*Path()` helpers just package those defaults too.

So the derivation logic is tested in `TestVHTLCTxContextDerivation`, but it isn’t actually applied by these public VHTLC accessors. My instinct would be to route these through the same `DeriveSequence` / `DeriveLockTime` path that `settlement.go` uses, probably keyed off `orderedNodes[leafIndex]`.

lib/arkscript/vtxo.go
```diff
@@ -0,0 +1,213 @@
...
+	default:
+		// Custom leaves default to no requirements.
+		info.RequiredSequence = 0xffffffff
+		info.RequiredLockTime = 0
+	}
+
+	return info, nil
+}
+
+// NewVTXOPolicy creates and validates a standard VTXO policy from the given
```
bhandras, line 46: codex: Small wording nit here: the “Leaf 0 / Leaf 1” wording reads like it is describing the canonical indices on the resulting `VTXOPolicy`, but after canonical sorting that isn’t actually true on this branch.

For a standard VTXO, the exit script currently sorts before the collab script, so `exitLeafIndex == 0` and `collabLeafIndex == 1`. The code handles that correctly by mapping scripts back to indices, but I think the doc comment should avoid fixed ordinal language here (or explicitly say these are the semantic leaves *before* canonical sorting).

lib/arkscript/spend_helpers.go
```diff
@@ -0,0 +1,267 @@
...
+	witnessStack[0] = MaybeAppendSighash(
+		timeoutSig, signDesc.HashType,
+	)
+	witnessStack[1] = signDesc.WitnessScript
+	witnessStack[2] = signDesc.ControlBlock
+
+	return witnessStack, nil
+}
+
+// NewVTXOSpendInfoFromPolicy derives the spend information for the
```
bhandras, line 246: codex: Small API-contract question here: this helper switches on `leafIndex` as `0 => collab` / `1 => exit`, which is not the canonical leaf order of `VTXOPolicy.Leaves`.

On this branch, a standard VTXO currently canonicalizes as `exitLeafIndex == 0`, `collabLeafIndex == 1`. So if a caller reads “specified leaf index from a VTXOPolicy” literally and passes a canonical index, they get the opposite spend path from what `policy.Leaves[index]` represents.

If this is intentionally preserving the old `scripts.NewVTXOSpendInfo` semantic ordering, I think the doc/name should probably say that explicitly so readers don’t confuse it with canonical `VTXOPolicy` indices.

lib/arkscript/psbt.go
```diff
@@ -0,0 +1,370 @@
...
+	// Verify we consumed all data.
+	if r.Len() > 0 {
+		return nil, fmt.Errorf("psbt: %d extra bytes after tap tree",
+			r.Len())
+	}
+
+	return leaves, nil
+}
+
+// EncodeConditionWitness serializes a hashlock preimage for PSBT storage.
```
bhandras, line 208: codex: I think the encode/decode pair is slightly inconsistent here.

`EncodeConditionWitness()` will currently serialize any preimage length, but `DecodeConditionWitness()` later hard-limits it to 520 bytes via `wire.ReadVarBytes(...)`. So this encoder can produce values that its own decoder refuses to round-trip.

That may be fine if the real intent is “hashlock witnesses are always <= 520 bytes”, but then I’d expect the bound to be enforced on encode too so the API stays symmetric and callers fail earlier.

lib/arkscript/tree.go
```diff
@@ -0,0 +1,258 @@
...
+	// Build the control block for this leaf.
+	controlBlock, err := p.buildControlBlock(leafIndex)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build control block: %w", err)
+	}
+
+	leaf := &p.Leaves[leafIndex]
+
+	return &SpendInfo{
+		WitnessScript:    leaf.Leaf.Script,
```
bhandras: codex: `SpendInfo()` is returning the compiled leaf script by reference here (`WitnessScript: leaf.Leaf.Script`), so callers can mutate the compiled policy after construction just by editing the returned `SpendInfo`.

I reproduced this with a one-leaf policy:
- call `info, _ := policy.SpendInfo(0)`
- mutate `info.WitnessScript[0]`
- `policy.Leaves[0].Leaf.Script` changes too
- `policy.RootHash` stays unchanged

That leaves the compiled policy internally inconsistent in the same way as the `BuildTree` leaf-slice aliasing issue.

Would it make sense to clone the witness script on the way out here so `CompiledPolicy` stays immutable after construction?

lib/arkscript/anchor.go
```diff
@@ -0,0 +1,28 @@
...
+		txscript.OP_DATA_2,
+		// The P2A identifier.
+		0x4e, 0x73,
+	}
+)
+
+// AnchorOutput returns a zero value P2A output.
+func AnchorOutput() *wire.TxOut {
+	return &wire.TxOut{
+		PkScript: AnchorPkScript,
```
bhandras: codex: Small immutability nit here:

`AnchorOutput()` is returning the package-global `AnchorPkScript` slice by reference, so a caller can accidentally mutate the shared global script by editing `txOut.PkScript`.

Given the rest of this package is trying to make policy/spend artifacts durable and deterministic, I wonder if it would be safer to clone the script on the way out here.

Another reasonable option would be to keep the global as immutable-by-convention and make `AnchorOutput()` defensive.

lib/arkscript/settlement.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	if lock < 0 || lock > int64(^uint32(0)) {
+		return 0, false
+	}
+
+	return uint32(lock), true
+}
+
+// parseScriptNumToken decodes the current tokenizer token as a minimally
```
bhandras, line 317: codex: Small canonicalness concern here:

The comment says this decodes a *minimally encoded* script number, but the implementation accepts any pushed bytes and just interprets them as a script num.

That means `parseAbsoluteLockTimePredicate()` will also accept non-minimal CLTV predicates like:
- `<01 0x01> OP_CHECKLOCKTIMEVERIFY OP_DROP`

Even though `AbsoluteLockTimeCondition()` only ever emits the minimal form.

If the goal is to recognize the canonical predicate shape, I think this probably wants an explicit minimal-encoding check here — or a helper from `txscript` that enforces it.

lib/arkscript/compose.go
```diff
@@ -0,0 +1,140 @@
...
+		a, b = b, a
+	}
+
+	return *chainhash.TaggedHash(chainhash.TagTapBranch, a[:], b[:])
+}
+
+// PolicyRoot returns just the policy root hash from a compiled policy.
+// This can be used when the caller needs to provide the root for external
+// composition.
+func PolicyRoot(policy *CompiledPolicy) chainhash.Hash {
```
bhandras, line 133: codex: Minor API-footgun here:

`ComposeWithSiblingRoot()` checks for `nil` policy explicitly, but `PolicyRoot()` will just panic on `policy.RootHash` if a caller passes `nil`.

Since this is a small exported helper, I wonder if it would be worth either:
- guarding `nil` here as well, for example by returning the zero hash
- or documenting that `nil` is not allowed

That would make the convenience API a bit less surprising for callers.

lib/arkscript/psbt.go
```diff
@@ -0,0 +1,370 @@
...
+	default:
+		buf[0] = 0xff
+		binary.LittleEndian.PutUint64(buf[1:9], val)
+		_, err := w.Write(buf[:9])
+
+		return err
+	}
+}
+
+// readCompactSize reads a compact size integer from the reader.
```
bhandras, line 360: codex: One canonical-encoding nit here:

`writeCompactSize()` always emits the minimal CompactSize form, but `readCompactSize()` will accept non-minimal encodings like:
- `0xfd 01 00` for the value `1`

Since this package otherwise talks about stable/canonical binary encodings, I wonder if the decoder should reject non-canonical CompactSize forms too so there is only one valid byte representation for a given tap-tree blob.

lib/arkscript/compose.go
```diff
@@ -0,0 +1,140 @@
...
+		c.InternalKey, c.CombinedRoot[:],
+	)
+}
+
+// SpendInfo returns the spend information for a leaf in the Ark policy
+// subtree. The control block's merkle proof will include the external root
+// as an additional sibling at the root level.
+func (c *ComposedPolicy) SpendInfo(leafIndex int) (*SpendInfo, error) {
+	// Get the base spend info from the Ark policy.
+	info, err := c.ArkPolicy.SpendInfo(leafIndex)
```
bhandras, line 43: codex: I think this drops tx-context requirements for composed non-standard policies.

What I observed:
- `ComposedPolicy` only holds a `*CompiledPolicy`.
- `SpendInfo()` therefore starts from `c.ArkPolicy.SpendInfo(leafIndex)`, which uses the default `(RequiredSequence=0xffffffff, RequiredLockTime=0)` values.

I reproduced this with a composed vHTLC policy:
- the `RefundWithoutReceiver` leaf comes back with `seq=0xffffffff, lock=0`
- but it should require a non-final sequence plus the refund CLTV locktime
- the unilateral CSV leaves similarly lose their required delays

For plain VTXO this happens to be fine because there is a separate typed wrapper. But if `ComposeWithSiblingRoot()` is meant to be generic over `arkscript` policies, I think it probably needs a way to preserve the semantic node / tx-context mapping rather than only the compiled leaf data.

lib/arkscript/vhtlc.go
```diff
@@ -0,0 +1,429 @@
...
+	// Build index mapping by matching scripts.
+	claimScript, _ := claimClosure.Script()
+	refundScript, _ := refundClosure.Script()
+	refundWithoutReceiverScript, _ := refundWithoutReceiverClosure.Script()
+	unilateralClaimScript, _ := unilateralClaimClosure.Script()
+	unilateralRefundScript, _ := unilateralRefundClosure.Script()
+	unilateralRefundWithoutReceiverScript, _ :=
+		unilateralRefundWithoutReceiverClosure.Script()
+
+	scriptToIndex := make(map[string]int, len(leaves))
```
bhandras: codex: One extensibility sharp edge here: all of these maps key only on `string(script)`, but tapleaf identity is really `(leafVersion, script)` — or equivalently the tapleaf hash.

That is fine for the current v0 builders because:
- every leaf uses `BaseLeafVersion`
- the scripts happen to be unique

But if this foundation is meant to carry leaf-versioned policies forward, `scriptToIndex` / `stn` will collapse distinct leaves that share script bytes.

Using the full tapleaf identity here would make this more future-proof and would also line up better with the earlier `CompareTo()` / leaf-version discussion.

lib/arkscript/vhtlc.go
```diff
@@ -0,0 +1,429 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &SpendPath{SpendInfo: info}, nil
+}
+
+// OrderedNodes returns the AST nodes in canonical leaf order,
+// matching the Leaves slice indices.
+func (p *VHTLCPolicy) OrderedNodes() []Node {
```
bhandras: codex: Similar immutability concern as `BuildTree()` / `SpendInfo()`:

This returns the internal `orderedNodes` slice by reference, so a caller can mutate the policy's canonical node mapping in place.

Even if today most callers only inspect it, this feels like another easy way for a supposedly compiled/deterministic policy object to drift after construction.

Would it make sense to return a shallow copy of the slice here instead?

lib/arkscript/psbt.go
bhandras: codex: One migration-story concern here: this introduces a second Ark-private tap-tree encoding that is different from the existing checkpoint/OOR blob in:
- `lib/tx/checkpoint/taptree.go`
- `lib/tx/oor/taptree.go`

The older path uses the `waddrmgr.Tapscript` TLV leaf encoding, while this helper uses:
- `[count][depth][leafVersion][scriptLen][script]...`

I don't think either format is inherently wrong. But since both are called “tap tree encoding” and both seem aimed at PSBT/durable-artifact use, it would really help to state explicitly whether this new format is meant to replace the older TLV blob or coexist with it during migration.

Right now it feels easy for later code to accidentally mix the two.

docs/policy_arkscript_review_guide.md
```diff
@@ -0,0 +1,516 @@
...
+    Amount:         100_000,
+    PolicyTemplate: vhtlcPolicy.Encode(),  // semantic, not raw pkScript
+}
+```
+
+The server validates this policy via `ValidatePolicy()` before accepting it
+into the round/OOR flow — confirming operator key presence on collab leaves and
+CSV gating on exit leaves.
+
+The checkpoint transaction commits to this output. The output's tap tree is
```
bhandras, line 302: codex: Small historical wording nit:

This sentence reads a bit like PSBT-based crash recovery is being introduced here, but the client already treated the Ark/checkpoint PSBT artifacts as the durable resume unit before `arkscript`.

My read is that the improvement in this PR is more specifically:
- the checkpoint PSBT becomes more self-contained
- it carries the tap-tree metadata directly
- resume no longer depends on the older sidecar / later-copying story in the same way

Tightening that wording could make the migration story easier to follow for reviewers.

lib/arkscript/compose_test.go
```diff
@@ -0,0 +1,289 @@
...
+	// Output keys should differ.
+	originalKey := policy.OutputKey()
+	composedKey := composed.OutputKey()
+	require.NotEqual(t,
+		originalKey.SerializeCompressed(),
+		composedKey.SerializeCompressed(),
+		"composed output key should differ from original")
+}
+
+// TestComposeWithSiblingRootSpendInfo tests that spend info includes the
```
bhandras, line 67: codex: I think the current composition tests are a little too VTXO-specific to exercise the generic API contract.

VTXO happens to work here because there is a separate typed wrapper that re-derives tx-context. But `ComposeWithSiblingRoot()` is exposed for arbitrary compiled policies.

A nice regression test to add alongside this one would be a composed vHTLC leaf that carries CSV / CLTV requirements (for example `RefundWithoutReceiver`), and then assert that the composed spend info preserves:
- the control-block extension
- the original `RequiredSequence`
- the original `RequiredLockTime` requirements

lib/arkscript/policy_template.go
```diff
@@ -0,0 +1,538 @@
...
+		if _, err := p.Leaves[i].Script(); err != nil {
+			return fmt.Errorf("compile policy leaf %d: %w", i, err)
+		}
+		nodes = append(nodes, p.Leaves[i].Node)
+	}
+
+	return ValidatePolicy(nodes, opts)
+}
+
+// Encode serializes the semantic policy template into a stable binary
```
bhandras: codex: I think the word `stable` may be overstating what this encoding guarantees today.

What the code does now:
- `Compile()` canonicalizes leaf order before building the tree
- `Encode()` preserves the caller's template order

I verified this with a standard 2-leaf VTXO template:
- swapping the two leaves gives the same compiled root hash / `pkScript`
- but different encoded `PolicyTemplate` bytes

If the intent is “one canonical binary representation per semantic policy,” this probably wants canonical leaf ordering at encode time too.

Otherwise, I think the docs should say more explicitly that the encoding preserves author order rather than semantic equivalence.

lib/arkscript/policy_template.go
```diff
@@ -0,0 +1,538 @@
...
+	nodeKindMultisig nodeKind = iota + 1
+
+	// nodeKindCSV encodes a CSV-gated child node.
+	nodeKindCSV
+
+	// nodeKindCondition encodes a generic predicate-gated child node.
+	nodeKindCondition
+)
+
+// LeafTemplate is the semantic representation of one policy leaf.
```
bhandras, line 40: codex: Architectural precision nit:

`LeafTemplate` currently captures only the AST node, not the tapleaf version, and every compile path turns it into `txscript.NewBaseTapLeaf(script)`.

So this semantic template is really a *base-leaf-version / v0* policy format today, not a fully general tapleaf representation yet.

That constraint is totally fine if intentional, but I think it would help to say so explicitly somewhere around this type (or in the reviewer guide), especially since a few other comments/docs in the PR talk in broader “policy foundation” terms.

lib/arkscript/policy_template_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	decoded, err := DecodeLeafTemplate(encoded)
+	require.NoError(t, err)
+
+	originalScript, err := leaf.Script()
+	require.NoError(t, err)
+
+	decodedScript, err := decoded.Script()
+	require.NoError(t, err)
+
```
bhandras, line 36: codex: This round-trip test is good, but I think the wording here overstates what is actually being pinned down.

Right now it proves:
- `decode(encode(x))` compiles to the same root / output key

But it does *not* prove:
- the encoding is canonical across semantically equivalent templates

In fact, swapping the two leaves of a standard VTXO template gives:
- the same compiled root / `pkScript`
- different `PolicyTemplate.Encode()` bytes

It may be worth either softening the comment or adding a second test that makes the intended contract explicit:
- canonical semantic encoding
- vs preserved author order

lib/arkscript/vhtlc_test.go
```diff
@@ -0,0 +1,365 @@
...
+
+		// Control block should be parseable.
+		ctrlBlock, err := txscript.ParseControlBlock(
+			info.ControlBlock,
+		)
+		require.NoError(t, err,
+			"failed to parse control block for leaf %d", i)
+		require.Equal(t, txscript.BaseLeafVersion,
+			ctrlBlock.LeafVersion)
+	}
```
bhandras, line 120: codex: I think this test is missing the most important part of the named-accessor contract: the tx-context requirements.

Right now it only checks that each accessor returns some witness script / control block, so it would still pass even though:
- `RefundWithoutReceiverSpendInfo()`
- `UnilateralClaimSpendInfo()`

currently drop their CLTV / CSV requirements and come back with the default:
- `RequiredSequence=0xffffffff`
- `RequiredLockTime=0`

It would be great to assert the expected sequence / locktime on the named accessors directly here so the regression is pinned at the API boundary callers actually use.

lib/arkscript/tree_test.go
```diff
@@ -0,0 +1,401 @@
...
+	"encoding/hex"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/lightninglabs/darepo-client/internal/testutils"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPolicyLeafCompareTo tests the canonical comparison of policy leaves.
```
bhandras, line 14: codex: I think the tree tests are still missing the exact edge case behind the `CompareTo()` / `SortLeaves()` leaf-version bug:
- two leaves with identical script bytes
- but different leaf versions

Right now all of the sorting tests use `NewBaseTapLeaf(...)`, so they never exercise whether canonical order remains stable once version differs.

A tiny regression case with the same script under `0xc0` and `0xc2` would pin down whether canonical ordering should compare:
- `(leafVersion, script)`
- or the full tapleaf hash

rather than raw script bytes alone.

bhandras:
codex: Thanks again for splitting this out — the focused PR is much easier to reason about, and after a fairly deep pass I think the direction is good.

My current take is:
- the standard `VTXO` / `Checkpoint` constructions look preserved
- the policy-first foundation still feels like the right long-term shape
- but I do not think this is merge-close yet, because there are still a few foundation-level correctness issues that later migrations would build on top of

**Blockers**
1. `SettlementPairsForParticipant()` needs to use canonical compiled indices consistently.
Line thread: `lib/arkscript/settlement.go`
Context: the current implementation mixes template-order indices with compiled-order spend-path derivation.
Risk: the reported leaf index can drift from the returned witness script / control block.

2. vHTLC named spend-info accessors need to preserve tx-context.
Line thread: `lib/arkscript/vhtlc.go`
Context: helpers like `RefundWithoutReceiverSpendInfo()` / `UnilateralClaimSpendInfo()` currently return default `RequiredSequence` / `RequiredLockTime` instead of the semantic leaf requirements.
Risk: callers using the typed API can build invalid transactions.

3. `ComposedPolicy.SpendInfo()` needs to preserve tx-context for non-standard policies.
Line thread: `lib/arkscript/compose.go`
Context: composition keeps compiled leaf data, but drops semantic CSV / CLTV requirements.
Risk: generic composition is currently wrong for policies like vHTLC.

4. Compiled objects should not be trivially mutable after construction.
Line threads:
- `lib/arkscript/tree.go`
- `lib/arkscript/vhtlc.go`
Context: `BuildTree()` aliases caller leaf slices, `CompiledPolicy.SpendInfo()` returns witness script by reference, and `OrderedNodes()` returns the internal slice by reference.
Risk: compiled/deterministic policy objects can drift after construction and become internally inconsistent.

5. Canonical ordering needs to account for leaf version, or the PR should explicitly constrain itself to base-leaf-version / v0 everywhere.
Line thread: `lib/arkscript/tree.go`
Context: `CompareTo()` / `SortLeaves()` currently compare only raw script bytes.
Risk: canonical ordering becomes input-order-dependent if two leaves share script bytes but differ in leaf version.

**Important**
1. The tap-tree encoding migration story should be stated explicitly.
Line thread: `lib/arkscript/psbt.go`
There are now two Ark-private tap-tree encodings in the codebase, and it is not yet clear whether the new one is intended to replace the older checkpoint/OOR TLV blob or coexist with it during migration.

2. The contract for `PolicyTemplate.Encode()` should be clarified.
Line thread: `lib/arkscript/policy_template.go`
Right now it is stable for a given template order, but not canonical across semantically equivalent policies.

3. Some decoder paths are looser than the package’s “stable/canonical encoding” story suggests.
Line threads:
- `lib/arkscript/settlement.go`
- `lib/arkscript/psbt.go`
Examples: non-minimal script numbers and non-minimal CompactSize encodings are accepted.

4. The docs should be tightened to match the actual guarantees.
Line threads:
- `docs/policy_arkscript_review_guide.md`
- `lib/arkscript/policy_template.go`
Examples: crash-recovery wording and the current base-leaf-version / v0 scope of the semantic template.

**Nits / Good Follow-Ups**
- `AnchorOutput()` could defensively clone its script.
- `PolicyRoot()` could guard `nil` or document that it does not.
- A few regression tests would make the next revision much easier to trust:
  - same-script / different-leaf-version sorting
  - composed vHTLC tx-context preservation
  - named vHTLC accessor tx-context preservation
  - semantic-equivalence vs encoded-byte-equivalence for `PolicyTemplate`

**Bottom line**
I’d summarize the current state as:
- architecture direction: good
- standard script compatibility: looks good
- main remaining risk: semantic correctness of the new foundational helpers

So I think this PR is promising, but not quite there yet. If the tx-context bugs, settlement-index bug, immutability leaks, and leaf-version canonicalness story get tightened up, I think it gets *much* closer very quickly.

lib/arkscript/validate.go
sputn1ck: not sure what you mean, as it checks the smallest found csv node against the passed in opts(which would be forwarded by the server to the client)

sputn1ck: The validation func to me seems both pretty complete for both client and server(client wants it's csv path, server wants the cosigner path)

lib/arkscript/settlement.go
sputn1ck: Fixed. Added a `templateToCanonical` mapping that resolves each template leaf's script to its canonical position in the compiled tree. All `spendPathForLeaf` calls and returned `AuthLeafIndex`/`ForfeitLeafIndex` now use canonical indices.

lib/arkscript/vtxo.go
sputn1ck: Fixed — went further: removed `LeafRole` entirely. Leaf classification is derivable from `ContainsKey(node, operatorKey)`, so storing it was redundant.

lib/arkscript/vhtlc.go
sputn1ck: Fixed. Named accessors now route through `CompiledPolicy.SpendInfoForNode(node)` which derives `RequiredSequence`/`RequiredLockTime` from the AST. Test added asserting correct values for all 6 paths.

lib/arkscript/tree.go
sputn1ck: Fixed. `BuildTree` rejects nil `internalKey` upfront. Went further: enforces the key must be `ARKNUMSKey` — every call site already passed NUMS, so this makes the invariant explicit.

lib/arkscript/tree.go
sputn1ck: Fixed. Input leaves are defensively deep-copied (script bytes + leaf version) before storing on `CompiledPolicy`.

lib/arkscript/anchor.go
sputn1ck: Fixed. `SpendInfo` now returns `bytes.Clone(leaf.Leaf.Script)`.

lib/arkscript/spend_path.go
sputn1ck: Fixed. Added explicit `> math.MaxUint32` bounds checks for `requiredSequence`, `requiredLockTime`, and CSV lock.

lib/arkscript/standard_vtxo.go
sputn1ck: Fixed. Added `exit.Lock == 0` rejection to match `StandardVTXOTemplate`'s non-zero invariant.

lib/arkscript/psbt.go
sputn1ck: Fixed. Added sanity bounds: `maxTapTreeLeaves=256`, `maxConditions=64`, `maxMultisigKeys=64`.

lib/arkscript/psbt.go
sputn1ck: Fixed. Added `r.Len() != 0` trailing bytes check.

lib/arkscript/psbt.go
sputn1ck: Fixed. `EncodeConditionWitness` now enforces the same `maxPreimageLen=520` limit as decode.

lib/arkscript/psbt.go
sputn1ck: Fixed. Each discriminant case now rejects values that could have been encoded with a smaller discriminant.

lib/arkscript/psbt.go
sputn1ck: Fixed. Updated doc to explicitly state this is an Ark-specific format, not BIP-371.

sputn1ck: Acknowledged. The existing TLV encoding in `lib/tx/` is for the current pipeline. The new format is for the policy-aware pipeline. Integration PR will document migration.

lib/arkscript/policy_template.go
sputn1ck: Fixed. Added empty predicate rejection in the decode path.

lib/arkscript/policy_template.go
sputn1ck: Fixed. Result is now sorted by x-only key bytes before returning.

lib/arkscript/tree.go
sputn1ck: Fixed. `CompareTo` now compares leaf version first, then script bytes. Test added with `0xc0` vs `0xc2` leaves.

lib/arkscript/compose.go
sputn1ck: Fixed. Returns zero hash for nil input.

lib/arkscript/settlement.go
sputn1ck: Fixed. `AnchorOutput` now returns a defensive copy.

lib/arkscript/vhtlc.go
sputn1ck: Fixed. `ClaimPath` now rejects preimages != 32 bytes (matching `OP_SIZE 32 EQUALVERIFY`).

lib/arkscript/vhtlc.go
sputn1ck: We validate 32-byte length (matching `OP_SIZE 32 EQUALVERIFY`). Full SHA256 validation would require storing the original hash. The script VM catches mismatches at execution time, so the length check is sufficient for early rejection.

lib/arkscript/vhtlc.go
sputn1ck: Fixed — went further: removed `OrderedNodes()` entirely. Callers now use `Template.Leaves[i].Node` directly.

lib/arkscript/validate.go
sputn1ck: Fixed. Updated godoc to accurately describe the typed AST walk over Multisig/CSV/Condition nodes.

lib/arkscript/validate.go
sputn1ck: Fixed. Tightened doc to say "lightweight prefilter" — should be combined with script VM execution.

lib/arkscript/compose.go
sputn1ck: Fixed. `BuildTree` now rejects any internal key that is not `ARKNUMSKey`. Invariant is explicit in the API.

lib/arkscript/compose.go
sputn1ck: Acknowledged. For VTXO the typed wrapper re-derives context via `SpendInfoForNode`. For vHTLC callers use it directly. Composed path is currently only used for TAP integration with standard VTXOs.

lib/arkscript/node_test.go
sputn1ck: Fixed. All node tests now assemble expected scripts manually using `txscript.NewScriptBuilder` and assert byte equality.

lib/arkscript/node_test.go
sputn1ck: Fixed — same approach, full script assembly + equality check.

lib/arkscript/node_test.go
sputn1ck: Fixed.

lib/arkscript/node_test.go
sputn1ck: Fixed across all node tests — no more Contains heuristics.

lib/arkscript/nums.go
sputn1ck: Done. Added `nums_test.go` with 3 tests: validity, documented hex match, determinism.

lib/arkscript/golden_test.go
sputn1ck: Fixed. Internal key checks now compare against `ARKNUMSKey.SerializeCompressed()` directly.

lib/arkscript/golden_test.go
sputn1ck: Fixed — all hardcoded NUMS hex replaced with the constant.

lib/arkscript/tree_test.go
sputn1ck: Fixed. Added `TestCompareToWithDifferentLeafVersions` with `0xc0` vs `0xc2` leaves. Also added nil/non-NUMS internal key tests and defensive copy tests.

lib/arkscript/vhtlc_test.go
sputn1ck: Fixed. `TestVHTLCNamedAccessors` now asserts `RequiredSequence` and `RequiredLockTime` for all 6 accessors.

lib/arkscript/validate_test.go
sputn1ck: Fixed. All test functions now have godoc comments.

lib/arkscript/validate.go
sputn1ck: The current validation covers the structural Ark invariants (operator presence, CSV gating, minimum delay). Semantic completeness beyond that belongs to the specific policy type constructor (e.g. `NewVHTLCPolicy` validates all fields).

sputn1ck: Agreed — the validation checks the smallest CSV node against the opts, which covers both client and server needs.

sputn1ck: Agreed, the validation is symmetric for both sides.

docs/policy_arkscript_review_guide.md
sputn1ck: Fixed. Removed Integration file table, trimmed Migration Boundaries to match foundation-only scope.

docs/policy_arkscript_review_guide.md
sputn1ck: Fixed. Updated to match actual wire shape: `version, condition_count, conditions..., witness_script, control_block, required_sequence, required_locktime`.

docs/policy_arkscript_review_guide.md
sputn1ck: Fixed. Updated example to use `sha256Condition` and added it to the helper table.

docs/policy_arkscript_review_guide.md
sputn1ck: Fixed. Clarified the improvement is that checkpoint PSBT now carries tap-tree metadata directly.

lib/arkscript/policy_template.go
sputn1ck: Fixed. Encoding preserves author order, not canonical. Two templates with same leaves in different order produce different bytes but identical compiled output.

lib/arkscript/policy_template.go
sputn1ck: Fixed. Added doc noting this is base-leaf-version (0xc0) only.

lib/arkscript/policy_template_test.go
sputn1ck: Acknowledged. The test validates `decode(encode(x))` semantic equivalence. Canonical encoding is intentionally not a goal — see updated Encode doc.

lib/arkscript/vtxo.go
sputn1ck: Fixed. Changed to Collab/Exit labels, noted canonical indices may differ.

lib/arkscript/spend_helpers.go
sputn1ck: Fixed. Added doc noting legacy semantic ordering (0=collab, 1=exit).

lib/arkscript/vhtlc.go
sputn1ck: Fixed. Expanded `VHTLCPolicy` godoc.

lib/arkscript/checkpoint.go
sputn1ck: Noted — will consider for final commit structure.

lib/arkscript/tree.go
sputn1ck: The parser uses `txscript.MakeScriptTokenizer` which follows Bitcoin push-data rules. Non-minimal encodings fail at script VM level. Nice-to-have but not a correctness issue since we only parse scripts we compiled.

lib/arkscript/vhtlc.go
sputn1ck: Acknowledged. Currently all leaves use `BaseLeafVersion` and scripts are unique. Added leaf version to `CompareTo` as a first step.

lib/arkscript/compose_test.go
sputn1ck: Acknowledged. Will add vHTLC composition tests when composed vHTLC support is wired up.

lib/arkscript/tree.go
sputn1ck: Fixed. `BuildTree` now sorts leaves canonically internally — single canonical ordering point. Callers no longer pre-sort.

lib/arkscript/golden_test.go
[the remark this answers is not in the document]
sputn1ck: 👍

lib/arkscript/tree_test.go
sputn1ck: Thanks!

sputn1ck: lmao claude said thanks, but it was codex who actually built it

lib/arkscript/compose.go
bhandras: codex: I re-checked this after the follow-up fixes, and I still don’t think this one is resolved yet.

The current implementation still starts from `c.ArkPolicy.SpendInfo(leafIndex)`, which returns the default tx-context from `CompiledPolicy.SpendInfo()`:
- `RequiredSequence = 0xffffffff`
- `RequiredLockTime = 0`

That means composed non-standard policies still lose their AST-derived requirements unless there is a typed wrapper that re-derives them later. The vHTLC named accessors are fixed, but `ComposedPolicy` itself still has no equivalent of `SpendInfoForNode(...)`, so a composed vHTLC leaf like `RefundWithoutReceiver` or the unilateral CSV leaves would still come back with the wrong sequence/locktime requirements.

I also checked the new composition coverage, and the vHTLC composition test currently only asserts that the control block is extended by 32 bytes; it does not assert preservation of `RequiredSequence` / `RequiredLockTime`.

So from my side this still looks like an unresolved generic-API issue. I think we either need to:
- preserve semantic node -> tx-context mapping through composition, or
- explicitly narrow/document `ComposeWithSiblingRoot()` as standard-VTXO-only for now.

sputn1ck: Agreed. Resolved more fundamentally: `SpendInfo` now contains only witness-level data (script + control block). `RequiredSequence`/`RequiredLockTime` moved to `SpendPath`.

`ComposedPolicy.SpendInfo()` no longer needs to propagate tx-context because `SpendInfo` does not carry it. The spending tx builder derives tx-context from the AST node via `CompiledPolicy.SpendPathForNode(node, conditions)`.

See commit `52c8e8a`.

lib/arkscript/node_test.go
sputn1ck: Done — all node tests now assemble expected scripts manually via `txscript.NewScriptBuilder` and assert byte equality. See commit `25eb520`.

lib/arkscript/node_test.go
sputn1ck: Done — all node tests now assemble expected scripts manually via `txscript.NewScriptBuilder` and assert byte equality. See commit `25eb520`.

lib/arkscript/node_test.go
sputn1ck: Done — all node tests now assemble expected scripts manually via `txscript.NewScriptBuilder` and assert byte equality. See commit `25eb520`.

lib/arkscript/node_test.go
sputn1ck: Done — all node tests now assemble expected scripts manually via `txscript.NewScriptBuilder` and assert byte equality. See commit `25eb520`.

lib/arkscript/nums.go
sputn1ck: Added `nums_test.go` with 3 tests: validity, documented hex match, determinism. See `25eb520`.

lib/arkscript/validate.go
sputn1ck: The validation checks collab leaf (operator key present), exit leaf (no operator), CSV gating on all exits, and minimum delay. This covers both server-side (operator cosigning) and client-side (CSV exit path) needs. We can tighten further if specific gaps are identified.

bhandras (approved):
LGTM, pending linter failure 🎉

PR #228, opened 2026-03-31
btcwbackend: add neutrino+btcwallet wallet mode

### LINKED ISSUE #149: darepod: add neutrino+btcwallet mode (wallet.type=btcwallet)
## Summary

Add a third wallet backend mode using btcwallet with neutrino as the
chain backend. This provides a fully in-process, SPV-based wallet that
doesn't require an external esplora server (unlike lwwallet) or a full
lnd node.

## Motivation

- lwwallet requires an esplora server to be running
- lnd mode requires a full lnd node
- neutrino+btcwallet gives a lightweight, fully self-contained option
  that connects directly to the Bitcoin P2P network

## Design

- `--wallet.type=btcwallet` config option
- Uses btcwallet for key management and UTXO tracking
- Uses neutrino (compact block filters / BIP 157/158) for chain sync
- Requires P2P peers to connect to (configurable)
- Implements the same `wallet.BoardingBackend`, `chainsource.ChainBackend`,
  and `wallet.ClientWallet` interfaces

## Implementation Notes

- New package `btcwbackend/` or extend `lwwallet/` with neutrino option
- Wire up in `darepod/server.go` alongside existing lnd/lwwallet branches
- Add config fields: `wallet.btcwallet.peers`, `wallet.btcwallet.datadir`

Roasbeef:
This PR introduces a third wallet backend mode (`wallet.type=btcwallet`) that
uses btcwallet with neutrino (BIP 157/158 compact block filters) as the chain
source. This provides a fully self-contained, SPV-based wallet that connects
directly to the Bitcoin P2P network—no external Esplora server or lnd node
required.

## Motivation

The existing wallet backends each carry an external dependency: `lnd` mode
requires a full lnd node, and `lwwallet` mode requires a running Esplora/
mempool.space instance. For users who want a lightweight, zero-dependency
setup, neutrino gives us BIP 157/158 compact block filter support over the raw
P2P network, which is exactly what lnd itself uses in its lightest
configuration.

## Design

The new `btcwbackend/` package implements the same three interfaces as the
existing backends:

- **`chainsource.ChainBackend`** — Uses lnd's `NeutrinoNotifier` for
  event-driven confirmation/spend/block notifications (no polling). Fee
  estimation is handled by `chainfee.WebAPIEstimator` with a configurable fee
  URL, since neutrino has no mempool visibility.

- **`wallet.BoardingBackend`** — Wraps btcwallet for key derivation and
  taproot script import. Unlike the lwwallet adapter (which bypasses btcwallet
  and queries Esplora directly for non-default scope UTXOs), this adapter uses
  btcwallet's native `ListUnspent`—after `ImportTaprootScript`, neutrino's
  compact block filter matching naturally detects matching outputs during sync.

- **`round.ClientWallet`** — The `Wallet` struct embeds `input.Signer` from
  btcwallet and exposes `DeriveNextKey`, satisfying the interface directly
  (same pattern as lwwallet).

The daemon wiring in `darepod/` adds `WalletTypeBtcwallet` arms to all six
wallet-type switch sites, plus a `tryAutoUnlockBtcwallet` function that reuses
the existing seed management and wallet unlock flow.

## Configuration

```
--wallet.type=btcwallet
--wallet.btcwallet_peers=host:port      # exclusive peers (no DNS seeding)
--wallet.btcwallet_addpeers=host:port   # additional persistent peers
--wallet.btcwallet_datadir=/path        # neutrino data dir
--wallet.feeurl=https://...             # fee estimation endpoint (required on mainnet)
--wallet.persist_filters=true           # write cfilters to disk
```

Closes #149.

## Test plan

- [ ] `go build ./btcwbackend/... ./darepod/...` compiles cleanly
- [ ] `make lint-changed-local` passes with zero issues
- [ ] Unit tests for btcwbackend (TODO: follow-up PR)
- [ ] Integration test on regtest with bitcoind peer (TODO: follow-up PR)
- [ ] Manual smoke test: start darepod with `--wallet.type=btcwallet` on
  signet, verify chain sync and wallet operations

### DIFF
--- a/btcwbackend/boarding_backend.go
+++ b/btcwbackend/boarding_backend.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
-	"sync"
 
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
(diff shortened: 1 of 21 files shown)


btcwbackend/boarding_backend.go
```diff
@@ -0,0 +1,212 @@
...
+func (b *BoardingBackendAdapter) ListUnspent(ctx context.Context,
+	minConfs, maxConfs int32) ([]*wallet.Utxo, error) {
+
+	// Treat maxConfs of 0 as "no upper bound" so callers that
+	// omit the parameter don't accidentally filter everything.
+	if maxConfs == 0 {
+		maxConfs = math.MaxInt32
+	}
+
+	addrs := b.SnapshotAddrs()
```
Roasbeef: This should be removed, instead returning all the addresses. On restart, we don't re-populate this, so it can cause us to miss outputs.

Roasbeef:
@claude review this

Roasbeef:
@claude Re the critical finding: the nil `s.chainBackend` passed to `ChainSourceActor` is not actually reachable in practice. The `walletReady` channel gates all wallet-dependent actors (`startWalletDependentActors` is deferred until `markWalletReady`), so no messages reach the chain source actor before the wallet is unlocked and `startBtcwallet` sets `s.chainBackend`. This mirrors how lwwallet handles the locked case. Even though lwwallet creates a standalone chain backend, nothing queries it before wallet-dependent actors start.

The address filter removal in `ListUnspent` (item #2) is addressed in the latest fixup. We now return all watched UTXOs from btcwallet directly without filtering by the in-memory `importedAddrs` map, since btcwallet already scopes to watched scripts.

GetTransaction docstring is also cleaned up in the same fixup.

FeeURL validation was already required unconditionally for btcwallet mode in an earlier commit.

sputn1ck (approved):
LGTM!

PR #229, opened 2026-04-01
serverconn+darepod: derive mailbox IDs from pubkeys, add Schnorr auth + mTLS

Roasbeef:
## Summary

This PR replaces the old arbitrary-string mailbox ID system with
cryptographically-derived identifiers and adds two layers of authentication to
the mailbox transport.

### The Problem

Previously, `LocalMailboxID` and `RemoteMailboxID` were plain config strings
(e.g., `"client-1"`, `"server-for-client-1"`). Nothing bound a client's mailbox
identity to any cryptographic material, which opened several attack vectors:

 - **Impersonation at registration**: A client could set its
   `LocalMailboxID` to any string, including another client's ID, and register
   as that client. The server's sender-stamping only kicks in _after_
   registration — it doesn't prevent the initial claim.

 - **Cross-client message reads**: The `Pull` and `AckUpTo` RPCs accept a
   `mailbox_id` parameter but had no access control. Any client could call
   `Pull(mailbox_id=X)` for an arbitrary `X` and read another client's messages.

 - **Configuration footgun**: Operators had to manually coordinate matching
   mailbox ID strings between client and server configs, with no feedback if
   they were misconfigured.

### The Fix: Three Layers

**1. Pubkey-derived mailbox IDs** — Mailbox IDs are now deterministically
computed as `hex(compressed_pubkey)`. The client uses its identity key
(`KeyFamilyNodeKey`, family 6, index 0) and the server uses its operator key
(`KeyFamilyMultiSig`, index 0). The `LocalMailboxID` and `RemoteMailboxID`
config fields are removed entirely.

The client resolves the server's operator pubkey via a direct gRPC
`GetInfo` call _before_ the mailbox transport is wired, sidestepping the
chicken-and-egg where `fetchOperatorTerms` normally goes over the mailbox.

**2. Schnorr auth signature** — On every outbound envelope, the client includes
an `x-mailbox-auth-sig` header carrying a Schnorr signature over
`sha256("mailbox-auth" || sender_pubkey || recipient_mailbox_id)`. The server
verifies this in `HandleUnknownClient` before completing registration. This
proves the client actually holds the secp256k1 private key for the mailbox ID
they claim.

**3. mTLS client certificate** — The client generates an ephemeral P-256 TLS
certificate with the secp256k1 identity pubkey in the Subject CN. The server
extracts this CN on every RPC call via `peer.FromContext` and enforces that
`Pull`/`AckUpTo` requests only access the authenticated client's own mailbox.

The Schnorr signature proves key ownership (one-time, at registration). The
mTLS enforces per-RPC access control (every call). Together they close both
the impersonation and cross-client read vectors.

### TODO

In `--no-tls` regtest mode, the mTLS layer is bypassed (no client cert
presented). The Schnorr auth signature still gates registration, but
`Pull`/`AckUpTo` are unguarded. A follow-up should either require TLS
unconditionally or add a server-side Pull guard using the registered client
map as a non-TLS fallback.

## Test Plan

- [x] Unit tests for `PubKeyMailboxID`, `SignMailboxAuth`, `VerifyMailboxAuth`,
  `ParseMailboxPubKey` (cross-server replay, bad hex, wrong key)
- [ ] Integration tests with `make itest` (requires server-side PR)
- [ ] Verify regtest `--no-tls` still works (Schnorr auth, no mTLS)


Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #232, opened 2026-04-02
multi: fix neutrino btcwallet UTXO detection, TxProof, and startup

Roasbeef:
## Summary

This PR fixes several critical issues with the neutrino-backed btcwallet
backend that prevented boarding flows from working end-to-end:

- **Boarding UTXOs invisible to wallet**: `ImportTaprootScript` used the
  custom `m/1017'/coinType'` key scope, but btcwallet's `chainntfns.go`
  skips `AddCredit` for non-default scopes. Switched to
  `KeyScopeBIP0086` to match lnd's behavior — neutrino now correctly
  detects and tracks boarding outputs via compact block filters.

- **TxProof construction always failed**: The block cache capacity was
  set to 20 (bytes, not blocks). Since even the smallest regtest block
  exceeds 20 bytes, every `GetBlock` call errored out, causing
  `buildBoardingTxProof` to return `None`. The server then rejected
  join requests with "TxProof is required when server has no chain
  source". Raised to 2 MiB.

- **Nil pointer panic on shutdown**: `RegisterConf`, `RegisterSpend`,
  and `RegisterBlocks` called `event.Cancel()` from both a deferred
  goroutine and the returned Cancel closure. lnd's
  `notificationDispatcher` panics on double-cancel (nil map lookup
  after delete). Wrapped in `sync.Once`.

- **Empty IdentityPubkey in GetInfo**: The btcwallet case was missing
  from the `GetInfo` RPC handler, leaving `IdentityPubkey` and
  `BlockHeight` empty. This broke `SendVTXO` tests that derive
  recipient keys from `GetInfo`.

- **Silent log suppression**: `btcwbackend.Subsystem` was missing from
  `allSubsystems`, causing all BTCW logs to route to `btclog.Disabled`.

- **Slow wallet unlock on restart**: Neutrino P2P connection happened
  inside the `UnlockWallet` RPC, causing timeouts when the connection
  manager hit backoff retries. Introduced `preStartNeutrino` to begin
  P2P sync early in daemon startup, and `NewWithNeutrino` to let the
  wallet reuse the pre-connected service.

- **Premature wallet-ready signal**: The sync wait goroutine checked
  `BestBlock height > 0` which could pass during header sync before
  btcwallet's recovery scan completed. Now uses `IsSynced()` which
  gates on both header sync and recovery completion.

## Test plan

- [x] All 8 boarding itests pass with `backend=btcwallet`
- [x] `TestSendVTXOIntegrationDryRunPreview` passes with `backend=btcwallet`
- [x] Both restart tests pass (RestartAfterRoundBroadcast, RestartAfterInputSigSent)
- [x] Systests pass (18/21, 3 failures are Docker network pool exhaustion)
- [x] Clean shutdown — no nil pointer panics

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -29,6 +29,8 @@ package may import from a higher layer.
 | [`chain`](chain/) | Bitcoind RPC utilities (package relay, `SubmitPackage`) |
 | [`lndbackend`](lndbackend/) | `BoardingBackend` implementation via LND's wallet kit |
 | [`lwwallet`](lwwallet/) | Lightweight in-process wallet (btcwallet + Esplora, no external LND) |
+| [`btcwbackend`](btcwbackend/) | Neutrino-backed wallet backend (btcwallet + compact block filters) |
+| [`walletcore`](walletcore/) | Shared wallet abstractions and boarding logic used by lwwallet and btcwbackend |
 | [`db`](db/) | SQLite/PostgreSQL persistence: boarding, rounds, VTXOs, OOR artifacts |
 | [`mailbox`](mailbox/) | Mailbox protocol primitives across three sub-packages (pb, rpc, conn) |
 | [`serverconn`](serverconn/) | Unified server connector: durable egress, ingress polling, unary RPC facade |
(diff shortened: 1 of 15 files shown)


Roasbeef:
@claude review this

PR #236, opened 2026-04-02
oor+round: fix checkpoint anchors and persisted client tree sigs

ellemouton:
## Summary

This PR pulls out two low-risk bug fixes that are needed for the unrolling
work, but are both fixes to behavior that is already wrong on `main` today.

It does two things:

- ensure OOR checkpoint transactions always include a valid anchor output and
  reject malformed checkpoint artifacts early
- propagate aggregated round signatures to extracted `ClientTrees` so the
  client-side trees that get persisted for unilateral exit are actually signed

## Root Cause

There were two separate bugs.

First, OOR checkpoint construction and validation were too permissive. That let
anchorless or malformed checkpoint artifacts exist, even though downstream OOR
and unrolling logic depends on the checkpoint anchor invariant.

Second, the round flow validated aggregated signatures on `VTXOTreePaths`, but
`ClientTrees` are extracted copies created earlier via
`ExtractPathForCoSigners`. Those copied trees never received the aggregated
signatures, so persisted client trees could remain unsigned even after the
round completed successfully.

### Checkpoint anchor fix

- make checkpoint builders always include the expected anchor output
- add checkpoint validation helpers for OOR package build/finalize/submit paths
- add unit coverage for anchorless checkpoint rejection and anchor invariants

### Client tree signature propagation fix

- after validating aggregated signatures on the main VTXO tree paths, submit
  the same signatures to the extracted `ClientTrees`
- verify those client trees are signed before progressing
- add a focused regression test that uses a real extracted client tree copy,
  which is the case that previously lost signatures

## Why This Is Needed For Unrolling

Unrolling depends on the client having a valid, signed recovery path.

- checkpoint artifacts need anchors so they are structurally valid and can be
  handled consistently by OOR/unrolling code
- persisted client trees need the aggregated signatures so unilateral-exit
  recovery transactions are actually broadcastable

These fixes are intentionally separated from the larger unrolling feature work
so they can land as standalone bug-fix PRs.

## Validation

- `go test ./round ./lib/tx/checkpoint ./lib/tx/oor`

### DIFF
--- a/lib/tx/checkpoint/build.go
+++ b/lib/tx/checkpoint/build.go
@@ -58,8 +58,9 @@ type Result struct {
 	TapTreeEncoded []byte
 }
 
-// BuildPSBT constructs an unsigned checkpoint PSBT that spends a VTXO input and
-// pays the entire input value to a checkpoint P2TR output.
+// BuildPSBT constructs an unsigned checkpoint PSBT that spends a VTXO input,
+// pays the entire input value to a checkpoint P2TR output, and appends a
+// zero-value anchor output.
 //
 // The checkpoint output pkScript is derived deterministically from:
 //
@@ -119,6 +120,7 @@ func BuildPSBT(policy scripts.CheckpointPolicy, in Input) (*Result, error) {
 		Value:    in.SpentVTXO.Output.Value,
 		PkScript: checkpointPkScript,
 	})
+	tx.AddTxOut(scripts.AnchorOutput())
 
 	pkt, err := psbt.NewFromUnsignedTx(tx)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


lib/tx/checkpoint/build.go
```diff
@@ -119,6 +120,7 @@ func BuildPSBT(policy scripts.CheckpointPolicy, in Input) (*Result, error) {
 		Value:    in.SpentVTXO.Output.Value,
 		PkScript: checkpointPkScript,
 	})
+	tx.AddTxOut(scripts.AnchorOutput())
```
bhandras, line 123: Oh nice catch!

bhandras (approved):
LGTM 🎉

Roasbeef (approved):
LGTM 🐔

PR #145, opened 2026-03-06
multi: add vhtlc package and fix arkscript import cycle

sputn1ck:
## Summary

- Adds `lib/vhtlc`: a typed 6-leaf taproot vHTLC policy for off-chain
  HTLC payments in the Ark protocol. The six spending paths are:
  - Claim (collab): HashLock + Multisig([receiver, server])
  - Refund (collab): Multisig([sender, receiver, server])
  - RefundWithoutReceiver (collab): CLTV + Multisig([sender, server])
  - UnilateralClaim (exit): CSV + HashLock + Checksig(receiver)
  - UnilateralRefund (exit): CSV + Multisig([sender, receiver])
  - UnilateralRefundWithoutReceiver (exit): CSV + Checksig(sender)
- Fixes an import cycle in `lib/arkscript/vtxo.go` by duplicating the
  ARK NUMS key hex constant locally instead of importing `lib/scripts`.

## Test plan

- [ ] `make unit pkg=./lib/vhtlc/... timeout=2m` — unit tests for vHTLC
- [ ] `make unit pkg=./lib/arkscript/... timeout=2m` — golden tests pass
- [ ] Systest: `TestVHTLCOORSendAndBobSweep` in darepo root repo

PR #249, opened 2026-04-13
darepod: add vHTLC custom input coverage

bhandras:
## Summary

This PR adds focused unit coverage for the vHTLC custom-input path that already exists on top of the policy-first `arkscript` stack.

Rather than reviving the old `lib/vhtlc` branch shape from `#145`, this small follow-up proves the modern client flow that survived into `policy-arkscript`:

- external vHTLC outputs can be passed through `SendOOR` custom inputs without consulting the local VTXO store
- store-backed custom inputs can reconstruct the same claim path from persisted descriptor state
- `BuildCustomTransferInputs` preserves the semantic policy template and decodes the serialized `SpendPath` into an OOR input whose `EffectiveSpendPath()` matches the expected vHTLC claim path

## Why this PR

The original vHTLC work was overtaken by the later policy-first migration. In the current codebase the relevant functionality lives in:

- `lib/arkscript/vhtlc.go`
- `lib/arkscript/spend_path.go`
- `darepod/wallet_ops.go`
- `daemonrpc/daemon.proto`

So this PR keeps the review surface honest and small: it does not try to resurrect the old package layout, it just adds the focused coverage that demonstrates the modern path is still working.

## Test Plan

- [x] `go test ./darepod ./lib/arkscript ./oor -count=1`
- [x] `make lint-local`

### DIFF
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -162,7 +162,7 @@ func (b *LNDBackend) BroadcastTx(ctx context.Context, tx *wire.MsgTx,
 
 	txHash := tx.TxHash()
 	b.logger(ctx).InfoS(ctx, "Broadcasting transaction via LND",
-		slog.String("txid", txHash.String()),
+		btclog.Hex("txid", txHash[:]),
 		slog.String("label", label))
 
 	err := b.broadcaster.PublishTransaction(ctx, tx, label)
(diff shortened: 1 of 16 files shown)


bhandras:
Closing this one because the focused vHTLC custom-input coverage has been folded directly into `#212`, where the policy-first daemon and custom-input functionality already lives.

That keeps the tests next to the implementation they validate and avoids carrying a tiny follow-up PR for functionality that is already part of the main `policy-arkscript` stack.

Issue #253, opened 2026-04-14
Design: tune DoS-protection caps across arkscript + durable TLV decoders


Roasbeef:
## Context

The PR 212 followup branch introduced bounded decode limits in two places to
close DoS-amplification vectors:

**arkscript (policy template decoding)**
- \`MaxPolicyTemplateBytes = 64 KiB\`
- \`MaxLeafTemplateBytes   = 16 KiB\`
- \`MaxPolicyLeaves        = 32\`
- \`MaxPolicyDepth         = 16\`
- \`MaxPolicyNodes         = 256\`
- \`MaxMultisigKeys        = 64\` (promoted from local const)

**oor/actor_durable_message.go (condition witness TLV)**
- \`maxConditionWitnessItems     = 64\`
- \`maxConditionWitnessItemBytes = 520\`  (Bitcoin MAX_SCRIPT_ELEMENT_SIZE)

These values are \"best guess\" — generous enough to not break known Ark
policy shapes (standard VTXO = 2 leaves, vHTLC = 6 leaves) while bounding
work on attacker-controlled input. We picked them without a detailed
workload analysis.

## What needs doing

1. **Audit real usage.** Instrument decoders with a histogram of observed
   sizes/depths/node-counts in staging / itests. Get a distribution of
   what legitimate policies look like in practice.

2. **Pick defensible numbers.** Once we have a distribution, set each
   cap at (say) 10x the observed 99th percentile so legitimate growth
   has room but attacker amplification is still gated.

3. **Decide: configurable vs hard-coded.**
   - Hard-coded is simpler, harder to accidentally weaken.
   - Configurable (via \`Config\` or build tag) lets operators tune for
     unusual workloads.
   Leaning toward hard-coded at the library layer, but exposing a
   per-RPC override on the admission path for policy size only.

4. **Cross-check against future extensions.** If we add more complex
   policy shapes (e.g. multi-hop vHTLC, multi-party settlements), do
   the depth/node caps still hold? Document the headroom and the
   upgrade path.

5. **Add metrics.** Emit counters for \"policy rejected: too deep\",
   \"policy rejected: too many nodes\", etc. so an operator can see if
   legitimate users are tripping a cap.

## Current placeholder behaviour

All caps reject fail-closed with a typed error. No silent truncation. The
tests added in the followup branch exercise each cap boundary.

## Related

- PR #212 H-2, H-5 findings (see \`.reviews/lightninglabs_darepo-client_PR_212_review.md\`).
[...]

Issue #252, opened 2026-04-14
Design: pubkey encoding strategy for arkscript policy templates


Roasbeef:
## Problem

The `arkscript` policy template currently serialises `Multisig` participant keys as 32-byte x-only via `schnorr.SerializePubKey`. On decode via `schnorr.ParsePubKey`, y-parity is synthesised to even. A compressed key that had odd parity will round-trip to a different key.

This surfaced in PR #212 as H-8: `db/vtxo_store.go:rowToDescriptor` unconditionally overwrote the stored compressed `OperatorPubkey` column with the lifted-from-policy value, silently flipping parity. `db/round_store.go` handles the same case by preferring the stored column and falling back to the policy only when the column is empty.

The parity-flip matters anywhere we compare 33-byte compressed forms (wire messages, DB indexes, key-locator lookups, \"is this my key?\" checks). It does NOT matter in tapscript output, which must be x-only (BIP-341/342).

## Options

**A — Encode 33-byte compressed in policy template.** Lossless round-trip. Policy grows ~1 byte per key reference (immaterial). Wire-incompatible change. Decoding still produces a valid `*btcec.PublicKey`.

But: some upstream callers legitimately have only x-only keys (e.g. keys extracted from a taproot context). Forcing them into 33-byte compressed requires synthesising a y-parity somewhere — we just move the problem from the encoder to the caller.

**B — Keep x-only in policy template, document the parity-lift behaviour, and audit every consumer to prefer the stored compressed column when both are available.** This is what `round_store` does; the H-8 bug was `vtxo_store` not doing it.

**C — Split the type: introduce `CompressedPubKey` for "reference" uses (DB columns, wire pubkey fields, key-locator-keyed maps) vs `XOnlyPubKey` for \"tap-output\" uses. Make `arkscript.PolicyTemplate` accept `CompressedPubKey` so the encoder just calls `.Bytes()` losslessly.**

## Questions

1. Are there callers that fundamentally can ONLY produce x-only keys into the policy template? If so, option A forces them to tag a default parity — the same problem we have today, just relocated.
2. Is there any BIP-341 context where we'd want the policy template itself stored as tap-leaf bytes (not semantic)? If so option A still works; we only decode back to Multisig for validation/round-trip.
3. Is there a case where DIFFERENT y-parities of the same x-only key are meaningful for the participant? (In Bitcoin the taproot-output key is always x-only, but identity keys across the rest of the system are compressed and distinct by parity.)

## Recommended next step

[...]

PR #255, opened 2026-04-14
arkscript: policy-first followup hardening

Roasbeef:
In this PR, we land the followup fixes flagged by the multi-agent code
review of #212. The parent PR moved the client to a policy-first
`arkscript` model, and the review surfaced a handful of hardening
opportunities in the admission surfaces, the durable-message decoders,
and the VTXO rehydrate paths. None of the findings are exploitable on a
deployed system (the parent PR hasn't shipped yet), but we want to land
the fixes on top of the same stack so the merged history carries them
as one unit.

See each commit message for the detailed rationale w.r.t the incremental
changes. The full review report is at
`.reviews/lightninglabs_darepo-client_PR_212_review.md` in the parent
branch, and the highest-severity findings addressed here are:

- **H-2** Decode-bomb vector in `arkscript.DecodePolicyTemplate`. A
  778 KB blob of 100k nested `Condition` nodes could consume ~7.8s of
  CPU on decode; every reachable call site was gated only by a 10 KB
  envelope cap or no cap at all. We now thread a shared depth /
  node-count budget through every recursive decoder and reject
  oversize blobs upfront.
- **H-3** Signature-oracle via `CustomOORInput`. Callers could supply
  an arbitrary `(pkScript, witness_script, control_block)` tuple and
  obtain a Schnorr signature under the VTXO owner key without any
  check that the control block actually committed to the declared
  pkScript. The new `SpendPath.VerifyBindsToPkScript` recomputes the
  output key from the supplied control block and asserts equality
  against the prevout pkScript. Wired into both the RPC boundary
  (`BuildCustomTransferInputs`) and the signer
  (`signCustomCheckpointPSBT`) for defense-in-depth.
- **H-4 + H-5** Admission validator was too permissive. `ValidatePolicy`
  silently accepted policies containing `Multisig{operator_only}`
  leaves (unilateral operator spend), and the daemon never passed a
  non-zero `MinExitDelay` so a 1-block CSV would slip through and
  break the forfeit incentive. We split `ValidatePolicy` into a
  structural validator and a stricter `ValidateStandardVTXOPolicy`
  that refuses a zero minimum fail-closed, and we walk every
  `Multisig` node to reject operator-only participants.
- **H-7** `tree.NewVTXODescriptor` helper was conflating the VTXO
  owner key with the MuSig2 tree-construction `CoSignerKey`. The
  production round path was already correct (it threads an
  ephemeral per-round `SigningKey` derived via
  `keychain.DeriveNextKey`), but the helper was an exported
  footgun. We require an explicit `signingKey` parameter and
  restore the \"ephemeral per round, never reused\" invariant in
  the docstring so a future caller can't silently collapse the
  MuSig2 aggregate-key set when one wallet contributes multiple
  VTXOs to the same round.
- **H-8** `db/vtxo_store.go` was unconditionally overwriting the
  stored compressed operator pubkey with the x-only-lifted value
  from the policy template, silently flipping y-parity. `round_store`
  already handled this correctly by preferring the stored column;
  we now match that pattern in `vtxo_store`. The deeper
  architectural question of whether to switch the policy template
  encoding to lossless 33-byte compressed is tracked in #252.
- **M-1** Concurrent `SendOOR` callers could race on the same
  caller-supplied custom input. The RPC server now holds a
  mutex-protected reservation map keyed by outpoint; collisions
  fail with `Aborted` rather than double-signing.
- **M-5** `NewFallbackSchnorrSigner` was dead code that silently
  swallowed primary-signer errors. Deleted.
- **M-7** Rehydrate now emits a structured warn-log when the stored
  `Expiry` column and the policy-derived `ExitDelay` disagree,
  surfacing drift that would otherwise be silently masked.
- **M-10** `oor/actor_durable_message.go` condition-witness decoder
  accepted 10 KB per item with no count cap, so a crafted durable
  blob could force multi-MB allocation during replay. Capped at 64
  items and 520 bytes per item (Bitcoin's
  `MAX_SCRIPT_ELEMENT_SIZE`). Encoder enforces the same caps so
  in-memory state can't drift from what the persisted form admits.

Also included: **H-9** (duplicate-checkpoint-txid rejection) was
verified preserved in `lib/tx/oor/submit.go` and the finalize
validator, no code change needed. The review had flagged it as
missing because the check moved files during the PR 212 refactor.

## Docs

The last commit adds `docs/arkscript_spec.md`, an RFC-style
specification of `lib/arkscript` covering the AST, policy template
binary encoding, canonical ordering, standard shapes (VTXO,
checkpoint, vHTLC), `SpendPath` witness contract, `ValidatePolicy`
invariants, and the security-considerations section cross-referenced
from this PR. Intended for external auditors and future implementors.

## Deferred design discussions

- [#252](https://github.com/lightninglabs/darepo-client/issues/252):
  pubkey encoding strategy. The lossy x-only encoding in
  `Multisig` is the root cause of the H-8 parity class. This PR
  fixes the immediate bug in `vtxo_store` but we want a design
  decision (switch to lossless 33-byte compressed, keep x-only with
  rehydrate discipline, or split the type) before sweeping the
  whole codebase.
- [#253](https://github.com/lightninglabs/darepo-client/issues/253):
  tuning the DoS caps introduced here. The numbers are generous
  first-guesses based on known Ark policy shapes (standard VTXO =
  2 leaves, vHTLC = 6 leaves); we want to instrument real workloads
  and resize to something like 10x the observed p99 before locking
  the numbers in.

## Base branch

This branches off #212 (\`policy-arkscript\`) because the fixes depend
on the policy-first types it introduces. Can be rebased onto \`main\`
once the parent lands.

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./lib/arkscript/... ./darepod/ ./oor/ ./round/ ./wallet/ ./db/ ./lib/tree/ ./lib/tx/... -count=1\`
- [x] New tests for every non-trivial fix: decode-cap boundaries,
  `ValidatePolicy` invariants, `NewVTXODescriptor` regression,
  `reserveCustomInputs` collision and atomicity, condition-witness cap
  round-trip.
- [ ] systest + itest sweep before merge.
- [ ] Reconcile against parent PR #212 rebase if the parent picks up
  new commits.

### DIFF
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -75,7 +75,7 @@ func NewRPCServer(server *Server) *RPCServer {
 // function reverses the claim and is safe to call on either success or
 // failure paths (e.g. via defer).
 func (r *RPCServer) reserveCustomInputs(
-	outpoints []wire.OutPoint) (release func(), err error) {
+	outpoints []wire.OutPoint) (func(), error) {
 
 	r.customInputLocksMu.Lock()
 	defer r.customInputLocksMu.Unlock()
@@ -94,7 +94,7 @@ func (r *RPCServer) reserveCustomInputs(
 		r.customInputLocks[outpoints[i]] = struct{}{}
 	}
 
-	release = func() {
+	release := func() {
 		r.customInputLocksMu.Lock()
 		defer r.customInputLocksMu.Unlock()
 
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -10,6 +10,7 @@ import (
 	"log/slog"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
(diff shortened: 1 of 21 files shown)


Roasbeef:
@claude review this

Roasbeef:
@codex review this

PR #256, opened 2026-04-14
ci: add nightly /doc-gardening cron workflow

Roasbeef:
In this PR, we add a nightly GitHub Actions cron that runs the project-scoped `/doc-gardening` skill via the existing `anthropics/claude-code-action@v1` integration, and opens a PR against master whenever the sweep finds drift in any per-package `CLAUDE.md`/`AGENTS.md`, in `ARCHITECTURE.md`, or in `docs/index.md`. Today the skill is only run on demand by humans, so doc drift accumulates silently between sprints and the next person to notice is usually whoever sees `make doc-check` red-X in CI. Running it nightly turns that drift into a small reviewable PR each morning instead of a quarterly cleanup pass.

The workflow shape: checkout w/ `BOT_GITHUB_TOKEN` (full history, since the skill walks `git log`/`git diff` against master), `setup-go` + `go mod download` to prime the cache for `go doc`, a shell step that closes any still-open auto-PRs from prior runs (so each sweep is authoritative against current master), then the Claude action with a focused prompt, and a `failure()` handler that opens a tracking issue if the run red-Xes overnight.

## Why bot token everywhere

We pass `BOT_GITHUB_TOKEN` both as the `actions/checkout` `token:` and as step-level `GH_TOKEN` on the Claude action. The first wires `git push` to the bot identity. The second makes `gh` (which honors `GH_TOKEN` over `GITHUB_TOKEN`) authenticate as the bot inside Claude's Bash subprocess. Both are required because PRs created with the default `GITHUB_TOKEN` don't trigger downstream workflow events, and we want `agent-docs-sync` and the rest of CI to fire on the nightly PR like any other change. We deliberately don't touch `GITHUB_TOKEN` itself, the action's internal bookkeeping uses it.

## Capability boundary

Since Claude is driving the full pipeline (sweep, commit, branch, push, `gh pr create`), the `--allowed-tools`/`--disallowed-tools` pair is the primary safety boundary. Allowed: `git` subcommands, `make doc-check`, `go doc`, the file/text primitives the skill needs (`find`, `cp`, `sed`, `awk`, etc), and `gh pr create`/`list`/`view`/`auth status`. Explicitly disallowed: `rm`, `gh api`, `gh secret`, `gh workflow`, `gh release`, `gh repo delete`, `gh auth login`/`logout`/`refresh`, `curl`, `wget`, `git push --force`, `git push -f`, `git push --delete`, `git reset --hard`, `git commit --amend`. Worst-case prompt-injection landing in some `CLAUDE.md` Phase 1 reads can do at most what an honest doc-gardening run can do, and we have master branch protection on the server side as a hard backstop against direct pushes.

## Prompt shape

The prompt tells Claude explicitly to read `.claude/skills/doc-gardening/SKILL.md` and execute Phases 1-7 with `scope=all`. We don't rely on slash-command auto-dispatch, since headless `claude -p` doesn't reliably surface project skills the way the REPL does. After each phase the model prints `PHASE N COMPLETE` so the Actions log is grep-able. On a clean sweep it prints `DOC-GARDENING NO-OP` and exits without committing. On a successful sweep it cuts a `doc-gardening/nightly-YYYY-MM-DD-${RUN_NUMBER}` branch, commits with `git commit -F .git/COMMIT_MSG`, pushes, opens the PR via `gh pr create --base master --head $BRANCH --label documentation --label automation`, and prints the resulting URL prefixed with `DOC-GARDENING PR: ` so it's easy to find in the log.

Note: the workflow uses an `automation` label that doesn't currently exist in this repo. It needs to be created (`gh label create automation --color c5def5 --description "Automated changes from CI bots"`) before the first nightly run, otherwise `gh pr create` inside the action will fail with "label not found".

## Observability

A `failure()` step opens a tracking issue on the repo if the workflow red-Xes for any reason (OAuth token expiry, `make doc-check` regression that the skill can't fix, runner OOM, etc, etc). The next-day Issue is the canary, no need to babysit cron at 06:00 UTC. Stale auto-PRs from prior days are closed by a shell step at the start of each run with `--delete-branch`, so the queue doesn't grow unbounded if nobody merges yesterday's sweep.

## Test plan

- [ ] Create the `automation` label in this repo (see note above) before first run.
- [ ] Manual `workflow_dispatch` after merge: watch for `PHASE 1 COMPLETE` through `PHASE 7 COMPLETE`, then either `DOC-GARDENING NO-OP` or `DOC-GARDENING PR: <url>`.
- [ ] On the sweep-with-changes path, confirm downstream CI fires on the resulting PR (proves the bot token plumbing is correct, not the default `GITHUB_TOKEN`).
- [ ] Verify `gh pr view <number> --json author` shows the bot identity, not the default actions user.
- [ ] Trigger a second `workflow_dispatch` the same day to confirm the stale-PR cleanup step closes yesterday's auto-PR.

### DIFF
--- a/.github/workflows/doc-gardening-nightly.yml
+++ b/.github/workflows/doc-gardening-nightly.yml
@@ -1,0 +1,188 @@
+name: Doc Gardening (Nightly)
+
+# Nightly cron that runs the /doc-gardening skill (see
+# .claude/skills/doc-gardening/SKILL.md) via the existing
+# anthropics/claude-code-action@v1 integration, and opens a PR against master
+# when the sweep produces any doc drift. A clean sweep exits silently.
+
+on:
+  schedule:
+    # 06:00 UTC daily. Both repos run at the same time; their PRs are
+    # independent so they don't contend.
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+concurrency:
+  # Serialize runs so a manual dispatch never races the scheduled one.
+  group: doc-gardening-nightly
+  cancel-in-progress: false
+
+permissions:
+  # Checkout uses BOT_GITHUB_TOKEN directly, so the default GITHUB_TOKEN
+  # only needs read access for the action's own bookkeeping.
+  contents: read
+  pull-requests: read
+
+jobs:
+  sweep:
+    name: Run /doc-gardening and open PR
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout (bot token)
+        uses: actions/checkout@v5
+        with:
+          # Use BOT_GITHUB_TOKEN so subsequent `git push` operations
+          # authenticate as the bot. This is required for the resulting PR
+          # to trigger downstream workflow events (agent-docs-sync, main,
+          # etc.) — the default GITHUB_TOKEN cannot create PRs that trigger
+          # other workflows.
+          token: ${{ secrets.BOT_GITHUB_TOKEN }}
+          # Full history: the doc-gardening skill runs git diff/log against
+          # master to detect stale packages. A shallow clone silently
+          # returns empty history and turns the sweep into a false no-op.
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Prime module cache
+        # Phase 7 of the skill runs `make doc-check`, which may shell out to
+        # `go doc` against third-party modules. Prime the cache so a cold
+        # runner doesn't hit network errors mid-sweep.
+        run: go mod download
+
+      - name: Close stale nightly PRs
+        env:
+          GH_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
+        run: |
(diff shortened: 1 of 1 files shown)


Roasbeef:
@code review this 

@claude review this

Roasbeef:
Both P0 (OIDC id-token requirement) and P1 (token routing via `github_token` input) addressed in fe9d2798.

Single fix: added `github_token: ${{ secrets.BOT_GITHUB_TOKEN }}` to the action's `with:` block. Looking at `anthropics/claude-code-action@v1`'s `src/github/token.ts`, `setupGitHubToken()` checks `OVERRIDE_GITHUB_TOKEN` (populated from `inputs.github_token`) first and returns it directly when set — short-circuiting the OIDC path that would otherwise fail without `id-token: write`. So providing the input resolves both: the action now uses our bot token for its internal git/gh operations, and the OIDC branch is never taken.

Kept `GH_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}` on the step env as defense-in-depth so `gh` inside Claude's Bash subprocess uses the bot token regardless of how the action plumbs its own env into the child process.

PR #261, opened 2026-04-16
lib/recovery + unrollplan: recovery proof graph and pure planner (2/5)

Roasbeef:
## Summary

**Part 2 of 5 in the stacked split of #235.** Adds two pure library
layers that the unroll subsystem builds on:

- `40582ca` — `lib/recovery: add recovery proof, session, and state
  packages`. The immutable recovery proof graph (`Proof`, `Node`,
  `NodeKind`) with topological layering and parent/child tracking. The
  recovery `Session` that tracks per-node broadcast/confirmation state
  and computes CSV maturity. `SessionState` with custom JSON marshaling
  for `map[chainhash.Hash]` keys. Proof binary codec for checkpoint
  persistence. Tree path extraction helpers. Comprehensive unit tests.

- `d8aa488` — `unrollplan: add pure dependency-resolution planner`.
  Given a recovery proof, durable progress state, and current block
  height, answers which proof transactions are ready to broadcast,
  which are blocked by unconfirmed parents, when the target becomes
  CSV-mature, and whether the sweep is needed. No I/O, no actors.
  Validates state consistency against the proof graph before planning.

## Forward-port from the original branch

The original commits imported the legacy `lib/scripts` helper package,
which `main` removed (PR #215). This branch uses `lib/arkscript.AnchorPkScript`
/ `AnchorOutput` instead. The fixup was folded back into Elle's
original commits so each commit compiles standalone — authorship is
preserved on both commits.

## Stack

| # | Branch | PR | Scope |
|---|--------|----|-------|
| 1/5 | `unroll-01-prep` | #260 | preparatory fixes + infra |
| **2/5** | `unroll-02-plan` | **this PR** | `lib/recovery` + `unrollplan` |
| 3/5 | `unroll-03-txconfirm` | (to come) | `txconfirm` actor |
| 4/5 | `unroll-04-core` | (to come) | vtxo + db + rpc + `unroll/` |
| 5/5 | `unroll-05-wire` | (to come) | daemon wiring + CLI |

Supersedes #235.

## Authorship

Both commits authored by @ellemouton.

## Test plan

- [x] `go test ./lib/recovery/... ./unrollplan/...`
- [x] `go vet ./...`
- [x] `go build ./cmd/...`
- [ ] CI: full unit + lint

PR #260, opened 2026-04-16
multi: unilateral-exit preparatory fixes and package-submission infra (1/5)

Roasbeef:
## Summary

**This is part 1 of 5 in a stacked split of #235.** The original PR
bundled 15 commits across five distinct concerns into a single review.
This PR isolates the three preparatory commits so they can land
independently of the unroll subsystem itself.

Scope:

- `188bd32` — `multi: populate round metadata on ClientVTXO at
  confirmation`. Fixes a VTXO-metadata race where callers reading
  between the round FSM's initial save and the VTXO manager's
  follow-up upsert would see zero `CommitmentTxID`, `BatchExpiry`, and
  `CreatedHeight`. Populates the fields from `BoardingConfirmed` in
  the first save; the manager's upsert becomes a no-op.

- `362588c` — `multi: improve durable actor and DB transaction failure
  logging`. Adds a `Log` option to `DurableActorConfig` so actors can
  surface warn/error through the caller's logger. Logs message type
  and delivery ID on Tell failures (makes silent nack loops visible).
  Adds `WarnS` to `TransactionExecutor` for begin/body/commit/retry
  exhaustion.

- `c2e1d51` — `multi: add package submission and harness
  infrastructure`. Adds `SubmitPackageRequest/Response` to
  `chainsource` with backend forwarding. Adds `SubmitPackage` to
  `LNDBackend` (via optional `PackageSubmitter` backed by bitcoind
  RPC). Adds Esplora package relay to `lwwallet`. Adds
  `PackageSubmitter` to `LNDBackendFromLndClientConfig` and daemon
  config. 15s timeout on `RegisterConfirmationsNtfn`, 10s on
  `ConfActor` backend registration. Exports
  `GetLNDClientConn`/`BitcoindRPCUser`/`BitcoindRPCPass` and adds
  `BitcoindPackageSubmitter` for itest package submission.

## Stack

| # | Branch | PR | Scope |
|---|--------|----|-------|
| **1/5** | `unroll-01-prep` | **this PR** | preparatory fixes + infra |
| 2/5 | `unroll-02-plan` | (stacked on 1) | `lib/recovery` + `unrollplan` |
| 3/5 | `unroll-03-txconfirm` | (stacked on 2) | `txconfirm` actor |
| 4/5 | `unroll-04-core` | (stacked on 3) | vtxo + db + rpc + `unroll/` |
| 5/5 | `unroll-05-wire` | (stacked on 4) | daemon wiring + CLI |

Supersedes #235.

[...]

PR #235, opened 2026-04-02
multi: add unilateral-exit subsystem

ellemouton:
## Summary

This PR adds the client-side unilateral-exit runtime and the daemon
plumbing around it.

The subsystem is split into three packages with distinct
responsibilities:

- `unrollplan`: pure dependency-resolution planner. Given a recovery
  proof, durable progress state, and current block height, it determines
  which proof transactions are ready to broadcast, which are blocked by
  unconfirmed parents, when the target becomes CSV-mature, and whether
  the sweep is needed. No I/O, no actors — callers own the durable State
  and the planner re-derives everything from first principles on each
  call.

- `txconfirm`: generic shared actor for getting transactions confirmed
  on-chain. Not tied to unrolling — any subsystem that needs "get this
  tx confirmed with CPFP" can reuse it. Handles anchor detection, CPFP
  child construction, package relay with individual-broadcast fallback,
  periodic fee bumping (non-terminal on failure so the original broadcast
  can still confirm via the active watch), and subscriber fan-out for
  deduped confirmation/failure notifications.

- `unroll`: durable per-target actor and thin registry. Each
  VTXOUnrollActor owns one target outpoint's lifecycle, delegates
  broadcasting to txconfirm, and drives the FSM through proof
  materialization → CSV wait → sweep. The UnrollRegistryActor manages
  job creation, dedup, boot restore, and terminal bookkeeping.

Supporting infrastructure added by earlier commits:

- `lib/recovery`: immutable proof graph, session state, and
  serialization for representing the target VTXO's reachable on-chain
  ancestry
- `db`: persisted unilateral-exit job store for tracking non-terminal
  and terminal unroll jobs, including sweep txid
- `daemonrpc`: Unroll, GetUnrollStatus, and GetBalance.onchain_wallet
  RPCs
- `vtxo`: ForceUnrollEvent in the VTXO lifecycle FSM and
  LazyChainResolver to bridge critical-expiry notifications into the
  unroll registry without a circular init dependency
- `chainsource`/`chainbackends`/`lwwallet`: package submission support
  across all wallet backends

Daemon wiring:

- Wallet adapters for LND, lwwallet, and btcwallet satisfying both
  txconfirm.Wallet and unroll.SweepWallet
- Shared TxBroadcasterActor for CPFP package submission
- UnrollRegistryActor with proof assembler and registry store
- Non-terminal job restore on boot via RestoreNonTerminal
- Critical-expiry routing: VTXO ExpiringNotification → lazy chain
[...]

PR #262, opened 2026-04-16
txconfirm: reusable tx confirmation actor with CPFP (3/5)

Roasbeef:
## Summary

**Part 3 of 5 in the stacked split of #235.** Adds `txconfirm`: a
generic shared actor that deduplicates confirmation requests by txid
and ensures transactions confirm on-chain. Any subsystem that needs
"get this tx confirmed" can use it — not tied to unrolling.

- `9fc7359` — `txconfirm: add reusable tx confirmation actor with CPFP`

Features:
- Automatic anchor detection and CPFP child construction
- Package relay with individual-broadcast fallback
- Periodic fee bumping (non-terminal on failure: the original
  broadcast is still live and the confirmation watch remains active,
  so the next bump waits the full interval before retrying)
- Subscriber fan-out for deduped confirmation/failure notifications
- protofsm-based lifecycle:
  `New → Broadcasting → AwaitingConfirmation → (FeeBumping loop) → Confirmed/Failed`

Key types: `TxBroadcasterActor`, `CPFPBroadcaster`, `EnsureConfirmedReq`.

Depends on the `SubmitPackage` chain-backend surface added in the
prep PR (#260).

## Forward-port from the original branch

Imports of `lib/scripts` were migrated to `lib/arkscript` (matching
main's removal of the legacy helper package). The fixup was folded
back into Elle's original commit.

## Stack

| # | Branch | PR | Scope |
|---|--------|----|-------|
| 1/5 | `unroll-01-prep` | #260 | preparatory fixes + infra |
| 2/5 | `unroll-02-plan` | #261 | `lib/recovery` + `unrollplan` |
| **3/5** | `unroll-03-txconfirm` | **this PR** | `txconfirm` actor |
| 4/5 | `unroll-04-core` | (to come) | vtxo + db + rpc + `unroll/` |
| 5/5 | `unroll-05-wire` | (to come) | daemon wiring + CLI |

Supersedes #235.

## Authorship

Commit authored by @ellemouton.

## Test plan

- [x] `go test ./txconfirm/...`
- [x] `go vet ./...`
- [x] `go build ./cmd/...`
- [ ] CI: full unit + lint

### DIFF
--- a/txconfirm/actor.go
+++ b/txconfirm/actor.go
@@ -84,6 +84,14 @@ type trackedTx struct {
 	fsm  *trackedTxStateMachine
 
 	subscribers map[string]actor.TellOnlyRef[Notification]
+
+	// confWatchRegistered reports whether a chainsource confirmation
+	// watch is currently active for this txid. It is flipped true by
+	// registerConfWatch on success and false by unregisterConfWatch on
+	// success. Terminal cleanup uses it to avoid redundant unregister
+	// round trips for entries whose watch was never registered (e.g.
+	// entries that failed during block-subscription setup).
+	confWatchRegistered bool
 }
 
 // confirmationObservedMsg routes a chainsource confirmation callback back into
@@ -404,6 +412,7 @@ func (a *TxBroadcasterActor) handleConfirmationObserved(ctx context.Context,
 	}
 
 	a.notifyConfirmed(ctx, entry, msg.blockHeight, msg.numConfs)
+	a.evictTerminal(ctx, entry)
 }
 
 // handleBlockObserved records a new best height and fee-bumps any eligible
@@ -657,19 +666,35 @@ func (a *TxBroadcasterActor) registerConfWatch(ctx context.Context,
 			NotifyActor: fn.Some(notifyRef),
 		},
 	).Await(ctx).Unpack()
+	if err != nil {
+		return err
+	}
 
-	return err
+	entry.confWatchRegistered = true
+
+	return nil
 }
 
-// unregisterConfWatch unregisters the confirmation watch for one tracked txid.
+// unregisterConfWatch unregisters the confirmation watch for one tracked
+// txid.
+//
+// The unregister request must supply the same fields that were used at
+// registration time — CallerID, Txid, PkScript, and TargetConfs — because
+// chainsource derives the sub-actor's service key by hashing all four
+// together. Omitting PkScript here (as an earlier revision of this file
+// did) produces a different service key and silently leaks the conf
+// sub-actor for every tracked txid.
 func (a *TxBroadcasterActor) unregisterConfWatch(ctx context.Context,
 	entry *trackedTx) error {
 
 	txid := entry.data.Txid
 	_, err := a.cfg.ChainSource.Ask(
 		ctx, &chainsource.UnregisterConfRequest{
-			CallerID:    a.confCallerID(entry.data.Txid),
-			Txid:        &txid,
+			CallerID: a.confCallerID(entry.data.Txid),
+			Txid:     &txid,
+			PkScript: append(
+				[]byte(nil), entry.data.ConfirmationPkScript...,
+			),
 			TargetConfs: entry.data.TargetConfs,
 		},
 	).Await(ctx).Unpack()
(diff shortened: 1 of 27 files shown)


Roasbeef:
@claude review this 

@codex review this

txconfirm/actor.go
```diff
@@ -0,0 +1,855 @@
...
+	if err != nil {
+		return false
+	}
+
+	lastBroadcastHeight := trackedTxLastBroadcastHeight(currentState)
+	if lastBroadcastHeight == 0 {
+		return false
+	}
+
+	return a.bestHeight-lastBroadcastHeight >=
```
Roasbeef, line 897: We should abstract this further, as it can eventually use the lnd fee function. https://github.com/lightningnetwork/lnd/blob/a8a3e13120ebd52050ea4c6b653e8182f62fbae9/sweep/fee_function.go#L120 

`sweep.NewLinearFeeFunction`. It takes into account the step, target, max, and then uses a linear function to bump progressively as needed.

txconfirm/actor.go
```diff
@@ -0,0 +1,855 @@
...
+		NumConfs:    numConfs,
+	}); err != nil {
+		a.log.WarnS(ctx, "Failed to deliver tx confirmation",
+			err, "txid", txid, "subscriber_id", subscriber.ID())
+	}
+}
+
+// notifyOneFailed delivers one terminal failure notification.
+func (a *TxBroadcasterActor) notifyOneFailed(ctx context.Context,
+	subscriber actor.TellOnlyRef[Notification], txid chainhash.Hash,
```
Roasbeef, line 1003: Nice pattern here re an abstract actor for subscriptions!

txconfirm/broadcaster.go
```diff
@@ -0,0 +1,637 @@
...
+const (
+	// DefaultMaxFeeRateSatPerVByte clamps fee estimates to a sane ceiling.
+	// On regtest or freshly synced nodes, the estimator can return wildly
+	// inflated rates.
+	DefaultMaxFeeRateSatPerVByte int64 = 100
+
+	// ChildVSizeEstimate is the estimated virtual size of a CPFP child with
+	// one anchor input, one confirmed fee input, and a single change
+	// output. Revisit this estimate if the child input composition changes.
+	ChildVSizeEstimate = 155
```
Roasbeef: We can use the actual fee estimator here instead: `input.TxWeightEstimator`.

txconfirm/broadcaster.go
```diff
@@ -0,0 +1,637 @@
...
+// FeeInput is a confirmed wallet UTXO selected for CPFP fee payment.
+type FeeInput struct {
+	// Outpoint identifies the wallet UTXO.
+	Outpoint wire.OutPoint
+
+	// Output is the output being spent for fees.
+	Output *wire.TxOut
+
+	// Confirmed indicates whether this UTXO is confirmed.
+	Confirmed bool
```
Roasbeef, line 124: 👍

txconfirm/broadcaster.go
```diff
@@ -0,0 +1,637 @@
...
+}
+
+// BroadcastRequest describes a signed transaction to broadcast.
+type BroadcastRequest struct {
```
Roasbeef, line 128: Would think this accepts a package? Or the idea is that we always spend an anchor if it's there.

txconfirm/broadcaster.go
```diff
@@ -0,0 +1,637 @@
...
+	}
+
+	deadline := time.Now().Add(2 * time.Second)
+
+	for {
+		utxos, err := b.cfg.Wallet.ListUnspent(ctx, 1, 9999999)
+		if err != nil {
+			return nil, fmt.Errorf("list unspent: %w", err)
+		}
+
```
Roasbeef, line 864: We should be exposing a `LeaseOutput` method via an interface to call once we finally select the input: https://github.com/lightningnetwork/lnd/blob/a8a3e13120ebd52050ea4c6b653e8182f62fbae9/lnrpc/walletrpc/walletkit.proto#L39-L53. 

We'd then need to unlock then as needed as well. 

Only the lnd+btcwallet backends need this, it'll be a noop for lwallet for now (can make an issue abouve that).

txconfirm/broadcaster.go
```diff
@@ -0,0 +1,637 @@
...
+			return i
+		}
+	}
+
+	return -1
+}
+
+// EstimatePackageFee computes the total package fee for one parent+child
+// submission at the given fee rate.
+func EstimatePackageFee(parentTx *wire.MsgTx,
```
Roasbeef: We can use `input.TxWeight` estimator here again.

Roasbeef:
@claude review this

@codex review this

PR #267, opened 2026-04-16
fix: preserve policy-backed custom inputs across darepod and lnd signing

bhandras:
## Summary

Isolates a core policy-backed input/signing fix: an lnd-backed darepod runtime could build a checkpoint for a policy-backed custom input, yet still persist an Ark PSBT whose client-side tapscript spend became invalid once the full submit path reconstructed the witness.

## Root Cause

There turned out to be a few small but compounding problems in the policy-backed path:

1. `GetInfo.IdentityPubkey` for lnd-backed daemons was still reporting the node pubkey rather than the derived proof/identity key that the indexer-side proof path actually uses.
2. Policy-backed custom transfer inputs could lose the exact settlement leaf they needed and fall back to a more generic collaborative owner path.
3. Custom Ark condition witness data was not preserved through Ark PSBT signing and later submit validation.
4. Tapscript witness reconstruction did not reorder script-spend signatures for `CHECKSIGVERIFY` chains.
5. The lnd remote signer path did not always include the key locator for the family-6 / index-0 identity key path, which made the newer lnd-backed signing path unreliable for these spends.
6. On `main`, the owned receive-script signer also needs a small fallback wrapper so script-scoped proofs and identity-key proofs can coexist cleanly.

Each piece on its own was subtle. Together they produced a failure where the persisted Ark PSBT already contained an invalid client signature for the real lnd-backed runtime.

## Commits

Organized as 6 fix commits (1:1 with the root-cause list above) followed by 3 review-driven clarification commits. Every commit builds and passes the targeted test suite on its own.

**Fixes:**

1. **`darepod: derive identity pubkey from wallet backend for lnd`** — route identity derivation through `deriveIdentityPubkey` for every wallet backend including lnd via `WalletKit.DeriveKey`; keep `LndIdentityPubkey` as the public node key.
2. **`darepod: preserve settlement leaf for policy-backed custom inputs`** — resolve the exact owner leaf per custom input from the policy template (exact match first, then the matching `SettlementPairsForParticipant` entry), and carry it on `TransferInput.OwnerLeafScript` / `OwnerLeafPolicy`. `spendPathsMatch` intentionally excludes runtime `Conditions` from branch identity.
3. **`arkscript/oor: preserve multi-item condition witness through sign+submit`** — generalize `EncodeConditionWitness` / `DecodeConditionWitness` to a bounded witness vector, persist the full condition slice during custom checkpoint and Ark PSBT signing, and reinsert those items between the script-spend signatures and the leaf script when rebuilding the tapscript witness.
4. **`oor: reorder tapscript script-spend sigs for CHECKSIGVERIFY chains`** — parse the leaf script for `CHECKSIG` / `CHECKSIGVERIFY` pubkeys and reorder script-spend signatures so witness items appear in reverse key order; tolerate optional checksig slots with empty placeholders.
5. **`lndbackend: propagate key locator on SignOutputRaw`** — bypass lndclient's higher-level helper and include the key locator whenever `Family != 0` or `Index != 0`, so the family-6 / index-0 identity path survives the signer RPC.
6. **`darepod: add fallback schnorr signer for owned receive scripts`** — introduce a wrapper that tries the owned-receive-script signer first and falls back to the identity signer. On fall-through the primary error is joined with the fallback via `errors.Join` so the original cause remains recoverable via `errors.Is` / `errors.As` rather than being silently dropped.

**Clarifications / follow-ups:**

7. **`oor: document condition-witness witness-layout assumption`** — call out that the current sigs-then-conditions order matches the vHTLC claim pattern; flag what changes for future policies that consume conditions above sigs.
8. **`darepod: document defensive fallback in findSettlementOwnerLeaf`** — every current Ark policy enumerates both auth and forfeit leaves in `template.Leaves`, so the exact-leaf loop always wins. The `SettlementPairsForParticipant` branch is defensive coverage for future policies whose auth path is derived; add a test that pins the condition-witness-through-resolution contract via a vHTLC claim carrying a preimage.
9. **`lndbackend: document lndclient key-locator workaround`** — TODO referencing the specific lndclient deficiency so the helper in commit 5 can be retired cleanly when upstream is fixed.

## Testing

- `go build ./...`
- `go test ./darepod ./lib/tx/oor ./oor ./lndbackend ./lib/arkscript -count=1`
- targeted: `TestBuildCustomTransferInputsUsesPolicyLeaf`, `TestBuildTaprootWitnessIncludesConditionWitness`, `TestFindSettlementOwnerLeafWithConditions`, `TestFallbackSchnorrSignerPreservesPrimaryError`
- `make lint-local` — clean

### DIFF
--- a/darepod/wallet_ops.go
+++ b/darepod/wallet_ops.go
@@ -1,6 +1,7 @@
 package darepod
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 
@@ -195,6 +196,11 @@ func BuildCustomTransferInputs(ctx context.Context,
 			}
 		}
 
+		var (
+			ownerLeaf       []byte
+			ownerLeafPolicy []byte
+		)
+
 		// Validate the policy template against Ark invariants
 		// when one is provided. This catches malformed policies
 		// before they reach the server.
@@ -251,11 +257,30 @@ func BuildCustomTransferInputs(ctx context.Context,
 					outpoint, err,
 				)
 			}
+
+			if len(ci.SpendPath) > 0 {
+				ownerLeaf, ownerLeafPolicy, err =
+					findSettlementOwnerLeaf(
+						template,
+						clientKey.PubKey,
+						operatorKey,
+						ci.SpendPath,
+					)
+				if err != nil {
+					return nil, fmt.Errorf(
+						"derive settlement owner leaf "+
+							"for %s: %w",
+						outpoint, err,
+					)
+				}
+			}
 		}
 
 		input := oor.TransferInput{
 			VTXO:               desc,
 			VTXOPolicyTemplate: ci.VtxoPolicyTemplate,
+			OwnerLeafScript:    ownerLeaf,
+			OwnerLeafPolicy:    ownerLeafPolicy,
 		}
 
 		if len(ci.SpendPath) > 0 {
@@ -293,4 +318,121 @@ func BuildCustomTransferInputs(ctx context.Context,
 	}
 
 	return inputs, nil
+}
+
+// findSettlementOwnerLeaf maps a custom auth spend path to the
+// operator-backed forfeit leaf that the later Ark tx must use for the
+// checkpoint output owner path.
+func findSettlementOwnerLeaf(template *arkscript.PolicyTemplate,
+	participant, operator *btcec.PublicKey,
+	rawSpendPath []byte) ([]byte, []byte, error) {
+
+	if template == nil {
+		return nil, nil, fmt.Errorf("policy template is required")
+	}
+
+	if participant == nil {
+		return nil, nil, fmt.Errorf("participant key is required")
+	}
+
+	if operator == nil {
+		return nil, nil, fmt.Errorf("operator key is required")
+	}
+
+	spendPath, err := arkscript.DecodeSpendPath(rawSpendPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("decode spend path: %w", err)
+	}
+
+	// Prefer the exact collaborative leaf when the caller is
+	// already spending an operator-backed branch such as the
+	// vHTLC claim/refund closures.
+	for _, leaf := range template.Leaves {
+		script, err := leaf.Script()
+		if err != nil {
+			return nil, nil, fmt.Errorf(
+				"compile settlement leaf: %w", err,
+			)
+		}
+
+		if !bytes.Equal(script, spendPath.WitnessScript) {
+			continue
+		}
+
+		encodedLeaf, err := leaf.Encode()
+		if err != nil {
+			return nil, nil, fmt.Errorf(
+				"encode settlement leaf: %w", err,
+			)
+		}
+
+		return bytes.Clone(script), encodedLeaf, nil
+	}
+
+	pairs, err := template.SettlementPairsForParticipant(
+		participant, operator,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, pair := range pairs {
+		if !spendPathsMatch(spendPath, pair.AuthPath) {
+			continue
+		}
+
+		for _, leaf := range template.Leaves {
+			script, err := leaf.Script()
+			if err != nil {
+				return nil, nil, fmt.Errorf(
+					"compile settlement leaf: %w", err,
+				)
+			}
+
+			if !bytes.Equal(
+				script,
+				pair.ForfeitPath.WitnessScript,
+			) {
+
+				continue
+			}
+
+			encodedLeaf, err := leaf.Encode()
+			if err != nil {
+				return nil, nil, fmt.Errorf(
+					"encode settlement leaf: %w", err,
+				)
+			}
+
+			return bytes.Clone(script), encodedLeaf, nil
+		}
+
+		return nil, nil, fmt.Errorf("forfeit leaf not found in policy")
+	}
+
+	return nil, nil, fmt.Errorf("no settlement pair matches spend path")
+}
+
+// spendPathsMatch reports whether two semantic spend paths describe the same
+// authenticated policy branch. Witness conditions are compared separately at
+// spend time and are not part of the branch identity.
+func spendPathsMatch(a, b *arkscript.SpendPath) bool {
+	switch {
+	case a == nil || b == nil:
+		return false
+
+	case !bytes.Equal(a.WitnessScript, b.WitnessScript):
+		return false
+
+	case !bytes.Equal(a.ControlBlock, b.ControlBlock):
+		return false
+
+	case a.RequiredSequence != b.RequiredSequence:
+		return false
+
+	case a.RequiredLockTime != b.RequiredLockTime:
+		return false
+	}
+
+	return true
 }
--- a/darepod/wallet_ops.go
+++ b/darepod/wallet_ops.go
@@ -369,6 +369,13 @@ func findSettlementOwnerLeaf(template *arkscript.PolicyTemplate,
 		return bytes.Clone(script), encodedLeaf, nil
 	}
 
+	// Defensive fallback: every current Ark policy enumerates both the
+	// auth and forfeit leaves in template.Leaves, so the loop above
+	// always resolves the caller's spend path. This second stage
+	// handles future policies whose auth path is derived rather than
+	// enumerated; spendPathsMatch intentionally ignores runtime
+	// condition witnesses so a claim-style spend with a preimage still
+	// matches its derived AuthPath.
 	pairs, err := template.SettlementPairsForParticipant(
 		participant, operator,
 	)
--- a/lib/tx/oor/submit_signature_validate.go
+++ b/lib/tx/oor/submit_signature_validate.go
@@ -172,6 +172,15 @@ func buildTaprootWitness(in psbt.PInput,
 			witness = append(witness, sigItems[i])
 		}
 
+		// Condition witness items are appended after the signatures
+		// and before the leaf script. This layout matches Ark policy
+		// leaves where the tapscript consumes signatures from the top
+		// of the stack and then consumes condition data pushed below
+		// them (e.g. vHTLC claim: <sig> <preimage> OP_SHA256 <hash>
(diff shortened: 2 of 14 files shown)


bhandras:
@claude review

darepod/wallet_ops.go
```diff
@@ -253,3 +277,126 @@ func BuildCustomTransferInputs(ctx context.Context,
...
+		return false
+	}
+
+	for i := range a.Conditions {
+		if !bytes.Equal(a.Conditions[i], b.Conditions[i]) {
+			return false
+		}
+	}
+
+	return true
```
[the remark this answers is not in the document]
bhandras: Fixed in `820eaf8`. `spendPathsMatch` now compares only the semantic branch definition (`WitnessScript`, `ControlBlock`, sequence, locktime) and intentionally ignores `Conditions`, because those are runtime witness items rather than part of the authenticated branch identity. I also added `TestSpendPathsMatchIgnoresConditionWitness` to pin that behavior.

lib/tx/oor/submit_signature_validate.go
```diff
@@ -192,6 +212,109 @@ func buildTaprootWitness(in psbt.PInput,
...
+			if len(prevData) == schnorr.PubKeyBytesLen {
+				keys = append(keys, bytes.Clone(prevData))
+			}
+		}
+
+		if len(data) > 0 {
+			prevData = bytes.Clone(data)
+		} else {
+			prevData = nil
+		}
```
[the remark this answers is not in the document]
bhandras: Fixed in `820eaf8`. `extractChecksigPubKeys` now keeps the most recent pushed data item across non-data opcodes instead of clearing it, so `CHECKSIG`/`CHECKSIGVERIFY` still see the correct pubkey even when opcodes like `OP_DUP` appear between the push and the checksig opcode. Added regression coverage in `TestOrderTaprootScriptSpendSignaturesSupportsMissingOptionalSignatures`.

lib/tx/oor/submit_signature_validate.go
```diff
@@ -192,6 +212,109 @@ func buildTaprootWitness(in psbt.PInput,
...
+
+		return ordered, nil
+	}
+
+	if len(keys) != len(sigs) {
+		return nil, fmt.Errorf(
+			"taproot signature count %d does not match leaf key count %d",
+			len(sigs), len(keys),
+		)
+	}
```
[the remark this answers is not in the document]
bhandras: Fixed in `820eaf8`. The ordering logic no longer requires `len(keys) == len(sigs)`. It now reconstructs witness items in reverse checksig order, inserts empty placeholders for optional signature slots, and errors only if a provided signature cannot be matched back to a checksig pubkey in the leaf. The same regression test covers the sparse-signature case.

oor/ark_sign.go
```diff
@@ -186,10 +187,29 @@ func signArkPSBTInput(signer input.Signer, arkPSBT *psbt.Packet,
...
+		return err
+	}
+
+	// Preserve condition witness items for custom spends so later submit
+	// validation can reconstruct the full tapscript witness after the
+	// operator attaches its signature.
+	if in.CustomSpend != nil && len(in.CustomSpend.Conditions) > 0 {
+		err = arkscript.PutConditionWitnessPSBTInput(
+			arkPSBT, inputIndex, in.CustomSpend.Conditions[0],
+		)
```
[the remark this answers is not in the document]
bhandras: Fixed in `820eaf8`. We now persist the full condition witness vector in the PSBT metadata instead of only `Conditions[0]`. That change also propagates through checkpoint signing and submit-path witness reconstruction, with coverage in `lib/arkscript/psbt_test.go` and `TestBuildTaprootWitnessIncludesConditionWitness`.

lib/tx/oor/submit_signature_validate.go
```diff
@@ -156,16 +158,34 @@ func buildTaprootWitness(in psbt.PInput,
...
+		for i := range sigItems {
+			witness = append(witness, sigItems[i])
+		}
+
+		conditionWitness, err := arkscript.GetConditionWitnessPSBTInput(
+			in,
+		)
+		switch {
+		case err == nil:
+			witness = append(witness, conditionWitness)
```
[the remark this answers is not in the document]
bhandras: Fixed in `820eaf8`. `buildTaprootWitness` now appends the full decoded condition witness slice when rebuilding the tapscript witness, rather than a single item. I also extended `TestBuildTaprootWitnessIncludesConditionWitness` to cover the multi-item case and added direct condition-witness round-trip tests in `lib/arkscript/psbt_test.go`.

bhandras:
Thanks — I addressed the substantive review items on the branch now at `c6c57b3`.

Fixed:
- `spendPathsMatch` no longer compares `Conditions`, so semantic branch matching is based only on the authenticated path definition.
- the PSBT condition-witness helpers now persist and recover the full witness vector rather than only a single item.
- `ark_sign` / `checkpoint_sign` / submit-path witness reconstruction now preserve multi-item condition witnesses end to end.
- `extractChecksigPubKeys` now keeps the last pushed pubkey across non-data opcodes.
- tapscript signature ordering no longer assumes `len(keys) == len(sigs)` and now tolerates optional checksig slots by reconstructing empty placeholders when needed.
- added direct regression coverage for spend-path matching, condition-witness round-tripping, multi-item witness reconstruction, and sparse optional-signature ordering.

I also cleaned up the follow-up lint fallout from the first push, so `make lint-local` and the targeted package tests are green locally on `c6c57b3`.

I did not widen scope further for the lower-severity observations around `fallbackSchnorrSigner`, the `signOutputRawWithLocator` duplication note, or the broader `GetInfo` behavior question. Those feel like reasonable follow-up cleanup/design discussions, but they are orthogonal to the core policy-backed input/signing correctness fix in this PR.

bhandras:
@claude re-review

bhandras:
@claude re-review after the split

darepod/wallet_ops.go
```diff
@@ -294,3 +319,120 @@ func BuildCustomTransferInputs(ctx context.Context,
...
+		if err != nil {
+			return nil, nil, fmt.Errorf(
+				"encode settlement leaf: %w", err,
+			)
+		}
+
+		return bytes.Clone(script), encodedLeaf, nil
+	}
+
+	pairs, err := template.SettlementPairsForParticipant(
```
Roasbeef, line 372: In which case wouldn't the loop above always find the exact script? Given that it looks at each witness script for a given leaf.

darepod/wallet_ops.go
```diff
@@ -251,11 +257,30 @@ func BuildCustomTransferInputs(ctx context.Context,
...
+					)
+				}
+			}
 		}
 
 		input := oor.TransferInput{
 			VTXO:               desc,
 			VTXOPolicyTemplate: ci.VtxoPolicyTemplate,
+			OwnerLeafScript:    ownerLeaf,
+			OwnerLeafPolicy:    ownerLeafPolicy,
```
Roasbeef, line 283: This looks to be the _actual_ bug fix here. AFAICT, the call to `SettlementPairsForParticipant` should never really be reached in practice.

darepod/wallet_ops.go
Roasbeef: Ah ok I see this clarification in the OP: 
> every current Ark policy enumerates both auth and forfeit leaves in template.Leaves, so the exact-leaf loop always wins. The SettlementPairsForParticipant branch is defensive coverage for future policies whose auth path is derived; add a test that pins the condition-witness-through-resolution contract via a vHTLC claim carrying a preimage.

lib/tx/oor/submit_signature_validate.go
```diff
@@ -211,6 +213,110 @@ func buildTaprootWitness(in psbt.PInput,
...
+	if len(sigs) == 0 {
+		return nil, fmt.Errorf("missing taproot script signatures")
+	}
+
+	keys, err := extractChecksigPubKeys(leafScript)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(keys) == 0 {
```
Roasbeef, line 233: Would this be an error if there're no keys at all? Given that a checksig operator _must_ declare at least one key.

lib/tx/oor/submit_signature_validate.go
```diff
@@ -211,6 +213,110 @@ func buildTaprootWitness(in psbt.PInput,
...
+		)
+	}
+
+	return ordered, nil
+}
+
+// extractChecksigPubKeys returns the x-only pubkeys that are immediately
+// consumed by CHECKSIG/CHECKSIGVERIFY opcodes in the given tapscript.
+func extractChecksigPubKeys(script []byte) ([][]byte, error) {
+	tokenizer := txscript.MakeScriptTokenizer(0, script)
```
Roasbeef, line 293: 👍 

Perfect use of the tokenizer here

lndbackend/client_wallet.go
```diff
@@ -128,6 +129,94 @@ func (c *ClientWallet) SignOutputRaw(tx *wire.MsgTx,
...
+			return nil, fmt.Errorf("sign descriptor %d is nil", i)
+		}
+
+		keyDesc := &signrpc.KeyDescriptor{}
+		if signDesc.KeyDesc.PubKey != nil {
+			keyDesc.RawKeyBytes = signDesc.KeyDesc.PubKey.
+				SerializeCompressed()
+		}
+
+		if signDesc.KeyDesc.KeyLocator.Family != 0 ||
```
Roasbeef, line 160: Alternatively we can also consider patching lndclient?

darepod/wallet_ops.go
```diff
@@ -369,6 +369,13 @@ func findSettlementOwnerLeaf(template *arkscript.PolicyTemplate,
 		return bytes.Clone(script), encodedLeaf, nil
 	}
 
+	// Defensive fallback: every current Ark policy enumerates both the
+	// auth and forfeit leaves in template.Leaves, so the loop above
+	// always resolves the caller's spend path. This second stage
```
Roasbeef, line 374: Nice was going to suggesst adding a comment here

Roasbeef (approved):
LGTM 🔫

PR #286, opened 2026-04-22
multi: thread dynamic fee awareness through board and send-vtxo

Roasbeef:
In this PR, we thread dynamic fee awareness through the client's boarding and send paths, so the operator fee the client reserves matches the amount the server's `validateOperatorFee` expects under a non-zero schedule.

Before this PR, both `Board` and `SendVTXO` deducted the legacy flat `terms.MinOperatorFee` (default 1000 sats) regardless of the operator's runtime fee schedule. That was fine when fees were universally zero, but it breaks in two directions the moment the server runs with an actual schedule: when the dynamic quote is cheaper than the flat fee, the client silently overpays and the excess lands on the operator's side as unattributed profit; when the dynamic quote exceeds the flat fee (as happens under a congestion spread, or when a larger base margin is configured), the server rejects the submission with `ErrOperatorFeeTooLow` and refresh just stops working.

The fix is to ask the operator. Both RPC handlers now call a small `quoteOperatorFee` helper that proxies to `EstimateFee` over the direct gRPC link, and they pass the returned `TotalFeeSat` through to the wallet actor in place of the legacy flat value. If `EstimateFee` is unreachable (nil `serverConn`, RPC failure), we log at warn and fall back to `terms.MinOperatorFee` so a boarding still goes through in a degraded mode rather than failing outright.

This is the client-side of [lightninglabs/darepo#263](https://github.com/lightninglabs/darepo/issues/263). The matching server PR is in the parent repo and bumps the submodule to the tip of this branch as its final commit. Both ship together.

## Dynamic fee quoting

In `darepod/rpc_fees.go`, `quoteOperatorFee` is a private helper that takes `(amountSat, isBoarding, remainingBlocks)` and returns a `btcutil.Amount`. It's deliberately boring: dial through the existing `s.serverConn`, call `arkrpc.ArkServiceClient.EstimateFee`, return `TotalFeeSat`. No caching, no retry, no local fee math. The operator is authoritative for what it will accept, and every cache layer just adds drift we'd have to paper over.

`Board` (in `rpc_server.go`) peeks the wallet's confirmed boarding balance via the existing `GetBoardingBalanceRequest` Ask, quotes `EstimateFee` for that amount, and passes the dynamic quote through `BoardRequest.MinOperatorFee`. There's a small TOCTOU window between the balance peek and the wallet's own intent-build call, but a new boarding arriving in that window can only push the effective amount higher, which the server still accepts (it only rejects under-quotes vs. its required fee).

`SendVTXO` does the same for the summed recipient amount, with `isBoarding=false` and `remainingBlocks=0`. The server's `MinRefreshDeltaBlocks` floor guarantees a stable liquidity fee regardless of the specific VTXO's remaining time, so we don't need to plumb per-VTXO block data through the RPC call.

Back-compat: under a zero schedule, `EstimateFee` returns `TotalFeeSat=0` and the whole flow reduces to today's legacy path. No flag, no gate.

## CLI error mapping

The daemon's `proxyUpstreamError` helper sanitizes upstream gRPC statuses before surfacing them to the CLI, so the code gets preserved but the message is rewrapped. That makes status-code matching brittle, so `mapFeeError` (in `cmd/darepocli/darepoclicommands/fee_errors.go`) matches on substring for the four load-bearing server-side sentinels: `ErrVTXOBelowMinViable`, `ErrOperatorFeeTooLow`, boarding-balance-too-small, and fee-calculator-not-configured. Each mapped error surfaces as a concise actionable line, with the original server message kept as parenthetical context so operators can still grep the server logs.

The `board` and `send inround` subcommands run their RPC error through `mapFeeError` before falling back to the generic wrap. `fee_errors_test.go` pins the exact substrings we depend on, so a future server-side message change fails the CLI test rather than silently routing around the mapper.

## Testing

Four new test files, all running as unit tests (no itest harness required):

In `round/fees_invariants_test.go`, rapid properties cover `computeClientOperatorFee`: non-negative clamp under any intent shape, conservation identity (`fee = inputs - outputs` when outputs don't exceed inputs), monotonicity (adding a boarding input never decreases the fee; adding an owned output never increases it), and the empty-intents zero-fee baseline. The generator caps per-entry amounts at 1B sat so running sums can't overflow int64 on realistic slice lengths.

In `round/refresh_psbt_size_test.go`, the PSBT cardinality invariant is locked in both as a canonical fixed-case test and as a rapid property: across random `(forfeit amount, fee amount)` draws, the number of forfeits, owned outputs, and leaves stays constant, and only per-output values move. A regression that added a fee-only output or dropped one would break this immediately.

In `darepod/congestion_roundtrip_test.go`, a `bufconn`-backed `fakeArkService` returns canned `EstimateFee` replies, and the client's `quoteOperatorFee` helper is driven through a baseline -> congestion-bumped sequence. The test asserts the helper returns the operator's `TotalFeeSat` verbatim (no local caching) and pins the `nil serverConn` guard so the helper surfaces a clean `Unavailable` error instead of panicking.

In `darepod/rpc_fees_e2e_test.go`, a similar bufconn setup exercises the daemon's `EstimateFee` RPC handler: every field (liquidity, on-chain share, margin, total, effective rate, min viable, below-dust flag) must proxy verbatim to the CLI caller, and the CLI-visible total must match the internal `quoteOperatorFee` number for the same inputs. This is the invariant that keeps the CLI confirmation prompt aligned with the actual booking.

## Test plan

- [x] \`make lint-native\`
- [x] \`go test ./darepod/... ./round/... ./cmd/darepocli/darepoclicommands/...\`
- [ ] \`make itest\` in the parent darepo repo once the server-side PR lands together

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_board.go
+++ b/cmd/darepocli/darepoclicommands/cmd_board.go
@@ -39,6 +39,13 @@ func board(cmd *cobra.Command, _ []string) error {
 
 	resp, err := client.Board(context.Background(), req)
 	if err != nil {
+		// Map well-known server-side fee rejections to a
+		// concise CLI message. Fall through to the generic
+		// error wrap if the cause is not a fee rejection.
+		if feeErr := mapFeeError(err); feeErr != nil {
+			return feeErr
+		}
+
 		return fmt.Errorf("board RPC failed: %w", err)
 	}
 
--- a/cmd/darepocli/darepoclicommands/cmd_send.go
+++ b/cmd/darepocli/darepoclicommands/cmd_send.go
@@ -111,6 +111,13 @@ func sendInRound(cmd *cobra.Command, _ []string) error {
 		context.Background(), req,
 	)
 	if err != nil {
+		// Map well-known server-side fee rejections to a
+		// concise CLI message. Fall through to the generic
+		// error wrap if the cause is not a fee rejection.
+		if feeErr := mapFeeError(err); feeErr != nil {
+			return feeErr
+		}
+
 		return fmt.Errorf("SendVTXO RPC failed: %w", err)
 	}
 
--- a/cmd/darepocli/darepoclicommands/fee_errors.go
+++ b/cmd/darepocli/darepoclicommands/fee_errors.go
@@ -1,0 +1,88 @@
+package darepoclicommands
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Error substrings used to recognize server-side fee rejections.
+// These match the sentinel messages in
+// darepo/rounds/validation.go so the client can surface a concise,
+// actionable CLI error instead of a raw gRPC status that leaks
+// internal percentages and numeric fields.
+//
+// Matching is on substring rather than status code because the
+// daemon wraps upstream errors through proxyUpstreamError, which
+// sanitizes the code and re-wraps the message.
+const (
+	errMsgOperatorFeeTooLow  = "operator fee is below minimum"
+	errMsgVTXOBelowMinViable = "VTXO amount is below minimum viable"
+	// errMsgBoardingTooSmall pins the wallet's exact "balance is
+	// too small after operator fee" rejection at
+	// wallet/wallet.go:1130. We deliberately do NOT match the
+	// shorter "boarding balance" substring because the Board RPC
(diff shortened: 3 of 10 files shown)


Roasbeef:
@claude review this 

@codex review this

Issue #269, opened 2026-04-16
unroll: verify deep OOR chain broadcast ordering (checkpoint→ark alternation)


Roasbeef:
## Summary

For preconfirmed VTXOs received through a chain of OOR transactions, the
unilateral exit must broadcast transactions in a specific alternating
order: `checkpoint_1 → ark_1 → checkpoint_2 → ark_2 → ...`. Each
checkpoint must confirm before the next Ark transaction can be broadcast,
because the Ark transaction spends the checkpoint output.

This issue tracks verifying that `lib/recovery` + `unrollplan` + `unroll/`
correctly handle OOR chains of depth >= 2, and adding explicit test
coverage for the ordering requirement.

## Spec Reference

ARK-05 §"Step 3: Broadcast Transaction Chain":

> **Important:** For OOR chains deeper than 1, the broadcasts MUST follow
> the alternating checkpoint→Ark pattern. Broadcasting all checkpoints
> first and then all Ark transactions would fail because each Ark
> transaction spends from its preceding checkpoint output.

## Current State

- `lib/recovery` (PR #261) represents the recovery proof graph with
  topological layering. Each `Node` has a `NodeKind` (Branch, Checkpoint,
  Ark, Leaf) and the proof graph tracks parent/child relationships.
- `unrollplan` (PR #261) is a pure planner that determines broadcast order
  based on dependency resolution and current chain state.
- `unroll/actor.go` (PR #263) drives the FSM through the planner's output.
- The integration test `TestUnilateralExitOORDerivedCompletion` (server
  PR #219) exercises Alice→Bob 1-hop OOR. This is depth=1 (one
  checkpoint + one ark tx), which doesn't exercise the alternation
  constraint.

### 1. Code Audit: Planner Ordering

Review `unrollplan/planner.go` to verify that when computing
`ReadyToBroadcast` transactions, it enforces:
- Checkpoint N must be confirmed before Ark N is marked ready.
- Ark N must be confirmed before Checkpoint N+1 is marked ready.
- The dependency graph in `lib/recovery` correctly encodes these
  relationships.

### 2. Depth-2+ OOR Chain Test

Create a test scenario:
```
Alice → Bob (OOR #1) → Carol (OOR #2)

Carol triggers unilateral exit:
  1. Broadcast VTXT path (root → leaf) for Alice's original VTXO
  2. Broadcast checkpoint_1 (Alice→Bob OOR checkpoint)
  3. Wait for checkpoint_1 confirmation
[...]

Issue #268, opened 2026-04-16
wallet: boarding UTXO unilateral recovery via CSV timeout path


Roasbeef:
## Summary

The client currently has no implementation for unilaterally reclaiming a
boarding UTXO when the operator is unresponsive. Boarding UTXOs use a
2-of-2 multisig with a CSV timeout fallback — after the timeout elapses,
the client can unilaterally spend the boarding UTXO back to their wallet
using only their own key.

This is listed as "Work Stream 4: Client Recovery" in the server-side
ChainResolver epic (darepo#57) and is a spec requirement in ARK-05.

## Spec Reference

ARK-05 §"Unilateral Exit Procedure" and §"Exit Capability":

> **Always maintain exit capability**: Never lose ability to exit unilaterally.

The boarding address script (from `lib/types/boarding.go`) has the structure:
```
<operator_key> OP_CHECKSIGVERIFY <client_key> OP_CHECKSIG  // collaborative
<client_key> OP_CHECKSIG OP_CSV <timeout>                  // timeout fallback
```

After `<timeout>` blocks from the UTXO's confirmation, the client can spend
via the timeout tapscript leaf using only their signing key.

## Current State

- `wallet/` manages boarding addresses and tracks boarding UTXO status. The
  `BoardingStatusExpired` status exists (`wallet/interfaces.go:248`),
  indicating the wallet already detects when a boarding UTXO's CSV timeout
  has elapsed.
- `lib/types/boarding.go` has `BoardingDescriptor` with the boarding address
  script construction.
- `db/sqlc/migrations/000002_boarding_tables.up.sql` persists boarding UTXOs.
- **No code constructs or broadcasts the timeout-path spend transaction.**
- The unroll subsystem (PRs #260-#264) handles VTXO unilateral exit but
  does not touch boarding UTXOs — they're a separate script type with a
  simpler recovery path.

### 1. Boarding Recovery Transaction Builder

```
func BuildBoardingRecoveryTx(
    boardingUTXO BoardingDescriptor,
    clientKey *btcec.PrivateKey,
    destAddr btcutil.Address,
    feeRate chainfee.SatPerVByte,
) (*wire.MsgTx, error):

    tx = wire.NewMsgTx(2)
    tx.AddTxIn(&wire.TxIn{
        PreviousOutPoint: boardingUTXO.Outpoint,
        Sequence:         boardingUTXO.CSVTimeout,  // enables CSV check
    })
[...]

PR #293, opened 2026-04-23
multi: quote operator fee on refresh for #269

Roasbeef:
In this PR, we close the client-side half of the refresh fee gap that
will be exposed once the server-side fix for
[lightninglabs/darepo#269](https://github.com/lightninglabs/darepo/issues/269)
lands. Today the client's refresh flows don't quote the operator fee at
all: the implicit operator fee on the round (sum of forfeit inputs minus
sum of new VTXO outputs) is zero. That's fine under a zero fee
schedule, but the server's extension of `validateOperatorFee` to refresh
and send paths turns the same silent-zero into a hard rejection under
any non-zero schedule, which for the auto-refresh path would push VTXOs
toward unilateral exit. We fix both refresh paths so the client's
implicit fee matches what the server's per-input `ComputeForfeitFee`
check expects.

Refresh on the client comes in two flavors and they travel through
entirely different paths. Both needed fixing.

## Manual refresh (daemon RPC → wallet actor)

For `darepocli refresh-vtxos`, the daemon now pre-quotes the operator
fee per target outpoint before handing the request to the wallet actor.
`RefreshVTXOs` resolves each VTXO via `vtxoStore.GetVTXO`, asks
`chainBackend.BestBlock` for the current tip, and calls
`quoteOperatorFee(amount, isBoarding=false, remainingBlocks=BatchExpiry - currentHeight)`
per input. The results land on a new outpoint-keyed
`OperatorFees map[wire.OutPoint]btcutil.Amount` field on
`wallet.RefreshVTXOsRequest`. `handleRefreshVTXOs` then deducts each
per-input fee from its matching new VTXO's `Amount`. The aggregate
identity `sum(forfeits) - sum(new_vtxos) = sum(per-input quotes)`
holds by construction, which is exactly what the server's
`validateOperatorFee` will compare against.

Outpoint-keyed (rather than index-parallel) so the wallet handler
looks up each input's fee directly without depending on the
slice-ordering invariant matching between the daemon and the wallet.
Degraded-mode policy mirrors `Board` and `SendVTXO`: on quote RPC
failure or operator-terms fetch failure, the helper falls back to
`terms.MinOperatorFee` per outpoint and logs at `WarnS`, so a partially
unreachable operator doesn't leave the refresh stuck at zero implicit
fee.

## Auto-refresh (VTXO actor → round actor, near batch expiry)

The auto-refresh path bypasses the wallet entirely. When a VTXO hits
`ExpiryStatusNeedsRefresh`, the FSM emits a `ForfeitRequest` on the
outbox, and the VTXO actor's outbox processor translates it to a
`round.RefreshVTXORequest` that goes through the manager to the round
actor. To make that translation fee-aware we add a
`RefreshFeeQuoter` hook on `VTXOActorConfig` and a matching field on
`ManagerConfig` that propagates into every spawned VTXO actor.

The daemon wires `autoRefreshFeeQuoter()`, a closure that wraps
`s.quoteOperatorFee` with an `s.fetchOperatorTerms` lookup for the
`MinOperatorFee` fallback. At auto-refresh time the actor reads the
VTXO's amount and computes remaining blocks as
`vtxo.BatchExpiry - state.LastCheckedHeight` (the last block epoch the
actor observed in `LiveState`), passes both to the quoter, and puts
the result on `round.RefreshVTXORequest.OperatorFee`. On the round
side, `buildVTXORequestFromRefresh` subtracts `req.OperatorFee` from
the new VTXO's `Amount` with a defensive clamp to zero so a buggy fee
never underflows the new output.

When the quoter is nil (tests, or operators running a zero schedule)
the auto-refresh emits `OperatorFee=0`, preserving pre-#269 behavior.

## Why `remainingBlocks` matters

Both paths pass the real remaining-blocks value to the quote, not
zero. The server's liquidity fee scales with how much of the VTXO's
lifetime is left (via `ComputeForfeitFee`). If the client quotes with
`remainingBlocks=0`, the server falls back to `SweepDelay` as the
default, which models a fresh full-cycle VTXO and over-quotes.
Over-quoting still validates (implicit fee >= expected) but it
over-charges the client on every refresh. Passing
`BatchExpiry - currentHeight` matches the server's own
`ExpiryHeight - currentHeight` computation at validation time, so the
quote and the validation see the same inputs.

## Test plan

- `TestProcessOutboxForfeitRequestQuotesFee` (vtxo) asserts the
  auto-refresh path threads the VTXO amount and
  `BatchExpiry - LastCheckedHeight` into the quoter and populates
  `OperatorFee` on the relayed request.
- `TestProcessOutboxForfeitRequest` (vtxo, existing) keeps covering
  the pre-#269 path: with no quoter configured, `OperatorFee` is 0
  and the forfeit input's `Amount` is unchanged.
- `TestRefreshDeductsPerInputOperatorFee` (wallet) drives a
  two-outpoint manual refresh with distinct per-input fees and
  asserts the aggregate implicit fee `sum(forfeits) - sum(new_vtxos)`
  equals the sum of per-input quotes.
- Full suite: `go test ./vtxo/... ./round/... ./wallet/... ./darepod/...`
  and `make lint-native` are both clean.

## Cross-reference

Client half of [lightninglabs/darepo#269](https://github.com/lightninglabs/darepo/issues/269).
The server half lands as part of a bundle on lightninglabs/darepo
covering #267 (chain-backed fee estimator),
#268 (at-cost batch sizing), and #269 (forfeit-path validation).
Once this PR merges, the server PR bumps the submodule pointer.

### DIFF
--- a/darepod/congestion_roundtrip_test.go
+++ b/darepod/congestion_roundtrip_test.go
@@ -26,19 +26,59 @@ type fakeArkService struct {
 	// changes.
 	response *arkrpc.EstimateFeeResponse
 
+	// estimateFeeErr, when non-nil, is returned from EstimateFee
+	// instead of the canned response. Used by degraded-mode tests
+	// to simulate an operator that is reachable for GetInfo but
+	// fails the fee quote.
+	estimateFeeErr error
+
+	// getInfoResponse is the canned GetInfo reply. Optional;
+	// tests that don't exercise the fetchOperatorTerms path can
+	// leave it nil, in which case GetInfo falls through to the
+	// embedded Unimplemented server and returns an error.
+	getInfoResponse *arkrpc.GetInfoResponse
+
+	// getInfoErr, when non-nil, is returned from GetInfo instead
+	// of the canned response. Used to exercise the degraded-mode
+	// fetchOperatorTerms-failed branch.
+	getInfoErr error
+
 	// lastRequest records the most recent EstimateFeeRequest
 	// so tests can assert what the client asked for.
 	lastRequest *arkrpc.EstimateFeeRequest
 }
 
-// EstimateFee records the request and returns the canned reply.
+// EstimateFee records the request and returns the canned reply
+// (or an error, if the test injected one).
 func (f *fakeArkService) EstimateFee(_ context.Context,
 	req *arkrpc.EstimateFeeRequest) (
 	*arkrpc.EstimateFeeResponse, error) {
 
 	f.lastRequest = req
 
+	if f.estimateFeeErr != nil {
+		return nil, f.estimateFeeErr
+	}
+
 	return f.response, nil
+}
+
+// GetInfo returns the canned reply (or an injected error). When
+// both fields are nil the call falls through to the embedded
+// Unimplemented server, which itself returns an error — useful
+// for tests that want fetchOperatorTerms to fail via the default.
+func (f *fakeArkService) GetInfo(ctx context.Context,
+	req *arkrpc.GetInfoRequest) (*arkrpc.GetInfoResponse, error) {
+
+	if f.getInfoErr != nil {
+		return nil, f.getInfoErr
+	}
+
+	if f.getInfoResponse != nil {
+		return f.getInfoResponse, nil
+	}
+
+	return f.UnimplementedArkServiceServer.GetInfo(ctx, req)
 }
 
 // newBufconnClient builds a *grpc.ClientConn wired to an
--- a/darepod/rpc_fees.go
+++ b/darepod/rpc_fees.go
@@ -8,6 +8,7 @@ import (
 	"github.com/lightninglabs/darepo-client/arkrpc"
 	"github.com/lightninglabs/darepo-client/daemonrpc"
 	"github.com/lightninglabs/darepo-client/db/sqlc"
+	"github.com/lightninglabs/darepo-client/vtxo"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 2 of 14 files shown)


Roasbeef:
@claude review this

@codex review this

Roasbeef:
Added a second commit `a02d0817` that extends the same per-input forfeit fee plumbing to the cooperative leave path so it doesn't break once the server-side #269 gate widens to `len(boardingInputs) > 0 || len(forfeitInputs) > 0`. Without this, every leave would be rejected with `ErrOperatorFeeTooLow`.

## Leave-path changes

- `wallet/messages.go`: add `OperatorFees map[wire.OutPoint]btcutil.Amount` to `LeaveVTXOsRequest`, mirroring the refresh request shape.
- `wallet/wallet.go`: derive per-leaf output value as `vtxo.Amount - OperatorFees[outpoint]` and build a fresh `*wire.TxOut` per leave (preserving the caller's pkScript). When `OperatorFees` is empty, fall back to the legacy single-`DestOutput` behavior so existing zero-fee callers keep working under a zero schedule.
- Reject negative fees and fees `>= vtxo.Amount` so a buggy quoter can't silently invert fee direction or emit garbage outputs.

Three new tests in `wallet/wallet_admission_test.go`:

- `TestLeaveDeductsPerInputOperatorFee` — two leaves with different per-input fees, asserts each leaf carries `vtxo.Amount - fee` and aggregate implicit fee equals the per-input quote sum.
- `TestLeaveRejectsNegativeOperatorFee`
- `TestLeaveRejectsFeeAtOrAboveAmount`

Both cooperative paths that produce forfeit inputs (refresh + leave) now thread explicit per-input fees matching the server's `ComputeForfeitFee` expectation.

@claude review this
@code review this

Issue #270, opened 2026-04-16
unroll: exit timing pre-flight guard before batch sweep deadline


Roasbeef:
## Summary

Before starting a unilateral exit, the client must compute whether there
is sufficient time to complete the full broadcast chain + CSV wait + sweep
before the batch sweep deadline (`T_e`). If the remaining time is
insufficient, the client should refuse the exit and alert the user rather
than starting a doomed unroll that wastes on-chain fees.

## Spec Reference

ARK-05 §"Exit Timing Requirements":

> **CRITICAL:** Unilateral exit MUST complete before batch sweep deadline.
>
> ```
> required_time = chain_length * avg_confirmation_time
>               + csv_delay * avg_block_time
>               + safety_margin
>
> assert current_height + required_time < batch_sweep_deadline
> ```
>
> If insufficient time remains:
> - **Alert user immediately**
> - **Use aggressive fees**
> - **Consider partial exit** (if multiple VTXOs)

## Current State

- `unrollplan/planner.go` (PR #261) computes which transactions are ready
  to broadcast based on current chain state, but does not evaluate whether
  the full sequence can complete before `T_e`.
- `unroll/registry.go` (PR #263) accepts unroll requests and spawns per-
  target actors. No pre-flight feasibility check.
- The VTXO descriptor stores `BatchExpiry` (the block height at which the
  operator can sweep), so the information is available.

### 1. Exit Feasibility Calculator

```
func CheckExitFeasibility(
    vtxo VTXODescriptor,
    currentHeight uint32,
    proofDepth uint32,       // number of txs in broadcast chain
    csvDelay uint32,         // t_e blocks
    avgConfTime uint32,      // estimated blocks per confirmation
    safetyMargin uint32,     // buffer blocks (default: 6)
) ExitFeasibility:

    requiredBlocks = proofDepth * avgConfTime + csvDelay + safetyMargin
    deadline = vtxo.BatchExpiry
    availableBlocks = deadline - currentHeight

    if availableBlocks < requiredBlocks:
        return ExitFeasibility{
            Feasible: false,
            Required: requiredBlocks,
            Available: availableBlocks,
            Deficit: requiredBlocks - availableBlocks,
        }
    return ExitFeasibility{Feasible: true, ...}
```

### 2. Registry Pre-Flight Check

Before spawning a `VTXOUnrollActor`, the registry should:

```
func (r *Registry) RequestUnroll(outpoint) error:
[...]

PR #296, opened 2026-04-24
daemonrpc+darepod+darepocli: wire cooperative leave end-to-end for #294

### LINKED ISSUE #294: darepod: wire cooperative leave end-to-end (RPC + CLI + fee quoter) for itest coverage
## Context

`wallet.LeaveVTXOsRequest` and `handleLeaveVTXOs` exist, and after #293 the
wallet-side fee plumbing (`OperatorFees map[wire.OutPoint]btcutil.Amount`)
is in place. But the cooperative leave path has no user-facing surface:

- No `LeaveVTXOs` RPC in `daemonrpc/daemon.proto`.
- No `darepocli leave` / `darepocli offboard` CLI command.
- No daemon-side `quoteLeaveOperatorFees` helper that pre-quotes the
  per-outpoint operator fee the way `quoteRefreshOperatorFees` does for
  refresh (see `darepod/rpc_server.go`).

As a result, `wallet.LeaveVTXOsRequest` is only constructed in tests today.

## Why now

The server's existing cooperative-leave coverage is only in
`darepo/systest/leave_e2e_test.go`. There is no `darepo/itest/leave_test.go`
because the itest pattern (see `darepo/itest/refresh_test.go`) drives a real
`darepod` via `daemonrpc`, and without a `daemonrpc.LeaveVTXOs` RPC on the
client we can't write that itest.

Once server-side #269 lands and widens `validateOperatorFee` to fire on
forfeit inputs, leave becomes fee-gated. A server itest that exercises a
full cooperative leave under a non-zero fee schedule is the right place
to catch regressions, but it's blocked on the client exposing the RPC.

## Proposed scope

**`daemonrpc/daemon.proto`**
- Add `LeaveVTXOs(LeaveVTXOsRequest) returns (LeaveVTXOsResponse)` mirroring
  the refresh RPC shape (OutpointSelection oneof + dry_run + response with
  queued outpoints + per-outpoint errors).

**`darepod/rpc_server.go`**
- Add `LeaveVTXOs` RPC handler. Parse outpoints, resolve per-outpoint
  destination pkScript (reuse `resolveRecipientOutput` or extend it for
  on-chain offboard destinations), call `quoteLeaveOperatorFees`, dispatch
  `wallet.LeaveVTXOsRequest` with populated `OperatorFees`.
- Add `quoteLeaveOperatorFees(ctx, targets) map[OutPoint]btcutil.Amount`
  mirroring `quoteRefreshOperatorFees`: fetch terms, fetch best block,
  call `quoteOperatorFee` per outpoint with the correct
  `remainingBlocks = BatchExpiry - currentHeight`, degraded-mode fall back
  to `MinOperatorFee`.

**`cmd/darepocli/darepoclicommands/`**
- Add `leave.go` (or `offboard.go`) with the CLI surface. Follows the
  existing `refresh-vtxos` command pattern.

**Testing**
- Unit tests for `quoteLeaveOperatorFees` degraded-mode branches (mirror
  `TestAutoRefreshFeeQuoter*` in `darepod/rpc_fees_test.go` from #293).
- Once landed here, open a follow-up in `darepo` to add
  `itest/leave_test.go` using the same harness pattern as
  `itest/refresh_test.go`.

## Reference

See fee-model.md (server repo): offboard is priced the same as refresh
(liquidity fee + on-chain share + margin). #269 on the server widens the
fee gate to every round with `len(boardingInputs) > 0 || len(forfeitInputs) > 0`,
so unlocking the client-side leave RPC without this quoter would trip
`ErrOperatorFeeTooLow` on every production leave.

#293 contains the wallet-side plumbing this issue depends on.

Roasbeef:
Closes #294.

## Summary

Wires `wallet.LeaveVTXOsRequest` (which already existed) through the full daemon stack so cooperative leave has a user-facing surface. Before this PR, `LeaveVTXOsRequest` was only constructed in tests; there was no daemonrpc method, no CLI, and no MCP tool — leave was an internal abstraction with no way to invoke it. The seal-time fee handshake (#270) already shipped, so the RPC layer does no client-side fee quoting; the server stamps the residual onto the `IsChange=true` leave output at seal time.

Five logical commits plus a fixup pass:

1. **`daemonrpc: add LeaveVTXOs RPC messages and service method`**
   New `LeaveDestination` oneof (`address` | `pk_script`), `LeaveVTXOsRequest` (`OutpointSelection`/`all` + `default_destination` + per-outpoint `destinations` map + `dry_run`), `LeaveVTXOsResponse`. Shape mirrors `RefreshVTXOs`; the per-outpoint override map unlocks batched offboards to distinct on-chain targets in one round.

2. **`wallet: allow per-outpoint destinations on LeaveVTXOsRequest`**
   Additive `DestOutputs map[wire.OutPoint]*wire.TxOut` field. `handleLeaveVTXOs` prefers overrides, falls back to singular `DestOutput`, surfaces a per-outpoint error if both are missing. Existing tests and the systest `TriggerVTXOLeave` helper keep working unchanged.

3. **`darepod: add LeaveVTXOs RPC handler`**
   Handler mirrors `RefreshVTXOs` but reorders pre-dispatch validation to run before the wallet-ready gate (malformed requests should always look like client bugs). New `resolveLeaveDestination` helper with an explicit `IsForNet` check — `btcutil.DecodeAddress` otherwise accepts a mainnet address under regtest `defaultNet`, which is a real footgun for a funds-moving RPC. The raw `pk_script` branch is length-capped (`txscript.MaxScriptSize`), class-whitelisted to standard forms (P2PKH/P2SH/P2WPKH/P2WSH/P2TR/P2PK/multisig/`OP_RETURN`), and rejects the BIP 431 P2A anchor pattern explicitly so a typo'd or hostile script can't silently burn funds. Stale destination keys (well-formed outpoints not in the selection) are rejected with `InvalidArgument` rather than silently dropped, and `--all --dry_run` enumerates live VTXOs into the preview so the user sees what would actually be queued.

4. **`darepocli: add vtxos leave subcommand`**
   `darepocli vtxos leave` follows the `vtxos refresh` pattern. `--all` is gated behind a y/N stdin confirmation unless `--yes` is passed. The confirmation gate runs after `parseRequest` returns so it covers both the flag-driven path and the `--json` path; previously the prompt lived inside the flag callback and an agent piping `{"all":true,...}` would silently bypass it. `--destination outpoint=addr` or `--destination outpoint=script:<hex>` supplies per-outpoint overrides. Flag parsing is extracted to `buildLeaveVTXOsRequest` so CLI and MCP stay in lock-step.

5. **`darepocli: wire leave into MCP and schema registry`**
   `vtxos_leave` MCP tool plus `vtxos.leave` schema registry entry. `make schema-check` passes.

## Test plan

- [x] `make unit pkg=./wallet/... timeout=5m` — Leave tests including `TestLeavePerOutpointDestinations`, `TestLeaveFallsBackToDefaultDestWhenMapMisses`, `TestLeaveRejectsMissingDestination`.
- [x] `make unit pkg=./darepod/... timeout=5m` — `rpc_leave_test.go` covers `resolveLeaveDestination` (address / pk_script / oversized / P2A / non-standard / OP_RETURN / cross-network / nil), every `LeaveVTXOs` pre-dispatch validation branch (all+overrides, extraneous destination keys, missing destination, malformed outpoint, malformed destination key, dry_run preview, all+dry_run preview).
- [x] `make unit pkg=./cmd/darepocli/... timeout=5m` — 18 tests in `cmd_vtxos_leave_test.go` covering every flag-combination validation path plus the JSON-bypass regression for the `--all` confirmation gate (`TestConfirmLeaveAllIfNeededJSONStillPrompts` and friends).
- [x] `make unit pkg=./daemonrpc/... timeout=5m`.
- [x] `make lint-native` (0 issues).
- [x] `make schema-check`.
- [x] `go build ./...` (clean for new code; unrelated `runtime.main_main` warning on root directory is pre-existing).
- [x] `darepocli vtxos leave --help` smoke.
- [ ] Regtest real-leave round-trip — deferred to the parent-repo itest follow-up so coverage lives where the harness pattern already is (`itest/refresh_test.go`).

## Follow-ups

- Parent-repo itest (`itest/leave_test.go`) driving this RPC end-to-end under a non-zero fee schedule.
- Parent-repo submodule pointer bump once this merges.

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_mcp.go
+++ b/cmd/darepocli/darepoclicommands/cmd_mcp.go
@@ -264,6 +264,44 @@ func registerMCPTools(s *mcp.Server,
 		return r, nil, err
 	})
 
+	// vtxos_leave — cooperative leave (offboard).
+	type vtxosLeaveArgs struct {
+		Outpoints    []string          `json:"outpoints,omitempty" jsonschema:"VTXO outpoint(s) to leave (txid:index)"`                                     //nolint:ll
+		All          bool              `json:"all,omitempty" jsonschema:"leave all live VTXOs"`                                                             //nolint:ll
+		Address      string            `json:"address,omitempty" jsonschema:"default on-chain destination address"`                                         //nolint:ll
+		PkScript     string            `json:"pk_script,omitempty" jsonschema:"default destination pk_script as hex; alternative to address"`               //nolint:ll
+		Destinations map[string]string `json:"destinations,omitempty" jsonschema:"per-outpoint override map; values are either an address or script:<hex>"` //nolint:ll
+		DryRun       bool              `json:"dry_run,omitempty" jsonschema:"validate without queuing"`                                                     //nolint:ll
+	}
+	mcp.AddTool(s, &mcp.Tool{
+		Name:        "vtxos_leave",
+		Description: "Queue VTXOs for cooperative leave (offboard)",
+	}, func(ctx context.Context,
+		req *mcp.CallToolRequest,
+		args vtxosLeaveArgs) (*mcp.CallToolResult, any, error) {
+
+		// Reuse the CLI's builder so CLI and MCP can't drift on
+		// destination parsing — a divergence would let the two
+		// surfaces offboard to subtly different scripts.
+		rpcReq, err := buildLeaveVTXOsRequest(
+			args.Outpoints, args.All,
+			args.Address, args.PkScript,
+			args.Destinations, args.DryRun,
+		)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		resp, err := client.LeaveVTXOs(ctx, rpcReq)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		r, err := mcpResult(resp)
+
+		return r, nil, err
+	})
+
 	// Register send tools in a separate function to keep each
 	// registration function under the line limit.
 	registerMCPSendTools(s, client)
--- a/cmd/darepocli/darepoclicommands/cmd_vtxos.go
+++ b/cmd/darepocli/darepoclicommands/cmd_vtxos.go
@@ -1,7 +1,9 @@
 package darepoclicommands
 
 import (
+	"bufio"
 	"context"
+	"encoding/hex"
 	"fmt"
 	"strings"
 
@@ -23,6 +25,7 @@ func newVTXOsCmd() *cobra.Command {
 	cmd.AddCommand(
 		newVTXOsListCmd(),
 		newVTXOsRefreshCmd(),
+		newVTXOsLeaveCmd(),
 	)
 
 	return cmd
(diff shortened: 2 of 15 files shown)


Roasbeef:
@claude review this

@codex review this

PR #305, opened 2026-04-27
timeout: add recurring-tick scheduler

Roasbeef:
In this PR, we extend the `timeout` package with a recurring-tick scheduler so the arkd round actor (and any future caller that needs cadence) can drive itself off an actor-mediated heartbeat instead of having to spawn its own goroutine. Motivation is lightninglabs/darepo#287: arkd's round actor is fully event-driven today, so an empty round (no clients ever joined) sits open until someone hits the admin `TriggerBatch` RPC. The consumer PR in lightninglabs/darepo (branch `rounds-periodic-tick`, link to follow) wires this scheduler into the rounds FSM as a new `TickEvent`.

Because recurring delivery semantics are different from the existing one-shot timeouts (the entry has to live across multiple fires, has its own goroutine, and stops differently), we split the actor's single timer map into two: `oneshots` for the existing `ScheduleTimeoutRequest` entries and `recurring` for the new `ScheduleRecurringTickRequest` entries. ID space is shared, so `CancelTimeoutRequest` works on either shape and re-scheduling against a live ID stops whatever was there before.

The four commits land the change incrementally; see each commit message for the why w.r.t the specific piece.

## What's new

`ScheduleRecurringTickRequest` carries an interval and a `TellOnlyRef[*TickFiredMsg]`. Each tick delivers a fresh `TickFiredMsg{ID, FiredAt}`, where `FiredAt` is the scheduler's observed fire time (under a fake clock that's the test's logical time, which makes property tests on monotonicity and exact tick count straightforward). First fire is at `Interval`, not immediate, matching `time.NewTicker` semantics.

`MapTickFired` mirrors `MapTimeoutExpired` so consumers can adapt `*TickFiredMsg` into their own actor message type without rolling their own `MapInputRef` boilerplate.

`Clock` is a small interface (`Now`, `AfterFunc`, `NewTicker`) that the actor now consumes. `RealClock` is the production impl and is wired implicitly by the existing no-arg `NewActor()` constructor, so the five existing callsites in this repo and in lightninglabs/darepo continue to compile unchanged. Tests use the new `NewActorWithClock(clock)` to drive the scheduler with a fake clock and assert exact tick counts.

## Tests

`actor_recurring_test.go` covers the canonical scenarios: exact tick count over a fixed window, cancel stops further fires (with `handleCancel` waiting for the goroutine to exit), re-scheduling against the same ID stops the previous ticker, mixed one-shot+recurring entries share an ID space, and scheduling a one-shot with an ID currently held by a recurring entry stops the recurring goroutine.

`property_test.go` adds a rapid-driven invariant: for any `(interval, totalAdvance, cancelAt)`, the number of `TickFiredMsg` delivered to the callback equals `floor(min(cancelAt, totalAdvance) / interval)`. 100 random iterations per invocation with intervals from 10ms-200ms and windows from 0-2s.

All seventeen tests in the package pass under `-race`.

## Cross-repo

This PR is half of a pair. Consumer is lightninglabs/darepo branch `rounds-periodic-tick`, which adds a new `TickEvent` to the rounds FSM, schedules it at round creation, gates seal on `>=1 client + SealPredicate`, and exposes `--rounds.tickinterval` (default 1m). Marking this draft until the consumer PR is up so reviewers can see the full picture.

## TODO

- [ ] Open consumer PR in lightninglabs/darepo and link here.
- [ ] Mark ready for review once consumer is up.


Roasbeef:
Consumer PR is up: lightninglabs/darepo#288 (rounds: add periodic TickEvent for round cadence). Marking that one draft until this lands so reviewers can see the pair.

Roasbeef:
@claude review this 

@codex review this

Roasbeef:
@code review 

@claude review

Issue #311, opened 2026-04-28
unroll: persist per-AncestryPath total_levels for cross-batch sanity checking


Roasbeef:
# Follow-up: persist per-AncestryPath total_levels on the unrolls control-plane row

## Context

PR #183 review feedback from @ellemouton called for persisting `total_levels`
on the `unrolls` table for cross-batch sanity checking — when the unroller
broadcasts a tree path it should know up front how many levels deep the path
goes so a partial broadcast (e.g. crash mid-flight) can be detected and
recovered without rewalking the whole proof graph.

With multi-input cross-round OOR (PR for issue #199) the producer side now
emits multiple tree-rooted ancestry fragments per VTXO. The persisted
`total_levels` therefore needs to become **per-fragment**, not a single
scalar — each commitment tree has its own depth and the unroller must be
able to detect partial progress on each independently.

## Scope

- Schema: extend `unrolls` (or split out a sibling table `unroll_ancestry_paths`)
  with one row per `(unroll_outpoint, fragment_index)` carrying
  `total_levels` for that fragment.
- Code: `client/unroll/db_store.go` — extend `UnrollRecord` to carry the
  per-fragment level counts, persist on insert, hydrate on
  `RestoreNonTerminal`.
- Tests: `client/unroll/db_store_test.go` — round-trip a multi-fragment
  record.

## Out of scope

- Replacing the existing single-tree path semantics in any other
  module (this is a strict additive change to `unrolls`).
- Changing the on-chain broadcast logic — the unroller already walks
  `LineageMaterial.TreePaths` per-fragment, so the only change is durable
  state.

## References

- PR #183 review thread.
- Issue #199 (multi-input cross-round OOR resolution).
- `client/unroll/CLAUDE.md` — multi-tree ancestry section.
- Server-side resolver: `indexer/vtxo_db.go::combineVirtualLineage`.

Issue #310, opened 2026-04-28
wallet: lineage-cap-aware OOR coin selection (cluster heuristic + pre-flight filter)


Roasbeef:
# Coin selection: prefer same-commitment cluster + lineage-cap pre-flight filter

## Context

Issue #199 introduced an operator-configurable lineage vbytes cap on OOR
submits. The cap is enforced server-side; clients receive the typed
`*oor.ErrLineageTooLarge` and the operator's value via
`OperatorTerms.MaxOORLineageVBytes` (mirrored from `GetInfo`).

The wallet's OOR coin selector currently picks any live VTXOs that satisfy
the target amount. Under the cap, naively picking inputs from many distinct
commitments can trip the cap unnecessarily — every additional commitment
adds the full tree-walk vbytes.

## Scope

Add to `client/wallet/` (OOR coin selection path):

1. **Greedy "prefer same-commitment cluster" heuristic.** When multiple
   live VTXOs satisfy the target amount, prefer combinations whose
   commitment txids overlap, minimizing the number of distinct trees
   the recipient would need to broadcast.

2. **Lineage-cap pre-flight filter.** Compute prospective lineage vbytes
   for the candidate selection (using the same `LineageVBytes` arithmetic
   the server enforces — the helper is already shared between client and
   server). Reject candidates whose vbytes exceed
   `OperatorTerms.MaxOORLineageVBytes`. Surface `*oor.ErrLineageTooLarge`
   with the same Reason format as the server side so callers can route on
   the typed error.

The typed `*oor.ErrLineageTooLarge` foundation is already in place via
`oor.ClassifySubmitError`; this issue builds the wallet-side avoidance
that lets a well-behaved client never hit a server cap rejection in
practice.

## Out of scope

- Replacing the server-side cap with a metered fee schedule (separate
  bitromortac follow-up).
- Surfacing the cap to the user-facing CLI / dry-run preview (separate
  UX issue).

## References

- Issue #199 (multi-input cross-round OOR resolution).
- `client/oor/errors.go` — typed error path.
- `client/lib/types/boarding.go::OperatorTerms.MaxOORLineageVBytes` —
  client-side mirror of operator cap.
- `indexer/lineage_vbytes.go` — server-side arithmetic that the client
  must mirror exactly to avoid client/server divergence.

PR #312, opened 2026-04-28
multi: resolve cross-round multi-input OOR end-to-end (closes #199)

### LINKED ISSUE #199: oor: multi-input Ark tx with inputs from different rounds breaks VTXO resolution
## Problem

When an Ark transaction spends multiple VTXOs from **different** batch/commitment transactions in a single OOR transfer, the recipient cannot materialize the received VTXO. The server-side indexer explicitly rejects this during lineage resolution:

```go
// indexer/vtxo_db.go — combineVirtualLineage()
if next.roundID != lineage.roundID {
    return nil, fmt.Errorf("OOR VTXO %v spans multiple rounds", outpoint)
}
if next.commitmentTxID != lineage.commitmentTxID {
    return nil, fmt.Errorf("OOR VTXO %v spans multiple commitments", outpoint)
}
```

However, the client-side OOR send flow has **no guard** against this — Alice can build and sign a multi-input Ark tx spending VTXOs from different rounds, the checkpoints get created and co-signed, but Bob's VTXO becomes irrecoverable.

This is a critical capability gap: users will naturally accumulate VTXOs across multiple rounds and need to consolidate them in OOR transfers. Blocking cross-round inputs makes the protocol significantly less useful.

## Scenario

1. Alice has VTXO #1 from round X and VTXO #2 from round Y
2. Alice creates an Ark tx spending both to pay Bob
3. Checkpoints are created (one per input) and co-signed by operator
4. Bob receives the OOR package
5. Bob queries indexer to materialize the VTXO
6. **Indexer errors**: "OOR VTXO spans multiple commitments"
7. **Bob cannot materialize the VTXO** — funds are stuck

## Key Insight: OOR VTXOs can have multi-tree ancestry

There are two kinds of VTXOs:

- **Round VTXOs** (tree leaves): always have exactly one parent batch/commitment tx. A \`tree.Tree\` correctly represents this — single \`BatchOutpoint\`, single tree hierarchy. This is fine.

- **OOR VTXOs**: created by Ark transactions that can spend multiple input VTXOs, each potentially from a **different** batch/commitment tx. An OOR VTXO's ancestry can therefore span **multiple trees from multiple batches**.

The data model issue is not that \`tree.Tree\` is wrong — it correctly represents a single tree with one batch parent. The issue is that **a VTXO created via OOR can have ancestry that requires multiple trees to fully represent**, and the current system only supports storing/returning a single \`TreePath\` per VTXO.

To unroll such a VTXO on-chain, ALL parent trees must be published (to get all input VTXOs on-chain) before the Ark transaction can be broadcast.

## Proposed Solution: Multi-tree ancestry support

**Priority: support cross-round multi-input Ark txs.** This is essential for UTXO consolidation and general usability.

### 1. Extend the VTXO data model

A VTXO needs to carry multiple tree paths when its ancestry spans multiple batches:

```protobuf
// In indexer.proto VTXO message
message AncestryPath {
    TreePath tree_path = 1;              // Tree from one batch/commitment tx
    bytes commitment_txid = 2;           // Which commitment this covers
    repeated uint32 input_indices = 3;   // Which Ark tx inputs this tree serves
}

// Replace single tree_path with:
repeated AncestryPath ancestry_paths = 18;  // One per distinct batch
```

For round VTXOs (single tree), this is just a single-element list. For cross-round OOR VTXOs, it contains one entry per distinct batch.

### 2. Update the indexer lineage resolution

In \`combineVirtualLineage()\`, instead of rejecting different rounds, **collect all distinct tree paths**:

- Group parent lineages by \`commitmentTxID\`
- For each group, resolve/merge the tree path (existing \`tryResolveCombinedRoundPath\` logic)
- Return all tree paths as an ordered collection
- Track which Ark tx inputs map to which tree path

### 3. Update the unroller

For multi-tree VTXOs, the unroller needs to:

1. **Phase 1**: Unroll ALL trees (one per batch) to get all input VTXOs on-chain
2. **Phase 2**: Broadcast checkpoint transactions (one per input)
3. **Phase 3**: Broadcast the Ark transaction (which spends all checkpoint outputs)
4. Wait for confirmation at each phase before proceeding

### 4. Short-term guard (until multi-tree is implemented)

Add a client-side guard in the OOR send flow to reject cross-round inputs, preventing funds from being silently stuck:

```go
// oor/transitions.go — during input selection
if vtxo1.CommitmentTxID != vtxo2.CommitmentTxID {
    return fmt.Errorf("cannot spend VTXOs from different rounds in single OOR transfer")
}
```

This is a stopgap — the long-term goal is full cross-round support.

## Same-round multi-input: works today

When both inputs are from the **same** round/commitment tx, \`combineVirtualLineage\` succeeds via \`tryResolveCombinedRoundPath()\` which merges the paths into a single spanning subtree within that batch's tree. This case is architecturally sound — single tree, single batch parent.

## Impact

- **Severity**: High — funds can be silently made irrecoverable (no client-side guard)
- **Usability**: Critical — blocking cross-round consolidation severely limits the protocol
- **Short-term**: Add client-side guard to prevent the issue
- **Long-term**: Implement multi-tree ancestry for full cross-round OOR support

## Related

- [#198 — unroller: support unilateral exit for OOR-received VTXOs](https://github.com/lightninglabs/darepo-client/issues/198) — broader OOR unrolling issue (single-tree case)
- [#200 — oor: add change output support for partial OOR sends](https://github.com/lightninglabs/darepo-client/issues/200) — related usability gap
- [PR #183](https://github.com/lightninglabs/darepo-client/pull/183) — unilateral exit implementation (round VTXOs only)

### Client (no guard)
- \`oor/transitions.go\` — OOR send FSM, input selection
- \`lib/tx/oor/build.go\` — \`BuildArkPSBT()\` accepts any inputs
- \`lib/tx/checkpoint/build.go\` — \`BuildCheckpointPSBT()\` per input
- \`vtxo/interfaces.go\` — \`Descriptor.TreePath\` is singular \`*tree.Tree\`
- \`unroller/unroll.go\` — assumes single tree path

### Server (rejects at resolution)
- \`indexer/vtxo_db.go\` — \`combineVirtualLineage()\` lines 319-326
- \`indexer/vtxo_db.go\` — \`resolveVirtual()\` recursive parent resolution
- \`indexer/service.go\` — \`VTXO\` response building

### Data model
- \`lib/tree/tree.go\` — \`tree.Tree\` correctly represents a single-batch tree
- \`arkrpc/indexer.proto\` — \`VTXO.tree_path\` is a single \`TreePath\` (needs to become a list for multi-batch OOR VTXOs)

Roasbeef:
In this PR, we resolve [#199](https://github.com/lightninglabs/darepo-client/issues/199): an OOR Ark transaction that spends VTXOs from different rounds (different commitment txids) used to silently strand the produced VTXO. The recipient could query it through the indexer just fine, but unilateral exit could not work because the on-the-wire shape only carried a single `tree_path` and the resolver dropped the cross-commitment ancestry on the floor with a graceful-degrade branch. With multi-input cross-round consolidation a real workflow, the silent stranding is a money-losing footgun.

The fix is end-to-end: a new repeated `AncestryPath` field on the indexer wire shape, a multi-element `Ancestry` slice on the client `vtxo.Descriptor` (with a side-table in the DB so paths up to ~100 KB don't get inlined into the row), a TLV migration on the OOR durable mailbox, and a multi-tree walk in the unroll descriptor resolver. We also bound the resulting DoS surface with an operator-configurable lineage vbytes cap that the server enforces at submit-validation time and surfaces to clients via `GetInfo`, plus a typed `OOR_REJECT_LINEAGE_TOO_LARGE` rejection so the future fallback (in-round payment when a submit overshoots the cap) can route on a code rather than string-match the human-readable reason.

This is a coordinated change with the server side; the server PR bumps the submodule pointer to this branch's HEAD. See each commit message for a detailed description w.r.t the incremental changes.

## Wire shape

The indexer's `arkrpc.VTXO` used to carry a singular `tree_path` and `tree_depth`. We replace those with a `repeated AncestryPath ancestry_paths` field, where each entry binds a `tree_path` to its `commitment_txid` and to the Ark tx input indices it serves. Round-direct VTXOs surface a length-1 slice; cross-commitment OOR VTXOs surface one entry per distinct contributing commitment. Consumers needing worst-case depth take `max(path.tree_depth)` across the slice.

`arkrpc.GetInfoResponse` gains `max_oor_lineage_vbytes` so the operator publishes its configured cap and clients can mirror it for pre-submit guarding (the wallet-side selector that consumes this is the follow-up tracked in #310).

`oorpb.SubmitPackageResponse` becomes a oneof success/rejection so a structured rejection carries both the typed `OORRejectCode` and the human-readable reason. Today the only code is `OOR_REJECT_LINEAGE_TOO_LARGE`; the enum is open.

## Client surface

`vtxo.Descriptor.TreePath` (singular) is replaced with `Descriptor.Ancestry []vtxo.Ancestry`. Helper methods `PrimaryAncestry()` and `MaxTreeDepth()` preserve the common-case ergonomics so callers that only need the deepest fragment or the first one don't have to iterate. `client/vtxo/expiry.go` switches to `MaxTreeDepth()` so the worst-case sweep horizon stays correct.

The OOR durable mailbox encodes `IncomingVTXOMetadata.Ancestry` via a TLV stream of primitive records (`encodeAncestryList` / `decodeAncestryList` and friends), the same template every other actor uses. New fragments and per-entry fields can be added additively without rewriting the existing record layout.

`oor.ClassifySubmitError(err)` maps a wire-level `*oorpb.SubmitRejectedError{Code: OOR_REJECT_LINEAGE_TOO_LARGE}` to the typed `*oor.ErrLineageTooLarge` so wallet callers can `errors.As()` directly without depending on proto types.

## Persistence

A multi-element ancestry can be a few kilobytes per fragment and up to multiple fragments per VTXO; an inline blob would balloon and force every list query to deserialize state it doesn't need. We add migration `000009_vtxo_ancestry_paths` which drops the old `tree_path` / `tree_depth` columns from `vtxos` and creates `vtxo_ancestry_paths`, a side table keyed by `(vtxo_outpoint_hash, vtxo_outpoint_index, path_order)` with FK CASCADE. The matching sqlc queries (`InsertVTXOAncestryPath`, `ListVTXOAncestryPaths`, `DeleteVTXOAncestryPaths`) extend `RoundStore` so the same transaction that creates a VTXO populates its ancestry. `LatestMigrationVersion` bumps from 8 to 9.

## Unroll

`client/unroll/descriptor_resolver.go` now iterates `desc.Ancestry` and appends every non-nil `TreePath` into `mat.TreePaths` (which was already plural, so the multi-tree machinery downstream was waiting for the producer side to catch up). `proof_assembler.go`'s nil-precondition switches from `desc.TreePath == nil` to `len(desc.Ancestry) == 0`.

## CI stabilization

The original push of this PR (and its server companion) saw CI green on the
LND backend but red on `btcwallet`+`lwwallet`. Tracing those failures
surfaced a stack of latent bugs in the actor / txconfirm / unroll /
btcwbackend layers that this revision also fixes. Each of them turned out
to be a real production bug exercised only by the multi-input
cross-round path under realistic regtest mining cadence.

`txconfirm`. Confirmation delivery to a durable subscriber used to drop
the notification on `ErrMailboxFull` and unconditionally evict the tracked
entry, silently losing the confirmation under transient SQLite WAL
contention. Tracked entries now stay alive until every subscriber
acknowledges and a later block tick retries delivery (`actor.go`
`notifyConfirmed`/`notifyFailed` return bool, plus
`retryTerminalNotifications`). The `parentBumpState` for CPFP fee bumps
also gains a `UsedFeeInputs` map so RBF children rebuild against the
parent's reserved fee outpoint instead of consuming a fresh wallet UTXO
on each bump — `btcwallet`/`lwwallet` drop a UTXO from `ListUnspent` the
moment any in-mempool tx spends it, which in the multi-tree case
exhausted the wallet within seconds. The terminal notification path
strips the sender's tx and cancellation via
`actor.WithoutTx(context.WithoutCancel(ctx))` so a downstream durable
subscriber's mailbox enqueue cannot deadlock against `txconfirm`'s open
DB tx (same cross-actor-tx pattern fixed earlier in this branch for
`VTXO -> ChainResolver` and the registry terminal handoff). And
`Evict`/`releaseFeeOutpoint` now snapshot lease outpoints synchronously
but release wallet leases best-effort off the actor goroutine, so a slow
uncancellable `wallet.ReleaseOutput` cannot block the shared `txconfirm`
actor for minutes.

`unroll`. Durable messages get explicit `Priority()` values:
`TxConfirmedMsg` / `TxFailedMsg` / `SpendObservedMsg` at +100,
`HeightObservedMsg` at -10, `GetStateRequest` at -20, so concrete
progress events drain ahead of lossy block churn and read-only status
probes when the mailbox is saturated. The registry's `handleGetStatus`
no longer enqueues a `GetStateRequest` into the child's durable mailbox
on every poll — it answers from cached pending/store records instead,
removing a polling-induced back-pressure source. `handleTerminated`
removes the child from the active map but defers the actual `child.Stop`
behind a drain probe so the child's durable mailbox can ack the inbound
terminal message before its actor ctx is cancelled (eliminates the
`ack message: context canceled` log noise on every clean COMPLETED).
Late events (`SpendObservedMsg`, `HeightObservedMsg`, `TxConfirmedMsg`)
that arrive after the FSM is already terminal are recognised in
`actor.go` and treated as idempotent no-ops rather than triggering
nack/replay loops. The terminal Tell from the child to the registry also
strips ctx + tx for the same reason as the txconfirm subscriber
notification.

`btcwbackend`. v3 (TRUC) parent+child packages cannot relay through
neutrino's individual `BroadcastTx` path because the parent is
non-relayable on its own (its absolute fee is below the mempool's
minimum relay fee). The harness was already injecting a
bitcoind-backed `chainbackends.PackageSubmitter`, but `darepod` only
wired it into the LND backend. We add the corresponding optional field
to `btcwbackend.Config` and use it in `SubmitPackage` (with the existing
individual-broadcast as a fallback for environments that genuinely have
no submitter wired), plus the one-line glue in
`darepod::startBtcwallet` that forwards `s.cfg.PackageSubmitter`. Tests
cover both the configured-submitter and nil-submitter paths.

`baselib/actor` + `db`. The `darepod.run()` cleanup defer chain only
stopped two specific actors (the OOR durable runtime and the unroll
registry), leaving every other registered actor racing against
`s.db.Close()` and producing a flood of `sql: database is closed` /
`Failed to lease message from mailbox` warnings at the tail of every
itest. We now invoke `s.actorSystem.Shutdown(shutdownCtx)` between the
runtime drain and the wallet/chain/db cleanup, so every actor drains
first. The unavoidable shutdown-time tx-begin and lease-loop errors are
demoted to debug via `isDBClosedError` / `isExpectedShutdownErr`
helpers. SQLite `busy_timeout` widens from 5s to 30s to tolerate the
natural multi-actor write contention bursts that under aggressive
mining used to surface upstream as `mailbox full`.

After this revision all four corners of the multi-input itest matrix
pass locally: `btcwallet` transfer 217s, `btcwallet` chained 225s,
`lwwallet` transfer ~3min, `lwwallet` chained 245s. LND was always green
on CI.
## Test plan

- [x] `make build` clean
- [x] `make unit pkg=./db/...` covers multi-Ancestry round-trip, upsert-replace semantics on re-save, and FK cascade on delete
- [x] `make unit pkg=./unroll/...` covers the descriptor resolver against multi-fragment, single-fragment, nil-fragment, ChainDepth=0 short-circuit, ChainDepth>0 fail-closed-without-store, and empty-ancestry rejection
- [x] `make unit pkg=./oor/...` round-trips the new TLV durable shape end-to-end
- [x] `make unit pkg=./vtxo/...` covers the new `Descriptor.Ancestry` shape
- [x] Server-side itest `TestOORIntegrationMultiInputTransfer` (lives in the server PR) drives Alice -> Bob cross-round multi-input with a tail Bob unroll. Pass locally on lnd/btcwallet/lwwallet
- [x] Server-side itest `TestOORIntegrationMultiInputChainedTransfer` (server PR) extends to Carol's `ChainDepth=2` unroll. Pass locally on lnd/btcwallet/lwwallet
- [x] Server-side itest `TestOORIntegrationLineageCapRejection` (server PR) pins the negative path

## Compatibility

This is an intentional protocol break (proto + DB schema). Coordinated upgrade window with the server. There are no shims to bridge an old and new shape — the resolver hard-errors when no fragment carries a tree path, so a regression cannot silently strand unilateral-exit data again.

## Follow-ups

Two follow-ups are filed and intentionally out of scope for this PR:

- #310 wires the cap-aware coin-selection heuristic on the wallet side so a well-behaved client never trips the server cap. The typed `*oor.ErrLineageTooLarge` foundation here is what that issue builds on.
- #311 persists per-`AncestryPath` `total_levels` on the unroll control-plane row for cross-batch sanity checking, picking up the per-fragment shape now that ancestry is multi-tree.


Roasbeef:
@claude review this 

@codex review this

PR #318, opened 2026-04-28
Make keyed OOR sends idempotent

bhandras:
## Summary

Use the persisted OOR idempotency key to make duplicate keyed sends return the existing local session instead of creating another one.

- check existing outgoing session handles before building a new OOR package
- return the existing session id when the same non-empty idempotency key is already known locally
- keep empty-key requests on the historical deterministic-session path
- cover the duplicate-key path with a test that changes the requested output but still returns the original session

## Edge Cases

This closes the local double-submit gap after a durable OOR session exists. If a caller retries a send with the same idempotency key after restart, the actor finds the restored outgoing session and returns that session id rather than creating a second transfer.

The lookup is by caller intent key, not amount, output script, selected input value, or generated Ark txid. That means two unrelated same-amount sends remain separate when they use different keys, while a retry of the same intent remains tied to the original session even if the retry request is reconstructed slightly differently.

If the caller crashes before the daemon has durably accepted a start message or created an OOR session, there is still no local session to find. In that case the retry with the same key creates the first session normally.

Terminal sessions also keep the key. A repeated call with the same key returns the terminal session instead of silently creating a new transfer; callers that want a fresh send should use a fresh key.

## Checks

- `go test ./oor ./darepod ./sdk/ark ./cmd/darepocli/darepoclicommands`
- `make lint`

### DIFF
--- a/oor/actor.go
+++ b/oor/actor.go
@@ -446,6 +446,19 @@ func (b *oorDurableBehavior) handleStartTransfer(ctx context.Context,
 		slog.Int("num_inputs", len(req.Inputs)),
 		slog.Int("num_recipients", len(req.Recipients)))
 
+	existingSessionID, found := b.findOutgoingSessionByIdempotencyKey(
+		req.IdempotencyKey,
+	)
+	if found {
+		b.logger(ctx).InfoS(ctx, "Returning existing OOR transfer",
+			slog.String("session_id", existingSessionID.String()),
+			slog.String("idempotency_key", req.IdempotencyKey))
+
+		return fn.Ok[ActorResp](&StartTransferResponse{
+			SessionID: existingSessionID,
+		})
+	}
+
 	// Build the deterministic submit package and start the session FSM.
 	// I/O is emitted as outbox messages.
 	session, outbox, err := NewSessionWithIdempotencyKey(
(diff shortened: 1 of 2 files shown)


oor/actor.go
```diff
@@ -506,6 +522,50 @@ func (b *oorDurableBehavior) handleStartTransfer(ctx context.Context,
...
+		if err != nil {
+			return SessionID{}, false, fmt.Errorf(
+				"summary for session %s: %w", sessionID, err,
+			)
+		}
+
+		if summary.IdempotencyKey == idempotencyKey {
+			return sessionID, true, nil
+		}
+	}
```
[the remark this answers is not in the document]
bhandras: Addressed. `findOutgoingSessionByIdempotencyKey` now directly iterates `b.sessions` and checks `sessionHandle.IdempotencyKey`, so it no longer sorts or builds session summaries for the lookup.

bhandras:
@claude please review

oor/actor.go
```diff
@@ -446,6 +446,19 @@ func (b *oorDurableBehavior) handleStartTransfer(ctx context.Context,
 		slog.Int("num_inputs", len(req.Inputs)),
 		slog.Int("num_recipients", len(req.Recipients)))
 
+	existingSessionID, found := b.findOutgoingSessionByIdempotencyKey(
```
Roasbeef, line 449: Why do we need this? If it's a replay, won't it re silently dropped at the actor level before it gets here?

bhandras: Yeah good q. i think the existing actor-level replay only catches the exact same package, since it dedupes by the derived ark txid/session id after rebuilding the package. The case i'm trying to cover here is a retry of the same higher-level intent where the request is reconstructed slightly differently, for example after a crash/restart coin selection can pick a different input. then the ark txid changes, and so in such cases we'd submit a second package. Test added to demonstrate exactly this.

bhandras:
@claude re-review after last push

PR #306, opened 2026-04-27
docs: nightly doc-gardening sweep 2026-04-27

litbot-9000:
## Summary

Automated nightly doc-gardening sweep against baseline `cca73c6` (last merged sweep).

- **New**: `sdk/swaps/CLAUDE.md` + `sdk/swaps/AGENTS.md` — documents the new Lightning↔Ark atomic swap SDK (vHTLC, `PayState`/`ReceiveState` FSMs, `DaemonConn`, `SwapServerConn`, durable SQLite-backed session store)
- **Updated**: `round`, `vtxo`, `oor`, `lib/types`, `darepod`, `sdk/ark`, `harness`, `baselib/actor`, `cmd/darepocli/darepoclicommands` — seal-time fee handshake (#270), `IsChange` on `VTXORequest`/`LeaveRequest`, CLTV-gated OOR spend paths, `mergeContexts` caller-context invariant, `electrsReadyTimeout`, swap CLI commands
- **ARCHITECTURE.md**: added `sdk/swaps` to Layer 3 table
- `make doc-check` passes ✓

## Test plan

- [ ] Verify `make doc-check` passes on CI
- [ ] Spot-check `sdk/swaps/CLAUDE.md` key types against source files in `sdk/swaps/`
- [ ] Confirm `ARCHITECTURE.md` Layer 3 table includes `sdk/swaps`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #328, opened 2026-04-30
actor: fix restart recovery writes

bhandras:
## Summary

Fix restart recovery write ordering across the actor paths involved in #306 and the parent darepo #308 CI failure.

Regular actor `Tell` now strips any caller transaction at the enqueue boundary with `actor.WithoutTx(ctx)`. This matters when a durable actor sends fire-and-forget work to a normal actor: the queued work must not inherit a DB transaction that may be committed or rolled back before the receiver handles it.

Durable actors still preserve `Tell` semantics after durable enqueue by processing delivery with the actor context, so caller cancellation after enqueue does not cancel behavior execution. `Ask` and `DurableAsk` continue to merge caller context because request deadlines and synchronous completion are part of that contract.

Round recovery also persists the actual commitment transaction input index for checkpointed boarding signatures. Restart replay must send the same input index that was signed, not the boarding intent's local slice position, otherwise recovery can replay a valid signature against the wrong input.

OOR finalize replay now completes local input-spend work before outgoing package persistence. The spend completion path is a synchronous `Ask`, so it intentionally keeps the durable actor transaction and lets the VTXO status update join that transaction instead of racing a second SQLite writer.

## Why

This is expected to address the root cause behind lightninglabs/darepo#306, where a refresh join reached the operator actor but failed validation while looking up a forfeit VTXO with `context canceled`.

The transaction-stripping follow-up matches the actor contract more directly: async `Tell` transfers work to the receiver, while synchronous `Ask` remains the path that can intentionally share caller-scoped transaction context.

The checkpointed input-index fix covers a restart race exposed by lightninglabs/darepo#308 CI: the client can shut down after persisting `InputSigSent` but before the original durable send reaches the operator, so replay has to be semantically equivalent to the original send.

The OOR ordering fix covers the follow-up #308 CI failure where OOR resumed after restart but the sender input stayed `spending`. Completing the VTXO manager update before package persistence avoids a SQLite writer-lock inversion during durable replay.

## Test Plan

- `go test ./oor ./vtxo ./baselib/actor ./db ./round`
- `GOGC=50 make lint`
- `go test -tags "systest" -v ./systest/... -run '^TestOORClientResumeAfterCoSign$' -timeout 10m` from the parent darepo branch
- `make itest backend=btcwallet icase=TestBoardingIntegrationRestartAfterInputSigSent` from the parent darepo branch

### DIFF
--- a/baselib/actor/actor.go
+++ b/baselib/actor/actor.go
@@ -324,14 +324,21 @@ func (ref *actorRefImpl[M, R]) Tell(ctx context.Context, msg M) error {
 		"actor_id", ref.actor.id,
 		"msg_type", msg.MessageType())
 
+	// Tell is fire-and-forget. If the caller is a durable actor currently
+	// processing inside a database transaction, do not retain that transaction
+	// in the queued envelope. The receiving actor cannot make async Tell
+	// processing atomic with the caller's transaction, and the transaction may
+	// already be committed or rolled back by the time the message is handled.
+	sendCtx := WithoutTx(ctx)
+
 	// Attempt to send the message to the mailbox. The mailbox's Send
 	// method handles context cancellation and actor termination internally.
 	env := envelope[M, R]{
 		message:   msg,
 		promise:   nil,
-		callerCtx: ctx,
+		callerCtx: sendCtx,
 	}
-	ok := ref.actor.mailbox.Send(ctx, env)
+	ok := ref.actor.mailbox.Send(sendCtx, env)
 
 	// If the send failed, determine the error and whether to route to DLO.
 	if !ok {
--- a/baselib/actor/caller_context_test.go
+++ b/baselib/actor/caller_context_test.go
@@ -252,3 +252,43 @@ func TestTellIgnoresCallerContextAfterEnqueue(t *testing.T) {
 		t.Fatal("Tell message was not processed despite being enqueued")
 	}
 }
+
+// TestTellStripsCallerTransactionBeforeEnqueue verifies that Tell does not
+// retain a durable actor's active transaction in the queued envelope. Tell
+// processing is asynchronous, so the caller's transaction may be committed or
+// rolled back before the receiving actor handles the message.
+func TestTellStripsCallerTransactionBeforeEnqueue(t *testing.T) {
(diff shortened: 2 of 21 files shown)


bhandras:
@claude review

bhandras:
@claude pls re-review. I added the symmetric Ask-side test you suggested, folded it into the existing commit, and re-ran `go test ./baselib/actor`.

bhandras:
Quick follow-up from review/chat: I also stripped `actor.WithoutTx(ctx)` at the non-durable `Actor.Tell` enqueue boundary.

`Tell` is fire-and-forget, so a receiving normal actor cannot make async processing atomic with the durable caller's transaction. Keeping that tx in the queued envelope can leave a stale handle around after commit/rollback.

`Ask` still preserves caller tx context intentionally; the existing test documents the durable-actor Ask-to-regular-actor DB-work case.

Re-ran `go test ./baselib/actor` and folded/pushed.

bhandras:
@claude pls re-review the latest small follow-up. I added `actor.WithoutTx(ctx)` at the non-durable `Actor.Tell` enqueue boundary so fire-and-forget sends do not retain a durable caller's active db transaction in the queued envelope. `Ask` still preserves caller tx context intentionally. Verified `go test ./baselib/actor`.

bhandras:
Rebased this branch onto the board-fanout client commit (`29413d1c`) so lightninglabs/darepo#308 can rebase onto current master without dropping the new boarding API surface. The actor-context fix is now on top of that commit at `8349d437`.

Verified:

```bash
go test ./baselib/actor
```

bhandras:
@claude pls re-review. I folded the CTO follow-up into the actor/restart fix: async normal-actor Tell strips caller tx context, synchronous OOR CompleteSpend stays in the durable tx, and OOR finalize replay now completes local input spend before package persistence so the VTXO status update does not race a second SQLite writer. Also updated the PR description. Verified go test ./oor ./vtxo ./baselib/actor ./db ./round, GOGC=50 make lint, make commitmsg-lint range=origin/main..HEAD, and the parent TestOORClientResumeAfterCoSign systest locally.

PR #332, opened 2026-05-01
oor: reduce durable submit deadline coupling

bhandras:
## Summary

This PR is the client-side OOR performance/liveness base from the high-concurrency arktest investigation.

The short version: OOR sends were timing out under stress while durable actor work, transport effects, signing effects, and VTXO-manager child-actor work were still queued behind each other. This branch keeps the durable state transitions durable, but moves slow/retriable side effects out of the hottest transaction paths, wakes outbox delivery immediately when work is available, and reduces repeated decode/manager pressure.

Main changes:

- durably queue OOR transport events through the actor outbox instead of sending them inline from the OOR actor turn
- move OOR signing work into a signing effect actor so wallet signing is no longer part of the core OOR actor critical path
- make stale replayed signing results no-ops once the OOR FSM has already advanced past the matching signing state
- detach the durable OOR submit context from RPC cancellation after input selection, while still making the RPC response respect the caller deadline
- bound detached-submit cleanup so a permanently stalled actor future cannot retain wallet/custom-input reservations forever
- strip inherited actor DB transactions from the durable submit path before sending across actor boundaries
- wake outbox publishers on enqueue so active same-process delivery is event-driven instead of waiting for the fallback poll interval
- relax durable mailbox/outbox fallback polling now that explicit wakeups handle the hot path
- cache decoded ancestry trees behind a bounded LRU so repeated VTXO ancestry reads do not re-decode the same committed tree blobs without retaining every historical tree forever
- bound refresh/forfeit VTXO child-actor asks so a stuck child cannot monopolize the shared VTXO manager admission path
- keep `oor/CLAUDE.md` and `oor/AGENTS.md` in sync after updating the OOR actor invariants

## Problem

High-concurrency arktest runs originally showed several symptoms that all looked like generic slow OOR payments from the outside:

- OOR `SendOOR` calls could hit the caller deadline even though durable actor work was still making progress
- transport submission lived too close to the durable actor transaction path
- signing work could extend the OOR actor turn instead of being treated as an effect
- durable outbox/mailbox loops relied too much on short polling intervals for latency
- repeated ancestry tree decoding showed up under stress
- refresh/forfeit child actors could hold the VTXO manager while unrelated OOR completion work waited behind it

The goal here is not to make the RPC ignore user deadlines. The RPC should still return when the caller context expires. The goal is to prevent cancellation, stale transactions, or slow side effects from corrupting or unnecessarily stretching the durable actor transition once OOR inputs have already been selected.

### OOR transport outbox

Transport events such as submit/finalize/ack are now durably enqueued and delivered to `ServerConn` after the OOR actor transition commits. The transport side effect runs outside the actor DB transaction and is retried through the actor delivery store.

This preserves the important durable ordering while keeping RPC transport work out of the actor transaction itself.

### Signing effect actor

OOR signing is moved into a signing effect actor. The OOR actor emits signing requests as effects, and the signing effect actor sends the resulting follow-up events back into OOR.

That gives signing the same durable/effect shape as transport work and avoids treating wallet signing as part of the core OOR actor state transition.

A durable signing request may be replayed after its result already reached the OOR actor. The OOR `DriveEventRequest` handler now treats stale duplicate `ArkSignedEvent`, `CheckpointsSignedEvent`, and matching signing `OutboxErrorEvent` messages as no-ops once the FSM has advanced past the corresponding signing state. That keeps signing-effect retry/replay idempotent without hiding fresh events that still belong to the current state.

### Submit context and actor transaction hygiene

The RPC server now detaches the durable OOR submit from caller cancellation after fresh inputs have been selected. The future wait still uses the caller context, so the RPC response can still deadline or cancel normally, but the durable actor submit is not killed by that response deadline.

If the caller stops waiting after the detached submit has been accepted, selected inputs remain owned by the in-flight actor work. Wallet locks and custom-input reservations are only released after the actor future actually completes with failure or a duplicate/existing-session response. This avoids making RPC timeout cleanup race against a still-running durable OOR transfer.

The response-wait cleanup path also has a long finite ceiling. If the detached actor future never completes, the cleanup waiter logs loudly after 10 minutes and releases the wallet/custom-input reservations rather than leaking a goroutine and held inputs indefinitely.

The submit path also uses the actor option that strips inherited DB transactions before crossing actor boundaries. That avoids passing a transaction that may be rolled back by the time downstream actor work uses it.

### Outbox wakeups and polling

The actor delivery store now wakes outbox publishers when work is enqueued. With explicit wakeups in the hot path, the durable mailbox/outbox fallback polling interval can be relaxed. This lowers idle DB pressure without adding same-process latency when new work is available.

### Ancestry tree cache

Decoded ancestry trees are cached by their serialized tree bytes. These trees represent committed round data and callers treat cached values as read-only. This avoids re-decoding the same tree blobs during repeated VTXO ancestry reads.

The cache is a fixed-size LRU, currently capped at 4096 entries. Eviction is opportunistic rather than correctness-driven: trees are immutable, age does not imply staleness, and a cache miss simply re-decodes the tree from durable storage. This keeps the optimization bounded for long-lived daemons and stress runs.

### VTXO manager child asks

Refresh/forfeit child-actor asks are bounded so one stuck child cannot hold the shared VTXO manager admission point indefinitely. Normal OOR spend reservation/completion still uses the caller deadline, so this does not introduce artificial short deadlines for healthy OOR payments.

## Telemetry

`SendOOR` keeps one low-noise success log with coarse phase durations such as input selection, change output construction, and OOR actor submission. Those fields are intended production telemetry for future stress/debug runs.

What was dropped from this PR was the temporary deep diagnostic instrumentation added during investigation, including the lower-level change-output breakdowns and the separate input-selection timing split. Those were useful while chasing the issue, but they are not needed in the final performance branch.

## Review Guide

Suggested review order:

1. `oor/actor.go`, `oor/actor_durable_message.go`, and `oor/signing_effect_actor.go` for the OOR outbox/effect shape and stale signing-result no-op handling.
2. `darepod/rpc_server.go` and `darepod/server.go` for submit context handling, bounded cleanup, effect actor wiring, and shutdown ordering.
3. `baselib/actor/*` and `db/actordelivery/*` for outbox publisher wakeups and durable polling behavior.
4. `db/ancestry_codec.go`, `db/vtxo_store.go`, and related tests for bounded ancestry tree caching.
5. `vtxo/manager.go` and `vtxo/manager_admission_test.go` for bounded child asks and rollback context behavior.
6. `oor/CLAUDE.md` and `oor/AGENTS.md` for the updated actor invariants.

## Validation

Local checks run on this branch:

```sh
go test ./db -run 'TestAncestryTreeCache|TestUpsertAncestryPaths' -count=1
go test ./darepod -run 'TestSendOOR|TestSubmittedOORCleanup|TestIsAwaitContextError' -count=1
go test ./oor -run 'TestOORClientActorDriveEvent|Test.*Signing|Test.*Outbox|Test.*Transport' -count=1
GOGC=50 make lint
```

Additional focused package coverage from the branch history/review cycle included:

```sh
go test ./darepod ./oor ./baselib/actor
go test ./db ./vtxo ./darepod ./oor ./baselib/actor ./db/actordelivery
go test ./indexer ./darepod
```

The parent arktest stress stack that consumes this pointer no longer showed OOR `DeadlineExceeded` or `SQLITE_BUSY` failures in the relevant high-concurrency run after this client branch plus the parent SQLite timeout/mailbox/indexer changes. Remaining unexpected failures were refresh-round related and are intentionally left visible for follow-up work.

### DIFF
--- a/baselib/actor/durable_actor.go
+++ b/baselib/actor/durable_actor.go
@@ -79,7 +79,9 @@ type DurableActorConfig[M TLVMessage, R any] struct {
 	HeartbeatInterval time.Duration
 
 	// PollInterval is how often to poll for new messages when empty.
-	// Default: 100ms.
+	// Same-process sends wake the mailbox immediately, so polling is only
+	// the fallback for missed wakes, restarts, and external enqueues.
+	// Default: 1s.
 	PollInterval time.Duration
 
 	// MaxAttempts is the default maximum delivery attempts.
@@ -115,7 +117,7 @@ func DefaultDurableActorConfig[M TLVMessage, R any](
 		TellRetryPolicy:   DefaultTellRetryPolicy,
 		LeaseDuration:     leaseDuration,
 		HeartbeatInterval: leaseDuration / 3,
-		PollInterval:      100 * time.Millisecond,
+		PollInterval:      time.Second,
 		MaxAttempts:       10,
 		CleanupTimeout:    5 * time.Second,
 		DeduplicationTTL:  24 * time.Hour,
--- a/baselib/actor/outbox_publisher.go
+++ b/baselib/actor/outbox_publisher.go
@@ -78,6 +78,9 @@ type OutboxPublisher struct {
 	// wg tracks the background goroutine.
 	wg sync.WaitGroup
 
+	// wake nudges the publisher when same-process outbox work commits.
+	wake chan struct{}
+
 	// startOnce ensures Run is only called once.
 	startOnce sync.Once
 
@@ -102,11 +105,18 @@ func NewOutboxPublisher(cfg OutboxPublisherConfig) *OutboxPublisher {
 		cfg.ClaimDuration = 30 * time.Second
 	}
 
-	return &OutboxPublisher{
+	p := &OutboxPublisher{
 		cfg:    cfg,
 		ctx:    ctx,
 		cancel: cancel,
+		wake:   make(chan struct{}, 1),
 	}
+
+	if registrar, ok := cfg.Store.(OutboxWakeRegistrar); ok {
+		registrar.RegisterOutboxWake(p.Wake)
+	}
+
+	return p
 }
 
 // Start begins the background publishing loop.
@@ -144,6 +154,9 @@ func (p *OutboxPublisher) run() {
 			return
 
 		case <-ticker.C:
+			p.publishBatch()
+
+		case <-p.wake:
 			p.publishBatch()
 		}
 	}
@@ -274,4 +287,13 @@ func (p *OutboxPublisher) deliverMessage(msg OutboxMessage) {
 // or when immediate delivery is needed after a transaction commits.
 func (p *OutboxPublisher) PublishPending() {
 	p.publishBatch()
+}
+
+// Wake asks the publisher to run a publish cycle soon. The signal is
+// best-effort because the periodic poll remains the durability fallback.
+func (p *OutboxPublisher) Wake() {
+	select {
+	case p.wake <- struct{}{}:
+	default:
+	}
 }
--- a/baselib/actor/outbox_publisher.go
+++ b/baselib/actor/outbox_publisher.go
@@ -20,7 +20,9 @@ type OutboxPublisherConfig struct {
 	System SystemContext
 
 	// PollInterval is how often to poll for pending outbox messages.
-	// Default: 100ms.
+	// Same-process outbox commits wake the publisher immediately, so polling
+	// is only the fallback for missed wakes and process restarts.
+	// Default: 1s.
 	PollInterval time.Duration
 
 	// BatchSize is the maximum number of messages to process per poll.
@@ -48,7 +50,7 @@ func DefaultOutboxPublisherConfig(
 		Store:               store,
 		Codec:               codec,
 		System:              system,
-		PollInterval:        100 * time.Millisecond,
+		PollInterval:        time.Second,
 		BatchSize:           100,
 		MaxDeliveryAttempts: 10,
 		ClaimDuration:       30 * time.Second,
@@ -93,7 +95,7 @@ func NewOutboxPublisher(cfg OutboxPublisherConfig) *OutboxPublisher {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	if cfg.PollInterval == 0 {
-		cfg.PollInterval = 100 * time.Millisecond
+		cfg.PollInterval = time.Second
 	}
 	if cfg.BatchSize == 0 {
 		cfg.BatchSize = 100
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -48,6 +48,14 @@ const (
 	// stream so a CLI disconnect does not cancel the unilateral-exit
 	// handoff.
 	manualUnrollAdmissionTimeout = 30 * time.Second
+
+	// submittedOORCleanupTimeout bounds the post-RPC cleanup waiter
+	// used after a detached OOR actor submit has been accepted but the
+	// caller stopped waiting for the response. The wait should normally
+	// end when the actor future completes or daemon shutdown fails
+	// pending futures; this ceiling prevents a pathological actor stall
+	// from retaining wallet/custom input reservations forever.
+	submittedOORCleanupTimeout = 10 * time.Minute
 )
 
 // RPCServer implements the daemon's gRPC DaemonService interface.
@@ -1427,6 +1435,18 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 	req *daemonrpc.SendOORRequest) (
 	*daemonrpc.SendOORResponse, error) {
 
+	startTime := time.Now()
+	var (
+		idempotencyDuration   time.Duration
+		resolveScriptDuration time.Duration
+		operatorTermsDuration time.Duration
+		policyResolveDuration time.Duration
+		inputSelectDuration   time.Duration
+		buildInputsDuration   time.Duration
+		changeOutputDuration  time.Duration
+		oorActorDuration      time.Duration
+	)
+
 	if err := r.requireWalletReady(); err != nil {
 		return nil, err
 	}
@@ -1442,9 +1462,11 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 	}
 
 	if req.GetIdempotencyKey() != "" && !req.DryRun {
+		phaseStart := time.Now()
 		key := req.GetIdempotencyKey()
 		sessionID, found, err :=
 			r.findOutgoingOORSessionByIdempotencyKey(ctx, key)
+		idempotencyDuration = time.Since(phaseStart)
 		if err != nil {
 			return nil, err
 		}
@@ -1465,7 +1487,9 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 	// Resolve the recipient's pkScript from the destination
 	// oneof. Pubkey destinations need operator terms to derive
 	// VTXO-compatible taproot outputs, so we pass a lazy fetcher.
+	phaseStart := time.Now()
 	pkScript, err := r.resolveOutputPkScript(ctx, req.Recipient)
+	resolveScriptDuration = time.Since(phaseStart)
 	if err != nil {
 		return nil, err
 	}
@@ -1494,7 +1518,9 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 	}
 
 	// Fetch operator terms for the checkpoint policy.
+	phaseStart = time.Now()
 	terms, err := r.server.fetchOperatorTerms(ctx)
+	operatorTermsDuration = time.Since(phaseStart)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal,
 			"unable to fetch operator terms: %v", err)
@@ -1505,20 +1531,25 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 		CSVDelay:    terms.VTXOExitDelay,
 	}
 
+	phaseStart = time.Now()
 	recipientPolicyTemplate, err := r.resolveOutputPolicyTemplate(
 		ctx, req.Recipient, pkScript, terms.PubKey,
 		terms.VTXOExitDelay,
 	)
+	policyResolveDuration = time.Since(phaseStart)
 	if err != nil {
 		return nil, err
 	}
 
 	var (
-		selectedInputs []oor.TransferInput
-		locked         *wallet.SelectAndLockVTXOsResponse
+		selectedInputs      []oor.TransferInput
+		locked              *wallet.SelectAndLockVTXOsResponse
+		customInputsRelease func()
+		releaseCustomInputs bool
 	)
 
 	if len(req.CustomInputs) > 0 {
+		phaseStart = time.Now()
 		// Custom inputs provided — bypass wallet selection and
 		// build TransferInputs from the specified VTXOs.
 		//
@@ -1532,7 +1563,10 @@ func (r *RPCServer) SendOOR(ctx context.Context,
 		// same outpoint. We gate that race here with an
 		// in-memory reservation keyed by outpoint; the
 		// reservation is released regardless of success or
-		// failure via defer.
+		// failure via defer, unless the detached OOR actor submit has
+		// already been accepted and the caller only stopped waiting for
+		// its response. In that case the release is deferred until the
+		// actor future actually completes.
 		customOutpoints := make(
 			[]wire.OutPoint, 0, len(req.CustomInputs),
 		)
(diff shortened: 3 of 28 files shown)


oor/actor.go
```diff
@@ -1476,13 +1480,113 @@ func (b *oorDurableBehavior) sendTransportEvent(ctx context.Context,
...
+
+	if b.cfg.DeliveryStore == nil {
+		return fmt.Errorf("delivery store must be provided")
+	}
+
+	if b.cfg.ServerConn == nil {
+		return fmt.Errorf("serverconn ref must be provided")
+	}
+
+	payload, err := serverconn.NewServerConnCodec().Encode(msg)
```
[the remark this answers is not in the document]
bhandras: fixed and folded. the OOR durable behavior now reuses package-level serverconn and signing-effect codecs instead of rebuilding them for every outbox enqueue.

bhandras:
@claude pls re-review. I addressed Gemini's codec reuse note by folding package-level serverconn/signing-effect codecs into the OOR outbox path, fixed the lint failures, and force-pushed the cleaned series. Verified go test ./oor -run 'Test.*Outbox|Test.*Signing|Test.*Transport' -count=1, GOGC=50 make lint, and commit message lint locally.

bhandras:
@claude pls re-review. I addressed the latest pass and force-pushed the folded series.

Changes since your review:
- signing effect actor now forwards every follow-up event instead of only `followUps[0]`
- `driveOutbox` builds the signing effect request once and passes it through
- docs/comments now spell out the post-commit transport outbox invariant, mixed codec registration/collision guard, ancestry cache immutability, and the negative rollback timeout behavior

Verified locally:
- `go test ./oor -run 'Test.*Outbox|Test.*Signing|Test.*Transport' -count=1`
- `GOGC=50 make lint`

oor/CLAUDE.md
```diff
@@ -237,9 +237,10 @@ resume semantics.
 - Point-of-no-return: when server co-signs checkpoint transaction(s).
 - After checkpoint signature, client must resume and obtain byte-identical
   co-signed PSBTs (deterministic construction).
-- Transport outbox events (submit, finalize, ack) are Tell'd to ServerConn
-  within the actor's DB transaction for atomic enqueue (same-DB tx joining via
-  `ExecTx`).
+- Transport outbox events (submit, finalize, ack) are durably enqueued, then
```
bhandras, line 240: We need this in AGENTS.md as well.

bhandras:
@claude pls re-review. I dropped the temporary `oor: trace change script allocation` commit from this PR after revisiting which diagnostics should actually land. The branch now ends at the OOR outbox/effect behavior fixes and does not include the change-output timing split. I also updated the PR description to match.

Verified locally:
- `go test ./oor -run 'Test.*Outbox|Test.*Signing|Test.*Transport' -count=1`
- `GOGC=50 make lint`

bhandras:
@claude pls re-review. I fixed the CLAUDE.md / AGENTS.md drift by making the OOR docs match again, kept the temporary timing-debug commit dropped, and re-ran commitmsg lint, the focused OOR tests, and GOGC=50 make lint locally.

bhandras:
@claude pls re-review. I addressed the local review findings and folded the fixes into the existing commits.

Changes since the last review:

- bounded the detached SendOOR cleanup waiter with a 10m ceiling, so a permanently stalled actor future logs loudly and releases wallet/custom-input reservations instead of leaking forever
- widened the caller-cancel detection so once the RPC context is done, either context.Canceled or context.DeadlineExceeded from the await path is treated as caller wait cancellation instead of an actor failure
- added regression coverage for detached cleanup timeout and await-cancel classification
- made stale replayed signing results no-ops once the OOR FSM has advanced past the matching signing state, covering ArkSignedEvent, CheckpointsSignedEvent, and matching signing OutboxErrorEvent
- added a drive-event regression test for stale signing replay
- changed the new custom-input test cleanup to defer release2()
- updated CLAUDE.md and AGENTS.md to say the durable poll fallback is 1s, and added a short rationale comment for the 5s VTXO child ask timeout

Validated locally:

```sh
go test ./darepod -run 'TestSendOOR|TestSubmittedOORCleanup|TestIsAwaitContextError' -count=1
go test ./oor -run 'TestOORClientActorDriveEvent|Test.*Signing|Test.*Outbox|Test.*Transport' -count=1
go test ./vtxo -run 'TestReserveForfeit|TestCompleteSpendUsesCallerDeadline|TestManager' -count=1
GOGC=50 make lint
```

oor/actor.go
```diff
@@ -1476,13 +1478,69 @@ func (b *oorDurableBehavior) sendTransportEvent(ctx context.Context,
...
+// serverconn actor for durable delivery to the server.
+func (b *oorDurableBehavior) sendTransportEvent(ctx context.Context,
+	msg OutboxEvent) error {
+
+	sendReq, err := b.buildTransportMessage(ctx, msg)
+	if err != nil {
+		return err
+	}
+
+	if b.cfg.TransportOutbox {
```
Roasbeef, line 1596: So then the benefit here is that just writing to the outbox directly is faster than enqueuing to the sererconn's durable mailbox as it lets our transaction exit quicker?

oor/signing_effect_actor.go
```diff
@@ -0,0 +1,433 @@
...
+	if !ok {
+		return fn.Err[actor.Message](fmt.Errorf(
+			"unknown signing effect message: %T", msg,
+		))
+	}
+
+	events := a.handleSigningRequest(ctx, req)
+	for _, event := range events {
+		// A durable signing request may be retried after this Tell has
+		// already reached OOR. The OOR DriveEventRequest handler
```
Roasbeef, line 330: 👍

oor/signing_effect_actor.go
```diff
@@ -0,0 +1,433 @@
...
+	outbox, err := req.toOutboxEvent()
+	if err != nil {
+		return []Event{&FailEvent{Reason: err.Error()}}
+	}
+
+	handler := &SigningOutboxHandler{
+		Signer: a.cfg.Signer,
+	}
+
+	followUps, err := handler.Handle(ctx, req.SessionID, outbox)
```
Roasbeef, line 359: Nice we're able to re-use that other layer of indirection here.

darepod/rpc_server.go
```diff
@@ -1689,7 +1756,22 @@ func (r *RPCServer) SendOOR(ctx context.Context,
...
+			operatorTermsDuration),
+		slog.Duration("policy_resolve_duration",
+			policyResolveDuration),
+		slog.Duration("input_select_duration",
+			inputSelectDuration),
+		slog.Duration("build_inputs_duration",
+			buildInputsDuration),
+		slog.Duration("change_output_duration",
+			changeOutputDuration),
+		slog.Duration("oor_actor_duration", oorActorDuration))
```
Roasbeef, line 1774: We can also export these as prom metrics. Works here vs reading from postgres as it's ephemeral data. We can also then export it as a histogram to generate percentile graphs, etc.

darepod/rpc_server.go
```diff
@@ -1802,6 +1884,87 @@ func (r *RPCServer) unlockSelectedVTXOsBestEffort(ctx context.Context,
...
+// after the RPC caller stopped waiting. Once the actor has actually completed,
+// cleanup follows the same rules as the synchronous path: wallet-selected
+// inputs are unlocked only on actor failure or duplicate/existing sessions,
+// while custom input reservations are released when the in-flight actor start
+// is no longer using the RPC-level double-use guard.
+func (r *RPCServer) cleanupSubmittedOORStart(ctx context.Context,
+	future actor.Future[oor.ActorResp],
+	locked *wallet.SelectAndLockVTXOsResponse, releaseCustomInputs func()) {
+
+	r.cleanupSubmittedOORStartWithTimeout(
```
Roasbeef, line 1909: Can just inline the body here?

baselib/actor/outbox_publisher.go
```diff
@@ -145,6 +155,9 @@ func (p *OutboxPublisher) run() {
 
 		case <-ticker.C:
 			p.publishBatch()
+
+		case <-p.wake:
```
Roasbeef, line 161: Nice happy medium here vs decreasing the ticker interval

db/actordelivery/store_impl.go
```diff
@@ -1355,7 +1402,10 @@ func (s *TxAwareActorDeliveryStore) ExecTx(
 
 	// Create a transaction-scoped queries object.
 	txQuerier := adsqlc.New(tx)
-	txStore := newTxActorDeliveryStore(txQuerier, s.clock, tx)
+	outboxEnqueued := false
+	txStore := newTxActorDeliveryStore(
+		txQuerier, s.clock, tx, &outboxEnqueued,
```
Roasbeef, line 1407: Don't _love_ the pointer pattern here, but it does the job nevertheless.

Roasbeef: Alternatives would be more invasive API wise prob

db/ancestry_codec.go
```diff
@@ -49,6 +53,145 @@ type ancestryStore interface {
...
+		tree: t,
+	})
+	c.trees[key] = elem
+
+	for len(c.trees) > c.maxEntries {
+		c.evictOldestLocked()
+	}
+}
+
+func (c *ancestryTreeCache) evictOldestLocked() {
```
Roasbeef: We can use the neutrino `cache` package here. I added it recently for cache Esplora block fetches: https://pkg.go.dev/github.com/lightninglabs/neutrino/cache

It's a sub module so it won't bring in a ton of deps.

baselib/actor/durable_actor.go
```diff
@@ -115,7 +117,7 @@ func DefaultDurableActorConfig[M TLVMessage, R any](
 		TellRetryPolicy:   DefaultTellRetryPolicy,
 		LeaseDuration:     leaseDuration,
 		HeartbeatInterval: leaseDuration / 3,
-		PollInterval:      100 * time.Millisecond,
+		PollInterval:      time.Second,
```
Roasbeef, line 120: 👍

vtxo/manager.go
```diff
@@ -763,9 +834,9 @@ func (m *Manager) handleReserveForfeit(ctx context.Context,
 	var reserved []wire.OutPoint
 	for _, op := range outpoints {
 		ref := m.actors[op]
-		result := ref.Ask(
-			ctx, &PendingForfeitEvent{},
-		).Await(ctx)
+		result := m.askForfeitVTXOActor(
```
Roasbeef, line 837: Hmm, but _shouldn't_ we continue to block here and not give up? Given we're about to do a refresh/transfer.

Roasbeef: An alternative here I think would be to make the RPCs more fire and forget. Then the user hits some other endpoint to keep track of what's going on (in this case, if a new round happened or not).

Roasbeef (approved):
Nice work!

The changes look really good. Also dig how we were able to re-use the extracted effet logic for the new handlers. 

Two minor comments:
  * We can use `neutrino/cache` for that new lineage cache 
  * Should we just make some/most RPCs fire and forget from the user/wallets PoV? IIUC that would let us side step some of those context issues, as maybe the server isn't connected by the time they do the RPC.

db/ancestry_codec.go
bhandras: yep, switched this to neutrino/cache's LRU and folded it into the cache commit. each cached tree counts as one entry, keeping the same 4096-entry bound while dropping the local list/map implementation.

oor/actor.go
bhandras: yep exactly. we still durably persist the handoff, but the OOR actor no longer waits on serverconn mailbox enqueue/transport work inside its turn. the outbox write commits with the actor-side transition, then the publisher wakes after commit and picks it up.

darepod/rpc_server.go
bhandras: yep, agreed. leaving metrics out of this branch, but the phase durations are now in one place so a follow-up can export them as histograms cleanly.

darepod/rpc_server.go
bhandras: agreed this can be cleaned up. leaving the RPC cleanup shape unchanged in this PR and punting that polish to the follow-up RPC cleanup pass.

vtxo/manager.go
bhandras: yeah, I think this points at a broader RPC model change. leaving the current bounded manager ask in this PR so a wedged child cannot monopolize the manager, and we can revisit true fire-and-forget/status polling as a follow-up once the operation-status surface is in place.

bhandras (commented):
Addressed the cache-package suggestion and left the RPC/fire-and-forget cleanup items scoped to follow-up.

bhandras:
@claude pls re-review. I addressed Roasbeef's cache suggestion by replacing the local ancestry tree LRU with neutrino/cache's LRU, kept the RPC cleanup/fire-and-forget semantics unchanged for a follow-up, and verified `make commitmsg-lint range=origin/main..HEAD`, `go test ./db -run 'TestAncestryTreeCache|TestUpsertAncestryPaths' -count=1`, `go test ./...`, and `GOGC=50 make lint` locally.

bhandras:
@claude re-review, see comments

bhandras:
@claude re review

PR #335, opened 2026-05-05
round, txconfirm: Forfeit response plumbing

ellemouton:
## Summary

Client-side companion to
[darepo#324](https://github.com/lightninglabs/darepo/pull/324)
(server-side broadcast of stored forfeit responses).

This PR pulls in the client-side changes the server's fraud-response
flow depends on:

- Round client tracks the VTXO amount and the connector amount as
  separate values when building / validating forfeit transactions, so
  the penalty output equals the full zero-fee input value (matches the
  server-side validation rule).
- Round client routes forfeit signature responses directly back to the
  collecting round actor with a context detached from the now-dead
  server-message handler, instead of relaying through the VTXO
  manager (whose 2-hop relay races a canceled context). An
  `ExpectedForfeits` fallback recovers RoundID drift visible in logs
  rather than dropping the signature.
- `txconfirm` sets the CPFP fee-input sighash so witnesses Core's
  mempool policy accepts.
- `txconfirm` tolerates the parent-already-broadcast case on the
  initial submit so a races-to-mempool parent does not abort the
  tracked tx.
- `tree` deduplicates connector leaf cosigners.

### DIFF
--- a/lib/tree/node.go
+++ b/lib/tree/node.go
@@ -73,10 +73,10 @@ func NewLeafNode(input wire.OutPoint, leaf LeafDescriptor,
 	error) {
 
 	// The cosigners for a leaf are the leaf owner and operator.
-	cosigners := []*btcec.PublicKey{
+	cosigners := UniqueCosigners([]*btcec.PublicKey{
 		leaf.CoSignerKey,
 		operatorKey,
-	}
+	})
 
 	outputs := []*wire.TxOut{
 		// The leaf output (VTXO or connector).
--- a/lib/tx/forfeit.go
+++ b/lib/tx/forfeit.go
@@ -67,11 +67,11 @@ type ForfeitTxContext struct {
 // BuildForfeitTx creates a 2-input (VTXO + connector), 2-output (penalty +
 // anchor) transaction.
 func BuildForfeitTx(vtxoOutpoint *wire.OutPoint, vtxoAmount btcutil.Amount,
-	connectorOutpoint *wire.OutPoint,
+	connectorOutpoint *wire.OutPoint, connectorAmount btcutil.Amount,
 	serverForfeitScript []byte) (*wire.MsgTx, error) {
 
 	return BuildForfeitTxWithContext(
-		vtxoOutpoint, vtxoAmount, connectorOutpoint,
+		vtxoOutpoint, vtxoAmount, connectorOutpoint, connectorAmount,
 		serverForfeitScript, ForfeitTxContext{
 			VTXOSequence: wire.MaxTxInSequenceNum,
 		},
@@ -82,7 +82,7 @@ func BuildForfeitTx(vtxoOutpoint *wire.OutPoint, vtxoAmount btcutil.Amount,
 // tx-level requirements for the VTXO input spend path.
 func BuildForfeitTxWithContext(vtxoOutpoint *wire.OutPoint,
 	vtxoAmount btcutil.Amount, connectorOutpoint *wire.OutPoint,
-	serverForfeitScript []byte,
+	connectorAmount btcutil.Amount, serverForfeitScript []byte,
 	ctx ForfeitTxContext) (*wire.MsgTx, error) {
 
 	switch {
@@ -98,6 +98,12 @@ func BuildForfeitTxWithContext(vtxoOutpoint *wire.OutPoint,
 	case vtxoAmount <= 0:
 		return nil, fmt.Errorf("vtxo amount must be positive, got %d",
 			vtxoAmount)
+
+	case connectorAmount < 0:
+		return nil, fmt.Errorf(
+			"connector amount must be non-negative, got %d",
+			connectorAmount,
+		)
 	}
 
 	tx := wire.NewMsgTx(3)
@@ -119,7 +125,7 @@ func BuildForfeitTxWithContext(vtxoOutpoint *wire.OutPoint,
 	})
 
 	tx.AddTxOut(&wire.TxOut{
-		Value:    int64(vtxoAmount),
+		Value:    int64(vtxoAmount + connectorAmount),
 		PkScript: serverForfeitScript,
 	})
 
--- a/round/actor.go
+++ b/round/actor.go
@@ -2118,8 +2118,13 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
 				a.log.DebugS(ctx, "Looking up VTXO actor by service key",
 					slog.String("outpoint", m.VTXOOutpoint.String()))
 
+				// The VTXO actor may process this request after the
+				// current server-message handler returns. Detach the
+				// context so its follow-up ForfeitSignatureResponse
+				// relay is not canceled before it reaches this round.
+				reqCtx := context.WithoutCancel(ctx)
 				err := serviceKey.Ref(a.cfg.ActorSystem).Tell(
-					ctx, &ForfeitRequestEvent{
+					reqCtx, &ForfeitRequestEvent{
 						RoundID:               m.RoundID,
 						ConnectorOutpoint:     m.ConnectorOutpoint,
 						ConnectorPkScript:     m.ConnectorPkScript,
@@ -2132,6 +2137,11 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
 					a.log.WarnS(ctx, "Failed to send forfeit request to VTXO actor",
 						err,
 						slog.String("outpoint", m.VTXOOutpoint.String()))
+
+					return fmt.Errorf(
+						"send forfeit request to VTXO actor: %w",
+						err,
+					)
 				}
 				a.log.InfoS(ctx, "Sent forfeit request to VTXO actor",
 					slog.String("outpoint", m.VTXOOutpoint.String()),
@@ -2456,7 +2466,6 @@ func (a *RoundClientActor) handleForfeitSignatureResponse(ctx context.Context,
 
 	roundIDStr := resp.RoundID
 
-	// Look up the round by its RoundID key string.
 	keyStr := RoundKeyStr(roundIDStr)
 	roundFSM, exists := a.rounds[keyStr]
 	if !exists {
--- a/round/harness_test.go
+++ b/round/harness_test.go
@@ -684,6 +684,15 @@ func generateTestKeyPair(t *testing.T) (*btcec.PrivateKey, *btcec.PublicKey) {
 
 const (
 	testExitDelay = uint32(144) // 1 day in blocks.
+
+	// testConnectorDustValue is the on-chain dust value of a connector
+	// output used in test fixtures. Mirrors the operator-published
+	// `DustLimit` of 546 sats configured in `newTestEnvironment`.
+	testConnectorDustValue = int64(546)
+
+	// testForfeitVTXOValue is the local VTXO amount used by forfeit-tx
+	// fixtures.
+	testForfeitVTXOValue = int64(50000)
 )
 
 // newTestVTXOTree creates a minimal valid VTXT tree for testing by
@@ -2033,9 +2042,9 @@ func (h *boardingTestHarness) newTestForfeitTx(
 		Sequence:         wire.MaxTxInSequenceNum,
 	})
 
-	// Output 0: Penalty to server.
+	// Output 0: Penalty to server (forfeited VTXO + connector dust).
 	tx.AddTxOut(&wire.TxOut{
-		Value:    50000,
+		Value:    testForfeitVTXOValue + testConnectorDustValue,
 		PkScript: serverForfeitScript,
 	})
 
--- a/round/transitions.go
+++ b/round/transitions.go
@@ -1519,6 +1519,13 @@ func (s *CommitmentTxReceivedState) ProcessEvent(
 			slog.Int("client_trees", len(clientTrees)),
 			slog.Int("vtxo_tree_count", len(s.VTXOTreePaths)))
 
+		forfeitMappings, err := populateForfeitMappingAmounts(
+			evt.ForfeitMappings, s.Intents.Forfeits,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("populate forfeit amounts: %w", err)
+		}
+
 		// Proceed to nonce generation. Forfeit mappings (if any) are
 		// carried forward through the MuSig2 signing states. Forfeit
 		// signatures are collected AFTER VTXO tree signing is complete,
@@ -1532,7 +1539,7 @@ func (s *CommitmentTxReceivedState) ProcessEvent(
 				Intents:              s.Intents.Clone(),
 				ClientTrees:          clientTrees,
 				BoardingInputIndices: boardingInputIndices,
-				ForfeitMappings:      evt.ForfeitMappings,
+				ForfeitMappings:      forfeitMappings,
 			},
 			NewEvents: fn.Some(ClientEmittedEvent{
 				InternalEvent: []ClientEvent{&GenerateNonces{}},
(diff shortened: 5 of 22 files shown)


vtxo/transitions.go
```diff
@@ -247,7 +248,8 @@ func (s *LiveState) handleForfeitRequest(
 	}
 
 	forfeitTx, err := tx.BuildForfeitTxWithContext(
-		&s.VTXO.Outpoint, s.VTXO.Amount,
+		&s.VTXO.Outpoint,
+		s.VTXO.Amount+btcutil.Amount(evt.ConnectorAmount),
```
ellemouton: rather add a "connectorAmount" param. dont use the vtxoAmount param since that changes the "truth"

round/actor.go
```diff
@@ -2102,20 +2102,31 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
 				a.log.DebugS(ctx, "Looking up VTXO actor by service key",
 					slog.String("outpoint", m.VTXOOutpoint.String()))
```
ellemouton, line 2120: this commit should not do 2 things. rather split it up for better review: 
1) fix the forfeit amount bug
2) then do the sig response change

round/actor.go
```diff
@@ -2446,6 +2457,27 @@ func (a *RoundClientActor) handleForfeitSignatureResponse(ctx context.Context,
...
+			if _, ok := collecting.ExpectedForfeits[resp.VTXOOutpoint]; ok {
+				roundFSM = candidate
+				exists = true
+
+				break
+			}
+		}
+	}
+
 	if !exists {
```
ellemouton: dont have 2 separate `!exists` blocks.

round/vtxo_messages.go
```diff
@@ -83,6 +83,11 @@ type ForfeitRequestEvent struct {
 	// collaborative leaf when the live output being settled uses
 	// a custom script policy.
 	ForfeitSpend *arkscript.SpendPath
+
+	// ResponseTarget receives the signed forfeit response. Round-originated
+	// requests set this to the round actor so the response does not depend
+	// on the VTXO manager relay path.
```
ellemouton: can you explain why the original reply path doesnt work/isnt reliable?

txconfirm/broadcaster_test.go
```diff
@@ -500,6 +508,36 @@ func TestBroadcasterHelperFunctions(t *testing.T) {
...
+	for i := range hash {
+		hash[i] = byte(i + 1)
+	}
+
+	script, err := txscript.NewScriptBuilder().
+		AddOp(txscript.OP_0).
+		AddData(hash).
+		Script()
+	if err != nil {
+		panic(err)
```
ellemouton: dont do panics in tests. rather pass in `t` and then do require.NoError

vtxo/transitions.go
ellemouton: Done — `BuildForfeitTxWithContext` now takes a separate `connectorAmount` parameter; the existing `vtxoAmount` stays as-is. Penalty output value is computed as `vtxoAmount + connectorAmount` inside the helper.

round/actor.go
ellemouton: Done — split into two commits:
- `round: Fix zero-fee forfeit amount validation`
- `round: Route forfeit signature responses directly`

round/actor.go
ellemouton: Done — extracted `lookupForfeitSignatureRound` so the keyed lookup and the ExpectedForfeits fallback live in one helper; `handleForfeitSignatureResponse` has a single `if roundFSM == nil` branch now.

round/vtxo_messages.go
ellemouton: Reworded the commit body to spell it out — both hops in the relay path (VTXO actor → manager → round) Tell with the original server-message handler's context. The VTXO actor signs the forfeit asynchronously, AFTER that handler has already returned, so by the time the relay messages need to be delivered the context is canceled and the message can be silently dropped at either hop. Threading a `ResponseTarget` lets the VTXO actor Tell the round actor's SelfRef directly with `context.WithoutCancel`, detached from the now-dead handler context.

txconfirm/broadcaster_test.go
ellemouton: Done — `p2trTestPkScript` and `makeWalletUTXO` now take `*testing.T` and use `require.NoError`. The closure inside `rewritingWallet.NewWalletPkScript` (which has no `t` in scope) now requires `changeScript` to be set at construction time and returns an error if it's empty; all 4 `rewritingWallet{}` instantiations pass `p2trTestPkScript(t)` explicitly.

lib/tx/forfeit.go
```diff
@@ -119,7 +125,7 @@ func BuildForfeitTxWithContext(vtxoOutpoint *wire.OutPoint,
 	})
 
 	tx.AddTxOut(&wire.TxOut{
-		Value:    int64(vtxoAmount),
+		Value:    int64(vtxoAmount + connectorAmount),
```
Roasbeef, line 128: Ah so they weren't exactly zero fee

round/harness_test.go
```diff
@@ -2035,7 +2035,7 @@ func (h *boardingTestHarness) newTestForfeitTx(
 
 	// Output 0: Penalty to server.
 	tx.AddTxOut(&wire.TxOut{
-		Value:    50000,
+		Value:    50546,
```
Roasbeef: 546 being the connector value? Could put it out into a contant.

round/transitions.go
```diff
@@ -2422,6 +2432,43 @@ func forfeitRequestMap(
...
+			return nil, fmt.Errorf("missing local forfeit request "+
+				"for %s", outpoint)
+		}
+		if req.Amount <= 0 {
+			return nil, fmt.Errorf("invalid local forfeit amount "+
+				"%d for %s", req.Amount, outpoint)
+		}
+
+		infoCopy := *info
+		infoCopy.VTXOAmount = req.Amount
```
Roasbeef, line 2465: Should double check that this doesn't break anything on the server side. Now we bind the fees at round seal time, so the amount that you request may not be the VTXO actually created (fees may be deducted from it).

round/actor.go
```diff
@@ -2102,20 +2102,31 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
 				a.log.DebugS(ctx, "Looking up VTXO actor by service key",
 					slog.String("outpoint", m.VTXOOutpoint.String()))
 
+				// The VTXO actor may process this request after the
+				// current server-message handler returns. Detach the
+				// context so its follow-up ForfeitSignatureResponse
+				// relay is not canceled before it reaches this round.
+				reqCtx := context.WithoutCancel(ctx)
```
Roasbeef, line 2125: If we clear out the cancel here, then do we still need response target? 

Alternatively should we just pass a background context here? Even if the caller gets a cancel req or w/e we _should_ still send thru the request to the vtxo actor IMO. 

Feels like adding this fast path is masking some issues re lack of rules of thumb re context passing/management.

txconfirm/broadcaster.go
```diff
@@ -1467,3 +1492,72 @@ func isPackageSubmissionUnsupported(err error) bool {
...
+// replacement") implies a competing tx is already in mempool by
+// definition, so the parent must be present even if not named.
+func isParentKnownChildFailed(parentTxid chainhash.Hash, err error) bool {
+	if err == nil {
+		return false
+	}
+
+	errStr := err.Error()
+
+	parentMarker := "txid=" + parentTxid.String() + ":"
```
Roasbeef: Lotta parsing code here for this error....lol

Can we re-use some of the lnd/btcwallet error mapping here? 

https://github.com/btcsuite/btcd/blob/c794769dc23fef28ed99b1fa3a1be92582559452/rpcclient/errors.go#L493-L498

Roasbeef: We try to keep this up to date as well, so then we won't end up in prior sitautions where bitcoind changes the error string and we misclasisfy the error. 

Also related: https://github.com/lightningnetwork/lnd/blob/6fd5b7bb27d7fedbded2a6dc910785414f81a56c/lnwallet/btcwallet/btcwallet.go#L1116-L1143

txconfirm/broadcaster.go
```diff
@@ -1467,3 +1492,72 @@ func isPackageSubmissionUnsupported(err error) bool {
...
+	"bad-txns-inputs-missingorspent",
+	"txn-mempool-conflict",
+}
+
+// isParentKnownChildFailed returns true when a SubmitPackage error
+// indicates that the only failure is on the CPFP child, with the parent
+// already broadcast by another path.
+//
+// Two flavours of error can surface this state. The package error from
+// btcwbackend / chainbackends.handlePackageResult joins per-tx results as
```
Roasbeef: Same here re reusing the existing error mapping functions we have. This also makes assumptions re what the error string will exactly look like which can be fragile.

lib/tree/node.go
```diff
@@ -73,10 +73,10 @@ func NewLeafNode(input wire.OutPoint, leaf LeafDescriptor,
 	error) {
 
 	// The cosigners for a leaf are the leaf owner and operator.
-	cosigners := []*btcec.PublicKey{
+	cosigners := UniqueCosigners([]*btcec.PublicKey{
```
Roasbeef, line 76: Why would there ever be a dup here?

Roasbeef (approved):

lib/tree/node.go
ellemouton: for connector trees as the operator is the only signer/key there.

lib/tx/forfeit.go
ellemouton: Right — penalty output value is `vtxoAmount + connectorAmount`, so the connector dust is rolled into the penalty rather than left as a separate output. Net fee on the forfeit tx itself is zero (anchor is unspendable until CPFP). This is unchanged behaviour on this PR; we just kept it explicit in the helper signature.

round/harness_test.go
ellemouton: Done — extracted `testConnectorDustValue` (546) and `testForfeitVTXOValue` (50000) constants; the penalty output is now built as `testForfeitVTXOValue + testConnectorDustValue`. Squashed into `round: Fix zero-fee forfeit amount validation`.

round/transitions.go
ellemouton: Good thing to double-check, but I think we're safe here:

`ForfeitRequest.Amount` is set in the wallet from the *existing* VTXO's stored on-chain amount (`vtxo.Amount` at `wallet/wallet.go:881`). Forfeits consume an existing VTXO as an *input*; seal-time fees under #270 are deducted from *output*-side slots only — exactly one `VTXORequest` or `LeaveRequest` carries `IsChange=true` and absorbs the residual at seal time. The forfeit input value itself is not adjustable at seal time.

The server also cross-checks: the round actor validates that the forfeit penalty output equals `connectorInfo.VTXOAmount + connectorInfo.ConnectorAmount` (transitions.go:1767), where `VTXOAmount` is taken from the server's local view of the VTXO. If the client's view ever diverged from the server's, the forfeit signature would be rejected.

The stale comment on `populateForfeitMappingAmounts` referencing "server-supplied amounts" actually motivated this PR — we want the local VTXO amount to drive validation, not echo whatever the server proto carries.

lib/tree/node.go
ellemouton: To add to ellemouton's note: connector trees use the operator key as the *only* leaf cosigner (and as one of the two operands fed into `UniqueCosigners` here at the leaf level). Without dedup, MuSig2 key aggregation hands you `MuSig2(opKey, opKey)` which fails downstream with non-deterministic key-coefficient errors. The dedup is what lets us use the same `NewLeafNode` path uniformly for VTXO trees (where leaf owner ≠ operator) and connector trees (where leaf owner == operator).

txconfirm/broadcaster.go
ellemouton: Done — refactored to use `rpcclient.MapRPCErr` instead of substring lists:

- New `chainbackends.PackageTxError` type wraps each per-tx package result; `Unwrap()` returns the result of `rpcclient.MapRPCErr(errors.New(reason))`, so each entry resolves to a `rpcclient.BitcoindRPCErr` sentinel.
- All three backends — `chainbackends/lnd.go`, `btcwbackend/chain_backend.go`, AND `lwwallet/esplora.go` — now build `*PackageTxError` per per-tx result and `errors.Join` them, instead of formatting strings.
- `isParentKnownChildFailed` walks the joined error tree (`chainbackends.WalkPackageTxErrors`) and classifies entries via `errors.Is` against `rpcclient.ErrTxAlreadyKnown` / `ErrTxAlreadyInMempool` / `ErrTxAlreadyConfirmed` (parent-known) and `ErrInsufficientFee` / `ErrMissingInputsOrSpent` / `ErrMissingInputs` / `ErrMempoolConflict` / `ErrConflictingTx` (child-failed). The hand-rolled marker lists are gone.
- The RBF "rejecting replacement" fallback is now properly gated: it only fires when the parent's per-tx entry is **absent** from the joined error (`!parentSeen`). Previously a parent echoed with a fatal reason (e.g. `bad-witness`) plus an unrelated child carrying the RBF marker would have been silently treated as "already broadcast". Added a regression test for that exact case.

One remaining carve-out: the "rejecting replacement; new feerate ... <= old feerate ..." RBF trace string is still kept as a substring fallback because upstream `rpcclient.MapRPCErr` doesn't yet map it to a typed sentinel. Documented inline in the `isParentKnownChildFailed` doc comment.

All squashed into `txconfirm: Tolerate parent-already-broadcast on initial submit` and the existing test rebuilt to construct `*PackageTxError` values directly via `chainbackends.NewPackageTxError` rather than synthetic strings.

txconfirm/broadcaster.go
ellemouton: Done — refactored to use `rpcclient.MapRPCErr` instead of substring lists:

- New `chainbackends.PackageTxError` type wraps each per-tx package result; `Unwrap()` returns the result of `rpcclient.MapRPCErr(errors.New(reason))`, so each entry resolves to a `rpcclient.BitcoindRPCErr` sentinel.
- All three backends — `chainbackends/lnd.go`, `btcwbackend/chain_backend.go`, AND `lwwallet/esplora.go` — now build `*PackageTxError` per per-tx result and `errors.Join` them, instead of formatting strings.
- `isParentKnownChildFailed` walks the joined error tree (`chainbackends.WalkPackageTxErrors`) and classifies entries via `errors.Is` against `rpcclient.ErrTxAlreadyKnown` / `ErrTxAlreadyInMempool` / `ErrTxAlreadyConfirmed` (parent-known) and `ErrInsufficientFee` / `ErrMissingInputsOrSpent` / `ErrMissingInputs` / `ErrMempoolConflict` / `ErrConflictingTx` (child-failed). The hand-rolled marker lists are gone.
- The RBF "rejecting replacement" fallback is now properly gated: it only fires when the parent's per-tx entry is **absent** from the joined error (`!parentSeen`). Previously a parent echoed with a fatal reason (e.g. `bad-witness`) plus an unrelated child carrying the RBF marker would have been silently treated as "already broadcast". Added a regression test for that exact case.

One remaining carve-out: the "rejecting replacement; new feerate ... <= old feerate ..." RBF trace string is still kept as a substring fallback because upstream `rpcclient.MapRPCErr` doesn't yet map it to a typed sentinel. Documented inline in the `isParentKnownChildFailed` doc comment.

All squashed into `txconfirm: Tolerate parent-already-broadcast on initial submit` and the existing test rebuilt to construct `*PackageTxError` values directly via `chainbackends.NewPackageTxError` rather than synthetic strings.

round/actor.go
ellemouton: yeah - removed the fast pass workaround. 


the reason it used WithoutCancel isntead of background is to keep context values

PR #337, opened 2026-05-05
sdk: validate onion-forwarded out-swap events

sputn1ck:
## Summary

- add the mailbox-delivered out-swap HTLC event RPC shape
- have the SDK pull per-receive HTLC events from mailbox and validate the forwarded onion locally
- keep receive auth as a single client key, persist receive metadata for resume, and bound receive-claim reconciliation

## Tests

- `go test -count=1 ./sdk/swaps ./swaprpc`

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_swap.go
+++ b/cmd/darepocli/darepoclicommands/cmd_swap.go
@@ -17,6 +17,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btclog/v2"
 	"github.com/lightninglabs/darepo-client/daemonrpc"
+	mailboxpb "github.com/lightninglabs/darepo-client/mailbox/pb"
 	sdkark "github.com/lightninglabs/darepo-client/sdk/ark"
 	"github.com/lightninglabs/darepo-client/sdk/swaps"
 	"github.com/lightningnetwork/lnd/keychain"
@@ -211,7 +212,7 @@ func swapReceive(cmd *cobra.Command, _ []string) error {
 
 	if verbose {
 		if err := printReceiveVerboseStart(
-			cmd, amount, session,
+			cmd, amount,
 		); err != nil {
 			return err
 		}
@@ -223,9 +224,11 @@ func swapReceive(cmd *cobra.Command, _ []string) error {
 	}
 
 	if verbose {
-		printReceiveVerboseFunding(
+		if err := printReceiveVerboseFunding(
 			cmd, outpoint, fundedAmount, session,
-		)
+		); err != nil {
+			return err
+		}
 	}
 
 	result, err := session.Claim(ctx, outpoint, fundedAmount)
@@ -243,25 +246,14 @@ func swapReceive(cmd *cobra.Command, _ []string) error {
 	return nil
 }
 
-// printReceiveVerboseStart prints the expected incoming vHTLC scripts for one
-// receive flow before funding is observed.
-func printReceiveVerboseStart(cmd *cobra.Command, amount int64,
-	session *swaps.ReceiveSession) error {
-
-	info, err := session.VHTLCInfo()
-	if err != nil {
-		return fmt.Errorf("describe receive vhtlc: %w", err)
-	}
-
+// printReceiveVerboseStart prints the expected receive amount before the
+// server's mailbox event has provided the concrete vHTLC script.
+func printReceiveVerboseStart(cmd *cobra.Command, amount int64) error {
 	fmt.Fprintf(cmd.OutOrStdout(),
 		"\nIncoming vHTLC\n"+
 			"  amount:        %d sat\n"+
-			"  output script: %x\n"+
-			"  claim script:  %x\n"+
 			"  waiting for:   funding via paid invoice\n",
 		amount,
-		info.PkScript,
-		info.ClaimScript,
 	)
 
 	return nil
@@ -270,18 +262,29 @@ func printReceiveVerboseStart(cmd *cobra.Command, amount int64,
 // printReceiveVerboseFunding prints the funded vHTLC acceptance and the
 // upcoming preimage sweep step.
 func printReceiveVerboseFunding(cmd *cobra.Command, outpoint string,
-	amount int64, session *swaps.ReceiveSession) {
+	amount int64, session *swaps.ReceiveSession) error {
+
+	info, err := session.VHTLCInfo()
+	if err != nil {
+		return fmt.Errorf("describe receive vhtlc: %w", err)
+	}
 
 	fmt.Fprintf(cmd.OutOrStdout(),
 		"\nAccepted incoming vHTLC\n"+
 			"  outpoint: %s\n"+
 			"  amount:   %d sat\n"+
+			"  output script: %x\n"+
+			"  claim script:  %x\n"+
 			"\nSweeping vHTLC\n"+
 			"  preimage: %x\n",
 		outpoint,
 		amount,
+		info.PkScript,
+		info.ClaimScript,
 		session.Preimage[:],
 	)
+
+	return nil
 }
 
 // swapPay executes the swap pay command.
@@ -530,11 +533,15 @@ func buildSwapClient(cmd *cobra.Command,
 	}
 
 	serverConn := swaps.NewGRPCSwapServerConn(swapConn)
+	mailboxClient := mailboxpb.NewMailboxServiceClient(swapConn)
 
 	client := swaps.NewSwapClientWithStore(
 		serverConn, arkClient, nil,
 		clientInvoiceGenerator(invoiceKey, daemonClient, chainParams),
 		store,
+	)
+	client.SetOutSwapEventReceiver(
+		swaps.NewMailboxOutSwapEventReceiver(mailboxClient, ""),
 	)
 
 	cleanup := func() {
--- a/sdk/swaps/client.go
+++ b/sdk/swaps/client.go
@@ -5,10 +5,12 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/ecdsa"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btclog/v2"
 	sdkark "github.com/lightninglabs/darepo-client/sdk/ark"
 	"github.com/lightningnetwork/lnd/invoices"
+	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/lightningnetwork/lnd/lntypes"
 )
 
@@ -145,6 +147,16 @@ type InvoiceCreator interface {
 		routeHint *RouteHint, expiry time.Duration,
 		preimage *lntypes.Preimage) (*invoices.Invoice, lntypes.Hash,
 		error)
+
+	// CreateInvoiceWithKey builds one signed invoice using the client's
+	// receive auth key. Receive swaps use this key as the invoice
+	// destination and later decode the forwarded final-hop onion with it.
+	CreateInvoiceWithKey(ctx context.Context,
+		amountSat btcutil.Amount, memo string,
+		routeHint *RouteHint, expiry time.Duration,
+		authKey keychain.SingleKeyMessageSigner,
+		preimage *lntypes.Preimage) (*invoices.Invoice, lntypes.Hash,
+		error)
 }
 
 // RouteHint describes a single hop hint for Lightning invoices.
@@ -193,6 +205,46 @@ type VHTLCConfig struct {
 	SwapServerPubkey []byte
 }
 
+// OutSwapHtlcEvent carries the server-funded HTLC metadata that the client
+// validates before revealing its invoice preimage.
+type OutSwapHtlcEvent struct {
+	// PaymentHash is the intercepted Lightning payment hash.
+	PaymentHash lntypes.Hash
+
+	// AmountSat is the amount funded by the server.
+	AmountSat int64
+
+	// OnionBlob is the raw final-hop onion blob forwarded by the server.
+	OnionBlob []byte
+
+	// VHTLCConfig contains the script parameters for the funded vHTLC.
+	VHTLCConfig VHTLCConfig
+}
+
+// OutSwapHtlcNotification carries one mailbox-delivered out-swap HTLC event
+// and an optional acknowledgement hook.
+type OutSwapHtlcNotification struct {
+	Event     *OutSwapHtlcEvent
+	AckCursor uint64
+	Ack       func(context.Context) error
+}
+
+// OutSwapEventReceiver waits for server-pushed out-swap mailbox events.
+type OutSwapEventReceiver interface {
+	WaitOutSwapHtlc(
+		ctx context.Context,
+		paymentHash lntypes.Hash,
+		mailboxPubkey *btcec.PublicKey,
+	) (*OutSwapHtlcNotification, error)
+
+	AckOutSwapHtlc(
+		ctx context.Context,
+		paymentHash lntypes.Hash,
+		mailboxPubkey *btcec.PublicKey,
+		cursor uint64,
+	) error
+}
+
 // InSwapConfig is returned by the server when creating an in-swap.
 type InSwapConfig struct {
 	// PaymentHash is the SHA-256 payment hash extracted from the
@@ -224,13 +276,13 @@ type InSwapConfig struct {
 // gRPC service. This allows the client to talk to the swap server
 // without importing the server module.
 type SwapServerConn interface {
-	// RequestChannelID asks the server for a route hint and
-	// the locked-in vHTLC configuration for this swap.
+	// RequestChannelID asks the server for a route hint for this swap.
 	RequestChannelID(
 		ctx context.Context,
 		vhtlcPubkey *btcec.PublicKey,
+		paymentHash lntypes.Hash,
 		expirySeconds uint32,
-	) (*RouteHint, *VHTLCConfig, error)
+	) (*RouteHint, error)
 
 	// CreateInSwap initiates an Ark->LN swap on the server.
 	CreateInSwap(
@@ -314,6 +366,28 @@ type DaemonConn interface {
 	AllocateReceiveScript(
 		ctx context.Context, label string,
 	) (*ReceiveInfo, error)
+
+	// ReceiveAuthKey returns the payment-scoped receive-auth public key.
+	ReceiveAuthKey(ctx context.Context,
+		paymentHash lntypes.Hash) (*btcec.PublicKey, error)
+
+	// SignReceiveAuthMessage signs one message with the payment-scoped
+	// receive-auth key.
+	SignReceiveAuthMessage(ctx context.Context,
+		paymentHash lntypes.Hash, message []byte,
+		doubleHash bool) (*ecdsa.Signature, error)
+
+	// SignReceiveAuthMessageCompact signs one message with the
+	// payment-scoped receive-auth key and returns a compact signature.
+	SignReceiveAuthMessageCompact(ctx context.Context,
+		paymentHash lntypes.Hash, message []byte,
+		doubleHash bool) ([]byte, error)
+
+	// ReceiveAuthECDH derives one Sphinx shared secret with the
+	// payment-scoped receive-auth key.
+	ReceiveAuthECDH(ctx context.Context,
+		paymentHash lntypes.Hash,
+		pubKey *btcec.PublicKey) ([32]byte, error)
 }
 
 // CustomInput aliases the Ark SDK's typed custom OOR input.
@@ -331,6 +405,7 @@ type SwapClient struct {
 	server     SwapServerConn
 	daemon     DaemonConn
 	invoiceGen InvoiceCreator
+	outEvents  OutSwapEventReceiver
 	store      *Store
 	log        btclog.Logger
 
@@ -342,7 +417,16 @@ type SwapClient struct {
 	refundLocktimeBuffer     uint32
 	claimRetryDelay          time.Duration
 	claimMaxAttempts         int
+	decodeOutSwapOnion       outSwapOnionDecoder
 	now                      func() time.Time
+}
+
+// SetOutSwapEventReceiver sets the mailbox event receiver used by
+// ReceiveViaLightning. Callers should configure this before starting receives.
+func (c *SwapClient) SetOutSwapEventReceiver(
+	receiver OutSwapEventReceiver) {
+
+	c.outEvents = receiver
 }
 
 // NewSwapClient creates a new swap client. The invoice creator may be nil for
@@ -366,10 +450,16 @@ func NewSwapClientWithStore(server SwapServerConn, daemon DaemonConn,
 		log = btclog.Disabled
 	}
 
+	var outEvents OutSwapEventReceiver
+	if receiver, ok := server.(OutSwapEventReceiver); ok {
+		outEvents = receiver
+	}
+
 	return &SwapClient{
 		server:                   server,
 		daemon:                   daemon,
 		invoiceGen:               invoiceGen,
+		outEvents:                outEvents,
 		store:                    store,
 		log:                      log,
 		waitPollInterval:         2 * time.Second,
@@ -380,6 +470,7 @@ func NewSwapClientWithStore(server SwapServerConn, daemon DaemonConn,
 		refundLocktimeBuffer:     defaultRefundLocktimeBuffer,
 		claimRetryDelay:          time.Second,
 		claimMaxAttempts:         10,
+		decodeOutSwapOnion:       decodeOutSwapOnion,
 		now:                      time.Now,
 	}
 }
--- a/sdk/swaps/invoice.go
+++ b/sdk/swaps/invoice.go
@@ -8,26 +8,30 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
-	"github.com/btcsuite/btcd/btcec/v2/ecdsa"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/lightningnetwork/lnd/graph/db/models"
 	"github.com/lightningnetwork/lnd/invoices"
 	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/lightningnetwork/lnd/lnrpc/invoicesrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/netann"
+	"github.com/lightningnetwork/lnd/routing/route"
 	"github.com/lightningnetwork/lnd/zpay32"
 )
 
 const (
-	// defaultInvoiceExpiry is the default expiry duration for
-	// swap invoices when none is specified by the caller.
+	// defaultInvoiceExpiry is the default expiry duration for swap invoices
+	// when none is specified by the caller.
 	defaultInvoiceExpiry = time.Hour
+
+	// defaultCLTVExpiry is the default CLTV expiry delta for invoices.
+	defaultCLTVExpiry = 40
 )
 
-type compactInvoiceSigner func([]byte) ([]byte, error)
-
-// InvoiceStore persists created invoices so they can be looked up
-// when HTLCs arrive.
+// InvoiceStore persists created invoices so they can be looked up when HTLCs
+// arrive.
 type InvoiceStore interface {
 	// AddInvoice stores a new invoice keyed by payment hash.
 	AddInvoice(
@@ -37,8 +41,8 @@ type InvoiceStore interface {
 	) (uint64, error)
 }
 
-// NewPreimage generates a cryptographically random 32-byte
-// preimage suitable for use as a Lightning payment preimage.
+// NewPreimage generates a cryptographically random 32-byte preimage suitable
+// for use as a Lightning payment preimage.
 func NewPreimage() (lntypes.Preimage, error) {
 	var preimage lntypes.Preimage
 	if _, err := rand.Read(preimage[:]); err != nil {
@@ -50,19 +54,19 @@ func NewPreimage() (lntypes.Preimage, error) {
 	return preimage, nil
 }
 
-// InvoiceGenerator creates properly signed Lightning invoices
-// with route hints pointing through the swap server using direct
-// BOLT-11 encoding and a SingleKeyMessageSigner.
+// InvoiceGenerator creates properly signed Lightning invoices with route hints
+// pointing through the swap server. It delegates invoice construction to lnd's
+// invoicesrpc.AddInvoice machinery so feature bits, payment address handling,
+// and invoice storage follow the same path as normal lnd invoices.
 type InvoiceGenerator struct {
-	signer      keychain.SingleKeyMessageSigner
+	invoiceCfg  *invoicesrpc.AddInvoiceConfig
 	chainParams *chaincfg.Params
 }
 
-// DirectInvoiceCreator creates signed BOLT-11 invoices directly from a private
-// key without depending on a full lnd invoice registry.
+// DirectInvoiceCreator is kept for source compatibility with older SDK users.
+// It now delegates to InvoiceGenerator instead of manually encoding invoices.
 type DirectInvoiceCreator struct {
-	privKey     *btcec.PrivateKey
-	chainParams *chaincfg.Params
+	generator *InvoiceGenerator
 }
 
 // MemoryInvoiceStore keeps invoices in memory for ephemeral callers such as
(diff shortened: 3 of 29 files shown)


swaprpc/swap.proto
```diff
@@ -28,17 +28,55 @@ message RequestChannelIdRequest {
...
     // paths.
     bytes client_vhtlc_pubkey = 2;
+
+    // payment_hash is the invoice payment hash. The swap server combines it
+    // with client_vhtlc_pubkey to derive a unique virtual channel ID without
+    // adding a separate auth identity.
+    bytes payment_hash = 3;
+
+    reserved 4, 5;
+    reserved "auth_pubkey", "client_mailbox_pubkey", "auth_signature";
```
Roasbeef: Why reserve here?

swaprpc/swap.proto
```diff
@@ -28,17 +28,55 @@ message RequestChannelIdRequest {
...
+// RequestChannelIdResponse returns the route hint for one receive negotiation.
 message RequestChannelIdResponse {
     // route_hint contains the hop hint the client should embed in the invoice.
     RouteHint route_hint = 1;
+}
+
+// OutSwapHtlcEvent describes one server-funded intercepted HTLC. The event is
+// intentionally raw from the Lightning side: the client must decode onion_blob
+// with the invoice/auth private key before it claims the Ark vHTLC.
+message OutSwapHtlcEvent {
```
Roasbeef, line 47: Added but not included in any other protos?

cmd/darepocli/darepoclicommands/cmd_swap.go
```diff
@@ -536,6 +536,7 @@ func buildSwapClient(cmd *cobra.Command,
 		clientInvoiceGenerator(invoiceKey, daemonClient, chainParams),
 		store,
 	)
+	client.SetReceiveAuthKey(invoiceKey)
```
Roasbeef: I think we should get this directly from lnd via an interface, as otherwise IIUC this won't be able to survive restarts. 

May be out of the scope of this PR however, in which case we can track with an error.

sdk/swaps/client.go
```diff
@@ -145,6 +146,16 @@ type InvoiceCreator interface {
...
 		preimage *lntypes.Preimage) (*invoices.Invoice, lntypes.Hash,
 		error)
+
+	// CreateInvoiceWithKey builds one signed invoice using the client's
+	// receive auth key. Receive swaps use this key as the invoice
+	// destination and later decode the forwarded final-hop onion with it.
+	CreateInvoiceWithKey(ctx context.Context,
+		amountSat btcutil.Amount, memo string,
+		routeHint *RouteHint, expiry time.Duration,
+		authKey *btcec.PrivateKey,
```
Roasbeef: Passing raw private key is an anti pattern. Should be done via an interface isntead.

sdk/swaps/client.go
```diff
@@ -394,3 +480,23 @@ func (c *SwapClient) currentTime() time.Time {
...
+// creating one lazily when the caller did not inject a durable key.
+func (c *SwapClient) receiveAuthPrivateKey() (*btcec.PrivateKey, error) {
+	c.receiveAuthMu.Lock()
+	defer c.receiveAuthMu.Unlock()
+
+	if c.receiveAuthKey != nil {
+		return c.receiveAuthKey, nil
+	}
+
+	key, err := btcec.NewPrivateKey()
```
Roasbeef: Same here re in memory private key

sdk/swaps/invoice.go
```diff
@@ -176,8 +201,33 @@ func (g *DirectInvoiceCreator) CreateInvoice(_ context.Context,
...
+		g.chainParams, func(msg []byte) ([]byte, error) {
+			digest := chainhash.HashB(msg)
+
+			return ecdsa.SignCompact(g.privKey, digest, true), nil
+		},
+	)
+}
+
+// CreateInvoiceWithKey creates one invoice signed by the supplied auth key.
+func (g *DirectInvoiceCreator) CreateInvoiceWithKey(_ context.Context,
```
Roasbeef: Does trying to re-use some/part/all of lnd's invoice manager make sense here? Or this is more specific

sdk/swaps/out_swap.go
```diff
@@ -647,6 +664,167 @@ func (s *ReceiveSession) waitForFunding(ctx context.Context) error {
...
+
+	payload, err := decoder(
+		authKey, s.PaymentHash, event.OnionBlob,
+	)
+	if err != nil {
+		return err
+	}
+
+	expectedMsat := lnwire.NewMSatFromSatoshis(s.amountSat)
+	if payload.amountToForward != expectedMsat {
```
Roasbeef, line 877: 👍

sdk/swaps/out_swap_mailbox.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+			if event.PaymentHash != paymentHash {
+				continue
+			}
+			ackCursor := env.GetEventSeq() + 1
+
+			return &OutSwapHtlcNotification{
+				Event: event,
+				Ack: func(ctx context.Context) error {
+					return r.ack(ctx, mailboxID, ackCursor)
```
Roasbeef: Do we need to worry about potential callers calling `ack` out of order? Like if one swap takes more time to process vs another.

sdk/swaps/invoice.go
sputn1ck: Ah, I was under the impression we actually did that, as the prototype used lnds invoices rpc server. gonna move to that.

sdk/swaps/out_swap.go
```diff
@@ -647,6 +664,167 @@ func (s *ReceiveSession) waitForFunding(ctx context.Context) error {
...
+	authKey *btcec.PrivateKey) error {
+
+	if authKey == nil {
+		return fmt.Errorf("receive auth key must be provided")
+	}
+
+	decoder := decodeOutSwapOnion
+	if s.client != nil && s.client.decodeOutSwapOnion != nil {
+		decoder = s.client.decodeOutSwapOnion
+	}
```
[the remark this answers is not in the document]
sputn1ck: Fixed. `decodeOutSwapOnion` is initialized by `NewSwapClientWithStore`, and the validation path now uses the configured decoder directly.

sdk/swaps/out_swap_mailbox.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, false, nil
+	}
+
+	body := env.GetBody()
+	if body == nil {
+		return nil, false, fmt.Errorf("out-swap event missing body")
+	}
+
+	var protoEvent anypb.Any
+	protoEvent = *body
```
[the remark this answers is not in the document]
sputn1ck: Already fixed by the current mailbox unwrapping path. The code now unwraps the `Any` with `UnmarshalNew` and converts the resulting `swaprpc.OutSwapHtlcEvent`.

sdk/swaps/store_sessions.go
```diff
@@ -361,23 +361,31 @@ func (s *ReceiveSession) persist(ctx context.Context) error {
...
+		swapServerPubKey = s.swapServerPubKey.SerializeCompressed()
+	}
+	clientPubKey := []byte(nil)
+	if s.clientPubKey != nil {
+		clientPubKey = s.clientPubKey.SerializeCompressed()
+	}
+	operatorPubKey := []byte(nil)
+	if s.operatorPubKey != nil {
+		operatorPubKey = s.operatorPubKey.SerializeCompressed()
+	}
```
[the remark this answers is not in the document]
sputn1ck: Fixed. Optional compressed public key serialization is now centralized in a helper, so receive persistence no longer repeats the nil-check blocks.

sdk/swaps/store.go
```diff
@@ -48,7 +48,7 @@ const (
 	defaultConnMaxLifetime = 10 * time.Minute
 
 	// LatestMigrationVersion is the latest swap-client schema migration.
-	LatestMigrationVersion uint = 1
+	LatestMigrationVersion uint = 2
```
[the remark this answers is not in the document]
sputn1ck: Fixed. Migration version 2 now exists. It adds the persisted receive-auth key table, and `LatestMigrationVersion` remains aligned with the highest migration file.

sdk/swaps/migrations/000001_swap_sessions.up.sql
```diff
@@ -31,6 +31,10 @@ CREATE TABLE IF NOT EXISTS receive_swaps (
     -- client_pubkey is the receiver/client key used in the expected vHTLC.
     client_pubkey BLOB NOT NULL,
 
+    -- payment_addr is the invoice MPP payment address the client validates
+    -- against the forwarded final-hop onion.
+    payment_addr BLOB NOT NULL DEFAULT x'',
+
```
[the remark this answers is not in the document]
sputn1ck: Leaving this in `000001` intentionally for this PR stack. The swap SDK schema is not in production yet, so we prefer keeping the first schema complete rather than carrying a temporary migration solely for an unreleased column. Migration 2 is reserved for the receive-auth key table added by this review pass.

sdk/swaps/out_swap.go
```diff
@@ -611,18 +601,45 @@ func (s *ReceiveSession) prepareInvoice(ctx context.Context) error {
...
 		s.clientPubKey = clientKey
 		s.operatorPubKey = operatorKey
-		s.swapServerPubKey = serverKey
-		s.vhtlcConfig = *vhtlcCfg
-		s.vhtlcPolicy = policy
-		s.vhtlcPolicyTemplate = policyTemplate
-		s.vhtlcPkScript = pkScript
+		s.paymentAddr = inv.Terms.PaymentAddr
 
 		return s.transition(receiveEventInvoiceCreated)
```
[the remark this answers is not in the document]
sputn1ck: This is covered by the current receive flow. Claim-side validation still requires the mailbox-delivered HTLC event to populate the vHTLC config before a claim can proceed, and the outdated line no longer reflects the current implementation.

swaprpc/swap.proto
sputn1ck: Removed. Those fields were speculative and not needed for the current wire shape.

swaprpc/swap.proto
sputn1ck: Correct, this is intentionally not a unary/streaming `SwapService` RPC. It is the payload type carried inside mailbox envelopes as a protobuf `Any`; the mailbox RPC metadata identifies it as the out-swap HTLC event route.

cmd/darepocli/darepoclicommands/cmd_swap.go
sputn1ck: Fixed. The CLI no longer injects its invoice key as the receive-auth key. The SDK now generates a local receive-auth key and persists it in the isolated swap DB, so receive invoice identity and onion decoding survive restarts.

sdk/swaps/client.go
sputn1ck: Fixed. The receive invoice path now takes the signer interface required by lnd’s invoice machinery instead of a raw private key. The default SDK receive-auth implementation hides the private key behind signer/ECDH interfaces.

sdk/swaps/client.go
sputn1ck: Fixed. The receive-auth key is persisted in the SDK swap DB when a store is configured. The in-memory key is now only the fallback for callers that construct a swap client without durable swap storage.

sdk/swaps/invoice.go
sputn1ck: Fixed. Invoice generation is back on the `invoicesrpc.AddInvoice` path with `AddInvoiceConfig`, `NodeSigner`, route hints, feature generation, best height, and invoice store plumbing. The direct zpay32 encoding path was removed from production invoice creation.

sdk/swaps/out_swap_mailbox.go
sputn1ck: We should not have cross-swap ack ordering issues because out-swap HTLC events are pulled from a per-payment mailbox. The mailbox ID is derived from the stable client mailbox key plus the invoice payment hash, so concurrent receive sessions do not share one `AckUpTo` cursor. The ack is also returned only after the matching event is found, and callers invoke it after validation and persistence.

swaprpc/swap.proto
sputn1ck: Not super sure about this one here / the mailbox usage in general.

sdk/swaps/out_swap_mailbox.go
sputn1ck: This probably should be changed? So no cursor per invoice right?

bhandras:
I did a pass on the current PR head (`666f4800`) and wanted to write this in plain-language form, because the shape of the PR is important and there are a few wiring/durability edges that are easy to miss.

## What this PR is trying to add

The main goal is to make Lightning-to-Ark receives safer by having the client validate the actual Lightning HTLC before it reveals/uses the invoice preimage to claim the Ark vHTLC.

The new flow, as I understand it, is:

1. The client creates a receive invoice using a stable receive-auth key.
2. The payer pays that invoice.
3. The swap server sees the incoming Lightning HTLC.
4. The server funds an Ark vHTLC.
5. The server sends the client a mailbox event with the raw final-hop onion and vHTLC config.
6. The client decodes the onion locally and checks that it matches the invoice it created.
7. Only then does the client claim the Ark vHTLC with the preimage.

That is the right direction: the client should not claim just because the server says “here is a vHTLC”; it should first prove locally that the Lightning payment was really for its invoice, with the expected amount and payment address.

## Findings

### 1. CLI receive does not wire the new mailbox event receiver

The SDK receive FSM now requires an `OutSwapEventReceiver` before it can progress past invoice creation. In `waitForFunding`, we return immediately if `s.client.outEvents == nil`.

But `cmd/darepocli/darepoclicommands/cmd_swap.go` currently builds only:

```go
serverConn := swaps.NewGRPCSwapServerConn(swapConn)

client := swaps.NewSwapClientWithStore(
    serverConn, arkClient, nil,
    clientInvoiceGenerator(invoiceKey, daemonClient, chainParams),
    store,
)
```

`GRPCSwapServerConn` does not implement `OutSwapEventReceiver`, and the CLI never calls `client.SetOutSwapEventReceiver(...)`.

So in real CLI usage, `swap receive` will likely:

1. create and print the invoice,
2. enter `WaitForFunding`,
3. fail with `out-swap event receiver is not configured`.

The new `MailboxOutSwapEventReceiver` looks like the intended production implementation, but right now it is only used in tests. The CLI needs to construct it from the mailbox edge/client and wire it into the swap client, probably with something like:

```go
eventReceiver := swaps.NewMailboxOutSwapEventReceiver(mailboxClient, "")
client.SetOutSwapEventReceiver(eventReceiver)
```

The exact mailbox client construction is the part that needs to match the existing serverconn/mailbox wiring.

### 2. Resume can get stuck if we ack the mailbox event before reaching `VHTLCFunded`

`waitForFunding` currently does this:

1. Pull mailbox event.
2. Validate/accept the event.
3. Persist the vHTLC policy/config/script.
4. Ack the mailbox event.
5. Wait for the vHTLC to show up in the Ark indexer.
6. Transition to `VHTLCFunded`.

The fragile case is a crash or shutdown between steps 4 and 6.

At that point, the DB may contain the accepted vHTLC config/script, but the session state is still `InvoiceCreated`. The mailbox event has also been acked, so on restart it may not be redelivered.

Then resume loads the session and calls `waitForFunding` again. But `waitForFunding` unconditionally waits for another mailbox event before it waits for the vHTLC. Since the original event was already acked, the resumed receive can sit waiting for an event it already accepted before the crash.

I think we need one of these fixes:

- if a resumed `InvoiceCreated` session already has a persisted vHTLC policy/script/config, skip the mailbox event wait and go straight to `waitForVHTLC`; or
- introduce a distinct durable state after the event is accepted, before the mailbox ack; or
- delay ack until the session has transitioned far enough that resume does not need the event again.

The important invariant is: after `AckUpTo`, the local state must be sufficient to continue without seeing that event again.

### 3. `swap receive --verbose` can panic before funding is observed

The CLI calls `session.VHTLCInfo()` immediately after `StartReceiveViaLightning` when `--verbose` is set.

Before this PR, that made sense because the client got the vHTLC config during setup and could build the expected vHTLC script immediately.

After this PR, the vHTLC policy is not built until the mailbox event arrives. So right after invoice creation, `s.vhtlcPolicy` is still nil. `VHTLCInfo()` then does:

```go
claimPath, err := s.vhtlcPolicy.ClaimPath(s.Preimage)
```

That can panic on nil.

Simple fix options:

- make `VHTLCInfo()` return a clear error if the HTLC event has not been accepted yet; and/or
- move the verbose vHTLC script printing until after `WaitForFunding`, when the event has been accepted and the policy exists.

### 4. Manual `Claim` from `InvoiceCreated` still exists, but no longer has enough data to work

`Claim(ctx, outpoint, amount)` still has a special case for `ReceiveStateInvoiceCreated`. It validates the outpoint/amount and transitions directly to `VHTLCFunded`.

That matched the old flow, where the vHTLC config/policy was already known at invoice creation time.

With this PR, the vHTLC config/policy comes from the mailbox event. So while still in `InvoiceCreated`, the session has no real `vhtlcConfig`, no `vhtlcPolicy`, and no `vhtlcPkScript` yet.

Practically, this means manual claim from `InvoiceCreated` either fails with confusing refund-locktime behavior because `RefundLocktime` is zero, or would later fail because the claim path cannot be built.

I think this path should either:

- fetch/accept the mailbox HTLC event first, then allow the manual outpoint/amount claim; or
- explicitly reject manual `Claim` from `InvoiceCreated` with a clear error like “wait for/accept the out-swap HTLC event first”.

## Test run

I ran:

```bash
go test -count=1 ./sdk/swaps ./swaprpc
```

and it passed.

The reason the test suite does not catch the missing CLI wiring is that the SDK tests use a fake `testSwapServerConn` that implements both `SwapServerConn` and `OutSwapEventReceiver`. That is useful for unit tests, but it hides the fact that the production `GRPCSwapServerConn` does not provide mailbox events by itself.

sputn1ck:
Addressed in 4541d8d, with the server submodule pointer updated in 64cc900. I also rebased the client branch onto current main and force-pushed both dependent branches.

The receive flow now persists a durable HTLCEventAccepted state after validating the mailbox event and before acking it. On resume, that state skips mailbox redelivery and continues waiting for the vHTLC funding output.

I also wired the CLI receive path to the mailbox-backed event receiver, moved verbose vHTLC script printing until after the HTLC event has been accepted, and made pre-accept VHTLCInfo/Claim calls return clear errors instead of using missing vHTLC metadata.

Validated with:
- make unit pkg=./sdk/swaps timeout=5m
- make unit pkg=./cmd/darepocli/darepoclicommands timeout=5m
- make commitmsg-lint commit=HEAD

swaprpc/swap.proto
bhandras: I think this is pointing at a real schema-discoverability issue.

The intent is that `OutSwapHtlcEvent` is not returned by a typed `SwapService` RPC directly. The swap server wraps it in a mailbox envelope body, and the client unwraps that mailbox `Any` back into `swaprpc.OutSwapHtlcEvent`.

Given that, the proto currently looks a little loose: the event message exists in `swap.proto`, but the proto file does not show the transport relationship. I think it would be clearer to make the mailbox payload explicit with a small wrapper, for example:

```proto
message SwapMailboxEvent {
    oneof event {
        OutSwapHtlcEvent out_swap_htlc = 1;
    }
}
```

Then the mailbox envelope body can carry `SwapMailboxEvent`, and the typed proto contract visibly says "swap mailbox events live here" instead of relying on readers to infer that `OutSwapHtlcEvent` is packed directly into `Any`. This should also give us a cleaner place to add future swap mailbox events without creating more standalone messages that are only indirectly referenced.

bhandras:
A few cleanup findings while this protocol/schema is still unreleased and easy to reshape:

## 1. Avoid storing the receive-auth private key in the swap DB

`sdk/swaps/migrations/000002_receive_auth_key.up.sql` stores the receive-auth private key directly:

```sql
private_key BLOB NOT NULL
```

Even though this is not a wallet spend key, it is still protocol authority: it signs receive invoices and is later used to decode/validate the forwarded onion for the mailbox-delivered HTLC event. I think we should avoid making the swap DB the owner of this raw private key.

Can we derive the receive-auth key from wallet/keychain material plus the payment hash, similar to Loop's HTLC key reveal/shared-key flow? Roughly:

```text
base wallet/keychain key + payment_hash -> shared secret -> receive auth key
```

Then the DB only needs durable context like the payment hash, and possibly a key locator/public key if the locator is not fixed. That keeps receive resume working across restart without persisting raw auth private key material. It also gives us per-receive auth keys instead of one long-lived client-level key, reducing blast radius if the DB is copied.

## 2. Make the mailbox event wrapper explicit

`OutSwapHtlcEvent` is currently a standalone proto message that is only used indirectly as the mailbox envelope `Any` body. That makes the proto surface harder to read because the file does not show where the event is transported.

I think we should wrap it explicitly, for example:

```proto
message SwapMailboxEvent {
    oneof event {
        OutSwapHtlcEvent out_swap_htlc = 1;
    }
}
```

Then the mailbox body carries `SwapMailboxEvent`, and the proto contract visibly says that swap mailbox events live here. This also gives us a cleaner place to add future swap mailbox events.

## 3. Remove unused server-observed event fields

`OutSwapHtlcEvent` currently includes:

```proto
string vhtlc_outpoint = 7;
uint64 vhtlc_amount_sat = 8;
```

The client maps these into the SDK event shape, but the receive accept path does not use them. The client derives the expected vHTLC script from the validated event config and then waits for its own indexer to observe the live vHTLC. That is the right trust model: the client should not need to trust a server-supplied outpoint/amount.

Since these are not part of the current validation path, I would remove them from the proto, SDK event struct, conversion code, and the server publisher.

## 4. Either validate or remove `channel_id` and `incoming_expiry_height`

The event also carries:

```proto
uint32 incoming_expiry_height = 3;
uint64 channel_id = 4;
```

Both are mapped into the SDK event type, but I do not see them being validated or used when the client accepts the HTLC event.

If `channel_id` is meant to bind the mailbox event back to the exact route hint that was placed in the invoice, then the client should persist the expected channel ID and verify it. If `incoming_expiry_height` is meant to enforce an incoming-HTLC safety window, then the client should enforce that rule.

If neither field has a concrete client-side validation rule, I would remove them from the event body while the protocol is unreleased. The server can still use channel ID internally for idempotency/message identity without exposing it as client-facing protocol data.

## 5. Remove compatibility/deprecation leftovers instead of reserving them

Because this flow is unreleased, I would avoid carrying deprecated/reserved proto or DB artifacts from intermediate PR shapes. If a field existed only in a prior iteration and never shipped, remove it rather than reserving or deprecating it.

Same for schema comments like the `intervention_reason` note saying the column name is kept narrow for compatibility with the first swap-session schema. If there is no compatibility burden yet, we can either rename now or simply remove the compatibility sentence.

The end state I would aim for is: the mailbox event only carries fields the client actually validates/uses, and the swap DB stores resume context but not raw receive-auth private key material.

bhandras:
This claim path is doing some non-trivial recovery-policy work now, and I think it would benefit from a few short comments around the key branches. The branch conditions are mechanically clear, but the reason each branch exists is easy to lose.

In particular, I think comments would help around:

1. The `claimSessionID != ""` branch: this is the durable boundary that the daemon accepted the claim spend, so after this point we should not submit another claim.

2. The restored `ClaimInitiated && !claimIntentRecordedInProcess` branch: a restarted session may have submitted a claim before the process stopped but failed to persist the returned session ID, so it does a fuller reconciliation to avoid duplicating that custom-input spend.

3. The `claimIntentRecordedInProcess` branch: for a fresh claim intent created in this process, the indexed-spend check is only a duplicate-send guard, so it is intentionally bounded to avoid a slow indexer consuming the caller's claim context before we submit the actual claim.

4. The refund-locktime guard before claim submission: once refund is mature, a new client claim becomes a late race with the server refund path rather than a clean recoverable receive.

5. The spent-without-preimage case: if the vHTLC was spent but the indexed package does not reveal our invoice preimage, that is not "claim not found yet"; it means another path consumed the vHTLC, so retrying the claim is wrong.

Something along these lines would make the intent much easier to preserve:

```go
// A persisted claim session ID is our durable boundary that the daemon
// accepted the claim spend. After this point we never submit another claim;
// completion can be recovered by session ID/indexer reconciliation.
if s.claimSessionID != "" {
        ...
}

// A restored ClaimInitiated session may have submitted the claim before the
// process stopped but failed to persist the returned session ID. Reconcile here
// so we prefer the already-indexed claim over issuing a duplicate custom-input
// spend.
if s.state == ReceiveStateClaimInitiated && !s.claimIntentRecordedInProcess {
        ...
}

// For a claim intent created in this process, the indexed-spend check is only a
// duplicate-send guard. Bound it so a slow indexer cannot consume the caller's
// claim context before we submit the actual claim.
if s.claimIntentRecordedInProcess {
        ...
}
```

This is not just style: these branches encode the crash/restart semantics of the receive claim path. A future cleanup could easily flatten them back into something that looks simpler but reintroduces duplicate-claim or late-refund races.

bhandras:
One more retry-boundary issue I noticed on the updated receive event flow:

## HTLC event ack failure is marked retryable, but there is no retry path after the event is accepted

`waitForHTLCEvent` does the important part in the right order: it validates and persists the mailbox event before acking it.

```go
if err := s.acceptOutSwapHtlcEvent(
        ctx, notification.Event, authKey,
); err != nil {
        return err
}

if notification.Ack != nil {
        if err := notification.Ack(ctx); err != nil {
                return newRetryableActionError(fmt.Errorf(
                        "ack out-swap HTLC event: %w", err,
                ))
        }
}
```

That fixes the dangerous crash-after-ack case. But it creates a smaller follow-up issue: if the ack fails after `acceptOutSwapHtlcEvent` succeeds, the receive state has already advanced to `HTLCEventAccepted`. On the next `Wait`/resume, the FSM starts from `HTLCEventAccepted`, so it skips `waitForHTLCEvent` entirely and goes straight to funding observation. That means the returned `retryableActionError` is not actually retryable with respect to the mailbox ack.

In simple terms, the sequence is:

```text
pull event
validate + persist event
transition to HTLCEventAccepted
ack fails
return retryable error
next resume skips event pull/ack
mailbox row remains unacked
```

This is not as dangerous as the original bug because the client can still resume and claim from the persisted event. But it can leave accepted events in the mailbox forever if the ack fails at this boundary.

I think we should either make this ack truly retryable or make it explicitly best-effort. The durable version could be one of:

- persist the mailbox `ackCursor` with the accepted event and retry ack from the `HTLCEventAccepted` state;
- add an explicit accepted-but-unacked state/boundary;
- have the accepted-event state re-pull the per-receive mailbox and ack the matching event before moving on to funding observation.

At minimum, I would add a test where `Ack` fails after `acceptOutSwapHtlcEvent` succeeds, then resume and assert the intended behavior. Right now the code calls it retryable, but the state machine no longer has a path that retries that ack.

sputn1ck:
Addressed in 510a9ed, with the server submodule pointer updated in 9af1f7c.

Agreed that the previous retryable error was not actually retryable after the state advanced to HTLCEventAccepted. The fix now persists the mailbox ack cursor with the accepted HTLC event. HTLCEventAccepted retries AckUpTo when that cursor is still pending, clears it only after the ack is durably complete, and then continues to funding observation.

I added regression coverage for: accept event -> persist HTLCEventAccepted -> ack fails -> resume -> retry ack -> clear cursor -> continue funding.

Validated with:
- make unit pkg=./sdk/swaps timeout=5m
- make unit pkg=./cmd/darepocli/darepoclicommands timeout=5m
- make lint-local
- make commitmsg-lint commit=HEAD

bhandras:
codex: I did a current-head pass on this PR after the latest fixes. The earlier CLI wiring and durable mailbox-ack retry issues look addressed now. A few findings still look worth cleaning up before this settles, especially because this protocol/schema is unreleased and still easy to reshape.

## 1. `LatestMigrationVersion` is still `2`, but only migration `000001` exists

`sdk/swaps/store.go` still has:

```go
LatestMigrationVersion uint = 2
```

But the receive-auth-key migration was removed in the latest wallet-derived-key commit, and the migrations directory now only contains:

```text
sdk/swaps/migrations/000001_swap_sessions.up.sql
sdk/swaps/migrations/000001_swap_sessions.down.sql
```

So this looks like the earlier migration-version mismatch came back after deleting `000002_receive_auth_key`. Since we are intentionally folding unreleased schema changes into `000001`, I think `LatestMigrationVersion` should go back to `1`.

I would also add a small store test that asserts the applied swap DB migration version matches `LatestMigrationVersion`, so this does not silently drift again when temporary migrations are added/removed during review.

## 2. The mailbox event is still not represented as an explicit proto wrapper

`OutSwapHtlcEvent` is still a standalone message in `swaprpc/swap.proto`. It is transported inside mailbox envelopes as an `Any`, but the proto file still does not show that relationship directly.

I think this is the source of the earlier “added but not included in any other protos?” confusion. The message is used, but only indirectly through mailbox transport, so the schema is hard to discover by reading `swap.proto` alone.

I would still prefer an explicit wrapper, for example:

```proto
message SwapMailboxEvent {
    oneof event {
        OutSwapHtlcEvent out_swap_htlc = 1;
    }
}
```

Then the mailbox body carries `SwapMailboxEvent`, and the proto contract visibly says “swap mailbox events live here”. That also gives us a clean extension point for future swap mailbox events instead of adding more standalone messages that are only referenced by convention.

## 3. `OutSwapHtlcEvent` still carries fields the client does not validate or use

The event still includes:

```proto
uint32 incoming_expiry_height = 3;
uint64 channel_id = 4;
string vhtlc_outpoint = 7;
uint64 vhtlc_amount_sat = 8;
```

These are converted into the SDK event shape, but the accept path still validates only the payment hash, amount, onion payload, and vHTLC config/script material. The client then observes the actual funded vHTLC itself through its indexer path.

That means `vhtlc_outpoint` and `vhtlc_amount_sat` are server-observed facts that the client does not rely on. I would remove them from the client-facing event.

For `channel_id` and `incoming_expiry_height`, I think we should either add concrete client-side validation rules or remove them too. For example:

- if `channel_id` is meant to bind the event back to the exact route hint in the invoice, persist the expected channel ID and verify it;
- if `incoming_expiry_height` is meant to enforce an incoming-HTLC safety window, enforce that before accepting the event.

If neither rule is needed in this PR, I would keep the protocol smaller and omit those fields while this is unreleased.

## 4. Wallet-derived receive auth key still needs stronger domain separation

Moving the receive-auth key out of the swap DB is the right direction. The current derivation is much better than storing `private_key BLOB` locally.

But the new daemon derivation currently uses lndclient’s generic shared-key material directly:

```go
lndSvc.Signer.DeriveSharedKey(
        ctx, lndclient.SharedKeyNUMS,
        lndclient.SharedKeyLocator,
)
```

and the local wallet paths do the equivalent keyring ECDH with the same `SharedKeyNUMS` / `SharedKeyLocator`. The SDK then interprets those returned 32 bytes as a secp private key for invoice signing and onion decoding.

Mechanically, `keyring.ECDH` is real ECDH, so that part is fine. The concern is that this is not clearly domain-separated for “darepo receive-auth key”, and it is still one stable client-level key rather than a per-receive key.

I think we should make the daemon RPC take the `payment_hash` and derive something domain-specific, roughly:

```text
base = wallet ECDH(receive-auth locator, receive-auth NUMS)
receive_auth_key = tagged_hash(
    "darepo/swap/receive-auth/v1",
    base || payment_hash,
)
```

Even if we keep using lndclient’s shared-key primitive as the base, I would still add a Darepo/swap-specific tagged hash before returning bytes to the SDK. That makes the exported key material narrow-purpose and avoids accidentally reusing the exact same secret as other shared-key consumers.

## 5. `DeriveReceiveAuthKey` returns private-key-equivalent bytes over the daemon RPC

The proto comment says:

```proto
// key is the 32-byte wallet-derived receive-auth private key used by
// receive swaps for invoice signing and onion decoding.
bytes key = 1;
```

That is accurate: once the SDK calls `btcec.PrivKeyFromBytes`, these bytes are private-key-equivalent authority for receive invoice signing and onion decoding.

This may be acceptable for the local daemon/SDK boundary in this PR, but I think we should narrow it as much as possible: make it per-payment-hash and domain-separated, and document that callers must not persist it. The cleaner long-term shape would be daemon-backed signing/ECDH so raw key bytes never leave the wallet boundary, but a per-swap derived key is already a much better boundary for this PR.

## 6. The claim recovery branches still need comments

The receive claim path now has several non-trivial recovery branches:

```go
if s.claimSessionID != "" { ... }

if s.state == ReceiveStateClaimInitiated &&
        !s.claimIntentRecordedInProcess { ... }

if s.claimIntentRecordedInProcess { ... }
```

The code is doing important crash/restart policy work here, but the reason for each branch is still not obvious from the code alone.

I would still add short comments explaining:

- a persisted `claimSessionID` means the daemon accepted the claim spend, so we should not submit another claim;
- a restored `ClaimInitiated` session without an in-process claim intent may have submitted before restart but failed to persist the returned session ID, so it needs fuller reconciliation;
- a fresh in-process claim intent only needs a bounded indexed-spend check so a slow indexer cannot consume the caller’s claim context before the actual claim is submitted;
- once the refund locktime is mature, a new receive claim is a late race with the server refund path;
- spent-without-preimage is terminal for retry purposes, not “claim not found yet”.

This is not just style. These comments preserve the crash/restart semantics that keep the receive claim path from submitting duplicate custom-input spends or racing a matured refund.

## 7. Stale unreleased-schema compatibility wording remains

`000001_swap_sessions.up.sql` still says:

```sql
-- intervention_reason is the durable terminal explanation stored when a
-- receive session fails or stops in NeedsIntervention. The column name is
-- kept narrow for compatibility with the first swap-session schema.
```

Since this schema is unreleased and we are intentionally editing `000001`, I would remove the compatibility sentence. There is no shipped schema compatibility burden here, so the comment is now misleading.

bhandras:
codex: Thanks for the latest push, this is much closer. The final tree is moving in the right direction, especially with the removal of the receive-auth key table and the payment-scoped receive-auth derivation. I think the remaining issue is mostly about making the stack reviewable and making the security boundary explicit.

The main thing I’d ask for is a history rework so that the PR never introduces durable receive-auth private-key storage in any commit. Right now commit `29e52b35` still adds the old shape:

- `receive_auth_keys`
- `private_key BLOB`
- `btcec.NewPrivateKey()`
- serialized receive-auth private key storage through sqlc queries

A later commit removes that, so the final tree is better, but the stack still contains an intermediate commit that stores private-key material durably. Since this is unreleased and we do not have DB/protocol compatibility constraints, I think we should fold/rewrite the stack so that table and those queries never appear at all.

One important security distinction: removing DB persistence is necessary but not sufficient if the intended boundary is “the receive-auth private key never leaves the daemon.” In the current final tree, the daemon still returns raw 32-byte receive-auth key material, and the SDK wraps that with `btcec.PrivKeyFromBytes`, stores the resulting private key in memory, and uses it directly for signing and Sphinx ECDH. That is better than durable SQL storage because it is payment-scoped and memory-only, but it still exposes private-key-equivalent material to the SDK process. If we want the stronger security property, the daemon API should expose pubkey/sign/ECDH operations instead of returning the scalar bytes.

Suggested final stack shape:

1. `swaprpc: add swap mailbox event envelope`

   Add the mailbox event protocol shape only:

   - `SwapMailboxEvent`
   - trimmed `OutSwapHtlcEvent`
   - only the fields the client actually needs:
     - `payment_hash`
     - `amount_sat`
     - `onion_blob`
     - `vhtlc_config`

   This commit should not touch receive-auth key storage, swap DB state, or the receive state machine.

2. `daemonrpc: add daemon-backed receive auth operations`

   Add the daemon API for receive auth in its final security shape.

   The daemon should not return the receive-auth private scalar. Instead, expose daemon-backed operations such as:

   - get the payment-scoped receive-auth public key,
   - sign a receive-auth message,
   - derive the Sphinx ECDH shared secret.

   That gives the SDK the interfaces it needs for invoice signing and onion decoding without exporting private-key-equivalent bytes.

3. `darepod: implement payment-scoped receive auth`

   Implement the daemon side:

   - validate `payment_hash` length,
   - derive from wallet-backed material,
   - domain-separate with a receive-auth tag,
   - include the payment hash in the derivation,
   - keep the private scalar inside the daemon,
   - expose only pubkey/signature/ECDH results.

   The key lifecycle should be: never persist it in SQL, never return it over RPC, and never cache it as a `btcec.PrivateKey` in the SDK process.

   This should include daemon-side coverage for:

   - invalid payment hash length,
   - same payment hash gives the same receive-auth pubkey,
   - different payment hashes give different receive-auth pubkeys,
   - signing verifies against the returned pubkey,
   - ECDH matches the expected Sphinx shared-secret construction.

4. `sdk/ark: expose daemon receive auth helpers`

   Add typed SDK wrappers around the daemon receive-auth RPCs:

   - `ReceiveAuthKey`
   - `SignReceiveAuthMessage`
   - `SignReceiveAuthMessageCompact`
   - `ReceiveAuthECDH`

   The wrapper should return public keys, signatures, and shared secrets, not private-key bytes.

5. `sdk/swaps: use daemon-backed receive auth`

   Add the swap-layer `ReceiveAuthKey` implementation in its final form.

   The important invariant here is that the SDK should not persist, generate, receive, or cache the receive-auth private key. The swap-layer key can adapt the daemon RPCs to the existing signer/ECDH interfaces used by invoice creation and onion decoding.

   This is also where I’d remove the public `SetReceiveAuthKey` override. The final design is payment-scoped, but `SetReceiveAuthKey` still installs one client-level key and bypasses payment-scoped derivation. If tests need injection, it should be internal/test-only or payment-hash-scoped.

6. `sdk/swaps: receive HTLC events from mailbox`

   Add the mailbox receiver and event decoding:

   - payment-scoped mailbox ID,
   - service/method/type filtering,
   - `SwapMailboxEvent` decoding,
   - `OutSwapHtlcEvent` extraction,
   - payment-hash validation,
   - mailbox ack hook.

7. `sdk/swaps: validate forwarded receive HTLC events`

   Extend the receive flow to consume and validate the mailbox HTLC event before revealing the preimage:

   - wait for the event after invoice creation,
   - decode the final-hop onion with the receive-auth key,
   - validate amount/payment address,
   - validate vHTLC config,
   - build/compare the expected vHTLC script/policy,
   - reject or defer manual claim attempts before the HTLC event has been accepted.

8. `sdk/swaps: persist accepted HTLC ack cursor`

   Add durable ack/retry state:

   - `pending_htlc_ack_cursor` in `receive_swaps`,
   - persist the cursor after accepting the HTLC event,
   - ack only after durable state is written,
   - retry ack on resume,
   - clear the cursor after successful ack.

   Since this is unreleased, this should be folded into the initial swap-session migration rather than adding compatibility migrations for a schema that never shipped.

9. `sdk/swaps: clarify receive claim reconciliation`

   Keep the useful comments/tests around the non-obvious restart cases:

   - persisted `claim_session_id`,
   - restored `ClaimInitiated` without a recorded session id,
   - bounded claim reconciliation,
   - already-spent vHTLC adoption.

Final invariants I’d like the stack to satisfy:

- no `receive_auth_keys` table in any commit,
- no `private_key BLOB` receive-auth storage in any commit,
- no `loadOrCreateReceiveAuthKey`,
- no `InsertReceiveAuthKey` / `GetReceiveAuthKey`,
- no production `btcec.NewPrivateKey()` fallback for receive auth,
- no daemon RPC that returns receive-auth private-key-equivalent scalar bytes,
- no SDK `btcec.PrivKeyFromBytes` wrapping of receive-auth key material in production code,
- no SDK in-memory map/cache of receive-auth private keys,
- no public client-level `SetReceiveAuthKey` override,
- comments say “payment-scoped” and do not describe receive auth as a stable client-level key,
- final tests cover mailbox HTLC validation, ack retry on resume, claim reconciliation, and receive-auth key scoping/security behavior.

The final tree is already much closer than the earlier version. I think the main cleanup now is to make the stack tell the same story as the final design: mailbox events plus payment-scoped receive auth, with private-key material kept out of the swap DB and behind the daemon boundary entirely.

bhandras (approved):
LGTM, very nice!! 🎉

PR #339, opened 2026-05-05
swaps: support same-Ark p2p vHTLCs

sputn1ck:
## Summary
- add swaprpc settlement typing and a dedicated in-Ark vHTLC event payload
- teach the SDK to receive either Lightning out-swap events or direct same-Ark vHTLC events
- validate direct p2p vHTLC policy construction without requiring an onion payload

## Testing
- make commitmsg-lint range=HEAD~3..HEAD
- git diff --check
- go test ./sdk/swaps -run TestAcceptInArkHtlcEventBuildsSenderReceiverPolicy -count=1
- go test ./sdk/swaps -run "^$"

Note: a full `go test ./sdk/swaps` was attempted after rebasing and hit `TestPaySessionResumeFundingGraceEventuallyRetries`; the focused p2p test and package compile check pass.

### DIFF
--- a/swaprpc/swap.proto
+++ b/swaprpc/swap.proto
@@ -6,6 +6,20 @@ option go_package = "github.com/lightninglabs/darepo-client/swaprpc";
 
 import "google/protobuf/timestamp.proto";
 
+// SettlementType identifies which settlement path backs a swap negotiation.
+enum SettlementType {
+    // SETTLEMENT_TYPE_UNSPECIFIED is accepted as Lightning for backward
+    // compatibility with older responses.
+    SETTLEMENT_TYPE_UNSPECIFIED = 0;
+
+    // SETTLEMENT_TYPE_LIGHTNING means the swap server bridges to Lightning.
+    SETTLEMENT_TYPE_LIGHTNING = 1;
+
+    // SETTLEMENT_TYPE_IN_ARK means sender and receiver settle with one Ark
+    // vHTLC inside the same Ark instance.
+    SETTLEMENT_TYPE_IN_ARK = 2;
+}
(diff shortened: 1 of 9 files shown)


sputn1ck:
@claude review
@codex review

swaprpc/swap.proto
```diff
@@ -66,9 +80,39 @@ message SwapMailboxEvent {
         // out_swap_htlc is emitted when the server has funded or accepted
         // funding for an out-swap HTLC.
         OutSwapHtlcEvent out_swap_htlc = 1;
+
+        // in_ark_htlc is emitted when a same-Ark sender has funded or
+        // accepted funding for a direct vHTLC.
+        InArkHtlcEvent in_ark_htlc = 2;
```
bhandras, line 86: imo we should add a mailbox-level test for the in-ark event too.

IIUC #337 made `SwapMailboxEvent` the explicit mailbox envelope, and this PR correctly adds `in_ark_htlc` as another oneof arm there. But the new test calls `acceptInArkHtlcEvent` directly, so it doesn't prove that the actual mailbox payload is `SwapMailboxEvent{in_ark_htlc: ...}` rather than a raw `InArkHtlcEvent`, nor that `WaitIncomingVHTLC` sets the ack cursor/hook correctly for this new event type.

Could we add a sibling of `TestMailboxOutSwapEventReceiverPullsAndAcks` that wraps an `InArkHtlcEvent` in `SwapMailboxEvent`, pulls it through `WaitIncomingVHTLC`, and asserts the `InArk` payload plus ack cursor?

bhandras (approved):

PR #341, opened 2026-05-06
lint: enable contextcheck

### LINKED ISSUE #273: lint: enable contextcheck and fix context-propagation bugs
## Background

After PR #266 ported golangci-lint to v2, `contextcheck` was left on the `disable:` list pending cleanup. A full run turned up **21 hits** and — given past incidents involving dropped context cancellation — several look like real bugs worth fixing before enabling the linter.

### Cluster A — likely real bugs (7 hits)

These are places where a live `context.Context` is in scope but the code creates a fresh `context.Background()` or fails to pass the inherited context down, breaking cancellation propagation:

| Location | Pattern |
|---|---|
| `btcwbackend/chain_backend.go:287` | Non-inherited new context |
| `btcwbackend/chain_backend.go:370` | Non-inherited new context |
| `btcwbackend/chain_backend.go:438` | Non-inherited new context |
| `lwwallet/boarding_backend.go:77` | `GetAddressUtxos` should take ctx param |
| `oor/signing_outbox_handler.go:65` | `handleCheckpointSignatures` should take ctx param |
| `serverconn/runtime.go:74` | `processDelivery` chain missing ctx propagation |
| `timeout/actor.go:69` | `handleSchedule` callback missing ctx |

Each of these breaks the cancellation chain: if the caller's context is cancelled (shutdown, timeout), the work in the callee keeps running against a fresh Background context. This is exactly the class of bug we've hit before where actors don't stop on system shutdown.

### Cluster B — intentional startup contexts (14 hits, all in `darepod/server.go`)

These are called from `run()` / server startup, before any request context exists. The contextcheck linter flags them because the functions involved (`NewActorSystem`, `NewActorRuntime`, `NewDurableActor`, various `Start` cascades) take a context that originates from `context.Background()` at process start.

| Location | Function chain |
|---|---|
| `darepod/server.go:477` | NewActorSystemWithConfig → NewActor |
| `darepod/server.go:518` | run$8 closure |
| `darepod/server.go:926` | New (chain backend) |
| `darepod/server.go:939` | Start → pollLoop → … → GetScriptUtxos |
| `darepod/server.go:1112` | Start |
| `darepod/server.go:1164` | NewWithNeutrino |
| `darepod/server.go:1166` | New → Start |
| `darepod/server.go:1172` | Start → Start → Start\$1 |
| `darepod/server.go:2254` | NewSqliteStore → ExecuteMigrations → RunMigrations |
| `darepod/server.go:2364` | New |
| `darepod/server.go:2447` | NewRuntime → NewDurableActor |
| `darepod/server.go:2457` | StartEgress → Start → process → processDelivery |
| `darepod/server.go:2675` | RegisterWithSystem → … → NewActor |
| `darepod/server.go:2872` | NewOORClientActor → NewDurableActor |

These get `//nolint:contextcheck // startup, no parent ctx` with a one-line comment explaining that the server lifetime context is the root.

## Proposed follow-up plan

1. Fix the 7 Cluster A bugs by plumbing the inherited context through.
2. Annotate the 14 Cluster B hits with `//nolint:contextcheck` + reason.
3. Remove `contextcheck` from `linters.disable` in `.golangci.yml`.
4. Verify `make lint-native` is clean.

## Other linters worth a similar follow-up

From the same v2 port evaluation:

- **`testifylint`** (31 hits). Of particular concern: **6 `go-require` hits** where `require.*` is used inside goroutines or HTTP handlers — `require` calls `t.FailNow()` which only works in the test goroutine, so these tests hang instead of failing cleanly:
  - `chainsource/chainsource_test.go:672`
  - `timeout/actor_test.go:338, 347`
  - `lwwallet/chain_backend_test.go:220, 305, 309`
  
  Plus 1 `useless-assert` in `indexer/client_test.go:142` that asserts a variable against itself (effectively a no-op test). The remaining 24 are stylistic migrations (`require.Equal` → `require.Len`/`require.Empty`/`require.Positive`/`require.ErrorIs`, etc.) that would improve failure messages.

- **`noctx`** (17 hits) — HTTP calls without context propagation. Same bug family as contextcheck.

- **`exhaustive`** extensions — currently relies on `default-signifies-exhaustive: true`; the few remaining explicit-case hits were addressed in #266.

- **`modernize`** (32 hits) — Go 1.22+ idioms (`maps.Copy`, `slices.Contains`, `min`/`max`). Pure cleanup, no bug-catching.

## Not recommended

These would dominate the PR without a clear safety payoff:

- `revive` (50 hits) — general style
- `unparam` (43 hits) — unused params  
- `protogetter` (50 hits) — proto getter migration

Context: measurements taken on commit 96f4e23d with the config changes from PR #266.

bhandras:
## Summary

Fixes #273.

This enables `contextcheck` in the main golangci-lint config and cleans up the current context-propagation findings on `main`.

The issue was originally filed with a smaller hit list, but the current tree had more contextcheck findings after later actor, wallet, and OOR changes. I split the cleanup into two buckets: places where cancellation really should propagate into lower-level work, and places where the code intentionally starts work owned by a registration, actor, timer, daemon component, or bounded async cleanup path.

### Esplora request contexts

The `lwwallet` Esplora helpers now accept `context.Context` and build HTTP requests with `http.NewRequestWithContext`.

That covers:

- tip height and tip hash reads
- block hash, block header, and raw block reads
- raw transaction reads
- script/address UTXO reads
- tx status and outspend reads
- fee estimates
- tx broadcast
- package relay submission
- `testmempoolaccept`

Request-scoped callers now pass their own context, so wallet, boarding, and chain-service work can stop promptly when the caller is cancelled or the daemon is shutting down. Long-lived internal paths, such as tip polling and registration checks, pass an explicit lifecycle root where the work intentionally outlives an individual request.

The Esplora HTTP gosec suppressions also now sit on the `http.Client.Do` calls where this repo's gosec runner actually reports G704, with the accurate "trusted URL" rationale.

### Esplora chain-service lifecycle context

`EsploraChainService` now captures the context passed to `Start` and uses that service lifecycle context for btcwallet chain-interface methods that do not accept their own `context.Context` parameter.

That includes block fetches, block hash/header lookups, block filtering, transaction broadcast, rescans, notification logging, and `TestMempoolAccept`.

This preserves the important property that these btcwallet callbacks are not tied to a short-lived caller context, while still allowing daemon shutdown to cancel in-flight Esplora HTTP calls instead of waiting for the HTTP client timeout.

### OOR context propagation

OOR shutdown and checkpoint-signing logs had a caller context available but fell back to `context.Background()`.

Those paths now reuse the inherited context so shutdown diagnostics, deadlines, and scoped logger values stay attached to the work that is actually waiting for shutdown or handling checkpoint-signing results.

### Actor-owned registration contexts

Round confirmation registrations now use the round actor lifecycle context captured in `Start`. This lets confirmation watches outlive the current actor receive turn, but still ties their submission to actor/daemon shutdown instead of an unbounded `context.Background()` root.

The fallback for test-only actor shells uses `context.WithoutCancel(ctx)` so cancellation is detached while context values for logs/tracing are preserved.

### Intentional context roots

Several `contextcheck` findings were intentional lifecycle boundaries. These now have targeted `//nolint:contextcheck` comments at the handoff site instead of keeping the linter disabled globally.

Examples include:

- chain notification registrations whose returned `Cancel` owns the forwarding goroutine lifetime
- actor startup paths where the actor owns its lifecycle after registration
- timer callbacks that fire after the scheduling actor turn has completed
- async cleanup and retry loops with their own owner or timeout
- daemon startup and shutdown roots
- test actors whose lifetime is owned by `t.Cleanup`

## Out Of Scope

This PR intentionally does not enable `noctx`, `testifylint`, or `intrange`.

The Esplora request-context work fixes the important HTTP context propagation covered by this PR, but the remaining `noctx`, `testifylint`, and `intrange` cleanups are separate policy changes and should stay in their own follow-ups.

## Validation

```sh

tools/custom-gcl run --enable-only=contextcheck --timeout=15m \
  --build-tags='test_postgres,test_sqlite,systest' \
  --max-issues-per-linter=0 --max-same-issues=0

go test ./lwwallet ./round ./timeout

GOGC=50 make lint-native
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -109,7 +109,6 @@ linters:
     - gochecknoinits
 
     # Disabled pending a larger cleanup PR.
-    - contextcheck
     - noctx
     - tparallel
     - unparam
--- a/btcwbackend/chain_backend.go
+++ b/btcwbackend/chain_backend.go
@@ -360,6 +360,8 @@ func (b *ChainBackend) handlePackageResult(ctx context.Context,
 // RegisterConf registers for confirmation notifications using
 // neutrino's chain notifier. The registration returns a
 // ConfRegistration with channels for receiving confirmation events.
+//
+//nolint:contextcheck // returned registration Cancel owns forwarder lifetime
 func (b *ChainBackend) RegisterConf(ctx context.Context,
 	txid *chainhash.Hash, pkScript []byte, numConfs uint32,
 	heightHint uint32,
@@ -443,6 +445,8 @@ func (b *ChainBackend) RegisterConf(ctx context.Context,
 
 // RegisterSpend registers for spend notifications using neutrino's
 // chain notifier.
+//
+//nolint:contextcheck // returned registration Cancel owns forwarder lifetime
 func (b *ChainBackend) RegisterSpend(ctx context.Context,
 	outpoint *wire.OutPoint, pkScript []byte,
 	heightHint uint32) (*chainsource.SpendRegistration, error) {
@@ -511,6 +515,8 @@ func (b *ChainBackend) RegisterSpend(ctx context.Context,
 
 // RegisterBlocks registers for new block notifications using
 // neutrino's chain notifier.
+//
+//nolint:contextcheck // returned registration Cancel owns forwarder lifetime
 func (b *ChainBackend) RegisterBlocks(
 	ctx context.Context) (*chainsource.BlockRegistration, error) {
 
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -266,6 +266,8 @@ func (b *LNDBackend) SubmitPackage(ctx context.Context,
 // RegisterConf registers for confirmation notifications using lnd's chain
 // notifier. The registration returns a ConfRegistration with channels for
 // receiving confirmation events.
+//
+//nolint:contextcheck // returned registration Cancel owns forwarder lifetime
 func (b *LNDBackend) RegisterConf(ctx context.Context,
 	txid *chainhash.Hash, pkScript []byte, numConfs uint32,
 	heightHint uint32,
@@ -338,6 +340,8 @@ func (b *LNDBackend) RegisterConf(ctx context.Context,
 // RegisterSpend registers for spend notifications using lnd's chain notifier.
 // The registration returns a SpendRegistration with channels for receiving
 // spend events.
+//
+//nolint:contextcheck // returned registration Cancel owns forwarder lifetime
 func (b *LNDBackend) RegisterSpend(ctx context.Context,
 	outpoint *wire.OutPoint, pkScript []byte,
 	heightHint uint32) (*chainsource.SpendRegistration, error) {
--- a/chainsource/block_epoch_actor.go
+++ b/chainsource/block_epoch_actor.go
@@ -158,6 +158,7 @@ func (a *BlockEpochActor) handleSubscribeBlocks(actorCtx context.Context,
 	// Register with the backend to receive block notifications. We do this
 	// before starting the goroutine so we can return an error to the
 	// caller if registration fails.
+	//nolint:contextcheck // actor root context owns registration lifetime
 	registration, err := a.cfg.Backend.RegisterBlocks(a.ctx)
 	if err != nil {
 		return fn.Err[EpochResp](fmt.Errorf(
--- a/chainsource/conf_actor.go
+++ b/chainsource/conf_actor.go
@@ -161,6 +161,7 @@ func (a *ConfActor) handleRegisterConf(actorCtx context.Context,
 	regCtx, regCancel := context.WithTimeout(a.ctx, 10*time.Second)
 	defer regCancel()
 
+	//nolint:contextcheck // actor root context owns registration lifetime
 	registration, err := a.cfg.Backend.RegisterConf(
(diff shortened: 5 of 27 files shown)


bhandras:
@claude pls review

bhandras:
@claude pls re-review.

Addressed the actionable review notes from Claude and Gemini:

- kept the Esplora gosec suppressions on the `http.Client.Do` lines where this repo's gosec runner reports G704, and updated the rationale to use the accurate G704 label
- made the recurring timeout actor suppressions use the inline reason style for consistency
- changed round confirmation registration to use the round actor lifecycle context captured in `Start`, with a `context.WithoutCancel(ctx)` fallback for test-only actor shells
- changed `EsploraChainService` to capture its `Start` context and use it for btcwallet chain-interface methods that do not receive a context, so shutdown can cancel in-flight Esplora HTTP calls
- avoided re-entering the service mutex in `NotifyReceived` while adding that context helper

Re-validated locally:

```sh
tools/custom-gcl run --enable-only=contextcheck --timeout=15m --build-tags='test_postgres,test_sqlite,systest' --max-issues-per-linter=0 --max-same-issues=0
go test ./lwwallet ./round ./timeout
GOGC=50 make lint-native
```

PR #342, opened 2026-05-06
docs: add context lifecycle review skill

### LINKED ISSUE #233: skill: detect goroutine context lifecycle bugs at review time
## Problem

We keep shipping bugs where a goroutine captures a parent context (`ctx`) that
gets cancelled when the caller returns, killing the goroutine prematurely. The
latest instance was `7c4778f` — the btcwallet sync goroutine used the
`UnlockWallet` RPC context, which dies as soon as the RPC handler returns,
leaving the wallet permanently unready.

This is a subtle and recurring class of bug. The pattern looks innocuous:

```go
func (s *Server) startSomething(ctx context.Context) error {
    // ... setup ...
    go func() {
        // BUG: ctx is the RPC context, cancelled when RPC returns
        select {
        case <-ctx.Done():
            return
        case <-ticker.C:
            // ... long-running work that must outlive the caller ...
        }
    }()
    return nil
}
```

The fix is always the same — use `context.Background()` or a server-scoped
context — but the bug is hard to spot in review because `ctx` is the natural
variable to reach for.

## Proposed Skill

A review-time or pre-commit skill that detects when a goroutine launched inside
a function captures the function's `context.Context` parameter and uses it for
cancellation (`ctx.Done()`, `context.WithCancel(ctx)`, etc.) when the goroutine
is clearly intended to outlive the caller (e.g., `go func()` inside a function
that returns before the goroutine completes).

### Detection heuristics

1. **Goroutine outlives caller**: Function launches `go func()` and returns
   without joining (no `wg.Wait()` or channel sync before return).
2. **Captured context**: The goroutine references `ctx` (or a derived context)
   from the parent scope.
3. **Cancellation sensitivity**: The goroutine selects on `ctx.Done()` or passes
   `ctx` to blocking calls.

### False positive mitigation

- If the function blocks until the goroutine completes (waitgroup, channel),
  the parent context is fine.
- If the context is explicitly documented as long-lived (e.g., server context),
  skip the warning.
- If the goroutine is short-lived (just forwarding a single result), skip.

### Prior art in this repo

Search for commits fixing this pattern — there have been several.

bhandras:
## Summary

Fixes #233.

This adds a review-time Claude skill for a recurring class of context lifecycle bugs: async work captures a short-lived caller context even though the work is owned by a daemon, wallet, actor, timer, registration, or cleanup path after the caller returns.

The original issue called out the btcwallet sync class of bug, where a goroutine launched from an RPC handler can accidentally use the RPC context and then stop as soon as the RPC returns. The new skill gives reviewers a repeatable checklist for spotting that shape before it lands.

### New context lifecycle review skill

Added `.claude/skills/context-lifecycle/SKILL.md`.

The skill covers:

- goroutines launched from request/RPC handlers
- timer callbacks and retry loops
- notification registrations and subscriptions
- actor handoffs through `Tell` and `Ask`
- daemon and wallet background workers
- bounded async cleanup paths
- test helpers that combine `t.Context()` with `t.Cleanup`

The central question is: who owns this work, and when should it stop?

The skill then asks reviewers to classify async handoffs as caller-owned, component-owned, registration-owned, timer-owned, or bounded cleanup work before deciding whether the caller context is correct.

### Search and review heuristics

The skill includes focused `rg` searches for the patterns that most often hide these bugs, including:

- `go func`
- `AfterFunc`
- context cancellation and deadline derivation calls
- actor `Tell` and `Ask`
- registrations and subscriptions

The search patterns use shell-ready ripgrep escaping and avoid the overly broad `Start(` query, which would bury the useful results in noise.

The skill also lists red flags and acceptable review outcomes so comments are consistent and actionable instead of vague “maybe use Background here” advice.

### Fix guidance

The skill documents preferred fixes:

- use a named owner context such as `daemonCtx`, `actorCtx`, `runCtx`, or `notifyCtx`
- pass component root contexts into constructors that start long-lived work
- derive registration contexts from the returned cancel path
- use Go 1.21's `context.WithoutCancel(ctx)` when work must survive caller cancellation but should preserve values such as trace IDs or scoped loggers
- use a finite timeout for detached cleanup
- derive test helper lifecycles from `context.WithCancel(context.Background())` and cancel from `t.Cleanup` when cleanup owns the work
- keep request contexts on work that really should stop with the request
- require ownership comments for intentional `contextcheck` suppressions

### Examples

The bad and better examples are now self-contained enough to show the ticker lifecycle, and the better example guards the background worker with `sync.Once` so the context fix does not imply that repeated RPC calls should spawn duplicate long-running workers.

### Agent docs link

The top-level `CLAUDE.md` and `AGENTS.md` now link to the new skill from a shared “Review Skills” section. Both files were updated with identical text so the repository’s agent-doc sync invariant stays intact.

## Validation

```sh

.github/scripts/check-agent-docs-sync.sh

make doc-check

GOGC=50 make lint-native
```

Note: `make doc-check` exits successfully on this branch. It currently prints an existing `db/CLAUDE.md` versus `db/AGENTS.md` divergence from `origin/main`, which this focused PR intentionally does not modify.

### DIFF
--- a/.claude/skills/context-lifecycle/SKILL.md
+++ b/.claude/skills/context-lifecycle/SKILL.md
@@ -1,0 +1,230 @@
+---
+name: context-lifecycle
+description: >
+  Review goroutines, timers, callbacks, actor handoffs, and async cleanup for
+  context lifetime bugs. Use when code starts work that may outlive the caller,
+  especially from RPC handlers, wallet startup, daemon startup, actors, or
+  notification registrations.
+argument-hint: "[package-path or file]"
+allowed-tools: Read, Grep, Glob, Bash(rg *)
+---
+
+# Context Lifecycle Review
+
+Use this skill to catch async work that accidentally captures a short-lived
+caller context.
+
+The failure mode is subtle: `ctx` is usually the nearest variable, so a
+goroutine, timer, callback, or actor handoff may inherit an RPC/test/request
+context even though the work must continue after that caller returns. When the
+caller exits, the context is cancelled and the background work can silently stop
+too early.
+
+## Core Question
+
+For every asynchronous handoff, ask:
+
+> Who owns this work, and when should it stop?
+
+The answer determines which context is correct.
+
+## Search
+
+Start with targeted searches instead of reading whole packages:
+
+```bash
+rg -n 'go func|AfterFunc|time\.AfterFunc|context\.AfterFunc' \
+  --glob '*.go' [path]
+
+rg -n 'Done\(|WithCancel\(|WithTimeout\(|WithDeadline\(' \
+  --glob '*.go' [path]
+
+rg -n '\.Tell\(|\.Ask\(|Register|Subscribe' \
+  --glob '*.go' [path]
+```
+
+Then inspect each hit in context.
+
+## Classify The Owner
+
+Classify each async handoff before suggesting a fix.
+
+### Caller-Owned Work
+
+The caller waits for the goroutine before returning, or the goroutine only
+performs request-scoped work.
+
+Using the caller context is usually correct.
+
+Examples:
+
+- a goroutine joined by `WaitGroup.Wait` before return
+- a worker that sends one result and the function waits on that result
+- a helper that should abort when the RPC deadline expires
+
+### Component-Owned Work
+
+The work is started by a caller but owned by a daemon, wallet, actor, or
+service after startup.
+
+Do not capture the RPC/request context. Use a component root context, a
+registration-owned context, or an explicit context with its own cancel path.
+
+Examples:
+
+- wallet sync loops started by `UnlockWallet`
+- daemon background workers started from an RPC
+- actor goroutines started during actor construction
+- server connection runtimes started by daemon setup
+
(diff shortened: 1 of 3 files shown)


bhandras:
@claude pls review

bhandras:
@claude pls re-review.

Addressed the actionable review notes from Claude and Gemini:

- fixed the ripgrep escaping so the commands are shell-ready and match the intended regexes
- removed the broad `Start(` search term so the search block stays high signal
- made the context search less dependent on the variable name `ctx`
- added Go 1.21 `context.WithoutCancel(ctx)` guidance for detached work that should preserve context values
- added test-helper lifecycle guidance for `context.WithCancel(context.Background())` plus `t.Cleanup`
- made both examples initialize and stop their ticker
- guarded the better example with `sync.Once` so repeated RPC calls do not imply duplicate long-running workers
- removed `Bash(go test *)` from the read-only review skill's allowed tools

Re-validated locally:

```sh
.github/scripts/check-agent-docs-sync.sh
make doc-check
GOGC=50 make lint-native
```

PR #348, opened 2026-05-07
fraud: add recipient-side OOR fraud response watcher

### LINKED ISSUE #345: Recipient-side fraud-response watcher (client/fraud)
## Background

The server-side fraud-response work on the operator branch (`forfeit-tx-response`) handles the operator's half of the fraud-response protocol: when a Spent or Forfeited VTXO is unilaterally exited on chain, the server broadcasts the persisted checkpoint or forfeit transaction and ratchets through follow-up arktxs as they confirm.

The protocol assumes the operator is the primary responder. ARK-00 §296–§315 acknowledges that **if the operator is offline or unresponsive, the recipient holding a preconfirmed VTXO must defend their own funds** — but the spec says only that recipients SHOULD "potentially manage checkpoint transaction broadcasts if the operator is offline." There is no defined flow chart, no timing budget, and no client implementation today.

Concretely, the client today does **none** of the proactive monitoring required for this defense:

- The only chainsource spend watch the client registers on ancestor outpoints lives in `client/unroll/actor.go` and only fires on the *target* of an actively-running unroll job. It does not watch parents of received preconfirmed VTXOs.
- There is no actor analogous to the server's `fraud/` package.
- Multi-tree ancestry (`desc.Ancestry` plural) compounds the gap: a multi-input OOR VTXO has N independent ancestor paths, all of which need to be defended in parallel.

The operator-side itests fake the missing client behavior via a harness `ForceBroadcastLineageTx` helper plus the terminal-tolerant `LocalProofAssembler.EnsureProofForHarness` path. That is fine for exercising the server PR; the present issue is to build the real client behavior.

## Problem

In the multihop preconfirmed case `A → B → C`:

```
vtxo_A ⟶ checkpoint_AB ⟶ arktx_AB ⟶ vtxo_B ⟶ checkpoint_BC ⟶ arktx_BC ⟶ vtxo_C
        (collab path)                        (collab path)
```

Once `vtxo_A` lands on chain via Alice's unilateral exit, **C is the only motivated party for every link in the recovery DAG above her** with the partial exception of checkpoint hops where the operator has aligned incentive (timeout claim of the predecessor VTXO).

| Hop | Aligned actor |
|-----|---------------|
| Broadcast checkpoint_AB | Operator (timeout-claims vtxo_A). C as backstop. |
| Broadcast arktx_AB | **Only C** — Bob's vtxo_B is "spent" in his books; he gains nothing. |
| Broadcast checkpoint_BC | Operator (timeout-claims vtxo_B). C as backstop. |
| Broadcast arktx_BC | **Only C** |
| Sweep vtxo_C after CSV | Only C |

If C does not monitor or react, an offline/unresponsive operator on any link causes silent loss of vtxo_C value at the next CSV expiry on Alice's exit.

## Goals

1. Detect on-chain materialization of any ancestor of a preconfirmed VTXO we hold.
2. Wait for the operator to broadcast the relevant checkpoint; only broadcast it ourselves at deadline.
3. Always broadcast our arktx hops promptly (no other party will).
4. Ratchet through every hop until our VTXO is on chain.
5. After our VTXO is materialized, perform CSV-then-wallet sweep.
6. Tear down monitors when our interest disappears (VTXO transitions to Spent or Forfeited).
7. Minimize wallet fee spend: every hop where the operator can act in our interest, defer until close to deadline.
8. Be correct under shared ancestry: multiple held preconfirmed VTXOs whose recovery DAGs overlap must share a single chainsource subscription per outpoint and a single broadcast attempt per checkpoint/arktx txid.
9. Be durable across restart.

## Non-goals

- Watching ancestors of *confirmed* VTXOs. Confirmed VTXOs have no ancestor risk.
- Re-implementing the legitimate unroll flow. The new package only reacts to *external* DAG materialization; it never broadcasts a recovery DAG of its own to start an exit.
- Operator → client liveness signaling (e.g. mailbox push of "checkpoint in flight"). Future work; for v1, recipients poll mempool/chain.
- Rewriting `client/unroll/sweep.go`. Either copy what's needed or extract a `client/sweep` package later if duplication grows.
- Reorg handling. See [Out-of-scope / future work](#out-of-scope--future-work).

### Scope

A watcher subscription exists for each preconfirmed (`Origin=OOR`) VTXO the client owns and is `Live`. Confirmed VTXOs and terminal VTXOs (Spent / Forfeited / Failed) carry no watcher.

### Topology

Frontier-style watch over the **single path** in the recovery DAG that ends at our VTXO. Mirror of the server's `batchwatcher` but single-path instead of fan-out. Initial frontier is the confirmed-on-chain ancestor at the top of our path (a batch-tx output we depend on); advance one node at a time as our path materializes.

Sibling VTXT branches are not registered, so we never observe them.

### Per-hop state machine

For each ratchet step:

1. **Idle / Watching frontier.** Spend watch on the current frontier outpoint. No timer.
2. **Spend observed.** Classify the spending tx by node type along our path:
   - **Inner VTXT-tree node confirmed.** Advance frontier to the next outpoint on our path. Re-enter Idle. No further action.
   - **VTXO leaf ancestor confirmed (= a previous-owner VTXO that we have a checkpoint for).** Enter `WaitingForCheckpoint` for that outpoint.
   - **Checkpoint confirmed.** Enter `MustBroadcastArkTx` for the matching arktx.
   - **Arktx confirmed and its on-path output is our VTXO.** Terminal: hand off to sweep flow.
   - **Arktx confirmed and its on-path output is an intermediate vtxo_X.** Advance frontier to vtxo_X; re-enter `WaitingForCheckpoint`.
3. **WaitingForCheckpoint(vtxo_X).**
   - Mempool/chain poll for the deterministic checkpoint txid (we hold the fully-signed PSBT; the txid is stable).
   - If observed → ride along; advance to "watching for confirmation."
   - If `current_height >= leaf_confirm_height + t_e − recipient_safety_margin` and still nothing observed → request single-flight broadcast of checkpoint_X via internal broadcast registry.
4. **MustBroadcastArkTx(arktx_X).**
   - Urgent regime: no aligned actor besides us. Operator's `t_c` claim window is open.
   - Request single-flight broadcast of arktx_X via internal broadcast registry. No wait timer.
5. **Sweep.** Once our VTXO is on chain (terminal of step 2), wait CSV, build sweep tx, broadcast, await confirmation. Mark VTXO terminal with a "fraud-response" discriminator distinguishing it from a client-initiated unroll.

### Timing

| Phase | Wait policy | Reason |
|-------|-------------|--------|
| WaitingForCheckpoint | Defer to `leaf_confirm_height + t_e − recipient_safety_margin` | Operator pays the fee out of recovered funds; recipient's fee is not recovered. Defer if at all possible. |
| MustBroadcastArkTx | Immediate | No other actor is aligned. Operator's `t_c` timeout claim races us. |
| Sweep | Wait CSV, then broadcast | Standard. |

`recipient_safety_margin` should be larger than the operator's recommended 6-block margin (ARK-04 §588) — the recipient is a backstop with potentially stale chain view and possibly intermittent connectivity. Configurable; suggested default 20–24 blocks.

### Single-flight broadcast + refcounted subscriptions

Three structurally identical refcount tables inside the watcher:

- `outpoint → (chainsource SpendActor handle, interested-VTXO set)`
- `checkpoint_txid → (broadcast job, interested-VTXO set)`
- `arktx_txid → (broadcast job, interested-VTXO set)`

For each:

- A new interested VTXO joining an existing entry is an in-memory map update only; no new chainsource subscription, no new broadcast attempt.
- An interested VTXO leaving (because it terminated, was onward-spent, was forfeited, etc.) is a refcount decrement.
- When the interested set hits empty, the underlying resource is released (`UnregisterSpendRequest` for outpoints; broadcast jobs are GC'd if not in flight).

This is the same single-flight + dedup shape the server's `fraud/` package uses (operator-side commit `fraud: Dedup in-flight forfeit broadcast jobs`).

### Cleanup invariant

A held preconfirmed VTXO has interests recorded in some subset of the three tables. When the VTXO transitions to **any** terminal-from-our-perspective state — `Spent` (onward OOR), `Forfeited`, materialization complete — every entry holding that VTXO in its interested set drops it and decrements refcount. No money at stake → no fees burned watching.

### Multi-input OOR

A preconfirmed VTXO with `len(desc.Ancestry) > 1` runs N independent ratchets, one per fragment, each with its own frontier and per-hop state machine. Each fragment's VTXO interest is registered independently into the three refcount tables; shared outpoints across fragments are deduplicated naturally by the same machinery that handles cross-VTXO sharing.

### Shared ancestry across multiple held VTXOs

Two preconfirmed VTXOs we hold may share segments of their recovery DAGs (e.g., both descend from the same arktx). The refcount tables make this transparent:

- Single chainsource subscription per shared outpoint.
- Single broadcast attempt per shared checkpoint/arktx txid.
- Per-VTXO sub-states advance independently when events fan out.
- Termination of one VTXO never tears down a subscription still needed by another.

### Persistence and restart

Per-VTXO sub-state is durable so restart can resume mid-ratchet. Concretely:

- Per-VTXO record: target VTXO outpoint, current frontier outpoint, current phase (Idle / WaitingForCheckpoint / MustBroadcastArkTx / Sweep / Terminal), timer-relevant heights.
- The three refcount tables are *not* persisted; on restart, rebuild them from the union of live preconfirmed VTXOs and re-register all subscriptions.

The fully-signed checkpoint and arktx PSBTs along the entire DAG are already persisted at OOR receive time (`AncestorPackages` in `client/oor/receive_session.go`; `submitArkPSBTRecordType` in OOR session storage). No new persistence is required for broadcast material.

### chainsource fan-out is in-package

`client/chainsource/spend_actor.go` is explicit that each `SpendActor` instance serves exactly one subscription and rejects duplicate registrations. Each `RegisterSpendRequest` carries one `NotifyActor`, and the `ChainSourceActor` factory spawns a brand-new `SpendActor` (and underlying backend subscription) per request. There is no built-in multi-subscriber fan-out.

The new watcher therefore owns the fan-out internally: exactly one `RegisterSpendRequest` per watched outpoint, with the watcher itself as `NotifyActor`, and an in-memory dispatch map distributing each `SpendEvent` to all interested per-VTXO sub-states.

### Package layout

New top-level package: **`client/fraud`**.

Mirrors the server's `fraud/` package both in name and structure:

- A singleton-per-daemon watcher actor.
- Per-VTXO sub-state records (durable).
- Refcount tables for spend subscriptions and broadcast jobs.
- Internal broadcast worker that builds CPFP children for checkpoint/arktx parents (TRUC v3 + ephemeral-anchor; production analogue of the operator-side `harness.ForceBroadcastLineageTx` test helper).

Sweep code initially copied from `client/unroll/sweep.go`. If duplication becomes meaningful, extract a `client/sweep` package. (Tracked in Future Work below; not required for v1.)

## Itest plan

The operator-side fraud itests on the server branch (`itest/fraud_response_test.go`) drive every flow with the *client* simulated by harness force-broadcast. The client-side mirror inverts that: silence the operator's fraud responder for the relevant hops and require the *real* client to react. The same harness primitives (`ForceBroadcastLineageTx`, `Bitcoind` RPC wrapper, in-process operator restart) carry over.

Concrete scenarios to land alongside the implementation, modeled on the existing server itests:

1. **Single-hop spent-VTXO checkpoint backstop.** `TestRecipientFraudCheckpointBackstop`. Alice OORs to Bob. Operator fraud responder disabled (via config mutator or by stopping the responder actor). Harness force-broadcasts Alice's source VTXO lineage onto chain. Advance the chain past `t_e − recipient_safety_margin`. Assert Bob's client broadcasts checkpoint_AB, then arktx_AB on confirmation, then sweeps vtxo_B after CSV. Wallet receives the swept funds.

2. **Single-hop deferral.** `TestRecipientFraudDefersToOperator`. Same setup but operator fraud responder enabled. Assert Bob's client *does not* broadcast checkpoint_AB and never consumes a wallet UTXO for one. After the operator's checkpoint and arktx confirm, Bob's client takes over only at the sweep step.

3. **Multihop ratchet.** `TestRecipientFraudMultihopRatchet`. Mirror of the server's `TestFraudResponseMultihopOORRatchet`. Setup A → B → C. Operator fraud responder disabled. Force vtxo_A on chain. Carol's client must broadcast checkpoint_AB, arktx_AB, checkpoint_BC, arktx_BC in order, then sweep vtxo_C.

4. **Mixed responsiveness.** `TestRecipientFraudPartialOperator`. A → B → C. Operator broadcasts checkpoint_AB and arktx_AB faithfully but is silent on the BC hop. Carol's client defers on the AB hop and acts on the BC hop.

5. **Shared ancestry single-flight.** `TestRecipientFraudSharedAncestrySingleFlight`. Construct two preconfirmed VTXOs Carol holds whose recovery DAGs share an ancestor outpoint. Force the shared ancestor on chain with operator silent. Assert exactly one chainsource subscription per shared outpoint and exactly one broadcast attempt per shared checkpoint/arktx txid (verify via the same `requireOnlyConfirmedSpender` shape the server itests use).

6. **Multi-input OOR.** `TestRecipientFraudMultiInputOOR`. Carol holds a single preconfirmed VTXO with `len(desc.Ancestry) == 2`. Force one fragment's ancestor on chain; assert that fragment's ratchet completes while the other fragment's watcher remains armed and idle. Optionally force the second ancestor too and assert both paths complete independently.

7. **Cleanup on onward-spend.** `TestRecipientFraudCleanupOnOnwardSpend`. Bob holds vtxo_B (preconfirmed). Bob OORs to Carol. Assert all spend subscriptions and any pending broadcast jobs Bob's watcher held for vtxo_B's ancestors are released; no further wallet activity from Bob's daemon when Alice's source VTXO is later force-broadcast.

8. **Restart mid-ratchet.** `TestRecipientFraudRestartMidRatchet`. Mirror of the server's `TestFraudRestartCheckpointMaturity` / `TestFraudRestartForfeitResponse`. Force ratchet to advance partway; restart the client daemon; assert it resumes from the persisted phase and completes.

These should live in `itest/recipient_fraud_test.go` and reuse existing harness helpers (`forceBroadcastLineageToOutpoint`, `waitForSpendOnChainOrMempool`, `waitForServerVTXOStatus`, etc.) where applicable.

## Open questions

1. **Where does `t_e` come from?** It is set per-policy at OOR time. Verify whether it is already exposed on the recipient's persisted descriptor / OOR artifacts, or whether plumbing is needed.
2. **SPV / esplora mempool visibility.** The wait-and-defer logic relies on observing the operator's checkpoint in the mempool to stand down. If the recipient's chain backend is esplora and does not surface mempool, we fall back to confirmation-only observation (slightly more aggressive trigger, marginally more wasted broadcasts). Verify which client backends apply.
3. **DoS surface.** A previous owner who broadcasts the recovery DAG but never broadcasts the unilateral-exit timeout-claim will repeatedly grief us into checkpoint backstops. Bounded by `recipient_safety_margin` and per-VTXO single-flight, but worth quantifying.
4. **Fee policy for our broadcasts.** Reuse `txconfirm`'s fee escalation, or set a stricter recipient-side cap?

## Out-of-scope / future work

- **Reorg handling.** Deferred. `chainsource.SpendActor` already supports re-org events (its `monitorSpend` loop continues monitoring after first delivery), so the hook exists, but walking the per-hop state machine back one phase on a checkpoint or arktx un-confirmation is non-trivial and not required for the v1 happy path. Track as a follow-up issue once the basic ratchet lands. v1 may simply log loud and stay in the most-recent-observed phase if a confirmation reorgs out.
- Operator → client mailbox push of "checkpoint in flight" so backstops stand down without polling.
- Coordination across multiple recipients of overlapping DAGs (e.g., both C and D in `A → B → C → D` redundantly broadcasting). Bounded grief; v1 accepts the duplication.
- Extracting a shared `client/sweep` package once duplication with `client/unroll/sweep.go` is meaningful.
- A spec PR adding "Recipient Fraud-Response Protocol" as a new section in ARK-04 or ARK-05 to formalize the wait-and-react timing window so server and client agree on it.

## Acceptance criteria

1. Receiving a preconfirmed VTXO arms a watcher; transitioning to any terminal state disarms it and releases all refcounted resources.
2. Force-broadcasting an ancestor leaf VTXO and *not* having the operator broadcast its checkpoint causes the recipient to broadcast the checkpoint just before `t_e − recipient_safety_margin`.
3. Force-broadcasting an ancestor leaf VTXO *with* operator response causes the recipient to defer (no recipient broadcast, no wallet UTXOs consumed).
4. After checkpoint confirms, the recipient broadcasts the matching arktx promptly (no wait).
5. Multihop scenario `A → B → C` end-to-end: with a partly-offline operator, C's watcher ratchets through every hop and recipient sweep confirms.
6. Two held preconfirmed VTXOs sharing an ancestor result in exactly one chainsource subscription per shared outpoint and exactly one broadcast attempt per shared checkpoint/arktx txid.
7. Multi-input OOR (N fragments) launches N parallel ratchets that complete independently.
8. Restart mid-ratchet resumes correctly: subscription graph rebuilt, in-flight broadcasts rejoined or re-issued under txid dedup, timers anchored on persisted heights.
9. The full itest suite described in [Itest plan](#itest-plan) passes — at minimum scenarios 1–4, 7, and 8 land with v1; 5 and 6 may follow incrementally if early-prototype constraints make them harder to land in a single PR.

## References

- ARK-00 §279, §296–§315 — recipient trust assumptions and the implicit recipient-as-backstop role.
- ARK-04 — operator-side fraud-response protocol (the symmetry the new package mirrors).
- ARK-03 §613–§618 — recipient monitoring SHOULDs.
- Server-side counterparts: operator `fraud/` package, `batchwatcher/` extensions, `oor/checkpoint_extract.go`.
- `client/unroll/CLAUDE.md` — multi-tree ancestry handling and `desc.Ancestry` plural.
- `client/chainsource/spend_actor.go` — single-subscription-per-actor invariant; informs the in-package fan-out design.

ellemouton:
## Summary

Adds the client-side recipient fraud-response defense for preconfirmed OOR VTXOs.

Receiving a live OOR VTXO arms passive ancestor spend watches. When any watched ancestor materializes on chain the watcher starts (or reuses) a durable unroll job under `TriggerFraudSpend`. The unroll job materializes the full proof DAG required for the target, defers checkpoint broadcasts to give the operator a window to broadcast first, broadcasts ark transactions promptly once their parents confirm, waits CSV, and sweeps to the wallet.

Fixes lightninglabs/darepo-client#345.

Paired with the server-side integration tests in lightninglabs/darepo#334 — the server PR's submodule pointer bumps to this branch's tip, and its recipient fraud itest suite (`TestRecipientFraud*`) exercises this client end-to-end.

## Details

- `fraud` is a slim passive package: it has **no** `txconfirm` dependency and never broadcasts checkpoint or ark transactions. It builds a `WatchPlan` from each live OOR VTXO's descriptor ancestry, deduplicates shared watch outpoints via refcounting, and on an observed ancestor spend fans out `unroll.EnsureUnrollRequest{Trigger: TriggerFraudSpend}` per affected target. Terminal VTXO transitions release passive watches so the watcher does not pay fees for VTXOs the wallet no longer owns.
- `unroll` owns full fraud-triggered recovery execution. Under `TriggerFraudSpend` it defers ready checkpoint nodes to a deadline computed from CSV delay and recipient safety margin, registers confirmation watches by deterministic txid, and broadcasts the checkpoint through `txconfirm` only if the deadline arrives without operator confirmation. Ark nodes are never deferred — once their checkpoint parents confirm, they go to `txconfirm` immediately. The deferred-checkpoint state is durable: `JobState.DeferredCheckpoints` round-trips through the actor's TLV checkpoint codec, and `ResumeEvent` reissues every persisted watch on restart.
- `oor` and `vtxo` gain optional `IncomingVTXOObserver` and `TerminalVTXOObserver` callback hooks. `darepod` wires these to the fraud watcher so incoming OOR materialization arms watches and terminal VTXO cleanup releases them, all without those packages taking a direct dependency on `fraud`.
- `txconfirm` package re-broadcasts now tolerate a deeply-confirmed parent that bitcoind dropped from its rejects cache. The `bad-txns-inputs-missingorspent` / `bad-txns-inputs-missing` combination on the parent plus a child-failure sentinel is treated as "parent already broadcast", letting the actor fall through to its existing confirmation watch instead of collapsing to `Failed`. This unblocks the fraud-triggered handoff where the operator's checkpoint confirms before the recipient deadline.

Mempool observation is intentionally **not** implemented for deferred checkpoints. Neutrino-backed clients cannot inspect the mempool, and supporting only some backends would create a silent two-mode behavior. The recipient deferral relies on chain confirmation plus the deadline backstop, which is uniform across backends.

## Validation

- `make lint-native` clean.
- `go test -race ./fraud ./unroll ./oor ./vtxo ./txconfirm ./chainsource` green.
- Full recipient fraud itest suite (`TestRecipientFraud*` in lightninglabs/darepo#334) passes against this branch: `BroadcastsArkAfterRestart`, `DefersToOperator`, `MultihopRatchet`, `PartialOperator`, `MultiInputOOR`, `CleanupOnOnwardSpend`.

## Design notes

See `docs/recipient_fraud_unroll_execplan.md` (added in the `darepod` commit) for the ExecPlan that motivated this branch's design pivot from "fraud package owns recovery execution" to "fraud is passive detection, unroll owns recovery DAG materialization under `TriggerFraudSpend`".

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -34,6 +34,7 @@ import (
 	"github.com/lightninglabs/darepo-client/daemonrpc"
 	"github.com/lightninglabs/darepo-client/db"
 	"github.com/lightninglabs/darepo-client/db/actordelivery"
+	"github.com/lightninglabs/darepo-client/fraud"
 	"github.com/lightninglabs/darepo-client/indexer"
 	"github.com/lightninglabs/darepo-client/ledger"
 	"github.com/lightninglabs/darepo-client/lib/actormsg"
@@ -280,6 +281,12 @@ type Server struct {
 		unroll.RegistryMsg, unroll.RegistryResp,
 	]]
 	unrollRegistry *unroll.UnrollRegistryActor
+
+	// fraudWatcherRef is the passive recipient-fraud watcher. It arms
+	// ancestry spend watches for received OOR VTXOs and starts unroll jobs
+	// only after a watched ancestor materializes.
+	fraudWatcherRef fn.Option[actor.ActorRef[fraud.Msg, fraud.Resp]]
+	fraudWatcher    *fraud.WatcherActor
 
 	// lazyChainResolver is the forwarding ref that connects the
 	// VTXO manager's critical-expiry path to the unroll manager.
@@ -745,6 +752,10 @@ func (s *Server) run(ctx context.Context, shutdownFn func()) error {
 
 		if s.unrollRegistry != nil {
 			s.unrollRegistry.Stop()
+		}
+
+		if s.fraudWatcher != nil {
+			s.fraudWatcher.Stop()
 		}
 
 		if s.oorSigningEffect != nil {
@@ -1778,7 +1789,7 @@ func (s *Server) startWalletDependentActors(ctx context.Context,
 	// 12. Register the unilateral-exit subsystem.
 	// -------------------------------------------------------
 	if err := s.initUnrollSubsystem(
-		ctx, chainSourceRef, vtxoManagerRef,
+		ctx, chainSourceRef,
 	); err != nil {
 		return err
 	}
@@ -3247,6 +3258,11 @@ func (s *Server) initVTXOManager(ctx context.Context,
 		LedgerSink:       fn.Some(ledger.NewSink(s.actorSystem)),
 		ChainResolver:    chainResolver,
 		RefreshFeeQuoter: s.autoRefreshFeeQuoter(),
+		TerminalVTXOObserver: func(ctx context.Context,
+			outpoint wire.OutPoint) error {
+
+			return s.untrackFraudVTXO(ctx, outpoint)
+		},
 	})
 
 	managerKey := actor.NewServiceKey[vtxo.ManagerMsg, vtxo.ManagerResp](
@@ -3443,6 +3459,11 @@ func (s *Server) initOORActor(ctx context.Context,
 		VTXOManager:     vtxoManagerRef,
 		VTXOStore:       vtxoStore,
 		LedgerSink:      fn.Some(ledger.NewSink(s.actorSystem)),
+		IncomingVTXOObserver: func(ctx context.Context,
+			descs []*vtxo.Descriptor) error {
+
+			return s.trackIncomingFraudVTXOs(ctx, descs)
+		},
 	})
 
 	// Wire the timeout callback ref using the registered service
@@ -4113,8 +4134,7 @@ func (w *btcwUnrollWallet) ReleaseOutput(_ context.Context, id wallet.LockID,
 func (s *Server) initUnrollSubsystem(ctx context.Context,
 	chainSourceRef actor.ActorRef[
 		chainsource.ChainSourceMsg, chainsource.ChainSourceResp,
-	],
-	vtxoManagerRef actor.TellOnlyRef[vtxo.ManagerMsg]) error {
+	]) error {
 
 	dbStore := db.NewStore(
 		s.db.DB, s.db.Queries, s.db.Backend(),
@@ -4196,6 +4216,11 @@ func (s *Server) initUnrollSubsystem(ctx context.Context,
 	s.unrollRegistry = registry
 	s.unrollRegistryRef = fn.Some(registry.Ref())
 
+	err := s.initFraudWatcher(ctx, chainSourceRef)
+	if err != nil {
+		return err
+	}
+
 	// 3. Restore non-terminal jobs from durable state.
 	if err := registry.RestoreNonTerminal(ctx); err != nil {
 		s.log.WarnS(ctx,
@@ -4225,6 +4250,78 @@ func (s *Server) initUnrollSubsystem(ctx context.Context,
 	s.log.InfoS(ctx, "Unroll subsystem initialized")
 
 	return nil
+}
+
+// initFraudWatcher creates the passive recipient-fraud watcher and restores
+// watches for live OOR VTXOs after daemon restart. The live VTXO set is
+// sourced from the VTXO manager (the authoritative runtime view) so the
+// watcher arms exactly the same set of VTXOs the manager has spawned
+// child actors for, rather than reading the store directly.
+func (s *Server) initFraudWatcher(ctx context.Context,
+	chainSourceRef actor.ActorRef[
+		chainsource.ChainSourceMsg, chainsource.ChainSourceResp,
+	],
+) error {
+
+	if !s.unrollRegistryRef.IsSome() {
+		return fmt.Errorf("unroll registry not initialized")
+	}
+	if !s.vtxoMgrRef.IsSome() {
+		return fmt.Errorf("VTXO manager not initialized")
+	}
+
+	//nolint:contextcheck // watcher owns its own root context lifecycle
+	watcher := fraud.NewWatcherActor(fraud.WatcherConfig{
+		ChainSource: chainSourceRef,
+		UnrollRef:   s.unrollRegistryRef.UnsafeFromSome(),
+		Log:         fn.Some(s.subLogger(fraud.Subsystem)),
+	})
+	s.fraudWatcher = watcher
+	s.fraudWatcherRef = fn.Some(watcher.Ref())
+
+	resp, err := s.vtxoMgrRef.UnsafeFromSome().Ask(
+		ctx, &vtxo.ListLiveDescriptorsRequest{},
+	).Await(ctx).Unpack()
+	if err != nil {
+		return fmt.Errorf("list live VTXOs for fraud restore: %w", err)
+	}
+	listResp, ok := resp.(*vtxo.ListLiveDescriptorsResponse)
+	if !ok {
+		return fmt.Errorf("unexpected VTXO manager response %T", resp)
+	}
+
+	return s.trackIncomingFraudVTXOs(ctx, listResp.Descriptors)
+}
+
+// trackIncomingFraudVTXOs arms recipient fraud watches for materialized OOR
+// VTXOs that still depend on preconfirmed ancestry.
+func (s *Server) trackIncomingFraudVTXOs(ctx context.Context,
+	descs []*vtxo.Descriptor) error {
+
+	if !s.fraudWatcherRef.IsSome() {
+		return fmt.Errorf("recipient fraud watcher not initialized")
+	}
+
+	return fraud.TrackVTXOs(
+		ctx, s.fraudWatcherRef.UnsafeFromSome(), descs,
+	)
+}
+
+// untrackFraudVTXO releases recipient fraud watcher interest for one VTXO.
+func (s *Server) untrackFraudVTXO(ctx context.Context,
+	outpoint wire.OutPoint) error {
+
+	if !s.fraudWatcherRef.IsSome() {
+		return nil
+	}
+
+	notifyCtx := context.WithoutCancel(ctx)
+
+	return s.fraudWatcherRef.UnsafeFromSome().Tell(
+		notifyCtx, &fraud.UntrackRequest{
+			TargetOutpoint: outpoint,
+		},
+	)
 }
 
 // unrollMaxFeeRate returns the configured max fee rate or zero to let
--- a/fraud/actor.go
+++ b/fraud/actor.go
@@ -1,0 +1,462 @@
+package fraud
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btclog/v2"
+	"github.com/lightninglabs/darepo-client/baselib/actor"
+	"github.com/lightninglabs/darepo-client/chainsource"
+	"github.com/lightninglabs/darepo-client/unroll"
+	"github.com/lightninglabs/darepo-client/vtxo"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+)
+
+const (
+	// Subsystem is the logging subsystem label for recipient fraud watches.
+	Subsystem = "CFRD"
+
+	// ServiceKeyName is the receptionist key for the recipient fraud
+	// watcher.
+	ServiceKeyName = "recipient-fraud-watcher"
+)
+
+// ServiceKey returns the actor service key for the recipient fraud
+// watcher. Callers that need to look the watcher up via the
+// receptionist should use this rather than constructing the key
+// themselves; it locks down a single concrete instantiation site.
+func ServiceKey() actor.ServiceKey[Msg, Resp] {
+	return actor.NewServiceKey[Msg, Resp](ServiceKeyName)
+}
+
+// WatcherConfig configures the recipient fraud watcher actor.
+type WatcherConfig struct {
+	// ChainSource provides passive ancestor spend watches.
+	ChainSource actor.ActorRef[
+		chainsource.ChainSourceMsg, chainsource.ChainSourceResp,
+	]
+
+	// UnrollRef starts durable unroll jobs for affected targets.
+	UnrollRef actor.ActorRef[unroll.RegistryMsg, unroll.RegistryResp]
+
+	// Log is an optional logger.
+	Log fn.Option[btclog.Logger]
+
+	// MailboxSize overrides the watcher's actor mailbox capacity.
+	// Zero or negative applies defaultWatcherMailboxSize. Sized to
+	// absorb a burst of per-target spend notifications during a
+	// chain reorganization without blocking the chainsource
+	// publisher.
+	MailboxSize int
+}
+
+// defaultWatcherMailboxSize is the fallback mailbox capacity applied
+// when WatcherConfig.MailboxSize is zero or negative. Sized large
+// enough to absorb a burst of per-target spend notifications during a
+// chain reorganization without blocking the chainsource publisher.
+const defaultWatcherMailboxSize = 64
+
+// trackedTarget records the watch plan a recipient VTXO is currently
+// armed under.
+type trackedTarget struct {
+	plan *WatchPlan
+}
+
+// WatcherActor is the recipient fraud watcher: a passive ancestor-spend
+// monitor that calls unroll.EnsureUnroll(..., TriggerFraudSpend) when
+// any watched ancestor of a tracked OOR VTXO is observed spent on
+// chain. It is both the public handle (Ref / Stop) and the
+// actor.ActorBehavior implementation; the runtime is driven through
+// the embedded actor.
+type WatcherActor struct {
+	cfg     WatcherConfig
+	log     btclog.Logger
+	selfRef actor.TellOnlyRef[Msg]
+
+	actor *actor.Actor[Msg, Resp]
+
+	// targets indexes the watch plan armed for each recipient VTXO.
+	targets map[wire.OutPoint]*trackedTarget
+
+	// watches collects the per-ancestor-outpoint refcounted state.
+	// Its set-of-targets per outpoint doubles as the reference count
+	// so the watcher never has to keep a separate counter + a
+	// pkScript map + a target-set map in sync.
+	watches ancestorWatches
+}
+
+// Ref returns the public watcher actor reference.
+func (w *WatcherActor) Ref() actor.ActorRef[Msg, Resp] {
+	if w == nil || w.actor == nil {
+		return nil
+	}
+
+	return w.actor.Ref()
+}
+
+// Stop stops the underlying watcher actor.
+func (w *WatcherActor) Stop() {
+	if w == nil || w.actor == nil {
+		return
+	}
+
+	w.actor.Stop()
+}
+
+// NewWatcherActor creates and starts the recipient fraud watcher actor.
+func NewWatcherActor(cfg WatcherConfig) *WatcherActor {
+	w := &WatcherActor{
+		cfg:     cfg,
+		log:     cfg.Log.UnwrapOr(btclog.Disabled),
+		targets: make(map[wire.OutPoint]*trackedTarget),
+		watches: newAncestorWatches(),
+	}
+
+	mailbox := cfg.MailboxSize
+	if mailbox <= 0 {
+		mailbox = defaultWatcherMailboxSize
+	}
+	w.actor = actor.NewActor(actor.ActorConfig[Msg, Resp]{
+		ID:          ServiceKeyName,
+		Behavior:    w,
+		MailboxSize: mailbox,
+	})
+	w.selfRef = w.actor.TellRef()
+	w.actor.Start()
+
+	return w
+}
+
+// Receive processes one watcher actor message.
+func (w *WatcherActor) Receive(ctx context.Context, msg Msg) fn.Result[Resp] {
+	var (
+		resp Resp
+		err  error
+	)
+
+	switch m := msg.(type) {
+	case *TrackVTXOsRequest:
+		resp, err = w.handleTrackVTXOs(ctx, m)
+
+	case *UntrackRequest:
+		resp, err = w.handleUntrack(ctx, m)
+
+	case *SpendObservedMsg:
+		resp, err = w.handleSpendObserved(ctx, m)
+
+	default:
+		err = fmt.Errorf("unknown fraud watcher message: %T", msg)
+	}
+
+	if err != nil {
+		return fn.Err[Resp](err)
+	}
+
+	return fn.Ok[Resp](resp)
+}
+
+// OnStop releases every active spend watch. A failure on one outpoint
+// must not abandon the rest, so each failure is logged and aggregated
+// into the returned joined error.
+func (w *WatcherActor) OnStop(ctx context.Context) error {
+	var errs error
+	for _, outpoint := range w.watches.outpoints() {
+		point, ok := w.watches.pointAt(outpoint)
+		if !ok {
+			continue
+		}
+		if err := w.unregisterSpendWatchPoint(ctx, point); err != nil {
+			w.log.WarnS(
+				ctx,
+				"Failed to unregister spend watch on stop",
+				err,
+				slog.String("outpoint", outpoint.String()),
+			)
+			errs = joinTrackError(errs, err)
+		}
+	}
+
+	return errs
+}
+
+// handleTrackVTXOs admits a batch of OOR VTXO descriptors and arms their
+// ancestor spend watches. Per-descriptor failures are aggregated rather
+// than short-circuiting the batch so one malformed descriptor cannot
+// disarm fraud defenses for the rest. A summary line is emitted so
+// daemon-startup callers (which use Tell and drop the returned error)
+// still see partial failures in the operator logs.
+func (w *WatcherActor) handleTrackVTXOs(ctx context.Context,
+	req *TrackVTXOsRequest) (Resp, error) {
+
+	if req == nil {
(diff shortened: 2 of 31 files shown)


fraud/actor.go
```diff
@@ -0,0 +1,684 @@
...
+			NotifyActor: fn.Some[actor.TellOnlyRef[chainsource.SpendEvent]](notify),
+		},
+	).Await(ctx).Unpack()
+	if err != nil {
+		return fmt.Errorf("register spend %s: %w", outpoint, err)
+	}
+	if _, ok := resp.(*chainsource.RegisterSpendResponse); !ok {
+		return fmt.Errorf("register spend %s response type %T",
+			outpoint, resp)
+	}
```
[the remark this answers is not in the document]
ellemouton: Acknowledged but punting to a follow-up. Switching every chainsource/txconfirm/unroll Ask in this actor to async-completion (Tell + self-message on future ready) is a substantive redesign that touches every side-effect path; doing it in this PR would balloon the diff and obscure the review of the watcher's state machine. The actor currently mitigates the head-of-line risk by sizing the mailbox at 64 and keeping per-Ask work bounded by chainsource latency. Tracking issue to follow.

fraud/actor.go
```diff
@@ -0,0 +1,684 @@
...
+			msg.Outpoint, msg.SpendingTx, msg.Height,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := w.executeActions(ctx, target, actions); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
ellemouton: Addressed. Refactored handleSpendObserved/TxConfirmed/TxFailed/HeightObserved through a shared `fanOut` helper that logs the per-target failure, continues the loop, and returns errors.Join over the aggregated errors so one bad target cannot block fraud response for the rest.

fraud/actor.go
```diff
@@ -0,0 +1,684 @@
...
+		return false, nil
+	}
+
+	result := acceptResp.Results[0]
+	if result.Accepted {
+		return false, nil
+	}
+
+	return chainsource.IsIgnorableMempoolRejectReason(result.Reason), nil
+}
```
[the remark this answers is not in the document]
ellemouton: Addressed. Deleted txKnownToMempool, the TestMempoolAccept arms in the actor/integration test fakes, the integrationChainRef.mempool field + setMempoolKnown helper, and the two tests that were exercising the removed stand-down (TestWatcherBehaviorSkipsKnownCheckpointBroadcast, TestWatcherActorRestartStandsDownForMempoolCheckpoint).

fraud/actor.go
```diff
@@ -0,0 +1,684 @@
...
+	log     btclog.Logger
+	selfRef actor.TellOnlyRef[Msg]
+
+	states map[wire.OutPoint]*State
+
+	spendRefs     map[wire.OutPoint]int
+	txWatchRefs   map[chainhash.Hash]int
+	broadcastRefs map[chainhash.Hash]int
+
+	lastTxFailure map[chainhash.Hash]string
```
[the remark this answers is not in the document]
ellemouton: Addressed. Removed the lastTxFailure field, its make() in newWatcherBehavior, and the writes/deletes in handleTxConfirmed and handleTxFailed. Per-target failures are now logged via the new fanOut helper instead.

fraud/actor.go
```diff
@@ -0,0 +1,684 @@
...
+	known, err := w.txKnownToMempool(ctx, action.Tx)
+	if err != nil {
+		return err
+	}
+	if known {
+		return nil
+	}
+
+	return w.broadcast(ctx, action)
+}
```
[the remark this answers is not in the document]
ellemouton: Addressed. Removed the txKnownToMempool stand-down — broadcastCheckpoint is gone and ActionBroadcastCheckpoint now collapses into the same w.broadcast path as ActionBroadcastArk. txconfirm tolerates already-known parents (see the txconfirm: Tolerate deeply-confirmed parent on package re-broadcast commit on this branch), so the recipient CPFP always gets registered.

ellemouton:
@claude review this

unroll/actor.go
```diff
@@ -1176,6 +1260,54 @@ func copyTx(tx *wire.MsgTx) *wire.MsgTx {
...
+	}
+
+	return result
+}
+
+// findDeferredCheckpoint returns the deferred checkpoint for txid if present.
+func findDeferredCheckpoint(checkpoints []DeferredCheckpoint,
+	txid chainhash.Hash) (DeferredCheckpoint, bool) {
+
+	for _, checkpoint := range checkpoints {
```
Roasbeef, line 1290: Non-blocking, can use a txid keyed map here for quicker loop up, same above re delete.

unroll/deferred_checkpoint_codec.go
```diff
@@ -0,0 +1,75 @@
...
+
+	var buf bytes.Buffer
+	sorted := copyDeferredCheckpoints(checkpoints)
+	count := uint32(len(sorted))
+	if err := binary.Write(&buf, binary.BigEndian, count); err != nil {
+		return nil, fmt.Errorf("write deferred checkpoint count: %w",
+			err)
+	}
+
+	for _, checkpoint := range sorted {
```
Roasbeef: Doesn't leave any room for later extension of `DeferredCheckpoint`. If this was a TLV itself, then we wouldn't have that worry. As it's a txid+int, should be able to use the existing primitive TLV typs for this.

unroll/fsm_logic.go
```diff
@@ -410,3 +438,140 @@ func readyTxids(frontier []unrollplan.TxFrontier) []chainhash.Hash {
...
+			job.DeferredCheckpoints, item.Txid,
+		); ok {
+
+			continue
+		}
+
+		deadline := checkpointBackstopHeight(
+			job.Height, env.Proof.CSVDelay(),
+		)
+		if deadline <= job.Height {
```
Roasbeef, line 498: Something we can do in a later pass, but we may want to build in a safety margin here. Eg if the expiry is actually in the very next block, then we may want to flag it as ready early. Or within N blocks. etc.

Roasbeef: Nvm, see the buffer margin!

fraud/actor.go
```diff
@@ -0,0 +1,367 @@
...
+	for outpoint := range w.watchRefs {
+		if err := w.unregisterSpendWatch(ctx, outpoint); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (w *watcherBehavior) handleTrackVTXOs(ctx context.Context,
```
Roasbeef: Missing `godoc` comments on this method and below, some attrs above as well

fraud/actor.go
```diff
@@ -0,0 +1,367 @@
...
+	UnrollRef actor.ActorRef[unroll.RegistryMsg, unroll.RegistryResp]
+
+	// Log is an optional logger.
+	Log fn.Option[btclog.Logger]
+}
+
+// WatcherActor wraps the recipient fraud watcher behavior.
+type WatcherActor struct {
+	ref   actor.ActorRef[Msg, Resp]
+	actor *actor.Actor[Msg, Resp]
```
Roasbeef: Why not just have the methds as part of this struct? Then you don't need the `watcherBehavior` internal split.

fraud/actor.go
```diff
@@ -0,0 +1,367 @@
...
+	a.actor.Stop()
+}
+
+// NewWatcherActor creates and starts the recipient fraud watcher actor.
+func NewWatcherActor(cfg WatcherConfig) *WatcherActor {
+	behavior := newWatcherBehavior(cfg)
+	watcher := actor.NewActor(actor.ActorConfig[Msg, Resp]{
+		ID:          ServiceKeyName,
+		Behavior:    behavior,
+		MailboxSize: 64,
```
Roasbeef: Hard coded const

fraud/actor.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+// WatcherActor wraps the recipient fraud watcher behavior.
+type WatcherActor struct {
+	ref   actor.ActorRef[Msg, Resp]
+	actor *actor.Actor[Msg, Resp]
+}
+
+// Ref returns the public watcher actor reference.
+func (a *WatcherActor) Ref() actor.ActorRef[Msg, Resp] {
```
Roasbeef: As we only have a single version of this, can instead expose a public func to make the service key, then look up as needed. Non-blocking.

fraud/actor.go
```diff
@@ -0,0 +1,367 @@
...
+	outpoint, target wire.OutPoint) error {
+
+	if targets := w.watchTargets[outpoint]; targets != nil {
+		delete(targets, target)
+		if len(targets) == 0 {
+			delete(w.watchTargets, outpoint)
+		}
+	}
+
+	if w.watchRefs[outpoint] <= 1 {
```
Roasbeef: Could perhaps encapsulate the map ref counting with a composite collection.

fraud/actor.go
Roasbeef: Also dig how compact and single purpose this sub-system is 👍

docs/recipient_fraud_unroll_execplan.md
```diff
@@ -0,0 +1,721 @@
+# Refactor Recipient Fraud Recovery Into Unroll
+
+This ExecPlan is a living document. The sections `Progress`,
+`Surprises & Discoveries`, `Decision Log`, and `Outcomes & Retrospective`
+must be kept up to date as work proceeds.
+
+This document follows the repository ExecPlan requirements in `PLANS.md`.
+
+## Purpose / Big Picture
```
Roasbeef: Committed by mistake.

darepod/server.go
```diff
@@ -3393,6 +3409,11 @@ func (s *Server) initOORActor(ctx context.Context,
 		VTXOManager:     vtxoManagerRef,
 		VTXOStore:       vtxoStore,
 		LedgerSink:      fn.Some(ledger.NewSink(s.actorSystem)),
+		IncomingVTXOObserver: func(ctx context.Context,
+			descs []*vtxo.Descriptor) error {
+
+			return s.trackIncomingFraudVTXOs(ctx, descs)
```
Roasbeef, line 3465: What about ensuring all the _existing_ VTXOs we have are watched for fraud on restart? The vtxo manager can handle this.

Roasbeef (changes requested):
Nice set of changes! I like how single purpose the `fraud` actor is. 

Main thing that jumped out is that I don't see where we watch for fraud again for all relevant outputs after we restart.

ellemouton:
@claude re review

unroll/deferred_checkpoint_codec.go
ellemouton: Done. Replaced the custom binary codec with a varint count + per-entry varint-length-prefixed TLV stream of primitive records (txid + deadline, odd TLV types). Adding a new field later only needs a new odd type;
  older decoders skip it. Three new tests cover round-trip, forward-compat (unknown odd record gets skipped), and short-txid rejection.

darepod/server.go
ellemouton: we had this in `initFraudWatcher`:

```
  liveVTXOs, err := vtxoStore.ListLiveVTXOs(ctx)
  if err != nil {
      return fmt.Errorf("list live VTXOs for fraud restore: %w", err)
  }
  return s.trackIncomingFraudVTXOs(ctx, liveVTXOs)
```

but it has been changed now so that it goes through the vtxo manager.

PR #355, opened 2026-05-09
multi: fold boarding-sweep flow into wallet actor

Roasbeef:
## Summary

Refactor the boarding-sweep recovery subsystem shipped in #350 from a
standalone darepod goroutine into the existing `wallet.Ark` actor, using
the abstractions the rest of the daemon already uses (chainsource for
spend events, txconfirm for broadcast/confirmation, the actor mailbox
for everything in-between). The DB schema, RPC wire shape, and
operator-facing CLI surface are unchanged.

A follow-up PR (stacked) extends the ledger package and emits the
boarding-sweep accounting entries from the wallet actor.

### Why

The original implementation owned its own goroutine + sync.Mutex, talked
to the chain backend via the raw `chainsource.ChainBackend` interface,
ran a 10-minute rebroadcast loop, and instantiated a fresh
`*db.BoardingWalletStore` on every RPC call. All of those are
duplicated work compared to the actor abstractions the rest of the
daemon already uses, and `wallet.Ark` was the natural owner: it
already drives boarding-UTXO confirmation polling and serves
`GetBoardingBalanceRequest`. Sweep is "what happens to a boarding UTXO
when it expires."

### What changed (commit-by-commit)

1. `walletcore: extract LockID, Utxo, and OutputLeaser` — break the
   wallet ↔ txconfirm import cycle.
2. `wallet: alias LockID, Utxo, and OutputLeaser to walletcore` — keep
   existing call sites compiling.
3. `txconfirm: switch to walletcore types and expose a service key` —
   txconfirm no longer imports wallet, and the wallet actor can now
   look up the broadcaster via `txconfirm.LookupRef`.
4. `wallet: add pure boarding-sweep builder helpers` — move the
   transaction-construction primitives + their tests in from darepod.
5. `wallet: add boarding-sweep store types and interface` — relocate
   the persistence-layer Go types.
6. `wallet: own the boarding-sweep flow inside the Ark actor` — the
   bulk of the refactor: messages, handlers (`handleSweepBoardingUTXOs`,
   `handleSweepSpendNotification`, `handleSweepTxNotification`,
   `handleResumeBoardingSweeps`, `handleListBoardingSweeps`), Ark
   struct extension, and `WithBoardingSweep` option. Adds
   `boarding_pending_sweep_sat` / `boarding_swept_sat` fields on
   `GetBoardingBalanceResponse`.
7. `db: alias boarding-sweep types to the wallet package` —
   `*db.BoardingWalletStore` now satisfies `wallet.BoardingSweepStore`
   through aliases, plus a `[]BoardingStatus`-shaped sweepable-status
   query.
8. `daemonrpc: add boarding sweep balance breakdown fields` —
   `boarding_pending_sweep_sat` and `boarding_swept_sat` on
   `GetBalanceResponse` so dashboards stop seeing transient
   sweep-in-flight balances disappear into a black hole.
9. `darepod: drop boarding-sweep watcher in favor of wallet actor` —
   delete `boarding_sweep_watcher.go` + tests, trim
   `boarding_sweep.go` to adapter factories, rewrite
   `rpc_boarding_sweep.go` as a thin Ask shim.

## Test plan

- [x] `go test ./wallet/... ./txconfirm/... ./darepod/... ./db/... ./ledger/...` — all green
- [x] `make lint-native` — 0 issues
- [ ] `make itest icase=BoardingSweep` (or whichever icase the
  PR #350 series shipped) — confirm the existing integration test
  still passes against the refactored implementation
- [ ] Manual smoke: `darepocli sweep --broadcast`,
  `darepocli sweep list`, restart daemon, verify pending sweep
  resumes and reaches confirmation


wallet/boarding_sweep_actor.go
```diff
@@ -0,0 +1,1215 @@
...
+	req := &txconfirm.EnsureConfirmedReq{
+		Tx:                   tx,
+		ConfirmationPkScript: pkScript,
+		Label:                BoardingSweepBroadcastLabel,
+		HeightHint:           heightHint,
+		TargetConfs:          1,
+		Subscriber:           subscriber,
+	}
+
+	ref := txconfirm.LookupRef(a.actorSystem)
```
bhandras, line 878: codex: `submitSweepConfirmer` should unpack the `*txconfirm.EnsureConfirmedResp` and treat `State == txconfirm.TxStateFailed` as an error. `txconfirm.handleEnsure` returns an OK actor response even after moving the tx into terminal failed state for conf-watch/register/broadcast failures, so the current code can let `handleSweepBoardingUTXOs` mark the sweep `published` and return success to the RPC caller before the async failure notification corrects the store. That regresses the `status=failed`/`failure_reason` surface from #350.

db/boarding_sweep_store.go
```diff
@@ -14,150 +15,46 @@ import (
...
-
-	// BoardingSweepInputStatusFailed means the input belonged to a sweep
-	// that failed before confirmation watching completed.
-	BoardingSweepInputStatusFailed = "failed"
+// Re-exports kept for backwards-compatible call sites inside db/.
+//
+// New code should reference the wallet package directly; these aliases let
+// existing internal helpers continue compiling without sweeping renames.
+const (
+	BoardingSweepStatusPending = wallet.
```
bhandras: codex: Static Checks are currently red because the formatter wants this alias block reflowed. The same CI log also shows import ordering drift in `wallet/wallet.go` and an extra EOF blank line in `wallet/boarding_sweep_actor.go`; please run the repo fmt target and commit the delta before merge.

wallet/wallet.go
```diff
@@ -128,6 +129,40 @@ type Ark struct {
...
+	// sweepSigner builds witnesses for boarding-timeout aggregate sweep
+	// transactions and allocates wallet-managed destination scripts.
+	// nil disables the sweep subsystem.
+	sweepSigner SweepSigner
+
+	// sweepChainParams is the chain network parameters used to validate
+	// caller-supplied sweep destination addresses.
+	sweepChainParams *chaincfg.Params
+
+	// pendingSweeps tracks in-flight aggregate boarding sweeps the
```
bhandras, line 151: codex: since these new maps now represent live chainsource spend subactors owned by the wallet actor, I think the shutdown ownership should be made explicit. If `wallet.Ark.Stop` can run without tearing down the whole actor system, it should cancel any active sweep spend watches before returning; if full actor-system shutdown always owns those children, a short comment/test would make that contract clear.

wallet/boarding_sweep_actor.go
Roasbeef: Added!

db/boarding_sweep_store.go
Roasbeef: Updated, also passes the new litner

wallet/wallet.go
Roasbeef: Addressed

Issue #207, opened 2026-03-19
unroller: add re-org awareness


ellemouton:
## Context

The VTXO tree unroller (PR #183) tracks confirmation state per-level to advance the broadcast and CSV wait state machine. Currently it does not handle chain reorganizations.

## Problem

If a re-org invalidates a confirmed tree level:
- `ConfirmedTxids` entries become stale
- The unroller may skip re-broadcasting invalidated levels
- CSV delay tracking may use an incorrect baseline height

## Proposed Solution

- Subscribe to re-org notifications from the chain source
- On re-org, roll back `ConfirmedTxids` for any txids whose confirmation height is above the re-org height
- Re-register confirmation watches for rolled-back txids
- Reset `LeafConfirmHeight` if the final level's confirmation was invalidated
- Transition from `AwaitingCSV` back to `Broadcasting` if needed

Tracking issue per Roasbeef's review on #183.

PR #410, opened 2026-05-13
unroll: Reorg-safe unilateral-exit subsystem

ellemouton:
## Summary

Makes the client-side unilateral-exit subsystem reorg-safe end-to-end. Today the unroll actor treats every confirmation as monotonic: if a proof-graph anchor or the sweep tx is reorged out (or vanishes during daemon downtime), the planner happily progresses off stale state and can broadcast a sweep against a target the chain no longer holds. This PR makes the entire stack reversible:

- **chainsource** gains reorg-aware conf/spend watches with reorg + finality events (`ConfReorgedEvent`, `ConfDoneEvent`, spend equivalents). The conf/spend sub-actors are multi-shot, and finality is synthesized height-based (`FinalityDepth`) so backends that drop the underlying signal (lndclient) still produce Done.
- **chainbackends** forwards lndclient reorg pings into the new event surface; bare correlation IDs only, since gRPC transport drops the depth/block-hash payloads.
- **txconfirm** confirmation watches become reorg-aware end-to-end with a new `Finalized` terminal state; `Confirmed` is no longer terminal until finality.
- **unroll** FSM gains:
  - Rollback semantics for `TxReorgedEvent` / `SpendReorgedEvent` that prune the reorged subtree from `ConfirmedTxids` + `InFlightTxids` and downgrade the sweep when the target is lost.
  - A new `AwaitingExternalSpendFinality` state with a persisted `ProvisionalExternalSpend` anchor so an external spend doesn't terminally fail the job until it's finalized.
  - `PhaseCompleted` is provisional until the sweep is `Finalized`; the registry only fires `UnrollTerminatedMsg` after finality.
  - Restart reconciliation against the canonical chain via a new `ChainReconciler` interface. The chainsource-backed implementation probes via future-mode RegisterConf/RegisterSpend with bounded timeouts, baked per-actor caller-IDs (target outpoint) so two restored actors probing a shared proof-graph ancestor cannot collide on chainsource service keys.

## Tests

- Unit: chainsource reorg-aware conf/spend actor tests, lndclient forwarder tests, txconfirm reorg lifecycle tests, unroll reorg-safety integration tests (proof-root reorg blocks downstream materialization, target reorg clears CSV maturity, sweep-confirmation reorg reverses Completed, external-spend reorg resumes actor, sweep stays provisional until finalized), restart-reconciliation tests (target gone, sweep gone, external-spend gone, target height refresh, combined offline-reorg restart), concurrent-probe service-key collision test.
- Systest (against bitcoind + lnd in regtest): full chainsource conf reorg round-trip, full txconfirm reorg round-trip including the new TxReorged event.

## Stacked on

Targets \`recipient-fraud-watcher-client\` because the two stacks both extend \`unrollplan.State\`, the unroll FSM, and the actor checkpoint codec. No functional dependency, but rebasing onto main would require manual conflict resolution in 7-8 files; keeping the stack until the fraud PR lands.

## Test plan

- [ ] CI green on unroll, chainsource, chainbackends, txconfirm, darepod unit suites
- [ ] CI green on systest (\`TestChainSourceConfReorgRoundTrip\`, \`TestTxConfirmReorgRoundTrip\`)
- [ ] \`make lint-native\` clean
- [ ] Reviewer sanity-check that the restart-reconciliation ordering invariant (reconcile runs before FSM session is bound; before block sub / spend watch / ResumeEvent reissue) holds in \`actor.go:ensureLoaded\`

---

- Closes lightninglabs/darepo-client#207 (unroller re-org awareness — full implementation lands in this PR)

### DIFF
--- a/chainsource/finality.go
+++ b/chainsource/finality.go
@@ -21,6 +21,16 @@ var finalityBlockSubscriptionBackoffs = []time.Duration{
 	2 * time.Second,
 }
 
+// finalityBlockSubscriptionAttemptTimeout bounds each individual
+// RegisterBlocks attempt. Without it a single hung RegisterBlocks call
+// (e.g. a wedged lndclient gRPC stream) would block the conf/spend
+// monitoring goroutine indefinitely — stalling Confirmed/Reorged/Done
+// delivery on that watch — since the retry schedule only bounds the gaps
+// between attempts, not the attempts themselves. 10s mirrors the per-call
+// registration timeout used in conf_actor.go's handleRegisterConf so the
+// whole file behaves consistently under a slow backend.
+const finalityBlockSubscriptionAttemptTimeout = 10 * time.Second
+
 // registerBlocksForFinality registers a block-epoch subscription used
 // to synthesize a Done signal at FinalityDepth past an observed
 // confirmation or spend. The call is retried with a short bounded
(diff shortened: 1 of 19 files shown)


levmi:
**Triage linkage check** — verified this PR's live-rollback reducer implements #207's full proposed solution (reorg notifications, roll back `ConfirmedTxids`, re-register conf watches, reset CSV-maturity baseline, backward state transition) and then some.

Suggest adding **`Closes #207`** so merge auto-closes it. (Heads-up: this is a draft stacked on a non-`main` base — the auto-close will only fire once it lands on the default branch.)

ellemouton:
Superseded by #897 as part of condensing the reorg-safety client stack (epic lightninglabs/darepo#454) from 12 PRs into 3. The commits are carried over unchanged; see #897. Branch retained as a backup.

PR #418, opened 2026-05-13
chainbackends: fix lnd fee rate units

bhandras:
## Summary

Fix the lndclient-backed fee estimator to preserve WalletKit's fee-rate units.

`lndclient.WalletKitClient.EstimateFeeRate` already returns a `chainfee.SatPerKWeight`. The adapter was treating that value as sat/vB and calling `FeePerKWeight()` on it, inflating the result by roughly 250x.

This exactly matches the signet symptom we saw while debugging boarding fees:

```text
WalletKit-ish rate:      31,774 sat/kW = 127 sat/vB
Buggy double conversion: 7,943,500 sat/kW = 31,774 sat/vB
Observed quote shape:    ~4.893M sat on-chain share
```

## Changes

- Return WalletKit's sat/kW value directly from `LndClientFeeEstimator.EstimateFeePerKW`.
- Keep the single sat/kW to sat/vB conversion at the `LNDBackend.EstimateFee` boundary.
- Add regression coverage for both boundaries so this cannot silently drift again.
- Log both sat/kW and sat/vB in the generic LND chain backend fee estimate log line.

## Tests

```text
go test ./chainbackends -count=1
```

### DIFF
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -119,6 +119,7 @@ func (b *LNDBackend) EstimateFee(ctx context.Context, targetConf uint32) (
 
 	b.logger(ctx).DebugS(ctx, "Fee rate estimated",
 		slog.Int("target_confs", int(targetConf)),
+		slog.Int64("sat_per_kw", int64(feePerKw)),
 		slog.Int64("sat_per_vbyte", int64(satPerVByte)),
 	)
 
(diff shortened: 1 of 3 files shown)


bhandras:
@claude pls review

bhandras:
Thanks for the review. CI is green and there are no open comments from Gemini. Could you please confirm whether this is ready to land?

@claude review

PR #422, opened 2026-05-13
multi: Reorg-aware chainsource + chainbackends + txconfirm

ellemouton:
## Summary

Foundation layer for reorg-safe chain observation. This PR makes the chain-event surface reversible end to end so downstream subsystems can observe a tx or spend moving through:

```
Observed -> Reorged -> Observed -> Done
```

For `txconfirm`, the equivalent lifecycle is:

```
TxConfirmed -> TxReorged -> TxConfirmed -> TxFinalized
```

The important semantic change is that the first positive observation is no longer terminal. A confirmation or spend can be reported, later leave the best chain, then be reported again. This is required for later client and server work where Ark may use a chain fact at low confirmation depth, likely one confirmation, while still remaining recoverable if that fact reorgs out.

## What `Done` / `Finalized` Means

`Done` and `TxFinalized` mean **policy finality at the configured `FinalityDepth`**, not absolute Bitcoin finality. They are not a claim that the tx can never be reorged under consensus.

The model is:

- `usability_depth`: when a higher layer is willing to act on a chain fact. For fast Ark UX this may be `1` confirmation.
- `FinalityDepth`: when this observation layer considers the watch complete enough to emit `Done` / `TxFinalized`. The production default wired here is `chainsource.DefaultFinalityDepth` (`6`).

So yes, we still care about reorgs after a one-confirmation observation. That is the main point of this PR: between first observation and policy finality, the watch stays alive and can emit `Reorged`, then a later positive observation again. Once `Done` / `TxFinalized` fires, low-level watchers may release resources. Later PRs will keep long-lived batch/VTXO meaning above this layer and can still reconcile or park on deeper-than-policy reorgs for still-relevant lineage.

## Why This PR Exists Before Batch/VTXO Reorg Safety

This PR intentionally does **not** decide whether a VTXO, batch, round, OOR receive, or unroll source is usable. It only observes chain facts.

Later PRs will add a `BatchCanonicalityManager` above `chainsource` that interprets these raw observations:

- batch tx confirms, reorgs out, reconfirms, or reaches policy finality;
- batch input outpoints are spent by conflicting txs;
- conflict spends reorg out or reach policy finality;
- effective batch expiry is recomputed when a batch reconfirms at a different height.

That later layer will own VTXO lineage states such as `available_provisional`, `limbo_reorg`, `limbo_conflict`, and `invalidated`. This PR is the observation substrate those states need.

## chainsource

- Conf/spend watch registrations now expose `Reorged` and `Done` channels alongside the existing `Confirmed`/`Spend` channel.
- New events: `ConfReorgedEvent`, `ConfDoneEvent`, `SpendReorgedEvent`, `SpendDoneEvent`.
- ConfActor and SpendActor are multi-shot: `Confirmed -> Reorged -> Confirmed -> Done` instead of terminate-after-first-event.
- `Done` is synthesized from configurable `FinalityDepth` height watermarks so backends that drop upstream Done signals, notably lndclient over gRPC, still complete the lifecycle.
- `RegisterBlocks` for finality synthesis is retried with bounded backoff before falling back to backend Done.

## chainbackends

- `LNDBackend` forwards lnd reorg and finality notifications into the chainsource event surface.
- `lndclient` adapter wires `WithReOrgChan` so the underlying gRPC stream is not torn down after the first event.

## txconfirm

- Confirmation watches are reorg-aware end to end.
- `TxConfirmed` is no longer terminal.
- New `TxReorged` and `TxFinalized` events.
- New `Finalized` terminal state.
- Reversible notifications (`TxConfirmed`, `TxReorged`, cached-confirmation replay) are fire-and-forget on bounded per-subscriber goroutines.
- Terminal notifications (`TxFinalized`, `TxFailed`) keep the reliable goroutine + timeout + idempotent-retry pattern.
- A backend `Done` for an entry not currently confirmed is logged and dropped rather than incorrectly promoting to `Finalized`.

## darepod

- Wires `chainsource.DefaultFinalityDepth` (`6`) into production `ChainSourceConfig`.
- This is required for lndclient-backed transports where finality synthesis is the practical `Done` source.

## Follow-Up Stack

This PR is the bottom of the reorg-safety stack.

Next PR (#410) adds unroll recovery-tx reorg handling:

- rollback of unroll proof/sweep anchors on reorg;
- provisional external-spend handling;
- restart reconciliation against the canonical chain;
- sweep-finality gating of `PhaseCompleted`.

Later PRs will add the batch/VTXO canonicality layer:

- long-lived batch tx and batch input watching;
- VTXO lineage availability and admission gates;
- round-born and OOR-received VTXO lineage registration;
- unroll source-lineage pause/resume;
- matching server-side canonicality and policy gates.

## Tests

- Unit tests for reorg-aware `chainsource` ConfActor and SpendActor lifecycles.
- Unit tests for height-based `Done` synthesis.
- `LNDBackend` forwarder reorg tests.
- Full `txconfirm` reorg lifecycle tests.
- `TestEnsureConfirmedDoneDuringReorgGapDropped` pins the Done-during-reorg-gap behavior.
- Systest coverage for chainsource and txconfirm reorg round trips.

## Test Plan

- [ ] CI green on chainsource, chainbackends, txconfirm, darepod unit suites
- [ ] `make lint-native` clean
- [ ] Reviewer sanity-check: `Done` / `TxFinalized` are policy-finality events at configured depth, not absolute consensus finality
- [ ] Reviewer sanity-check: this PR remains an observation layer only; batch/VTXO interpretation is intentionally deferred
- [ ] Reviewer sanity-check: bare reorg correlation IDs are intentional because gRPC-backed transports cannot reliably preserve richer reorg metadata

### DIFF
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -27,6 +27,20 @@ import (
 // unexported there and so must be matched by value.
 const lndNeutrinoBroadcastMsg = "broadcast-unverified"
 
+// reorgSignalBufferSize bounds the buffered reorg notifications forwarded
+// from lnd's notifier to a chainsource registration. The forwarder sends
+// to this channel with a blocking send, so a full buffer head-of-line
+// blocks delivery of unrelated Confirmed / Done events on the same
+// forwarder goroutine. A reorg burst (e.g. a multi-block reorg emitting
+// one NegativeConf per disconnected block) can produce several signals
+// back-to-back; sizing the buffer to absorb a typical reorg depth keeps
+// the forwarder moving. The signal is coalescing — the consumer
+// re-queries chain state on any reorg — so exact depth need not be
+// preserved; the buffer only needs to be deep enough to avoid stalling,
+// not to record every event. Eight comfortably covers realistic reorg
+// depths while staying negligible in memory.
+const reorgSignalBufferSize = 8
+
 // TxBroadcaster is a minimal interface for broadcasting transactions. This
 // allows LNDBackend to work with both lnwallet.WalletController (in-process
 // lnd) and lndclient wrappers (remote lnd via gRPC).
@@ -330,37 +344,98 @@ func (b *LNDBackend) RegisterConf(ctx context.Context, txid *chainhash.Hash,
 	// context is released.
 	notifyCtx, cancel := context.WithCancel(context.Background())
 
-	// Create a channel to convert lnd's TxConfirmation to our type.
+	// Create channels to convert lnd's confirmation lifecycle to our
+	// backend-agnostic types. NegativeConf carries a reorg depth that
+	// the lndclient gRPC transport cannot preserve, so we forward the
+	// forwarder's sequence number instead (see below).
 	confChan := make(chan *chainsource.TxConfirmation, 1)
+	reorgChan := make(chan uint64, reorgSignalBufferSize)
+	doneChan := make(chan struct{}, 1)
 
 	go func() {
+		// seq is a per-registration monotonic counter stamped onto
+		// every Confirmed and Reorged signal in the order this single
+		// forwarder goroutine observes them. Confirmed and Reorged
+		// leave on separate channels, so a select over both at any
+		// downstream hop (here, and again in the chainsource conf
+		// actor) cannot recover their order; the shared sequence lets
+		// the final consumer apply highest-seq-wins and discard a
+		// stale signal that lost a cross-channel race. This goroutine
+		// is the single authoritative ordering point — whatever order
+		// it reads lnd's channels in is the order the consumer honors.
+		var seq uint64
+		// Defers run in LIFO order. event.Cancel() must run first so
+		// the upstream notifier stops writing to its internal
+		// channels before we cancel notifyCtx (which any in-flight
+		// downstream sends are still using) and finally close the
+		// outgoing chans. Reversing this order would race the
+		// upstream notifier against closed channels.
 		defer close(confChan)
+		defer close(reorgChan)
+		defer close(doneChan)
 		defer cancel()
 		defer event.Cancel()
 
-		select {
-		case lndConf, ok := <-event.Confirmed:
-			if !ok {
+		for {
+			select {
+			case lndConf, ok := <-event.Confirmed:
+				if !ok {
+					return
+				}
+
+				seq++
+				conf := &chainsource.TxConfirmation{
+					BlockHash:   lndConf.BlockHash,
+					BlockHeight: lndConf.BlockHeight,
+					TxIndex:     lndConf.TxIndex,
+					Tx:          lndConf.Tx,
+					Block:       lndConf.Block,
+					Seq:         seq,
+				}
+
+				select {
+				case confChan <- conf:
+				case <-notifyCtx.Done():
+					return
+				}
+
+			case _, ok := <-event.NegativeConf:
+				if !ok {
+					event.NegativeConf = nil
+					continue
+				}
+
+				seq++
+				select {
+				case reorgChan <- seq:
+				case <-notifyCtx.Done():
+					return
+				}
+
+			case _, ok := <-event.Done:
+				if !ok {
+					event.Done = nil
+					continue
+				}
+
+				select {
+				case doneChan <- struct{}{}:
+				case <-notifyCtx.Done():
+					return
+				}
+
+				return
+
+			case <-notifyCtx.Done():
 				return
 			}
-
-			conf := &chainsource.TxConfirmation{
-				BlockHash:   lndConf.BlockHash,
-				BlockHeight: lndConf.BlockHeight,
-				TxIndex:     lndConf.TxIndex,
-				Tx:          lndConf.Tx,
-				Block:       lndConf.Block,
-			}
-
-			confChan <- conf
-
-		case <-notifyCtx.Done():
-			return
 		}
 	}()
 
 	return &chainsource.ConfRegistration{
 		Confirmed: confChan,
+		Reorged:   reorgChan,
+		Done:      doneChan,
 		Cancel: func() {
 			cancel()
 			event.Cancel()
@@ -393,39 +468,97 @@ func (b *LNDBackend) RegisterSpend(ctx context.Context, outpoint *wire.OutPoint,
 	// Keep spend delivery alive independently of the actor request context.
 	notifyCtx, cancel := context.WithCancel(context.Background())
 
-	// Create a channel to convert lnd's SpendDetail to our type.
+	// Create channels to convert lnd's spend lifecycle to our
+	// backend-agnostic types. lnd's spend Reorg carries no payload, so
+	// we forward the forwarder's sequence number instead (see below).
 	spendChan := make(chan *chainsource.SpendDetail, 1)
+	reorgChan := make(chan uint64, reorgSignalBufferSize)
+	doneChan := make(chan struct{}, 1)
 
-	// Start a goroutine to convert and forward the spend.
+	// Start a goroutine to convert and forward spend lifecycle events.
 	go func() {
+		// seq is a per-registration monotonic counter stamped onto
+		// every Spend and Reorged signal in the order this single
+		// forwarder observes them. Spend and Reorged leave on separate
+		// channels, so a downstream select cannot recover their order;
+		// the shared sequence lets the final consumer apply
+		// highest-seq-wins and discard a stale signal that lost a
+		// cross-channel race. This goroutine is the single
+		// authoritative ordering point.
+		var seq uint64
+		// Defer order is LIFO: event.Cancel() stops the upstream
+		// notifier first, then cancel() ends notifyCtx, and only
+		// then are the outgoing channels closed. This avoids a race
+		// where the upstream notifier writes to a freshly-closed
+		// downstream channel.
 		defer close(spendChan)
+		defer close(reorgChan)
+		defer close(doneChan)
 		defer cancel()
 		defer event.Cancel()
 
-		select {
-		case lndSpend, ok := <-event.Spend:
-			if !ok {
+		for {
+			select {
+			case lndSpend, ok := <-event.Spend:
+				if !ok {
+					return
+				}
+
+				// Convert to our type.
+				seq++
+				spend := &chainsource.SpendDetail{
+					SpentOutPoint: lndSpend.SpentOutPoint,
+					SpenderTxHash: lndSpend.SpenderTxHash,
+					SpendingTx:    lndSpend.SpendingTx,
+					SpenderInputIndex: lndSpend.
+						SpenderInputIndex,
+					SpendingHeight: lndSpend.SpendingHeight,
+					Seq:            seq,
+				}
+
+				select {
+				case spendChan <- spend:
+				case <-notifyCtx.Done():
+					return
+				}
+
+			case _, ok := <-event.Reorg:
+				if !ok {
+					event.Reorg = nil
+					continue
+				}
+
+				seq++
+				select {
+				case reorgChan <- seq:
+				case <-notifyCtx.Done():
+					return
+				}
+
+			case _, ok := <-event.Done:
+				if !ok {
+					event.Done = nil
+					continue
+				}
+
+				select {
+				case doneChan <- struct{}{}:
+				case <-notifyCtx.Done():
+					return
+				}
+
+				return
+
+			case <-notifyCtx.Done():
 				return
 			}
-
-			// Convert to our type.
-			spend := &chainsource.SpendDetail{
-				SpentOutPoint:     lndSpend.SpentOutPoint,
-				SpenderTxHash:     lndSpend.SpenderTxHash,
-				SpendingTx:        lndSpend.SpendingTx,
-				SpenderInputIndex: lndSpend.SpenderInputIndex,
-				SpendingHeight:    lndSpend.SpendingHeight,
-			}
-
-			spendChan <- spend
-
-		case <-notifyCtx.Done():
-			return
 		}
 	}()
 
 	return &chainsource.SpendRegistration{
-		Spend: spendChan,
+		Spend:   spendChan,
+		Reorged: reorgChan,
+		Done:    doneChan,
 		Cancel: func() {
 			cancel()
 			event.Cancel()
--- a/chainbackends/lndclient_adapters.go
+++ b/chainbackends/lndclient_adapters.go
@@ -17,6 +17,16 @@ import (
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 )
 
+// lndRegistrationTimeout bounds how long a conf/spend registration call into
+// lndclient may block before we give up and return an error. lnd can be slow
+// to answer a RegisterConfirmationsNtfn / RegisterSpendNtfn while it is busy
+// processing a fresh block, and a registration that hangs forever would pin
+// the chainsource sub-actor's Receive call and back-pressure the factory
+// actor onto every other in-flight registration. Fifteen seconds is well
+// above lnd's normal under-load response time yet short enough that a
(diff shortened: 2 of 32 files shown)


ellemouton:
@claude review this

ellemouton:
@claude re review

litbot-9000:

chainbackends/lndclient_adapters.go
```diff
@@ -236,10 +255,44 @@ func (n *LndClientChainNotifier) RegisterConfirmationsNtfn(txid *chainhash.Hash,
 		}
 	}()
 
+	// Bridge lndclient's payload-less reorg ping onto the chainntnfs
+	// NegativeConf channel that downstream code consumes. lndclient does
```
Roasbeef: Should we fix this? So `lndclient` forwarding the true information?

chainbackends/lndclient_adapters.go
```diff
@@ -248,14 +301,50 @@ func (n *LndClientChainNotifier) RegisterConfirmationsNtfn(txid *chainhash.Hash,
 func (n *LndClientChainNotifier) RegisterSpendNtfn(outpoint *wire.OutPoint,
 	pkScript []byte, heightHint uint32) (*chainntnfs.SpendEvent, error) {
 
+	// Ask lndclient to keep the spend stream alive past the first
```
Roasbeef, line 318: Another area we should fix lndclient?

chainsource/conf_actor.go
```diff
@@ -275,17 +311,47 @@ func (a *ConfActor) monitorConfirmation() {
...
+			}
+
 		case _, ok := <-a.registration.Reorged:
 			if !ok {
 				a.registration.Reorged = nil
 				continue
 			}
 
 			a.deliverConfReorged(lastEvent)
+			// The previous confirmation is no longer on the
```
Roasbeef, line 388: Style nit: extra newline above

chainsource/conf_actor.go
```diff
@@ -22,6 +22,20 @@ type ConfActorConfig struct {
...
+	// height-based finality synthesis entirely; in that case the actor
+	// only fires ConfDoneEvent when the backend's own Done channel
+	// fires (e.g. an in-process lnd notifier). Non-zero values close
+	// the lndclient transport gap, where the gRPC layer does not
+	// surface lnd's internal "past reorg-safety depth" signal.
+	//
+	// The depth is counted inclusively (a tx confirmed at height H is
+	// at depth 1; height-based finality fires once the actor observes
+	// a block at H + FinalityDepth - 1). The conventional choice is 6.
+	FinalityDepth uint32
```
Roasbeef, line 38: > The lndclient gRPC transport does not surface lnd's internal "past
reorg-safety depth" signal: ConfRegistration.Done is allocated but
never written to over that transport. Consumers (txconfirm and the
unroll actor) that gate terminal eviction on Done can therefore
retain per-tx state indefinitely after a confirmation in production
deployments.

Given this, do we need that extra logic in the earlier commits? What if we updated `lndclient` to continue to deliver at this point?

chainsource/conf_actor.go
```diff
@@ -215,45 +257,153 @@ func (a *ConfActor) monitorConfirmation() {
...
+			if !ok {
+				a.registration.Reorged = nil
+				continue
+			}
+
+			a.deliverConfReorged(lastEvent)
+			// The previous confirmation is no longer on the
+			// canonical chain; reset the depth counter so the
+			// next re-confirmation starts a fresh window.
+			a.confirmHeight = 0
```
Roasbeef, line 394: Possible that with rapid fire blocks (also depends on esplora behavior if it'll notify each block or just the new one) that there might be stale blocks in the epoch. So we should drain them here.

chainbackends/lnd.go
```diff
@@ -305,37 +305,84 @@ func (b *LNDBackend) RegisterConf(ctx context.Context, txid *chainhash.Hash,
 	// context is released.
 	notifyCtx, cancel := context.WithCancel(context.Background())
 
-	// Create a channel to convert lnd's TxConfirmation to our type.
+	// Create channels to convert lnd's confirmation lifecycle to our
+	// backend-agnostic types. NegativeConf carries a reorg depth that
+	// the lndclient gRPC transport cannot preserve, so we drop the value
+	// and forward a bare struct{} signal.
 	confChan := make(chan *chainsource.TxConfirmation, 1)
+	reorgChan := make(chan struct{}, 1)
```
Roasbeef: Belt and suspenders increase the buffer here to hadnle repid reogs?

chainbackends/lndclient_adapters.go
ellemouton: Agreed this is the right long-term fix — the lndclient gRPC transport drops the `NegativeConf` reorg depth, so we forward a bare `struct{}`. The height-based finality synthesis in this PR is the client-side workaround that needs no lnd/lndclient release. I'll file a follow-up to enrich lndclient to forward the true reorg info; once it does, the synthesis becomes redundant and can be disabled via `FinalityDepth=0`.

chainbackends/lndclient_adapters.go
ellemouton: Same as above — tracking an lndclient follow-up to forward the real signal here; the bare `struct{}` is the transport-limited stopgap until then.

chainsource/conf_actor.go
ellemouton: Already gone — gofmt collapsed it during the rebase; no double blank remains in that select.

chainsource/conf_actor.go
ellemouton: The synthesis is the pragmatic fix that works against today's lndclient without an lnd release. Updating lndclient to deliver `Done` over gRPC is the cleaner long-term path, but it's a separate lnd/lndclient change + release cycle.

I kept the synthesis (gated by `FinalityDepth`, default 6) so the gap is closed now, and it's deliberately a no-op when `FinalityDepth=0` — so once lndclient delivers `Done` natively we just set `FinalityDepth=0` and the synthesis path goes dormant with no code removal. Filing the lndclient follow-up.

chainsource/conf_actor.go
ellemouton: Done — added `drainToLatestEpoch`, which non-blockingly coalesces any epochs already queued behind the one we dequeued and evaluates finality against the most recent height only (in both `ConfActor` and `SpendActor`). It also reports a mid-drain channel close so the caller parks its receive. Added unit tests for coalescing, the empty case, nil-skip, and the closed-channel signal.

chainbackends/lnd.go
ellemouton: Done — the reorg channel is now sized from a documented named const (`reorgSignalBufferSize = 8`) on both the conf and spend forwarders. The forwarder's send is blocking, so the deeper buffer keeps a reorg burst (one `NegativeConf` per disconnected block) from head-of-line-blocking unrelated `Confirmed`/`Done` deliveries on the same goroutine; the signal is coalescing so exact depth need not be preserved.

txconfirm/actor.go
```diff
@@ -1124,21 +1320,67 @@ func (a *TxBroadcasterActor) handleTerminalNotifyResult(ctx context.Context,
...
+			BlockHeight: blockHeight,
+			NumConfs:    numConfs,
+		})
+		if err != nil {
+			a.log.WarnS(ctx, "Failed to deliver TxConfirmed",
+				err, "txid", entry.data.Txid,
+				"subscriber", subscriber.ID())
+		}
+	}
+}
```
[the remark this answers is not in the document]
ellemouton: Already addressed on the branch: reversible deliveries (`notifyConfirmed`/`notifyReorged`) go through `notifyReversibleAsync` — a per-subscriber goroutine bounded by `reversibleNotifyTimeout` — and the legacy/replay path (`attachExistingSubscriber`) uses `notifyOneTerminal` (goroutine + timeout + async completion). No subscriber `Tell` runs on the actor's synchronous path, so a slow or stuck subscriber can't pin the loop. Dropped reversible events are recovered by the next lifecycle transition. This is codified as a package invariant in the txconfirm CLAUDE.md.

chainsource/conf_actor.go
```diff
@@ -215,45 +257,138 @@ func (a *ConfActor) monitorConfirmation() {
...
+				if err != nil {
+					log.WarnS(a.ctx, "Failed to register "+
+						"block epoch subscription "+
+						"for height-based finality",
+						err)
+				} else {
+					a.blockReg = reg
+					blockEpochs = reg.Epochs
+				}
+			}
```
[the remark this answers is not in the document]
ellemouton: Addressed by `registerBlocksForFinality` (finality.go), which retries the block-epoch subscription with a bounded backoff (3 attempts over ~2.6s) specifically to avoid leaking the sub-actor on a transient registration failure. After exhaustion it logs at WARN and falls back to the backend's own `Done`; the residual permanent-failure case is operator-visible and can grow a re-arm path later if it proves real.

chainsource/spend_actor.go
```diff
@@ -215,13 +262,82 @@ func (a *SpendActor) monitorSpend() {
...
+				if err != nil {
+					log.WarnS(a.ctx, "Failed to register "+
+						"block epoch subscription "+
+						"for height-based finality",
+						err)
+				} else {
+					a.blockReg = reg
+					blockEpochs = reg.Epochs
+				}
+			}
```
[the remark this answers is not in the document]
ellemouton: Same as the `ConfActor` — the `SpendActor` arms via the shared `registerBlocksForFinality` bounded-retry helper, so a transient block-subscription failure no longer leaks the sub-actor.

ellemouton:
Addressed the automated review in 5321ad4 + bc869df (pushed to `reorg-safe-chainsource`).

**🔴 Cross-channel ordering hazard at the lndclient bridge** — fixed. lndclient drives the whole confirmation lifecycle off a single ordered gRPC receive loop, but the adapter split it across two channels (the confirmation channel and the `WithReOrgChan` ping) and forwarded each on its own goroutine. That let the downstream `ConfActor` `select` consume a re-`Confirmed` before the `Reorged`: it would set `confirmHeight` to the new height, a later `Reorged` would reset it to `0`, and with no further `Confirmed` coming the height-based `Done` synthesis never fired — leaking the watch + txconfirm FSM for the lifetime of the daemon.

Fix: forward the whole lifecycle through a **single** goroutine (`forwardOrderedReorg` / `forwardOrderedSpendReorg`) that drains a pending reorg with priority before any confirmation and hands every event off with a blocking send. The single downstream consumer now observes events strictly in lndclient's emission order (reorg before re-confirmation), closing the race. The spend path had the identical two-goroutine split and gets the same single-forwarder treatment.

**🟡 Unbounded `RegisterBlocks` attempt in finality synthesis** — fixed. `registerBlocksForFinality` bounded the backoff *between* attempts but ran each `RegisterBlocks` on the actor root context with no per-call timeout, so a wedged lndclient stream could block the monitoring goroutine indefinitely. Each attempt is now wrapped in a bounded context (`finalityBlockSubscriptionAttemptTimeout`, 10s, mirroring `handleRegisterConf`). The subscription's own lifetime is unaffected — the adapter derives the stream from its own background context, so cancelling the attempt context only bounds the registration call.

**Verification:**
- New deterministic unit tests (`lndclient_ordering_test.go`) pre-load both source channels and assert the reorg is delivered before the confirmation/spend.
- `TestChainSourceConfReorgRoundTrip` + `TestTxConfirmReorgRoundTrip` reorg systests still pass against real bitcoind+LND.
- `build` / `vet` / `fmt` / `lint-changed-local` clean.

ellemouton:
Addressed the second review pass (the 🔴 ordering hazards + the medium/low findings). Pushed 6 commits to `reorg-safe-chainsource`.

### The ordering fix is now correct for *both* interleavings (`101168f`)

You were right that the priority-drain only papered over the common case. I worked through the two interleavings for a single watched tx:
- **reorg → re-confirm** (common): reorg-first is correct.
- **confirm → reorg-of-that-block, no re-confirm** (rare): reorg-first is *wrong* — it leaves a stale confirm winning and synthesizes a false `Done` for a tx that's gone.

"Reorg-first" is a heuristic, not ordering. And it can't be fixed at any single forwarder hop because `Confirmed`/`Reorged` are separate channels re-split at every hop. So instead I **stamp a per-registration monotonic sequence at the single `LNDBackend` forwarder** (the one place that sees the whole lifecycle in arrival order) onto `TxConfirmation`/`SpendDetail` and the `Reorged` signal, and the `ConfActor`/`SpendActor` apply **highest-seq-wins**, discarding any event whose seq doesn't exceed the highest applied. Delivery interleaving downstream is now irrelevant — both cases resolve to the highest-sequence outcome. `Seq 0` = a backend that never reorgs (always-apply), preserving existing behavior. Added deterministic actor tests for both interleavings (stale-reorg-discarded, stale-confirm-discarded). The lndclient adapter forwarders no longer inject a reorg-first bias.

This makes the consumer the authoritative ordering point, so the in-process `lnd.go` forwarder finding is covered too — no per-hop priority-drain needed.

### Other findings

- **🔴 `notifyReorged` before/without initial `TxConfirmed`** (`02b1f52`): skip `pendingConfirmed` subscribers in `notifyReorged` — they get live state on their eventual reliable initial delivery.
- **🟠 reversible delivery trailing terminal** (`02b1f52`): added a per-entry `sealed` flag set when terminal delivery begins; each fire-and-forget reversible goroutine checks it immediately before its `Tell` and skips, so a reversible spawned just before finality can't resurrect released bookkeeping.
- **🟠 `registerBlocksForFinality` blocking the select loop** (`b17729f`): arming now runs in a short-lived goroutine that hands the registration back on a channel (or cancels it if the actor exits first), gated so at most one armer runs; the loop stays responsive to Reorged/Done/ctx.Done during the ~30s retry window.
- **🟡 spend actor missing the single-shot guard** (`ce5ef78`): added the `reorgAware` gate so a non-reorg-aware actor-mode spend watch exits after the first spend (matching ConfActor + the documented contract) and doesn't arm an unwanted block sub.
- **🟡 conf-watch leak on terminal `OnStop`** (`02b1f52`): terminal entries now unregister their watch too, guarded by `confWatchRegistered`.
- **🟡 reorg-only watch torn down at finality** (`3eb6f61`): documented as intended (past safety depth there's nothing left to observe).
- **🟡 boarding-sweep `MarkBoardingSweepFailed`** (`3ae1bb5`): added a no-op-when-already-resolved guard so a spurious `TxFailed` can't roll back a confirmed sweep's irreversible ledger legs.

The stale-`lastEvent`-after-reorg nit is folded into the seq commit (lastEvent is cleared on reorg).

**Verification:** new unit tests for the seq ordering (both interleavings) + the boarding-sweep guard; `chainsource`/`txconfirm` race suites, `chainbackends`/`wallet` suites all green; `build`/`vet`/`fmt`/`lint-changed-local` clean; commit messages lint-clean.

ellemouton:
@claude please re-review. The 6 new commits (`bc869df..3eb6f61c`) address your second review pass:

- The 🔴 cross-channel ordering hazard is now fixed at the root with **seq-stamped lifecycle ordering** (per-registration monotonic sequence stamped at the `LNDBackend` forwarder; conf/spend actors apply highest-seq-wins), replacing the reorg-first heuristic — correct for both reorg-then-reconfirm and confirm-then-reorg.
- 🔴 `notifyReorged` now skips `pendingConfirmed` subscribers.
- 🟠 per-entry `sealed` flag stops reversible deliveries trailing the terminal; finality arming moved off the select loop.
- 🟡 spend-actor single-shot guard, terminal conf-watch unregister on stop, reorg-only-watch teardown doc, and the boarding-sweep already-resolved guard.

Please focus on the seq-ordering approach in `101168f` (is highest-seq-wins sound given the forwarder is the single ordering point?) and the `sealed`-flag race window in `02b1f52`.

ellemouton:

PR #423, opened 2026-05-13
wallet+darepod: persist Board RPC, replay on restart (#416)

### LINKED ISSUE #416: non-persistent round watching
A `darepod` restart forgets boarding sessions.

```
darepocli board --no-tls                                                                                                                                                                                                                                                                              
{                                                                                                                                                                                                                                                                                                                                                     
  "status":  "registered",                                                                                                                                                                                                                                                                                                                            
  "vtxo_count":  1                                                                                                                                                                                                                                                                                                                                    
}                                               
```

```
darepocli rounds watch --no-tls                                                                                                                                                                                                                                                                       
{                                                                                                                                                                                                                                                                                                                                                     
  "round":  {                                                                                                                                                                                                                                                                                                                                         
    "round_id":  "019e20b3-2956-7c6f-a6df-d81e06dc0864",                                                                                                                                                                                                                                                                                              
    "state":  "ROUND_STATE_REGISTRATION_SENT",                                                                                                                                                                                                                                                                                                        
    "is_temp":  false,                                                                                                                                                                                                                                                                                                                                
    "vtxos":  [],                                                                                                                                                                                                                                                                                                                                     
    "commitment_txid":  "",                                                                                                                                                                                                                                                                                                                           
    "commitment_height":  0,                                                                                                                                                                                                                                                                                                                          
    "creation_time":  "0",                                                                                                                                                                                                                                                                                                                            
    "last_update_time":  "0",                                                                                                                                                                                                                                                                                                                         
    "input_outpoints":  [],                                                                                                                                                                                                                                                                                                                           
    "output_outpoints":  [],                                                                                                                                                                                                                                                                                                                          
    "failure_reason":  ""
  }
}
```

RESTART daemon:
```
darepocli rounds watch --no-tls
```


```
2026-05-13 09:43:55.067 [INF] ROND: Loaded active rounds from database count=0
2026-05-13 09:43:55.067 [INF] ROND: Round client actor started
2026-05-13 09:43:55.067 [INF] DRPD: Round actor registered and started
2026-05-13 09:43:55.067 [DBG] ACTR: Actor registered with system actor_id=txconfirm service_key=txconfirm
2026-05-13 09:43:55.067 [INF] DRPD: Unroll subsystem initialized
2026-05-13 09:43:55.067 [DBG] ACTR: Actor registered with system actor_id=oor-timeout service_key=oor-timeout
2026-05-13 09:43:55.067 [DBG] OORC: Starting durable actor actor_id=oor-signing-effect
2026-05-13 09:43:55.067 [INF] OORC: OOR signing effect actor started actor_id=oor-signing-effect
2026-05-13 09:43:55.067 [INF] OORC: Creating OOR client actor actor_id=oor-client
2026-05-13 09:43:55.069 [DBG] OORC: Starting durable actor actor_id=oor-client
2026-05-13 09:43:55.069 [INF] OORC: OOR durable actor started actor_id=oor-client
2026-05-13 09:43:55.069 [INF] OORC: OOR actor registered with receptionist actor_id=oor-client
2026-05-13 09:43:55.069 [INF] DRPD: OOR client actor started
2026-05-13 09:43:55.069 [DBG] ACTR: Actor registered with system actor_id=incoming-vtxo-handler service_key=incoming-vtxo-handler
2026-05-13 09:43:55.069 [INF] DRPD: Incoming VTXO handler started
2026-05-13 09:43:55.069 [INF] DRPD: Wallet-dependent actors started
2026-05-13 09:43:55.069 [INF] SVRC: Loaded ack checkpoint actor_id=serverconn-02499e33f5fa270606ec54895c3f210e0421b33af53c1244a0ea427b980cd0c11e pull_cursor=1927 dispatch_committed_to=1927 ack_target=1927 ack_committed_to=1927
2026-05-13 09:43:55.069 [INF] SVRC: Ingress loop starting mailbox_id=02499e33f5fa270606ec54895c3f210e0421b33af53c1244a0ea427b980cd0c11e
2026-05-13 09:43:55.076 [INF] OORC: Handling restart message has_checkpoint=true
2026-05-13 09:43:55.076 [INF] OORC: Restoring sessions from checkpoint checkpoint_version=2 num_outgoing_snapshots=0 num_incoming_snapshots=0
2026-05-13 09:43:55.076 [INF] OORC: Resuming restored sessions num_sessions=0
2026-05-13 09:43:55.076 [INF] OORC: Restart complete num_sessions=0
```

Roasbeef:
In this PR, we fix [#416](https://github.com/lightninglabs/darepo-client/issues/416): a `darepocli board` call that returned `registered` was silently dropped on daemon restart. The user saw a confirmed boarding balance, the CLI happily reported the request was accepted, and the next `darepocli rounds watch` came back empty because the round actor only checkpoints state at the InputSigSent point of no return. Every prior step lived in memory, so a restart between admission and seal wiped the in-flight registration and the user's intent with it.

The shape of the fix is to persist the user's Board intent (not the round FSM) and replay it on the next start. A single-row `pending_board_requests` table records the request as soon as `handleBoard` admits it, and the daemon's startup path asks the wallet to re-issue `TriggerBoardMsg` once the round actor is registered. When confirmed boarding intents have already transitioned to Adopted/Swept (the prior round actually completed before restart), the wallet clears the row instead of replaying. See each commit message for a detailed description w.r.t the incremental changes.

This PR pairs with [lightninglabs/darepo#XXX](https://github.com/lightninglabs/darepo) which teaches the server's `IntentCollectingState` to treat a same-client re-join as a replacement rather than rejecting it as `client already registered`. Both sides are needed: the client-side replay alone bounces off the operator while the pre-restart registration is still alive in the round's IntentCollectingState, and the server-side replacement alone doesn't help if the client never reissues. The integration test in the paired darepo PR drives the full restart-recovery scenario end to end.

## Ordering invariant

The single subtle thing in `handleBoard` is the order: we persist the row _before_ telling the round actor, not after. The persist-first ordering closes the only window where a crash could leave the round actor holding an in-memory intent with no on-disk marker. Under the new ordering, every crash window converges: pre-persist crashes are a no-op, post-persist pre-Tell crashes replay on next start, post-Tell crashes also replay on next start, and a Tell failure after persist leaves the row in place so the next startup retries. If persistence itself fails we return the error rather than silently admitting an unrecoverable request.

## Replay through `handleBoard` (not a synthetic Tell)

The replay path re-issues `BoardRequest` through the same `handleBoard` entry that the live RPC uses, rather than synthesising a `TriggerBoardMsg` directly. Two reasons. First, this keeps the persistence write inside the same handler, so a replay followed by another crash before seal leaves a fresh pending row with an updated `requested_at_unix` timestamp. Second, any future invariant added to `handleBoard` (extra pre-flight checks, side effects, etc.) automatically applies to the recovery path without a second integration point.

## Schema choice

The migration uses `PRIMARY KEY (id)` + `CHECK (id = 1)` in a separate clause rather than the inline `INTEGER PRIMARY KEY` form. The SQLite -> Postgres replacer (in `db/migrate/migrations.go`) rewrites the inline form to `BIGSERIAL PRIMARY KEY`, which would otherwise allocate an unused sequence under Postgres. We declare the constraint explicitly to side-step that.

## Test plan

- [x] `make unit pkg=./db case=TestPendingBoardRequestRoundTrip` pins down the store contract (empty -> None, upsert durable, second upsert replaces, clear is idempotent).
- [x] `make unit pkg=./wallet ./db ./darepod` (all green, no regressions).
- [x] `make lint-native` clean.
- [x] Restart-recovery itest in the paired darepo PR passes end to end (board -> restart -> replay observable -> manual `TriggerBatch` -> confirmed VTXO).
- [x] Existing restart itests (`TestBoardingIntegrationRestartAfterRoundBroadcast`, `TestBoardingIntegrationRestartAfterInputSigSent`) still pass.

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -1798,6 +1798,22 @@ func (s *Server) startWalletDependentActors(ctx context.Context,
 	}
 
 	// -------------------------------------------------------
+	// 13b. Replay any persisted Board RPC the user issued before the
+	//      last shutdown. Like resumeBoardingSweeps, this Ask MUST run
+	//      AFTER the round-client actor has registered with the
+	//      receptionist (step 11) — the replay's downstream
+	//      TriggerBoardMsg dispatch goes through the actor system,
+	//      and a Tell against an unresolved service key is a silent
+	//      drop. Driving the replay from wallet.Ark.Start would race
+	//      the registration and leave the recovered Board orphaned.
+	// -------------------------------------------------------
+	if err := s.replayPendingBoardRequest(ctx, walletRef); err != nil {
+		s.log.WarnS(ctx, "Failed to replay pending Board request",
+			err,
+		)
+	}
+
+	// -------------------------------------------------------
 	// 14. Register the OOR client actor.
 	// -------------------------------------------------------
 	if err := s.initOORActor(ctx, vtxoManagerRef); err != nil {
@@ -1805,6 +1821,28 @@ func (s *Server) startWalletDependentActors(ctx context.Context,
 	}
 
 	s.log.InfoS(ctx, "Wallet-dependent actors started")
+
+	return nil
+}
+
+// replayPendingBoardRequest Asks the wallet actor to replay any
+// persisted Board RPC across daemon restart. Called once during
+// startup, after the round-client actor has registered with the
+// receptionist, so the wallet's handleBoard can resolve the round
+// actor via the service-key router without racing.
+//
+// A failure here does not block daemon startup: a fresh Board RPC by
+// the user overwrites the pending rows, and a future restart re-tries
+// the replay. Returning the error lets the caller decide whether to
+// surface it.
+func (s *Server) replayPendingBoardRequest(ctx context.Context,
+	walletRef actor.ActorRef[wallet.WalletMsg, wallet.WalletResp]) error {
+
+	future := walletRef.Ask(ctx, &wallet.ReplayPendingBoardRequest{})
+	result := future.Await(ctx)
+	if result.IsErr() {
+		return fmt.Errorf("ask replay pending board: %w", result.Err())
+	}
 
 	return nil
 }
@@ -3042,6 +3080,7 @@ func (s *Server) initWalletActor(ctx context.Context,
 		wallet.WithBoardingSweep(
 			s.boardingSweepStore, sweepSigner, s.chainParams,
 		),
+		wallet.WithClock(s.clk),
 	)
 	walletKey := actor.NewServiceKey[
 		wallet.WalletMsg, wallet.WalletResp,
--- a/db/sqlc/migrations/000013_pending_board_request.up.sql
+++ b/db/sqlc/migrations/000013_pending_board_request.up.sql
@@ -1,0 +1,31 @@
+-- Records the user's explicit Board RPC intent so a daemon restart between
+-- Board admission and round seal does not silently drop the request.
+--
+-- Rows are keyed by the confirmed boarding outpoint (outpoint_hash,
+-- outpoint_index) that the original Board call admitted: one row per intent
+-- the call covered. This binds each pending request to the specific UTXOs
+-- the user pointed at, so a replay can never rebind a stale target_vtxo_count
+-- to a fresh, unrelated boarding deposit. Rows are cleared when the matching
+-- boarding intent transitions out of Confirmed (e.g. via the round-state
+-- checkpoint that flips the intent to Adopted), so the cleanup runs in the
+-- same SQL transaction as the state change rather than via a cross-actor
+-- callback.
+--
+-- target_vtxo_count and requested_at_unix are denormalised onto every row
+-- belonging to the same Board call. A subsequent Board call upserts the
+-- value of target_vtxo_count for every still-Confirmed outpoint; rows for
+-- already-adopted outpoints are unaffected because they have already been
+-- cleared.
+CREATE TABLE IF NOT EXISTS pending_board_requests (
+    outpoint_hash BLOB NOT NULL,
+    outpoint_index INTEGER NOT NULL,
+
+    target_vtxo_count INTEGER NOT NULL DEFAULT 0,
+
+    requested_at_unix BIGINT NOT NULL,
+
+    PRIMARY KEY (outpoint_hash, outpoint_index),
+
+    CHECK (target_vtxo_count >= 0),
+    CHECK (requested_at_unix > 0)
+);
--- a/wallet/messages.go
+++ b/wallet/messages.go
@@ -531,6 +531,15 @@ type BoardRequest struct {
 	// TargetVTXOCount is the requested number of boarded VTXOs. Zero means
 	// one output, preserving the legacy single-VTXO board behavior.
 	TargetVTXOCount uint32
+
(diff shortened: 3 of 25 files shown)


Roasbeef:
@claude review this

@codex review this

db/sqlc/migrations/000013_pending_board_request.up.sql
```diff
@@ -0,0 +1,22 @@
+-- Records the user's explicit Board RPC intent so a daemon restart between
+-- Board admission and round seal does not silently drop the request.
+-- Single-row semantics: id is fixed to 1 so the latest Board call always
+-- overwrites the prior one. The wallet writes a row from handleBoard once
+-- TriggerBoardMsg has been accepted by the round actor, and re-reads on
+-- startup; if any confirmed boarding intents remain the wallet re-issues
+-- TriggerBoardMsg before the daemon goes hot.
+--
+-- Note: we declare PRIMARY KEY separately rather than via
+-- `INTEGER PRIMARY KEY` to side-step the SQLite→Postgres replacer that
```
Roasbeef: Why is this needed at all? No other tables need to declare primary keys like this.

wallet/messages.go
```diff
@@ -568,6 +568,60 @@ func (m *BoardResponse) MessageType() string {
...
+func (m *ReplayPendingBoardRequest) walletMsgSealed() {}
+
+// ReplayPendingBoardResponse reports what the replay path observed: whether a
+// row was found, whether confirmed boarding intents were available to board,
+// and whether TriggerBoardMsg was actually re-dispatched. This lets the
+// daemon log a precise restart-recovery summary without a second DB read.
+type ReplayPendingBoardResponse struct {
+	actor.BaseMessage
+
+	// Found is true iff a row was present in pending_board_requests when
```
Roasbeef: Why have these two bools at all? If no pending requests were found, then we simply don't need to send the message.

darepod/server.go
```diff
@@ -1833,6 +1845,59 @@ func (s *Server) resumeBoardingSweeps(ctx context.Context,
...
 
+// replayPendingBoardRequest Asks the wallet actor to re-issue any persisted
+// Board RPC after restart. Called once during startup, after the round actor
+// has been registered, so the wallet's TriggerBoardMsg resolves a live
+// round-actor ref via the service-key router.
+//
+// A failure here does not block daemon startup: the next user-issued Board
+// RPC will replay the row, and the round actor's existing handleTriggerBoard
+// is idempotent against rejoining the same assembling round.
+func (s *Server) replayPendingBoardRequest(ctx context.Context,
```
Roasbeef: Simpler to either do this at start up in the wallet, or to do a self tell.

Roasbeef:
@claude

> Simplest fix: clear the row immediately after a successful handleBoard call in handleReplayPendingBoardRequest. The downside (daemon crashes post-replay, pre-seal → user re-issues) is the same behaviour as today (pre-PR). Clearing on replay is strictly better than the status quo for all cases except crash-during-round-after-replay.

We _do_ want the behavior here where on restart if a user wanted to board then it happens. Perhaps a middle ground is adding a new field/bool to the RPC/CLI, so then users can opt into persistent boarding.

Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #424, opened 2026-05-14
[codex] unroll: use conservative proof watch height

ellemouton:
## Summary

Use the earliest chain height as the confirmation height hint for unroll proof-graph transactions and deferred checkpoint watches. The restart reissue path routes through the same proof-node confirmation helper, so reattached watches use the same conservative hint.

Target spend watches and final sweep confirmation watches keep their existing tighter hints.

## Root Cause

Fraud-triggered unroll can discover proof transactions after another actor or test flow has already mined them. The previous current-height hint caused lnd confirmation subscriptions to miss historical confirmations, which could strand the unroll actor waiting for callbacks that never arrived.

A target descriptor creation height is also not conservative enough for every proof node, because roots and intermediate OOR checkpoint ancestors can confirm before the target VTXO is created.

## Validation

- `go test ./unroll`
- `make fmt-changed`
- `make lint-changed-local`
- `make itest icase=TestRecipientFraudMultihopRatchet` from the parent `darepo` checkout

### DIFF
--- a/unroll/actor.go
+++ b/unroll/actor.go
@@ -396,6 +396,12 @@ func safeTxOutPkScript(tx *wire.MsgTx, index uint32) ([]byte, error) {
 	return append([]byte(nil), tx.TxOut[index].PkScript...), nil
 }
 
+// proofNodeHeightHint is the earliest safe confirmation height hint for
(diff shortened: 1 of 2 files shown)


ellemouton:
This is a tiny ollow up to #348 so will merge if CI passes

PR #430, opened 2026-05-14
rpc: add grpc-gateway for daemon APIs

sputn1ck:
## Summary

- add grpc-gateway v2 code generation from external YAML configs
- generate daemon, ark, mailbox, and swap client gateway handlers
- start an enabled-by-default daemon HTTP/JSON gateway
- register swap client handlers when the swap runtime is enabled

## Verification

- `go test ./... -run '^$'`
- `go test ./darepod -run 'TestDefaultConfigEnablesGateway|TestConfigValidateAllowsDisabledGateway|TestConfigValidateAllowsInjectedRPCListener' -count=1`
- `git diff --check`

`make lint-native` was attempted twice after rebasing onto `origin/main`, including once with `LINT_WORKERS='--concurrency=1'`, but the local process was killed with SIGKILL during `custom-gcl run` before reporting lint findings.


sputn1ck:
@claude review pls

darepod/server.go
```diff
@@ -917,6 +918,27 @@ func (s *Server) run(ctx context.Context, shutdownFn func()) error {
 		return err
 	}
 
+	s.gateway = newGatewayServer(
+		s.cfg.RPC.Gateway, lis.Addr().String(), s.rpcServer, s.cfg,
```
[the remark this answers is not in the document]
sputn1ck: This is addressed in the current branch: the endpoint passed to grpc-gateway is normalized with `gateway.NormalizeEndpoint`, so wildcard binds are dialed through loopback.

darepod/server.go
```diff
@@ -917,6 +918,27 @@ func (s *Server) run(ctx context.Context, shutdownFn func()) error {
...
+		s.cfg.RPCGatewayRegistrars, s.log,
+	)
+	if err := s.gateway.Start(ctx); err != nil {
```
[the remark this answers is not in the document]
sputn1ck: Addressed in the latest push: if gateway startup fails after the daemon gRPC listener is opened, the listener is now closed before returning the error.

darepod/gateway_server.go
```diff
@@ -0,0 +1,152 @@
...
+
+	go func() {
+		g.log.InfoS(ctx, "HTTP gateway listening",
+			"addr", listener.Addr(),
+		)
+
+		err := g.httpSrv.Serve(listener)
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
+			g.log.ErrorS(ctx, "HTTP gateway error", err)
+		}
```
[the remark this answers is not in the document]
sputn1ck: Partially addressed in the latest push: the gateway serve goroutine is now tracked and joined during shutdown. Runtime `Serve` errors are still logged rather than treated as daemon-fatal because startup/registration errors are already returned synchronously from `Start`, while later listener errors occur after the owning daemon is running.

darepod/gateway_server.go
```diff
@@ -0,0 +1,152 @@
...
+				UnmarshalOptions: protojson.UnmarshalOptions{
+					DiscardUnknown: true,
+				},
+			},
+		),
+	)
+
+	dialOpts := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
```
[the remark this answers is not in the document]
sputn1ck: Leaving this unchanged for now: the daemon currently starts its local gRPC server without TLS (there is still a TODO next to that server setup), so the in-process gateway has to dial it with insecure local credentials. If/when the daemon RPC server wires `RPC.TLSCertPath`/`TLSKeyPath` into `grpc.Creds`, the gateway dial options should be updated in the same change.

sputn1ck (commented):
Review responses for grpc-gateway follow-ups.

sdk/walletdk/types.go
```diff
@@ -2,6 +2,7 @@ package walletdk
 
 import (
 	"io"
+	"net/http"
 	"time"
 
 	"github.com/lightninglabs/darepo-client/darepod"
```
bhandras: Blocking for the stated WASM/browser-wallet goal: `walletdk` still imports the embedded daemon stack unconditionally via `darepod`, so browser consumers cannot import this package even when they only want `Connect(... TransportREST)`. I verified `GOOS=js GOARCH=wasm go test -c ./sdk/walletdk` fails through `darepod -> db/migrate -> modernc.org/sqlite -> modernc.org/libc` build constraints, while `./rpc/restclient` compiles for WASM. The embedded-only surface (`Config`, `DefaultConfig`, `Start`, `DaemonConfig`, and the daemon runtime helpers/stubs) needs non-js build tags or a separate package so the REST/browser client API stays js-safe.

bhandras (commented):
I reviewed this with the browser/WASM wallet goal in mind. The REST transport itself is heading in the right direction, but I found two issues that should be addressed before relying on this from a browser wallet.

gateway/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+// BrowserHeaders wraps an HTTP handler with browser-friendly CORS headers.
+func BrowserHeaders(next http.Handler, metadataHeaders ...string) http.Handler {
+	allowedHeaders := []string{
+		"content-type",
+		"authorization",
+	}
+	allowedHeaders = append(allowedHeaders, metadataHeaders...)
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
```
bhandras: Because this gateway exposes wallet-control RPCs to browsers, wildcard CORS is a real authority grant. With the default gateway enabled on localhost, any web page a user visits can POST to `http://localhost:10031/v1/daemon/...` once this returns `Access-Control-Allow-Origin: *`; generated routes include wallet actions like `init-wallet`, `unlock-wallet`, and `send-vtxo`. For a browser wallet, this should be restricted to trusted origins and/or require an unguessable auth token enforced by the gateway or a fronting proxy, and the default local daemon path should fail closed.

darepod/gateway_server.go
```diff
@@ -0,0 +1,163 @@
...
+
+			return fmt.Errorf("register optional gateway "+
+				"handlers: %w", err)
+		}
+	}
+
+	g.listener = listener
+	g.cancel = cancelRegister
+	g.httpSrv = &http.Server{
+		Handler:           gateway.BrowserHeaders(mux),
```
bhandras: Heads-up for whenever the auth fix lands here: `gateway.BrowserHeaders(mux)` is called with no `metadataHeaders` arguments, so the CORS allowlist is just `content-type, authorization`. The moment the wasm wallet has to attach any per-RPC token outside `Authorization` (an `x-darepod-auth` cookie-replacement, an internal-only marker mirroring `x-mailbox-auth-sig` on the operator gateway, etc.), browsers will block the preflight until that header name is added here. Worth wiring `gateway.BrowserHeaders(mux, "x-...-auth")` in the same change that introduces the auth so the CORS allowlist tracks the auth surface, instead of discovering it later as a CORS-only failure in the browser.

rpc/restclient/stream.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return &StreamClient[T]{
+		resp: resp,
+		dec:  json.NewDecoder(resp.Body),
+		name: name,
+		new:  newMsg,
+	}
+}
+
+// Recv returns the next message from the REST stream.
```
bhandras, line 42: Flagging this specifically for the wasm/browser-wallet path: server-streaming over chunked JSON depends on the HTTP client surfacing bytes as they arrive, and the Go `GOOS=js` http roundtripper goes through `fetch`. Recent Go versions can hook into Fetch's `ReadableStream`, but behavior varies by browser and isn't universally enabled — historically `fetch` resolves only after the full response body is buffered, in which case `Recv()` here will block until the stream ends rather than producing incremental messages.

The long-running streams generated through this path (`WatchRounds`, `SubscribeSwaps`, `WalletEntry`) are exactly the ones a wallet UI wants to subscribe to live. Worth doing one manual `WatchRounds` test from a real wasm build before we claim streams are usable in the browser; if Fetch buffering bites, we should plan a polling shape for those RPCs rather than ship a transport that quietly behaves differently from the gRPC one.

rpc/restclient/client.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+	if md, ok := metadata.FromOutgoingContext(ctx); ok {
+		for key, values := range md {
+			if strings.HasSuffix(strings.ToLower(key), "-bin") {
+				continue
+			}
+			for _, value := range values {
+				req.Header.Add(key, value)
+				req.Header.Add("Grpc-Metadata-"+key, value)
```
bhandras: Why double-write each metadata entry — once with the raw key and once as `Grpc-Metadata-`+key? grpc-gateway's default matcher consumes the prefixed form, the operator gateway's custom `HeaderMatcher` consumes the raw form, and emitting both means values like `Authorization` are sent twice — once as the standard HTTP header and once as `Grpc-Metadata-Authorization`. That can interact badly with intermediate proxies that try to validate, strip, or rewrite the standard form (k8s ingress, auth middleware, anything looking at `Authorization` for its own purposes).

Would prefer to pick one convention: keep the raw form, since `BrowserHeaders` already allows the raw `authorization` and the operator-side matcher special-cases the raw key explicitly, and drop the prefixed duplicate.

rpc/restclient/client.go
```diff
@@ -0,0 +1,268 @@
...
+	headers    http.Header
+}
+
+// New creates a REST RPC client transport for one grpc-gateway base address.
+func New(addr string, opts ...Option) *Client {
+	baseURL := addr
+	if !strings.HasPrefix(baseURL, "http://") &&
+		!strings.HasPrefix(baseURL, "https://") {
+
+		baseURL = "http://" + baseURL
```
bhandras: `restclient.New("daemon.example.com:1234")` silently becomes `http://daemon.example.com:1234`. Fine for local dev, but for a hosted wasm wallet pointed at a remote daemon this trips mixed-content protections in the browser and, more importantly, sends any auth header we eventually add over plaintext HTTP without the caller realizing the scheme was inferred.

I'd flip the default to `https://` and only fall back to `http://` when the address resolves to a loopback host (or, simpler, require the caller to pass an explicit scheme and reject schemeless inputs). The current behavior is a footgun that won't surface until someone deploys against a real hostname.

bhandras (commented):
Another pass with the wasm/browser-wallet goal in mind. Four follow-ups inline; none are blocking on top of the two earlier ones about the embedded-stack import and the wildcard CORS authority grant.

cmd/darepod/main.go
```diff
@@ -122,19 +122,7 @@ func newRootCmd() *cobra.Command {
...
-	)
-	f.String(
-		"server.tlscertpath", cfg.Server.TLSCertPath,
-		"path to ark server TLS certificate",
-	)
-	f.Bool(
-		"server.insecure", cfg.Server.Insecure,
-		"disable TLS for the server connection (dev only)",
-	)
+	registerArkServerFlags(f, cfg)
```
bhandras, line 125: nit: commit message is weirdly formatted

sputn1ck: Addressed while autosquashing the review fixes; the rewritten commit message now uses normal wrapped paragraphs.

sdk/walletdk/types.go
sputn1ck: Addressed: the embedded walletdk config/runtime files are now !js-only. GOOS=js GOARCH=wasm go test -c ./sdk/walletdk passes, and js builds keep the REST Connect surface while gRPC connect returns a clear unsupported error.

rpc/restclient/client.go
sputn1ck: Addressed: outgoing metadata now emits only the raw header form; the Grpc-Metadata-* duplicate was removed and covered by a test.

bhandras (approved):
Re-reviewed the updated head with the WASM/browser-wallet path in mind. The prior blocking walletdk import issue is fixed: js/wasm builds no longer pull in darepod/sqlite, and `GOOS=js GOARCH=wasm go test -c ./sdk/walletdk` now passes. The gateway CORS behavior is also fail-closed by default with explicit allowed origins, REST metadata no longer double-emits `Grpc-Metadata-*`, and schemeless non-loopback REST addresses now default to HTTPS.

Local checks run:
- `GOOS=js GOARCH=wasm go test -c -o /tmp/walletdk-pr430.test.wasm ./sdk/walletdk`
- `GOOS=js GOARCH=wasm go test -c -o /tmp/restclient-pr430.test.wasm ./rpc/restclient`
- `go test ./gateway ./rpc/restclient ./sdk/walletdk ./darepod -run 'TestBrowserHeaders|TestRESTClient|TestGatewayStatusError|TestDefaultConfigEnablesGateway|TestConfigValidate|TestConfigValidateAllowsDisabledGateway|TestConfigValidateAllowsInjectedRPCListener' -count=1`
- `go test ./... -run '^$'`
- `git diff --check origin/main...HEAD`

No further blocking findings from me.

gateway/gateway.go
bhandras: Resolved cleanly by the `BrowserHeaders` rewrite. Wildcard `Access-Control-Allow-Origin: *` is gone, per-origin echo with `Vary: Origin`, no-Origin requests still pass through for CLI / server-to-server, and `validateGatewayAllowedOrigins` rejects `*` and `""` so the wildcard can't sneak back in via config. Default `AllowedOrigins` is nil → browser callers fail closed. Thanks.

sdk/walletdk/types.go
bhandras: Verified the split locally: `GOOS=js GOARCH=wasm go test -c ./sdk/walletdk` produces a valid wasm binary and `GOOS=js GOARCH=wasm go build ./sdk/walletdk ./rpc/restclient` is clean. The `connect_grpc_js.go` stub returning a clear pointer to `TransportREST` is a nice touch. Thanks.

rpc/restclient/client.go
bhandras: Addressed: `defaultBaseURL` + `isLoopbackAddress` infer `http://` only for `localhost`, IPv4 loopback, and IPv6 loopback (including `[::1]:port`), and `https://` for everything else; explicit schemes pass through. Thanks.

darepod/config.go
```diff
@@ -609,17 +681,68 @@ func (c *Config) Validate() error {
...
+
+// validateGatewayAllowedOrigins rejects wildcard CORS grants on wallet-control
+// APIs.
+func validateGatewayAllowedOrigins(origins []string) error {
+	for _, origin := range origins {
+		switch origin {
+		case "", "*":
+			return fmt.Errorf("rpc.gateway.allowedorigins must "+
+				"contain explicit origins, got %q", origin)
+		}
```
bhandras, line 728: Nice fail-closed default and the `""`/`*` rejection. One gap worth closing while we're here: the validator accepts any non-empty non-`*` string, but a browser's `Origin` header is always `scheme://host[:port]` with no path/query/fragment. If a user writes `wallet.example` (no scheme) or `https://wallet.example/` (trailing slash) the daemon happily accepts it on startup, then every browser request silently 403s at runtime with no clue why.

A `url.Parse` check that requires non-empty `Scheme` and `Host`, and rejects non-empty `Path`, `RawQuery`, `Fragment` would catch the typical misconfigurations at startup instead of at runtime. Same applies to trailing `/`, which browsers strip.

bhandras (commented):
Re-reviewed since `a76cfe5`. The four actionable items from the previous pass are all in, plus the bonus `BrowserHeaders` rewrite that closes the wildcard-CORS authority grant — wallet-control routes now fail closed for any browser origin not on the explicit allowlist. Verified the wasm build locally (`GOOS=js GOARCH=wasm go test -c ./sdk/walletdk` produces a valid 32MB wasm binary, gateway/walletdk/config tests green).

One small gap noted inline on the origin validator; nothing blocking.

bhandras (approved):
LGTM 🎉

bhandras:
Three follow-up notes I'd like tracked after this lands; none of them block the PR.

- `validateGatewayAllowedOrigins` correctly rejects `""` and `*` but accepts schemeless or path-bearing strings. A `url.Parse` check requiring non-empty `Scheme`/`Host` and rejecting `Path`/`RawQuery`/`Fragment` would surface misconfigs at startup instead of as silent 403s at runtime.
- Whenever the first per-RPC daemon auth header lands, please update `gateway.BrowserHeaders(mux, allowedOrigins, "x-...-auth")` in the same change — the signature already supports it, but browsers will block the preflight until that header name is on the CORS allowlist.
- Before the wasm wallet leans on `WatchRounds`, `SubscribeSwaps`, or `WalletEntry`, please do one manual smoke test from a real `GOOS=js` build. Go's wasm http roundtripper goes through `fetch`, and incremental chunk delivery via `ReadableStream` is browser-dependent — if `Recv()` ends up blocking until stream end we'll want a polling shape for those RPCs rather than ship a transport that quietly behaves differently from gRPC.

PR #431, opened 2026-05-14
wasm: add walletdk browser demo

sputn1ck:
## Summary

Adds the walletdk browser demo implementation branch:

- OPFS-backed SQLite for daemon, swap, and lwwallet wallet DB paths in WASM
- grpc-gateway browser adapters for Ark, mailbox, and swap RPC seams
- static walletdk demo UI with create, unlock, balances, address, receive, send, swaps, and activity surfaces
- GitHub Pages build/deploy workflow
- committed Playwright smoke harness for Chromium create, address, reload, unlock, identity persistence, swapruntime startup, and OPFS verification

## Validation

- `make wasm-wallet-demo-browser-test`
- `make unit pkg=./db timeout=5m`
- `make unit pkg=./sdk/swaps timeout=5m`
- `make unit pkg=./sdk/walletdk timeout=5m`
- `env GOOS=js GOARCH=wasm go test -c -o /private/tmp/db.test.wasm ./db`
- `env GOOS=js GOARCH=wasm go test -c -o /private/tmp/swaps.test.wasm ./sdk/swaps`
- `env GOOS=js GOARCH=wasm go test -c -tags swapruntime -o /private/tmp/walletdk.test.wasm ./sdk/walletdk`
- `env GOOS=js GOARCH=wasm go build -tags swapruntime -o /private/tmp/walletdk.wasm ./cmd/walletdk-wasm`
- `make commitmsg-lint range=90888222..HEAD`

## Remaining follow-up

This is a draft because Playwright MCP was not exposed in the local session, the Pages workflow has not run from `main`, and full receive/send behavior still needs validation against a live SwapDK test instance.


Roasbeef:
Superseded by #803, which rebuilds the wasm wallet bindings against current `main`. This branch had gone stale: the OOR actor set was rewritten (the global signing-effect actor that the inline OOR path wrapped is gone), the `walletdk.Client` send path moved to a prepare/dispatch split, and `sdk/walletdk` was reorganized around the `walletdkrpc` rename.

The new PR drops the inline OOR actor entirely (the durable registry + per-session actors run under `js/wasm` as-is), drives the browser bridge through the `sdk/walletdk/mobile` JSON facade so it can't drift from the `Client` API, and leaves out the app-specific bits (React demo, Playwright, Pages workflow). Let's move review over to #803.

cc @jamaljsr

PR #432, opened 2026-05-14
Correlate early round quotes by accepted inputs

bhandras:
## What changed

This makes `JoinRoundQuote` self-correlating by carrying the accepted boarding and VTXO outpoints that were already present on the join acknowledgement path.

On the client side, an out-of-order quote can now be matched to the pending temp-keyed round before the separate admission envelope arrives. The actor re-keys the round from the quote, synthesizes the admission event for the FSM, then processes the quote. A later duplicate admission for the same round is treated as an idempotent watermark.

## Why

Mailbox delivery does not guarantee that the join acknowledgement arrives before the seal-time quote. Without correlation data on the quote itself, a client can buffer the quote under the assigned round id while its local FSM is still temp-keyed, then miss the server's quote timeout window.

## Validation

- `go test ./round ./rpc/roundpb`
- `go test ./darepod -run TestRunWithContextServesInjectedListener -count=1`
- `go test ./sdk/ark -run TestStartEmbeddedUsesBufconnTransport -count=1`

`go test ./...` was also run; it hit timing-sensitive startup/readiness failures in the two isolated tests above, and both passed when rerun individually.

### DIFF
--- a/round/actor.go
+++ b/round/actor.go
@@ -794,6 +794,23 @@ func (a *RoundClientActor) findAssemblingRound() *RoundFSM {
 func (a *RoundClientActor) findRoundByOutpoints(
 	boardingOutpoints, vtxoOutpoints []wire.OutPoint) *RoundFSM {
 
+	roundFSM, ambiguous := a.findRoundByOutpointsDetailed(
+		boardingOutpoints, vtxoOutpoints,
+	)
+	if ambiguous {
+		return nil
+	}
+
+	return roundFSM
+}
+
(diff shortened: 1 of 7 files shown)


bhandras:
Closing this as superseded by the correlation-key FIFO work in #435 and the server-side round message keying in lightninglabs/darepo#437.

The original issue was quote-before-join-ack delivery for the same client/round. With client+round correlation keys, those messages now share a durable FIFO lane, so the mailbox layer preserves the join-ack-before-quote ordering without adding accepted-input correlation fields to `JoinRoundQuote`.

PR #334, opened 2026-05-02
docs: nightly doc-gardening sweep 2026-05-02

litbot-9000:
Nightly automated documentation sweep (2026-05-02).

**Changes:** Added `rpc/oorpb/CLAUDE.md` and `rpc/oorpb/AGENTS.md` — the only package with hand-written Go source (`payloads.go`) but no per-package docs. All sqlc-generated and proto-only packages were correctly skipped.

**Workflow run:** https://github.com/lightninglabs/darepo-client/actions/runs/0

Please review the updated `rpc/oorpb/CLAUDE.md` diff to verify the Key Types, Relationships, and Invariants sections are accurate.

PR #434, opened 2026-05-14
harness: retry bitcoind container start on RPC probe failure

ellemouton:
## Summary

Add a per-attempt RPC health probe to `startBitcoind` and retry the whole container reconstruct (up to 3 times) when the probe fails. Catches a CI flake where the bitcoind container reports running and answers its first JSON-RPC call, then becomes unresponsive before any user-driven RPC arrives.

## Root cause

In lightninglabs/darepo PR #334 (CI run [25870709078](https://github.com/lightninglabs/darepo/actions/runs/25870709078)) the same test (`TestBoardingIntegrationSingleClientSubsequentRounds`) failed on all three backends — lnd, lwwallet, and btcwallet — with:

```
rpc request failed: listwallets: Post \"http://127.0.0.1:<port>\":
dial tcp 127.0.0.1:<port>: connect: connection refused
```

The operator's block epoch stayed at the harness initial height (107) for the whole test, never observing the mined blocks from `FundOperatorLND`'s `Generate(6)`. So bitcoind either never finished initializing or died after accepting the very first JSON-RPC call. `startBitcoind` happily returned and the deep `connection refused` only surfaced many seconds later, inside the test's first `Faucet` call.

## Fix

Wrap `startBitcoind`'s per-attempt body in a retry loop:

- `attemptStartBitcoind` does docker run + initial RPC ready + wallet ensure + a sustained-health probe.
- `probeBitcoindHealthy` spends `bitcoindStartProbeDuration` (5s) repeatedly calling `getblockcount`. Returns on the first error, otherwise nil. Short enough to keep happy-path setup snappy, long enough to observe the death window seen in CI.
- On per-attempt failure the container is torn down, `BitcoindRPC` et al. are reset to zero so no stale port leaks into a later attempt, and `startBitcoind` backs off with jitter before retrying up to `maxBitcoindStartRetries` (=3) times.

Initial readiness moves to a new non-fatal `waitForBitcoindReady`, and wallet ensure has a non-fatal sibling `bitcoindEnsureWalletErr`. The retry loop needs error returns so a `require.Eventually`-style fatal cannot mark the test failed before the next attempt can run. The original `bitcoindEnsureWallet` stays for the mid-test `Faucet` path, which has no recovery option and should still fail loudly.

## Validation

- `go build ./...` — pass
- `go test ./harness/...` — pass
- `make lint-native` — pass (0 issues)
- From the parent `darepo` checkout: `make itest backend=lnd icase=TestBoardingIntegrationSingleClientSubsequentRounds` — PASS (99.83s and 103.19s across two runs)

The CI run against PR #334 that exposed the flake has been re-triggered with the original (pre-fix) submodule pointer to confirm baseline. If you'd like to test this fix against the full itest suite, bumping #334's client submodule to this branch will exercise it across all three backends.

## No public API change

`startBitcoind` keeps its `require.NoError` behavior on exhausted retries, so callers still see the same fatal signal when the harness truly cannot bring bitcoind up.

### DIFF
--- a/harness/harness.go
+++ b/harness/harness.go
@@ -93,6 +93,28 @@ const (
 	// starting a container when Docker fails to bind a randomly assigned
 	// host port due to a race with parallel test execution.
 	maxPortBindRetries = 15
+
+	// maxBitcoindStartRetries is the maximum number of times the harness
+	// rebuilds the bitcoind container when post-start RPC probing
+	// fails. The CI signature this guards against is a container whose
+	// port forward exists but whose bitcoind process either never bound
+	// the RPC socket or crashed during init: subsequent RPC calls hit
+	// "connection refused" against an apparently-running container.
+	// Three attempts is enough to recover from a transient docker
+	// scheduling glitch under heavy parallel load without masking a
+	// genuine bug in the bitcoind image.
+	maxBitcoindStartRetries = 3
+
+	// bitcoindStartProbeDuration is the wall-clock window the post-start
+	// RPC probe spends repeatedly polling bitcoind to confirm it stays
+	// reachable. The existing waitForBitcoind already gates on the
+	// first successful getblockchaininfo response; this extra probe is
+	// what catches the failure mode where bitcoind answers the first
+	// poll and then dies before the harness's first user-driven RPC
+	// call. 5 seconds is short enough to keep test setup snappy and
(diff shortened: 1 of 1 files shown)


Roasbeef (approved):
LGTM 🫛

PR #435, opened 2026-05-14
multi: per-correlation-key FIFO claim ordering for durable mailbox

Roasbeef:
## Summary

In this PR, we close a structural reorder window in the durable actor
mailbox by adding a per-correlation-key FIFO claim invariant, then stamp
the matching keys on the client-to-server OOR and round outbox messages
so the invariant fires on production traffic. The framework primitive
lands first, then the SQL claim + schema, then the per-message stamping
for the two FSMs that need it. See each commit message for the per-step
rationale.

## The bug

When a Tell to the serverconn durable mailbox transiently failed (e.g.
a gRPC blip on `Edge.Send`), the actor runtime nacked the message with
the default exponential backoff (1s, 2s, 4s, 8s, 16s) by pushing
`available_at` into the future. The existing claim SQL orders by
`(priority DESC, available_at ASC, created_at ASC)`, so a later-enqueued
message in the same mailbox with a smaller `available_at` could overtake
the in-backoff predecessor on the next `LeaseNextMailboxMessage`. Any
pair of logically-sequenced outbox events sharing a per-client mailbox
could reorder. The concrete bite is round: a transient send failure on
`SubmitNoncesRequest` would let the follow-up `SubmitPartialSigRequest`
for the same round overtake it, and the server would see partial sigs
before nonces, outside the protocol contract.

## Approach

We add a `CorrelationKey() string` method to the `actor.Message`
interface with a default empty value on `BaseMessage`. The durable
mailbox reads the key at enqueue and stamps it on a new
`mailbox_messages.correlation_key` column (NULL when empty). The claim
SQL gains a `NOT EXISTS` anti-join: a keyed row is claim-eligible only
if no earlier same-key, same-mailbox, retry-budget-remaining row exists.
"Earlier" is the row's UUIDv7 `id`, which embeds a millisecond timestamp
plus a per-generator tiebreaker so two messages from the same producer
are strictly orderable even when they fall in the same second-granularity
`created_at` bucket.

Unkeyed rows (NULL `correlation_key`) skip the anti-join entirely and
keep the legacy global `available_at` order. They are not blocked by
keyed lanes, and they do not block them. The default is opt-in: every
existing `TLVMessage` implementer keeps the empty key from `BaseMessage`,
so the schema change is back-compat for in-flight rows and the migration
is nullable with no backfill.

A filtered composite index on `(mailbox_id, correlation_key, id) WHERE
correlation_key IS NOT NULL` keeps the anti-join probe an index seek
under load. Head-of-line blocking is bounded to the correlation key, not
the mailbox; per-mailbox consumers are already strictly serial so this
does not regress throughput.

## Stamping

Round stamps `roundCorrelationKey(roundID)` on its seven
client-to-server outbox messages: `JoinRoundRequest`,
`JoinRoundAcceptOutbox`, `JoinRoundRejectOutbox`, `SubmitNoncesRequest`,
`SubmitPartialSigRequest`, `SubmitForfeitSigRequest`,
`SubmitVTXOForfeitSigsToServer`. A `JoinRoundRequest` with an empty
`RoundID` (fresh join, no server-assigned id yet) falls back to the
unkeyed lane; it's the first message for the round and has nothing
earlier to sequence against.

OOR stamps `sessionCorrelationKey(sessionID)` on its three
client-to-server outbox messages: `SendSubmitPackageRequest`,
`SendFinalizePackageRequest`, `SendIncomingAckRequest`. Submit and
finalize derive the session id from the embedded `ArkPSBT` via
`sessionIDFromArk`; `SendIncomingAckRequest` already carries an explicit
`SessionID`. A malformed PSBT falls back to the unkeyed lane (the
message would error out at `ToProto` anyway).

The `"oor/"` and `"round/"` namespace prefixes keep the two key spaces
disjoint so a session id can never accidentally collide with a round id.

## Wrapper hop and dead-letter gap (review followups)

A multi-agent code review of the original five commits surfaced two
issues, both addressed in the same branch ahead of opening the PR.

The first was that the OOR and round outbox messages were stamped on
the inner concrete type but were then wrapped in
`serverconn.SendClientEventRequest` before reaching the durable mailbox.
`SendClientEventRequest` embeds `actor.BaseMessage`, so its
`CorrelationKey()` returned `""` and the inner key was lost at the
wrapper hop. Every wrapped row would have landed in the unkeyed lane and
the per-key invariant would never have fired on production traffic. The
fix adds a `CorrelationKey()` method on the wrapper that forwards the
inner `ServerMessage`'s key via a structural type assertion, so any
concrete message that opts in is supported without coupling
`ServerMessage` to the actor interface.

The second was that the anti-join did not filter exhausted predecessors.
A same-key row whose `attempts == max_attempts` but which had not yet
been physically deleted (e.g. a crash window between `MoveToDeadLetter`
and `DeleteMailboxMessage` in `handlePoisonMessage`) would permanently
block every later same-key message even though the outer eligibility
predicate already excluded it from being re-leased. The anti-join now
carries `AND m2.attempts < m2.max_attempts` so its predicate agrees
with the outer eligibility predicate.

## Tests

Two layers cover the invariant. `db/actordelivery/correlation_key_test.go`
drives the SQL claim path directly with a test clock: the canonical
reorder reproduction (`TestPerKeyFIFOBlocksOvertakeOnNack`), cross-key
independence, unkeyed-lane orthogonality, ack-unblocks-key, mailbox
isolation, and the exhausted-predecessor regression test that fails
without the dead-letter-gap fix.

`db/actordelivery/durable_actor_reorder_test.go` exercises the same
invariant through a real `DurableActor` with a behavior that errors on
the first attempt and succeeds on retry, asserting emission order
survives transient failure end to end.

For the wrapper-forwarding fix,
`serverconn/actor_correlation_key_test.go` covers the four wrapper
branches (keyed inner, unkeyed inner, non-implementer post-decode, nil
`Message`) plus an integration test that drives a stamped
`SendClientEventRequest` through the `DurableActor` and asserts the
inner key reaches `EnqueueParams.CorrelationKey` at the store layer.
Without the forward the captured params carry an empty string and the
test fails.

## Test plan

- [x] `make unit pkg=./db/actordelivery/...`
- [x] `make unit pkg=./baselib/actor/...`
- [x] `make unit pkg=./serverconn/...`
- [x] `make lint-native`
- [x] Verified `TestPerKeyFIFOExhaustedPredecessorDoesNotBlockSuccessor`
      fails on the pre-fix SQL and passes with the anti-join attempts
      filter in place.
- [x] Verified `TestSendClientEventRequestPlumbsCorrelationKeyToStore`
      requires the wrapper forwarding hop to pass (empty inner key would
      land in the unkeyed lane).

### DIFF
--- a/baselib/actor/CLAUDE.md
+++ b/baselib/actor/CLAUDE.md
@@ -38,6 +38,20 @@ crash-safe at-least-once delivery with exactly-once deduplication.
 - `ChannelMailbox[M, R]` — In-memory channel-based mailbox (non-durable, for lightweight actors).
 - `Mailbox[M, R]` — Interface for actor message queues: `Send(ctx, env) error` (blocking, returns `ErrMailboxClosed` or `ErrActorTerminated` on failure), `TrySend(env) error` (non-blocking), `Receive(ctx) iter.Seq[envelope]`, `Close()`, `IsClosed() bool`, `Drain() iter.Seq[envelope]`. `Send` previously returned `bool`; it now propagates the exact failure cause so callers can distinguish `ErrMailboxClosed`, `ErrActorTerminated`, and `context.Canceled`/`context.DeadlineExceeded` without inspecting actor state.
 - `isExpectedShutdownErr(err) bool` — Internal helper that classifies errors as expected during teardown: context cancellation/deadline, closed DB handle ("sql: database is closed", "sql: connection is already closed", "use of closed network connection"). Used by the lease loop to demote shutdown-path failures to debug instead of warn-flooding test artifacts at itest tail.
+- `Message.CorrelationKey() string` — Per-message FIFO key consumed by the
+  durable mailbox's claim path. Non-empty keys participate in per-key FIFO:
+  a message is claim-eligible only when no earlier same-key message
+  (compared by UUIDv7 `id`) exists in the same mailbox, even if the
+  earlier message is in retry backoff. Empty (the default on
+  `BaseMessage`) means the message is unkeyed and uses the existing
+  global `available_at` claim order. The override site is the concrete
+  message struct (e.g. `clientconn.ClientMessage` types in `rounds`),
+  not the framework — the framework just plumbs the value through
+  `EnqueueParams.CorrelationKey`.
+- `EnqueueParams.CorrelationKey` — Per-enqueue override stamped into the
+  `mailbox_messages.correlation_key` column. Populated automatically from
+  `msg.CorrelationKey()` by `DurableMailbox.Send`. A zero (empty) value
+  preserves the legacy unkeyed claim semantics.
 
 ## Relationships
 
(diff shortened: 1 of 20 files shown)


Roasbeef:
@claude review this 

@codex review this

Roasbeef:
Addressed review feedback in fixups (force-pushed onto `origin/main`):

**@chatgpt-codex-connector P1 (correlation key not surviving outbox CDC decode):**
You're right — when `TransportOutbox` is enabled, `sendTransportEvent` encodes `SendClientEventRequest` into the actor outbox, `OutboxPublisher` decodes it (inner becomes `*rawServerMessage`), and the first enqueue into the serverconn mailbox is where `CorrelationKey()` is consumed. The structural assertion misses on `rawServerMessage` so the key was being dropped. Fixed by persisting the inner key in its own TLV record (new `correlationKeyRecordTLV = tlv.TlvType8`) at Encode and caching the raw bytes on the wrapper at Decode. `CorrelationKey()` now falls back to the cached value when the structural assertion misses. Backward compat: old outbox blobs without the new TLV record decode to an empty cached key (same lane behavior they had pre-fix). New tests cover the Encode→Decode→`CorrelationKey()` roundtrip and the unkeyed-roundtrip case.

**@claude minor: stale `db/actordelivery/migrations/CLAUDE.md` `LatestMigrationVersion = 1`:**
Fixed — bumped the doc to `LatestMigrationVersion = 2` and described what migration 000002 added.

**@claude test gap: no test for actively-leased predecessor blocking successor:**
Added `TestPerKeyFIFOActiveLeasedPredecessorBlocksSuccessor` which pins the invariant that an actively-leased same-key predecessor (lease_until in the future, retry budget remaining) blocks any later same-key message even though it isn't in backoff. Guards against accidental future loosening of the anti-join (e.g. adding a `lease_until` filter).

**@gemini-code-assist on `fraud/actor.go`:** out of scope for this PR — `fraud/actor.go` isn't changed here, those comments would attach to the original fraud-watcher PR (#348). Happy to follow up separately.

**@chatgpt-codex-connector on `fraud/tracker.go` (Tell vs Ask) and `unroll/actor.go` (unregister conf watches):** also out of scope — pre-existing code from #348. Separate follow-up.

PR #436, opened 2026-05-14
p-models: add durable actor mailbox model

Roasbeef:
## Summary

This PR adds a formal model for the durable actor mailbox claim semantics,
stacked on top of #435 (`mailbox-correlation-key-fifo`). The goal is to make
that fix precise enough that we can both explain the production failure mode and
keep checking future mailbox changes against an ideal specification.

P is a state-machine modeling language for asynchronous systems. We describe the
system as machines that exchange events, then let the checker explore many event
schedules looking for assertion failures, monitor violations, deadlocks, and
other illegal states. That is a good fit for the durable actor mailbox because
the important invariant is distributed across enqueue, lease, nack, retry,
lease-expiry, and ack transitions rather than isolated in one function.

The new model lives under `p-models/durableactor/` and covers the core mailbox
state machine:

- enqueue with duplicate-id idempotence
- lease/claim with priority and availability ordering
- per-correlation-key FIFO blocking
- ack and nack ownership by lease token
- retry backoff, retry exhaustion, and dead-letter removal
- lease expiry and later re-claim
- mailbox isolation and cross-key independence

It includes two claim modes:

- `LegacyAvailableAtOrder`, which matches the old behavior that ordered eligible
  rows by priority and `available_at`.
- `PerCorrelationKeyFIFO`, which models the fixed behavior where a live earlier
  same-mailbox/same-correlation-key row blocks later rows for that key.

## Why this helps

The issue seen in the wild was a distributed-systems ordering bug: a message that
had already been observed and nacked into backoff was no longer claim-visible,
so a later message for the same logical protocol stream could be claimed first.
That allowed the durable actor mailbox to violate the protocol's per-session or
per-round ordering assumptions even though each individual row transition looked
locally valid.

The model captures the intended global invariant instead: within a mailbox and
correlation key, live messages must be claimed in enqueue/id order. Availability,
backoff, and lease state can delay the head of the key, but they must not let a
successor overtake it.

This gives us a place to reason about the durable actor mailbox from first
principles rather than only testing specific SQL query shapes. It should help us
catch future issues around retry scheduling, leases, priority ordering,
dead-letter handling, multi-mailbox isolation, and any new mailbox behavior that
could accidentally weaken per-key ordering.

## Counterexample for the original bug

The model includes an intentional negative testcase,
`tcMailboxLegacyReorderCounterexample`, that checks the ideal same-key FIFO
property against the legacy ordering mode. It reproduces the exact class of bug
from production:

1. enqueue `msg-1` for correlation key `k`
2. lease `msg-1`
3. nack `msg-1` into backoff
4. enqueue `msg-2` for the same key
5. legacy ordering claims `msg-2` while `msg-1` is still live

That testcase is intentionally excluded from the default green suite, but it can
be run directly to show that P finds the counterexample.

## Bridge into the implementation

This also adds a Go replay bridge under `p-models/durableactor/bridge`. The
bridge loads checked-in JSON traces and replays them against the real
`db/actordelivery` SQLite store using the production migrations and
`LeaseNextMessage` path.

That bridge is important because it keeps the formal model tied to the code we
ship. As we add model-generated traces or new hand-written traces, the same
scenarios can exercise the actual SQL/store implementation rather than only the
P abstraction.

The initial traces cover:

- same-key FIFO across nack/backoff
- active lease blocking of a same-key successor
- cross-key independence
- retry-exhausted predecessor not blocking a successor
- dead-lettered predecessor removal unblocking a successor
- lease token ownership and lease expiry
- duplicate enqueue idempotence

## Docs and workflow

The PR also adds:

- `p-models/README.md`, which explains what P is, why it is useful here, and how
  the model/bridge workflow should be extended.
- `p-models/AGENTS.md` and `p-models/CLAUDE.md`.
- `p-models/durableactor/README.md`, with durable actor mailbox-specific run
  instructions and semantics.
- `p-models/durableactor/AGENTS.md` and `p-models/durableactor/CLAUDE.md`.
- `p-models/scripts/check.sh`.

The script compiles the P model, runs the default green model suite, and runs the
Go bridge tests. Generated P artifacts are ignored via `.gitignore`.

## How to run

```shell
./p-models/scripts/check.sh
```

That command compiles `p-models/durableactor/infra.pproj`, runs the default
`tcMailboxCorrelationKeyFIFO` P testcase, and then runs
`go test ./p-models/durableactor/bridge`.

To demonstrate that the model would have found the original same-key reorder
bug, run the intentional negative testcase:

```shell
p check PGenerated/PChecker/net8.0/MailboxInfraModels.dll \
  --testcase tcMailboxLegacyReorderCounterexample \
  --schedules 1 \
  --max-steps 200
```

That check should report one bug because it checks the ideal same-key FIFO
property against the legacy claim rule.

## Validation

- `./p-models/scripts/check.sh`
- `p check PGenerated/PChecker/net8.0/MailboxInfraModels.dll --testcase tcMailboxLegacyReorderCounterexample --schedules 1 --max-steps 200` finds the expected intentional bug
- `go test ./p-models/durableactor/bridge`
- `make unit pkg=./db/actordelivery/...`
- `make unit pkg=./baselib/actor/...`
- `make unit pkg=./serverconn/...`
- `make fmt-changed`
- `make lint-native`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,6 +155,51 @@ jobs:
         run: GOGC=50 make lint
 
   ########################
+  # P formal model check
+  ########################
+  p-model:
+    name: P durable mailbox model
+    runs-on: [self-hosted]
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v5
+
+      - name: Setup go ${{ env.GO_VERSION }}
+        uses: ./.github/actions/setup-go
+        with:
+          go-version: '${{ env.GO_VERSION }}'
+          key-prefix: p-model
+          use-build-cache: 'no'
+
+      # The self-hosted ARC runners cannot write to the default
+      # /usr/share/dotnet, so install the SDK under the writable runner temp
+      # dir. setup-dotnet honours DOTNET_INSTALL_DIR and adds it to PATH and
+      # DOTNET_ROOT for later steps.
+      - name: Setup .NET 8
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '8.0.x'
+        env:
+          DOTNET_INSTALL_DIR: ${{ runner.temp }}/dotnet
+
+      - name: Install P checker
+        run: |
+          dotnet tool install --global P --version 3.0.4
+          echo "$HOME/.dotnet/tools" >> "$GITHUB_PATH"
+        env:
+          DOTNET_INSTALL_DIR: ${{ runner.temp }}/dotnet
+
+      # Compiles the P model, model-checks the green safety suite
+      # (tcMailboxCorrelationKeyFIFO) and the liveness property
+      # (tcMailboxLiveness), then replays the model traces against the real
+      # actordelivery SQLite store via the Go bridge. GOWORK is disabled so the
+      # bridge builds against the module exactly as an external consumer would.
+      - name: Check P model and bridge conformance
+        run: ./p-models/scripts/check.sh
+        env:
+          GOWORK: off
+
+  ########################
   # cross compilation
   ########################
   cross-compile:
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,7 @@ go.work.sum
 .claude/worktrees/
 .claude/.substrate-plan-context
 .sessions/
+
+# P model generated artifacts.
+/PGenerated/
+/PCheckerOutput/
--- a/p-models/AGENTS.md
+++ b/p-models/AGENTS.md
@@ -1,0 +1,29 @@
+# p-models
+
+This tree contains executable P models and bridge checks. Keep model-specific
+files under a named subdirectory such as `durableactor/`; keep shared runner
+scripts under `scripts/`.
+
+## Commands
+
+| Command | Purpose |
+|---------|---------|
+| `./p-models/scripts/check.sh` | Compile the durable actor model, run P checks, then run the Go bridge |
+| `p compile -pp p-models/durableactor/infra.pproj` | Compile only the durable actor P project |
+| `p check PGenerated/PChecker/net8.0/MailboxInfraModels.dll --testcase tcMailboxCorrelationKeyFIFO` | Run the default durable mailbox P test case |
+| `go test ./p-models/durableactor/bridge` | Replay checked-in traces against the real Go store |
+
+## Layout
+
+- `durableactor/` — durable actor mailbox model, tests, traces, and bridge.
+- `scripts/` — top-level orchestration scripts.
+- `PGenerated/` and `PCheckerOutput/` are generated at repo root and ignored.
+
+## Rules
+
+- Models should encode ideal contracts first, then implementation profiles.
+- Keep known-bad or counterexample tests as separate test cases so the default
+  suite stays green.
+- Bridge tests should exercise real Go code where possible instead of
+  reimplementing production semantics.
+- Add traces for every model scenario that should be replayed against Go.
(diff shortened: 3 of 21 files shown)


Roasbeef:
@claude reviiew this

litbot-9000:

Roasbeef:
@claude review this

Roasbeef:
@claude thanks — this was a sharp pass. All of your findings (plus a related issue your "monitor doesn't catch the bug" thread surfaced) are now addressed in `2a958c8b`. The branch was also rebased onto current `main` so the review is in context with what's actually on the branch.

### Substantive — the monitor didn't catch the production failure class

Fixed, and it turned out to be worse than "weak": **the monitor wasn't being checked at all.** The `// activated globally by P` comment was wrong — P only checks a spec monitor when a test case attaches it with `assert <spec> in { ... }`, which none of the test cases did. So the old `SameKeyClaimsNeverGoBackwards` was dead code on top of being too weak.

Changes:
- Replaced it with **`SameKeyFIFOClaimsRespectLiveHead`**, which reconstructs the live per-lane row set from the enqueue/claim/removal stream and, on every keyed claim, asserts no earlier-id row in the same `(mailbox_id, correlation_key)` lane is still live (present + retry budget). attempts are tracked exactly as `LeaseMailboxRow` updates them, so an exhausted predecessor correctly stops blocking — matching the SQL anti-join's `m2.attempts < m2.max_attempts`.
- The events now carry the announcing mailbox machine (`this`) so the monitor namespaces state per `DurableMailboxSpec` instance — without that, the green suite's many mailbox instances reuse row ids `(1, 10, 100)` and the monitor produced false positives.
- Attached the monitor via `assert ... in` on every claim-bearing test case.
- Added **`tcMailboxMonitorCatchesLegacyReorder`**: it replays the legacy reorder with **no in-machine assertion**, so the bug it finds is raised *solely* by the monitor. It finds the bug; the green suite stays clean.

### Medium — dead-letter token gate

Fixed. `eDurableMailboxDeadLetter` now removes by id with no token check, matching production `MoveToDeadLetter(ctx, id, reason)`. Added **`TestDurableMailboxSpec_DeadLetterByIdRemovesExhaustedRow`**, which walks exactly the path you flagged as unreachable: lease → exhaust → nack (which clears the token) → dead-letter by id → row gone → successor claimable. Under the old token gate this returned `MailboxOpTokenMismatch`; now it passes.

### Minor — `created_at` tiebreaker

Confirmed against the real query: `ORDER BY m.priority DESC, m.available_at ASC, m.created_at ASC` (`db/actordelivery/queries/mailbox.sql`). So `created_at` is correct; the divergence was the model's trailing `id` axis, which the SQL doesn't have. Added a comment on `RowOrdersBefore` documenting that `id` is a model-only determinism tiebreaker (the SQL leaves equal-`created_at` ties unordered under `LIMIT 1`), and a README note that traces keep `created_at` unique within a lane so model and SQL stay congruent.

### Minor — no liveness property

Added **`MailboxKeyedWorkEventuallyDrains`** (hot/cold spec) plus a `MailboxLivenessDriver` that enqueues a same-key pair + a cross-key row and drains by lease-and-ack. Checked by **`tcMailboxLiveness`**. A model where a row could never be claimed would leave the monitor hot forever; today it drains to cold.

### Nit — duplicate-enqueue idempotence

Added an `expect_duplicate` field on the enqueue op; `replayEnqueue` now asserts the no-op explicitly with a duplicate-specific message instead of relying only on the downstream lease, and the duplicate trace sets it.

### Validation

- `./p-models/scripts/check.sh` → `tcMailboxCorrelationKeyFIFO` 0 bugs, `tcMailboxLiveness` 0 bugs, bridge tests `ok`
- `tcMailboxMonitorCatchesLegacyReorder` and `tcMailboxLegacyReorderCounterexample` both find the expected bug
- `gofmt`/`go vet` clean on the bridge

Mind taking another look?

PR #437, opened 2026-05-15
multi: Harden CI flake paths

Roasbeef:
## Summary

- keep anchor-parent broadcasts retryable when CPFP setup fails and the direct parent fallback is rejected
- add a txconfirm regression test for the retryable direct-reject fallback

## Context

This branch is rebased on the latest `origin/main`, which already includes the bitcoind startup/readiness retry from #434. The remaining diff here is the txconfirm hardening that prevents a direct fallback rejection from turning a CPFP setup failure into a terminal tracked-tx failure.

## Validation

- `make -C client fmt-changed`
- `go test ./txconfirm` in `client/`
- `make -C client lint-native`


Roasbeef:
@claude review this 

@codex review this

PR #365, opened 2026-05-12
walletdk: add embedded wallet SDK and TUI

sputn1ck:
## Summary

- add `sdk/walletdk`, a wallet-shaped facade over an embedded darepod runtime
- wire daemon-owned swap runtime mailbox events for walletdk receive/send flows
- add a Bubble Tea `walletdk-tui` for manual wallet, swap, accounting, and log testing
- document basic walletdk integration patterns for developers and LLMs

## Test Plan

- `make fmt-changed-check`
- `make commitmsg-lint range=origin/main..HEAD`
- `go test -tags swapruntime ./sdk/walletdk ./cmd/walletdk-tui ./swapclientserver`
- `make lint`

PR #440, opened 2026-05-15
swapwallet: add daemon-side walletrpc subserver + darepocli wallet subtree

Roasbeef:
In this PR, we add a new `swapwallet` package and a `walletrpc` gRPC
subserver that exposes a simplified, swap-vocabulary-free wallet API on
top of the running darepod daemon. The goal is to hide the swap concept
from first-party CLI and daemon users entirely: today you have to know
about swap-in vs swap-out, choose a payment direction, manually
`darepocli swap resume` after a daemon restart, and the CLI itself ends
up managing an ephemeral `btcec.NewPrivateKey` invoice signing key. With
walletrpc compiled in, you just `darepocli wallet send <invoice>` or
`darepocli wallet send <bcrt1...>` and the daemon owns every detail
underneath.

This PR builds on top of #365 (walletdk SDK + TUI) and inherits PR #339's
server-driven same-Ark p2p vHTLC routing: invoice sends always go through
`StartPay` and the swap server transparently picks same-Ark p2p vs real
Lightning per its own routing. The wallet layer never surfaces that
distinction. See each commit message for a detailed description w.r.t the
incremental changes.

## Integration model

`walletrpc` is the single build tag that integrates `swapwallet` into the
main daemon. Without the tag, `darepod` is byte-for-byte identical to
today: no new RPC surface, no new package-level imports, no extra
background goroutines. With `-tags walletrpc,swapruntime` (the walletrpc
tag requires swapruntime because swapwallet composes the daemon-owned
swap subsystem), the daemon gains the `WalletService` RPC and the
`swapwallet.Runtime` lifecycle owner. The paired-file pattern mirrors
`swapruntime{,_stub}.go`: `cmd/darepod/walletrpc.go` carries the real
implementation, `walletrpc_stub.go` carries a no-op `configureWalletRPC`,
and `main.go` calls the hook unconditionally next to `configureSwapRuntime`.

The walletrpc registrar runs after the swapclientserver registrar so the
swap subserver has already published its in-Go `Backend` handle (a new
`darepod.SwapBackend` interface) on `cfg.Swap.Backend` by the time
`swapwallet.Register` reads it. The walletrpc layer flips
`cfg.Swap.SuppressResume` so the wallet runtime becomes the sole owner of
the unified resume sweep; the swap subserver skips its own. Default
behavior is unchanged for swapruntime-only builds because SuppressResume
defaults to false.

## Lifecycle ownership

`swapwallet/runtime.go` owns the full swap lifecycle in-process. On
`Register` it runs a synchronous resume sweep *before* the gRPC server
begins accepting calls, so the daemon never surfaces a PENDING entry it
is not actively driving. A `deadlineWatcher` goroutine consults a
configurable wallet-level deadline (default 30 min) and overlays stuck
entries as `FAILED / failure_reason="timed_out"` so a hung swap never
leaves the user surface waiting. A `monitorLoop` consumes
`SwapService.SubscribeSwaps` in-process through a small `inlineSwapStream`
shim that adapts the gRPC server-stream interface into a Go callback,
normalizes each update into a flat `WalletEntry`, applies the overlay,
and fans the result to `SubscribeWallet` subscribers. The loop is
resilient: transient subscribe failures back off exponentially
(500ms → 30s) and only rootCtx cancellation exits it. Every background
goroutine anchors to the daemon rootCtx, never to RPC-call contexts, so
an RPC client disconnect can never cancel in-flight work.

## API shape

`WalletEntry` is the single flat row type every wallet history surface
returns. It collapses every backing FSM into `PENDING / COMPLETE / FAILED`
and drops every internal correlator (session_id, round_id, settlement_type,
mailbox subtype) so the user surface stays uniform across SEND, RECV,
DEPOSIT, and EXIT rows. A `failure_reason` field carries one human string;
no coded enum tree.

`Send` dispatches by destination oneof. BOLT-11 invoices route through
`swapclientserver.StartPay`. Onchain addresses route through the existing
`LeaveVTXOs` cooperative-exit RPC (`daemonrpc/daemon.proto:91`); no new
proto or handler additions on the daemon side, swapwallet just composes
over what's already there. We coin-select VTXOs whose sum covers the
requested amount; `amt=0` sweeps all live VTXOs. LeaveVTXOs has full-VTXO
sweep semantics so the recipient may receive slightly more than `amt_sat`.
That's documented as a v1 limitation with a follow-up OOR pre-split being
the obvious refinement.

`Recv` composes `swapclientserver.StartReceive`. The invoice is signed
with a daemon-managed payment-scoped auth key resolved through PR #337's
receive-auth surface; the wallet layer never generates or holds raw keys.
We also harden the daemon-side `daemonInvoiceGenerator` to assert that the
no-key `CreateInvoice` path can no longer be hit silently: a small
`daemonAuthOnlyInvoiceCreator` wrapper rejects it with an explicit error,
so a future regression fails loudly instead of falling back to ephemeral
plumbing-key signing.

`List` merges the swap subserver's `ListSwaps`, the daemon's unified
`ListTransactions` (ledger + boarding_sweeps already joined server-side),
and the runtime's overlay into one `WalletEntry` stream sorted by
`updated_at` descending and paginated. A canonical-id intent map indexed
by payment_hash, txid, outpoint, and address projects future ledger rows
onto the same id the caller saw at Send / Deposit / Recv submit time, so
polling the history surface produces stable correlation across the
lifecycle.

## CLI

`darepocli wallet send|recv|list|deposit|status` lives alongside the
existing `wallet create|unlock|balance|newaddress` admin subcommands.
Users get one wallet vocabulary: backend admin and everyday spend/receive
live side by side under one parent. The new subcommands are gated by the
`walletrpc` build tag; a no-op `addWalletRPCSubcommands` hook in the
unflagged build keeps the parent compilable. Each new subcommand is a
thin gRPC client wrapper. No key generation, no BOLT-11 parsing beyond a
cheap prefix sniff to pick the destination oneof, no SQLite access, no
manual swap-resume.

The legacy `swap *` commands stay reachable as the `swapdirect` build's
standalone testing path; we add an inline comment in `cmd_swap.go`
redirecting production users to the daemon-backed path.

## Test plan

- [x] `make rpc` regenerates `rpc/walletrpc/*.pb.go` cleanly via Docker.
- [x] Build matrix: default, `swapruntime`, `walletrpc,swapruntime`, `swapdirect` all build.
- [x] `go test -tags walletrpc,swapruntime ./swapwallet/` (26 tests covering router dispatch, recv, history merge + normalization + overlay + canonical-id projection + pagination, runtime deadline watcher, monitor fan-out + canonical-id projection + transient-failure recovery + root-cancel exit, service Deposit/Balance/Status composition).
- [x] `go test -tags swapruntime ./swapclientserver/` (existing tests pass; new tests for the auth-only invoice wrapper and the exported `ResumePending`/`SwapBackend` contract).
- [x] `go test ./darepod/` (no regressions).
- [x] Production-readiness greps: zero `btcec.NewPrivateKey` or `secp256k1.GenerateKey` inside `swapwallet/` or `cmd_wallet_rpc.go`; zero internal correlator names leaking from the proto into the user surface.
- [ ] End-to-end with a live swapdk-server and two operators on the same Ark, exercising the seven scenarios in the plan (recv+external-pay, send-invoice-LN, send-invoice-p2p, send-onchain, restart-resume, deadline-timeout, deposit-balance). Needs a setup we don't have in-CI; will run manually before merge.


sputn1ck:
@Roasbeef pushed some commits on top to make walletdk be the native wrapper over walletrpc

Roasbeef:
@claude review this 

@codex review this

Roasbeef:
@claude review this 

@codex review this

PR #449, opened 2026-05-15
doc: add sample darepod config

### LINKED ISSUE #446: doc: sample daemon configuration `sample-darepod.conf`
Ideally we have a `sample-darepod.conf`.

bhandras:
## Summary

Fixes #446.

This PR adds a `sample-darepod.conf` modeled after lnd's sample config pattern, with every daemon config key commented out and set to the current default value. The sample is intentionally written as top-level `key=value` entries, because darepod now reads `~/.darepod/darepod.conf` as a simple properties-style config file.

The daemon previously had comments describing config-file precedence but never loaded a config file. This PR wires `darepod` to load the default config file when it exists, ignore a missing default file for first-run ergonomics, and fail if the user explicitly passes a missing `--configfile` path or `DAREPOD_CONFIGFILE` path. Flags and environment variables remain higher-priority than file values.

## Sample Config Gate

Adds `make sample-conf-check`, modeled on lnd's `sample-conf-check`, and runs it in CI static checks. The checker reflects `darepod.Config` `mapstructure` keys, includes daemon-only flags such as `bitcoind.*`, accounts for explicit aliases like `oor.limits.max-checkpoints -> oor.limits.maxcheckpoints`, and verifies that `sample-darepod.conf` contains the expected default value for each option. This should make future daemon config additions fail CI unless the sample config is updated at the same time.

## Tests

- `make fmt-changed`
- `make sample-conf-check`
- `go test ./cmd/darepod ./darepod`
- `make lint-changed-local`
- `make tidy-module-check`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,6 +108,14 @@ jobs:
           GOWORK: off
 
       ########################
+      # Sample configuration check
+      ########################
+      - name: Check sample darepod config
+        run: make sample-conf-check
+        env:
+          GOWORK: off
+
+      ########################
       # SQLC code gen check
       ########################
       - name: Docker image cache
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: sqlc sqlc-check migrate-create migrate-up migrate-down gen
-.PHONY: lint lint-source lint-local lint-source-local lint-changed-local lint-native build-native-linter local-custom-gcl install-custom-gcl docker-tools fmt fmt-changed fmt-check fmt-changed-check tidy-module tidy-module-check schema-check doc-check
+.PHONY: lint lint-source lint-local lint-source-local lint-changed-local lint-native build-native-linter local-custom-gcl install-custom-gcl docker-tools fmt fmt-changed fmt-check fmt-changed-check tidy-module tidy-module-check schema-check doc-check sample-conf-check
 .PHONY: ast-lint ast-grep-fix
 .PHONY: unit unit-cover unit-race unit-swapruntime check-go-version build install clean release
 .PHONY: build build-swapruntime build-swapclient rpc install install-swapruntime help clean-networks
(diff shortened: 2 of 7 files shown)


bhandras:
@claude review

Context for review: this PR fixes #446 by adding `sample-darepod.conf`, making the daemon load `~/.darepod/darepod.conf` when present, and adding a CI gate that checks sample config coverage/defaults against `darepod.Config` mapstructure keys plus daemon CLI flags. One deliberate choice is that the sample file uses top-level `key=value` entries instead of lnd-style INI sections, because the daemon reads this file directly and the dotted keys need to map cleanly into Viper/mapstructure.

bhandras:
@claude thanks, fixed and folded into the single commit. Changes since your review:

- `readPropertiesConfig` now strips whitespace-prefixed inline comments from values, so `network=regtest # local development` loads as `regtest`.
- Added `TestReadConfigFileRejectsMissingEnvPath` to cover the `DAREPOD_CONFIGFILE` missing-path branch.
- The sample config checker now rejects any live non-comment line in `sample-darepod.conf`, so accidental uncommented defaults cannot slip through.
- Added comments explaining the checker's literal flag-registration assumption and the two-pass help/source flag validation.
- `readConfigFile` now stores the expanded `configfile` path back into Viper after `~` expansion.

Re-ran locally:
- `make fmt-changed`
- `make sample-conf-check`
- `go test ./cmd/darepod ./darepod`
- `make lint-changed-local`
- `make tidy-module-check`

Please re-review.

bhandras:
@claude review

Re-review request after folded fixes for the previous Claude feedback. Since the first review, this commit now strips whitespace-prefixed inline comments in `readPropertiesConfig`, adds `TestReadConfigFileRejectsMissingEnvPath`, rejects live non-comment lines in `sample-darepod.conf`, documents the sample checker regex/two-pass assumptions, and stores the expanded `configfile` path back into Viper. Local checks listed in the previous comment are green.

PR #456, opened 2026-05-15
serverconn: sign TLS leaf binding for mailbox auth (#448)

ellemouton:
Closes lightninglabs/darepo#448.

## Summary

Adds the client-side half of the #448 TLS-binding mTLS gap fix.

- Introduces `SignMailboxTLSBind` / `VerifyMailboxTLSBind` in
  `client/serverconn`. The signature is BIP-340 over a tagged
  (`mailbox-tls-bind`) digest of `senderPubKey || tlsLeafSPKI`, using
  the mailbox secp256k1 key.
- `darepod.dialServer` stamps the binding signature as a gRPC metadata
  header (`x-mailbox-tls-bind-sig`) on outbound mailbox calls so the
  server can verify, on first-contact Send, that the mailbox-key
  holder chose the TLS leaf the server observes.

Without this header an attacker who captured a victim's signed first
Send could replay the bytes over a TLS connection backed by an
attacker-controlled leaf and have their fingerprint bound to the
victim's mailbox ID, defeating the post-registration fingerprint
defense added in lightninglabs/darepo#443.

## Server side

Server-side enforcement lives in lightninglabs/darepo#443, which also
folds in the #448 fix to keep the rollout atomic (a soft-rollout flag
`MailboxConfig.RequireTLSBindingSig` gates strict enforcement).

## Test plan

- [x] new unit tests for `SignMailboxTLSBind` / `VerifyMailboxTLSBind`
  (wrong-key, wrong-leaf, malformed, replay across leaves)
- [x] `darepod.dialServer` stamps the header end-to-end

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -219,6 +219,20 @@ type Server struct {
 	// so response envelopes can include it without re-computing.
 	authSigHex string
 
+	// tlsLeafSPKI is the DER-encoded SubjectPublicKeyInfo of the
+	// P-256 client TLS leaf certificate this daemon dialed with.
+	// It is captured during dialServer and used to compute the
+	// secp256k1 → TLS-leaf binding signature that the server
+	// verifies against the leaf it observes on the connection
+	// (issue #448). Empty when TLS is disabled (Server.Insecure).
+	tlsLeafSPKI []byte
+
+	// tlsBindSigHex caches the hex-encoded Schnorr signature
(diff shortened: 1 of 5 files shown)


ellemouton:
Superseded by consolidated PR https://github.com/lightninglabs/darepo-client/pull/459. Closing to reduce CI load.

Issue #128, opened 2026-02-26
lwwallet: add re-org detection to chain backend


sputn1ck:
## Context

The `lwwallet.ChainBackend` polling loop detects new blocks by comparing
tip heights but doesn't handle chain reorganizations. If a re-org occurs:

1. Confirmation registrations that were fulfilled based on a now-stale
   block won't be re-evaluated (the registration is deleted after
   fulfillment).
2. Spend registrations similarly won't detect that a spend was reverted.
3. Block epoch subscribers won't receive `BlockDisconnected` notifications.

The `EsploraChainService` has the same limitation — it sends
`BlockConnected` notifications but never `BlockDisconnected`, so
btcwallet's internal state can diverge from the chain on a re-org.

## Proposed Solution

1. **Detect re-orgs** in the poll loop by comparing the previous tip hash
   with the new tip's parent hash. If they don't match, walk back to find
   the fork point.
2. **Re-evaluate fulfilled registrations** by keeping a bounded history
   of recently fulfilled conf/spend registrations that can be re-checked
   after a re-org.
3. **Send `BlockDisconnected` notifications** via EsploraChainService so
   btcwallet can roll back its internal state.

## References

- PR #106 review comment by @Roasbeef (chain_backend.go:392): "What about
  re-orgs?"
- PR #106 review comment by @bhandras (chain_backend.go:383): "I don't
  think this is correct as is, as on confirmation we'll remove conf and
  spend registrations."

ellemouton:
Closing: verified against the current tree.

Fixed by #895 (multi: reorg-aware chain observation (chainsource + backends)).

[...]

PR #461, opened 2026-05-15
multi: reorg-aware lwwallet/Esplora chain backend (C0b)

### LINKED ISSUE #107: lwwallet: production hardening gaps
## Context

The `lwwallet` package (#106) provides a lightweight wallet for Ark client participation. While architecturally sound for its purpose, a comparison with btcwallet/lnd identified several gaps that should be addressed before production use.

### 1. Reorg awareness in chain backend

**Current**: The Esplora-based chain backend does not detect when a previously confirmed transaction becomes unconfirmed due to a chain reorganization.

**Impact**: If a boarding transaction confirms in block N, then a reorg removes that block, the wallet still considers the tx confirmed. This can lead to the client attempting round participation with a VTXO backed by an unconfirmed boarding tx.

**Proposed fix**: Track the last-seen tip hash. On each poll cycle, compare the new tip hash against the previous. If the tip hash changes but the height doesn't increase (or decreases), detect the reorg and re-check all pending/recently-confirmed registrations. At minimum, re-query confirmation status for transactions confirmed in the last N blocks.

### 2. Persistent key index tracking

**Current**: The keyring tracks `nextIndex` per key family in memory only. `GetState()`/`RestoreState()` are available for manual persistence, but if the process crashes between `DeriveNextKey()` and the caller persisting the result, the index counter is lost.

**Impact**: On restart, the keyring might reuse a previously derived index (if `RestoreState` is called with stale data) or skip ahead (if the caller persists aggressively). Either way, there's a window for index desync.

**Proposed fix**: Store `nextIndex` per family in the existing client SQL database (a simple `keyring_state` table with `family` and `next_index` columns). Update atomically with the operation that uses the derived key.

### 3. Secure key zeroing

**Current**: Derived private keys are never explicitly zeroed in memory after use. The `LocalSigner.SignOutputRaw`, `signWalletInputs`, and MuSig2 session methods all hold private keys that persist in memory until GC.

**Impact**: In a memory dump or side-channel attack, private keys could be recovered. btcwallet uses `internal/zero` to scrub key material.

**Proposed fix**: Zero private key bytes after use in signing paths. Use `btcwallet/internal/zero.Bytea32()` or equivalent. Key areas:
- `LocalSigner.SignOutputRaw` after signing
- `LocalSigner.ComputeInputScript` after signing
- `Wallet.signWalletInputs` after signing each input
- MuSig2 session creation (after storing context)
- `Keyring.DerivePrivKey` callers

### 4. Fee estimation fallback

**Current**: If the Esplora `/fee-estimates` endpoint fails or returns empty data, `EstimateFeePerVByte` returns an error with no fallback.

**Impact**: Temporary Esplora API issues could block transaction creation entirely.

**Proposed fix**: Add a configurable default fee rate (e.g., 2 sat/vB) used when the API call fails. Log a warning when falling back.

## Non-issues (intentional design)

These were identified in the comparison but are **not** gaps for the Ark client use case:

- **No walletdb/BoltDB**: The client already has its own SQL database. Adding BoltDB would be a second persistence layer with no benefit.
- **No wtxmgr transaction lifecycle**: The client doesn't manage channels or need full transaction history. VTXO state is tracked at the protocol level.
- **No waddrmgr encryption**: Seed encryption is the caller's responsibility. The wallet receives a pre-decrypted seed. This is documented in `Config.Seed`.
- **No rescan/recovery**: Ark protocol state is stored server-side. Client recovery is protocol-level, not wallet-level.
- **No multiple chain backends**: Esplora is the target backend for lightweight clients by design.

ellemouton:
## Summary

Makes the lwwallet/Esplora-backed \`chainsource.ChainBackend\` satisfy the
same reorg-aware contract the LND-backed adapter ships in #422:

- \`ConfRegistration\` lifecycle: \`Confirmed → Reorged → Confirmed → Done\`.
- \`SpendRegistration\` lifecycle: \`Spend → Reorged → Spend → Done\`.
- \`Done\` is synthesized at the chainsource actor layer from block epochs;
  the backend keeps registrations alive past first positive event.

Without this, any darepod running on the lightweight wallet backend was
silently reorg-blind below the chainsource layer — meaning all the
reorg-handling work in #410 (unroll rollback) and the extras stack
(round Option B, boarding-sweep lifecycle, fraud invariants) would never
observe a reorg event for lwwallet deployments.

This is workstream **C0b** from epic lightninglabs/darepo#454. Branched
off \`reorg-safe-chainsource\` (#422), independent of #410 and the extras
stack — can land in parallel.

## Changes

- \`tip_poller.go\`: bounded \`height → hash\` ring buffer
  (\`DefaultHashHistorySize = 100\`), sibling \`EventServer[*ReorgEvent]\`
  reachable via \`SubscribeReorgs\` / \`BestBlockAndSubscribeAll\`. Three-case
  poll cycle catches same-height hash drift via \`GetBlockHashByHeight\` at
  the current tip and deeper reorgs via raw-header \`PrevBlock\` continuity
  on the first new height.
- \`chain_backend.go\`: per-registration \`stateWatching ↔ statePositive\`
  state machine with cached last block hash. Reorg handler matches
  cached hash against \`ReorgEvent.Disconnected\`, fires \`Reorged\`
  non-blocking, resets state, and re-checks so a re-confirmation /
  re-spend can fire in the same handler turn. \`Cancel()\` is idempotent.
- \`chain_backend_reorg_test.go\`: 7 new tests covering same-height drift,
  conf reorg round-trip, conf reorg with eviction (no re-confirm), spend
  reorg round-trip, deeper multi-height reorg, cancel cleanup with
  goroutine accounting, and registration-stays-alive after first event.
- \`AGENTS.md\` / \`CLAUDE.md\`: removed the "same-height reorgs invisible"
  invariant; documented the new reorg-aware contract.

## Test plan

- [x] \`go test -race ./lwwallet/... ./chainsource/... ./txconfirm/... ./chainbackends/...\`
- [x] Each commit builds and \`go vet\`s clean in isolation.
- [x] \`make lint-native\` clean (0 issues).
- [x] \`make fmt-changed-check\` clean.
- [x] \`make commitmsg-lint\` clean across all 4 commits.

## Out of scope

- BatchCanonicalityManager (C2/C3 — separate workstream).
- Server-side batchwatcher (S1/S2 — separate epic).
- Probe-vs-registration API split (deliberately deferred).
- \`SpendDetail.SpendingBlockHash\` chainsource public API addition
  (worked around with an internal cache; possible follow-up).

## Related

- Epic: lightninglabs/darepo#454
- Closes: lightninglabs/darepo-client#128 (lwwallet reorg detection)
- Part of lightninglabs/darepo-client#107 (item 1: Esplora reorg awareness; gaps 2-4 — key-index persistence, key zeroing, fee fallback — remain, so this PR must NOT auto-close #107)
- Depends on: #422 (chainsource reorg-aware substrate)

### DIFF
--- a/lwwallet/chain_backend.go
+++ b/lwwallet/chain_backend.go
@@ -16,6 +16,7 @@ import (
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/btcsuite/btclog/v2"
 	"github.com/lightninglabs/darepo-client/chainsource"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
 	"golang.org/x/sync/singleflight"
 )
 
@@ -42,6 +43,29 @@ const (
 // raise Esplora load over the default poll cadence.
 const recheckHeartbeatInterval = 60 * time.Second
 
+// regState is the lifecycle state of a conf/spend registration.
+//
+// stateWatching: no positive event has been delivered yet (or the
+// last positive event was reorged out). The next checkSingle... that
+// finds the tx confirmed / outpoint spent will emit Confirmed /
+// Spend and transition to statePositive.
+//
+// statePositive: a Confirmed / Spend event has been delivered and the
+// associated block-hash is cached on the registration. The next
+// checkSingle... that does NOT find the tx confirmed / outpoint
+// spent in the same block leaves the registration alone (the
+// chainsource actor handles Done synthesis at FinalityDepth via
+// block epochs). A reorg event that names the cached block-hash in
+// its Disconnected set fires Reorged, resets cached state, and
+// transitions back to stateWatching so the next re-check can fire
+// Confirmed / Spend again on the new chain.
+type regState uint8
+
+const (
+	stateWatching regState = iota
+	statePositive
+)
+
 // confRegistration tracks a pending confirmation registration within the
 // polling loop.
 type confRegistration struct {
@@ -63,8 +87,37 @@ type confRegistration struct {
 	// confChan is the channel to send the confirmation on.
 	confChan chan *chainsource.TxConfirmation
 
+	// reorgChan is the channel that fires when a previously delivered
+	// confirmation is reorged out of the canonical chain. Buffered to
+	// 1; never written by the backend before stateWatching has been
+	// re-entered. The value is the chainsource ordering sequence; this
+	// backend does not stamp sequences (it emits conf/reorg from a
+	// single ordered handler goroutine), so it always sends 0, which
+	// the chainsource actor treats as always-apply.
+	reorgChan chan uint64
+
+	// doneChan is allocated for API symmetry with the chainsource
+	// contract. The Esplora-backed backend does not write to it: the
+	// chainsource ConfActor synthesizes Done at FinalityDepth from
+	// block epochs once the backend stops re-firing events.
+	doneChan chan struct{}
+
 	// cancelCh signals that this registration has been cancelled.
 	cancelCh chan struct{}
+
+	// regMu guards the fields below that mutate over the
+	// registration's lifetime. It is held only briefly during state
+	// transitions; channel sends are performed without it.
+	regMu sync.Mutex
+
+	// state is the current lifecycle state.
+	state regState
+
+	// lastBlockHash is the BlockHash of the last delivered
+	// confirmation when state == statePositive. Used to detect when
+	// a reorg's Disconnected hash list invalidates this
+	// registration's last event.
+	lastBlockHash chainhash.Hash
 }
 
 // spendRegistration tracks a pending spend registration within the
@@ -82,8 +135,34 @@ type spendRegistration struct {
 	// spendChan is the channel to send the spend detail on.
 	spendChan chan *chainsource.SpendDetail
 
+	// reorgChan is the channel that fires when a previously delivered
+	// spend is reorged out of the canonical chain. Buffered to 1.
+	reorgChan chan uint64
+
+	// doneChan is allocated for API symmetry with the chainsource
+	// contract. The Esplora-backed backend does not write to it: the
+	// chainsource SpendActor synthesizes Done at FinalityDepth.
+	doneChan chan struct{}
+
 	// cancelCh signals that this registration has been cancelled.
 	cancelCh chan struct{}
+
+	// regMu guards the fields below.
+	regMu sync.Mutex
+
+	// state is the current lifecycle state.
+	state regState
+
+	// lastSpenderHash is the SpenderTxHash of the last delivered
+	// spend when state == statePositive.
+	lastSpenderHash chainhash.Hash
+
+	// lastSpendingBlockHash is the block hash that the last
+	// delivered spend confirmed in, parsed from the same
+	// outspend response that confirmed the spend. Reorg
+	// comparison against ReorgEvent.Disconnected uses this so
+	// the spend-watch path does not need an extra HTTP round-trip.
+	lastSpendingBlockHash chainhash.Hash
 }
 
 // blockRegistration tracks a block epoch subscription.
@@ -205,7 +284,8 @@ func (b *ChainBackend) Start() error {
 		}
 	}
 
-	height, hash, _, sub, err := b.tipPoller.BestBlockAndSubscribe()
+	height, hash, _, chainSub, err :=
+		b.tipPoller.BestBlockAndSubscribeChain()
 	if err != nil {
 		return fmt.Errorf("subscribe to tip poller: %w", err)
 	}
@@ -216,7 +296,7 @@ func (b *ChainBackend) Start() error {
 	b.mu.Unlock()
 
 	b.wg.Add(1)
-	go b.handleTipEvents(sub)
+	go b.handleChainEvents(chainSub)
 
 	b.log.InfoS(context.Background(), "Chain backend started",
 		slog.Int("tip_height", int(height)),
@@ -388,11 +468,20 @@ func (b *ChainBackend) SubmitPackage(ctx context.Context, parents []*wire.MsgTx,
 }
 
 // RegisterConf registers for confirmation notifications of a transaction.
+// The registration is reorg-aware: the returned ConfRegistration's
+// Reorged channel fires when a previously delivered confirmation is
+// reorged out of the canonical chain, and the Confirmed channel may
+// fire again when the tx re-confirms on the new chain. The Done
+// channel is allocated but never written by this backend; the
+// chainsource ConfActor synthesizes Done at its configured
+// FinalityDepth using block epochs.
 func (b *ChainBackend) RegisterConf(ctx context.Context, txid *chainhash.Hash,
 	pkScript []byte, numConfs uint32, heightHint uint32,
 	includeBlock bool) (*chainsource.ConfRegistration, error) {
 
 	confChan := make(chan *chainsource.TxConfirmation, 1)
+	reorgChan := make(chan uint64, 1)
+	doneChan := make(chan struct{}, 1)
 	cancelCh := make(chan struct{})
 
 	reg := &confRegistration{
@@ -402,6 +491,8 @@ func (b *ChainBackend) RegisterConf(ctx context.Context, txid *chainhash.Hash,
 		heightHint:   heightHint,
 		includeBlock: includeBlock,
 		confChan:     confChan,
+		reorgChan:    reorgChan,
+		doneChan:     doneChan,
 		cancelCh:     cancelCh,
 	}
 
@@ -424,11 +515,7 @@ func (b *ChainBackend) RegisterConf(ctx context.Context, txid *chainhash.Hash,
 	)
 
 	cancelFn := func() {
-		close(cancelCh)
-
-		b.mu.Lock()
-		delete(b.confRegs, id)
-		b.mu.Unlock()
+		b.cancelConfReg(id, reg)
 	}
 
 	// Run an immediate single-shot check scoped to JUST this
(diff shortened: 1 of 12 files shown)


lwwallet/tip_poller.go
```diff
@@ -349,44 +556,337 @@ func (t *TipPoller) poll() {
...
-		// subscribe.Server is shutting down; log and exit so
-		// we do not advance the cached tip past an event we
-		// failed to fan out.
-		if sendErr != nil {
+		if !ok {
+			continue
+		}
+
+		disconnected = append(disconnected, hash)
+	}
```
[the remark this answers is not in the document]
ellemouton: Addressed in b749924 — single critical section, ascending iteration removes the sort, and prune-walk reuses the same `[forkHeight+1, cachedTipHeight]` range so the height list goes away entirely.

lwwallet/chain_backend.go
```diff
@@ -579,11 +749,55 @@ func (b *ChainBackend) runSpendOneShot(id uint64, reg *spendRegistration) {
...
+	// the relevant ones — correctness is preserved.
+	var spendingBlockHash chainhash.Hash
+	if detail.SpendingHeight > 0 {
+		hash, err := b.esplora.GetBlockHashByHeight(
+			context.Background(), detail.SpendingHeight,
+		)
+		if err == nil {
+			spendingBlockHash = hash
+		}
+	}
```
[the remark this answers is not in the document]
ellemouton: Addressed in 2db83e6 — `checkSingleSpend` now returns the spending block hash parsed from `outspend.Status.BlockHash` alongside the `SpendDetail`, and `deliverSpendIfNew` consumes it directly. The extra `GetBlockHashByHeight` round-trip is gone.

levmi:
**Triage linkage check** — verified this PR fully implements all three of #128's proposed fixes (poll-loop reorg detect, re-evaluate fulfilled registrations, `BlockDisconnected`), each with test coverage. Note the current `- Closes: lightninglabs/darepo-client#128` line won't trigger auto-close — the `Closes:` (with colon) form isn't a valid keyword — so #128 won't close on merge as written.

Suggestions:
- Use a parseable **`Closes #128`**.
- For #107, only item 1 of 4 is addressed here → **`Refs #107`** (not Closes), so the remaining items stay open.

lwwallet/tip_poller.go
```diff
@@ -38,6 +46,62 @@ type TipBlock struct {
...
 
+// ReorgEvent describes a chain reorganization observed by the
+// TipPoller. The poller emits one ReorgEvent every time it detects
+// that one or more previously broadcast blocks are no longer in the
+// canonical chain. Disconnected hashes are listed in canonical
+// (low-height first) order; Connected blocks are the new tip's path
+// from ForkHeight+1 onward, also in canonical order. Either slice may
+// be empty: a same-height hash-replacement reorg has Disconnected of
+// length 1 and Connected of length 1.
+type ReorgEvent struct {
```
Roasbeef, line 57: 👍

lwwallet/tip_poller.go
```diff
@@ -38,6 +46,62 @@ type TipBlock struct {
...
+// embedded EventServer delivers updates in SendUpdate order to a
+// single subscriber goroutine, so a downstream consumer that
+// dispatches on event type from one channel observes the same order
+// the TipPoller emitted. That is the load-bearing property
+// downstream needs to emit BlockDisconnected before BlockConnected
+// for the replacement chain (btcwallet's disconnectBlock requires
+// it; chainsource finality synthesis requires it too).
+type ChainEvent struct {
+	Reorg *ReorgEvent
+	Tip   *TipBlock
```
Roasbeef, line 98: Are these fields always preseent? If so, can ditch the pointer perhaps.

Roasbeef: Or is it more of an either/fn situation?

lwwallet/tip_poller.go
```diff
@@ -170,12 +282,60 @@ func (t *TipPoller) Start() error {
...
+				"Tip poller history seed fetch failed",
+				err,
+				slog.Int("height", int(h)),
+			)
+
+			break
+		}
+
+		t.mu.Lock()
+		t.recordHashLocked(h, liveHash)
```
Roasbeef, line 335: Can use the lnd wrapepr around sync map here (type safe sync map)

lwwallet/tip_poller.go
```diff
@@ -257,6 +445,85 @@ func (t *TipPoller) BestBlockAndSubscribe() (int32, chainhash.Hash, time.Time,
...
+
+	reorgSub, err := t.reorgs.Subscribe()
+	if err != nil {
+		sub.Cancel()
+
+		return 0, chainhash.Hash{}, time.Time{}, nil, nil,
+			fmt.Errorf("subscribe to reorg events: %w", err)
+	}
+
+	return t.tipHeight, t.tipHash, t.tipTime, sub, reorgSub, nil
```
Roasbeef, line 473: Can return the first 3 vars as an independent struct

lwwallet/tip_poller.go
```diff
@@ -58,18 +122,46 @@ type TipPoller struct {
...
 	tipHash   chainhash.Hash
 	tipTime   time.Time
 
+	// recentHashes maps a recent canonical-chain height to the hash
+	// the poller broadcast for it. It is bounded to historySize
+	// entries: on every insert we prune any entry whose height is
+	// more than historySize below the current tip. The buffer lets
+	// reorg detection walk back to the fork point using the cached
+	// hashes rather than re-fetching the entire pre-fork chain.
+	recentHashes map[int32]chainhash.Hash
```
Roasbeef, line 163: Hmm, not really a ring buffer persay. For that it'd be a fixed array with an index mod pointer into the array. You'd take a height, then map that relative to the array. 

Just food for thought re strict semanatics

lwwallet/tip_poller.go
```diff
@@ -298,23 +579,58 @@ func (t *TipPoller) poll() {
...
-	// matches the behavior of the per-component pollers that
-	// preceded the unified TipPoller and has historically been
-	// acceptable for lwwallet's confirmation-target use case
-	// (downstream callers re-check status against Esplora on every
-	// tip event, so a stale hash at height N converges within one
-	// extra tip advance). Documented here so it is not filed as a
-	// regression by a future reader.
-	if newHeight <= oldHeight {
+	switch {
+	case newHeight < oldHeight:
```
Roasbeef, line 586: Or it's a re-org that results in a shorter, but higher work chain.

lwwallet/tip_poller.go
```diff
@@ -298,23 +579,58 @@ func (t *TipPoller) poll() {
...
+		// The remote claims an older height than we have. This
+		// is almost always a transient indexer hiccup; treat it
+		// as no-op and re-check on the next tick.
 		return
+
+	case newHeight == oldHeight:
```
Roasbeef, line 657: Should test this behavior and the above against a live Eslpora instane

Roasbeef: We should also be able to cover all the scenarios as we envision with unit tests here

lwwallet/chain_backend.go
```diff
@@ -637,17 +837,230 @@ func (b *ChainBackend) RegisterBlocks(_ context.Context) (
...
+		// Fallback: the registration may have delivered against
+		// a block older than the poller's seeded hash history
+		// (e.g. on a fresh daemon where RegisterConf landed an
+		// immediate historical positive). The reorg's
+		// disconnected set is bounded by recentHashes, so a
+		// reorg deep enough to invalidate that historical block
+		// would not appear here. Re-query canonical status and
+		// compare against the cached block hash; if the tx is
+		// no longer confirmed in that block, treat as reorg.
+		current := b.checkSingleConf(reg, currentHeight)
```
Roasbeef: So if this return `nil`, then we fall thru thinking it's a re-org IIUC. I think we want to thread thru a proper error here or diff enum state (or fn.result) to be able to distinguish this state.

lwwallet/tip_poller.go
ellemouton: Kept the two-pointer struct here for now. The "exactly one of `Reorg`/`Tip` is non-nil per event" invariant is documented and the producer/consumer ordering is unit-tested, and the only two consumers (`ChainBackend` + `EsploraChainService`) sit on the load-bearing reorg-before-tip ordering path, so I'd rather not churn it inside this PR. Happy to convert it to an `fn.Either[*ReorgEvent, *TipBlock]` in a focused follow-up if you'd prefer the type-enforced version.

ellemouton: Re: the either/fn idea — agreed that's the cleaner encoding of the discriminated union. Holding off in this PR for the reason above (touches the ordering-critical consumers); will do the `fn.Either` conversion as a standalone follow-up.

lwwallet/tip_poller.go
ellemouton: `recentHashes` is guarded by `t.mu` together with `tipHeight`/`tipHash`/`tipTime` and is mutated via a range-and-prune under that same lock, so a standalone typed sync.Map would split the tip state across two synchronization domains and complicate the prune. Kept it as a mutex-guarded map for that reason.

lwwallet/tip_poller.go
ellemouton: Agreed it'd read better. The `(height, hash, time)` triple recurs across `BestBlock` / `BestBlockAndSubscribe` / `…All` / `…Chain` plus several cross-package callers, so introducing a `BestBlock` struct ripples fairly wide. Left it out to keep this PR focused; can do it as a standalone cleanup.

lwwallet/tip_poller.go
ellemouton: Good catch — it's a prune-on-insert bounded map, not a ring buffer. Renamed the comments (and the CLAUDE.md reference) to "bounded history map".

lwwallet/tip_poller.go
ellemouton: Fixed. `poll()` now compares the live hash at the reported height against the cached one when the remote tip is *lower*, so a reorg onto a shorter higher-work chain is routed through `handleReorg` instead of being treated as transient lag.

While adding coverage I also found and fixed a latent walk-back panic: on a forward reorg whose new tip extends past the cached tip, the old code mistook the first uncached height for the fork point and underflowed the `Disconnected` slice cap (`makeslice` panic in the poll loop). Added unit tests for all four scenarios (shorter, same-height, deeper-forward, transient-lag).

lwwallet/tip_poller.go
ellemouton: Left the live-Esplora exercise as a manual/integration follow-up — it needs a real endpoint plus a controllable reorg, so it doesn't fit the unit suite. Covered the logic with deterministic unit tests against the stubbed chain instead (see below).

ellemouton: Done — added TipPoller-level reorg unit tests covering the shorter higher-work chain, same-height replacement, deeper forward reorg, and the transient-lag no-op. There were none before.

lwwallet/chain_backend.go
ellemouton: Fixed. Added `confirmedBlockHash`, a tri-state lookup: `Some(hash)` when the tx is still confirmed (independent of `numConfs`), `None` when definitively unconfirmed, and a non-nil error when canonical status can't be determined right now. `reorgConfReg` now only fires `Reorged` on a definitive negative or a divergent block hash; on a transient backend error it leaves the registration in `statePositive` so a later tip/reorg event re-evaluates, instead of firing a spurious reorg.

litbot-9000:

ellemouton:

PR #468, opened 2026-05-18
[experimental] per-subsystem SQL durability (client)

bhandras:
# exploration: per-subsystem SQL durability (alternative shape to durable-actor framework)

## What this is

A narrower durability model for in-process subsystems: each FSM owns its state in domain SQL tables, with per-subsystem `<x>_effects` tables holding pending side effects. A state advance and the effect row that triggers the next step commit in the same transaction. Workers resume from domain tables on restart instead of replaying durable-actor mailboxes or decoding generic checkpoint payloads.

This is **not a proposal to merge**. It's an architecture exploration — a working alternative shape made concrete enough to reason about with the code in front of us, rather than in the abstract.

The durable-actor design is intentional and solves real problems: durable delivery, retries, leases, codec evolution, restart recovery. The framework's assumptions are right at distributed boundaries. The question this branch asks is narrower: **for subsystems that run inside the same Go process and already need durable FSM state, can we preserve the same restart-safety properties with a simpler, more inspectable model?**

A paired branch on the server repo completes the cross-repo picture; this PR is the client-side half.

---

## What changed

**One invariant:** a state advance and the effect row that triggers the next step commit in the same transaction.

**Effect tables per subsystem** (`client_round_effects`, `oor_client_effects`, `unroll_effects`, `wallet_effects`) with a common shape:
- `effect_type TEXT CHECK IN (...)` — typed enum, not opaque
- `status TEXT CHECK IN ('pending','claimed','done','dead')`
- Lease, owner, token columns for worker claiming
- Domain-stable idempotency key

**Typed domain columns** in place of TLV payloads. The clearest example is MuSig2 nonces: `client_round_nonce_state(pub_nonce, sec_nonce)` rather than a TLV bundle in a `payload BLOB`. Persist-before-send becomes one INSERT.

**Workers poll effect tables and switch on `effect_type`**. Producers don't know the worker exists; the only contract is the row.

**Framework files removed in this branch** (`baselib/actor/durable_*.go`, `baselib/actor/delivery_*.go`, `baselib/actor/outbox_publisher.go`, `baselib/actor/restart.go`, `baselib/actor/ask_response.go`, the `db/actordelivery` package). A `make sql-durability-check` CI gate uses `rg` to prevent re-introduction of:
- `DurableActor | DurableMailbox | DeliveryStore | RestartMessage | OutboxPublisher` in production code
- `WorkflowStore | workflow_sessions | workflow_artifacts | workflow_outbox` anywhere
- `payload BLOB` columns in migrations

**Operator runbook** at `docs/operator-runbook.md` — per-subsystem SQL queries for "what's stuck?", "what's retrying?", "what's in state X?".

Full design notes: [SQL Durability Refactor — Design](https://gist.github.com/bhandras/31802d5626877f423533e8ad15b9f605) (secret gist).

---

## What this doesn't touch

Worth saying upfront because the diff is large: nothing visible outside the process changes.

- **Wire protocol with the server**: unchanged.
- **gRPC interfaces**: unchanged.
- **Cross-process delivery semantics over the mailbox**: unchanged. The client's `mailbox_egress` (outbound to server) and `mailbox_ingress_cursors` (ack progress for inbound from server) are preserved.
- **Behaviour visible to a user** (when there are users): unchanged.

This is purely about how in-process client subsystems persist their FSM state and dispatch their own side effects. Nothing crosses a process boundary differently than it did before.

---

## What the framework is solving for, and where its apparatus pays off here

There's a structural observation worth surfacing that I think is actually the underlying *why* of this branch — more so than any single concrete concern like TLV or operator queries.

The durable-actor framework is built around distributed-systems semantics. Every Tell carries the cost of assuming: messages can be lost in flight, recipients can be offline, workers can die holding leases, sender and receiver can be on different code versions across an asymmetric deployment, codecs may need to evolve in the wild. From that assumption follows the apparatus — durable mailbox row, codec versioning, at-least-once delivery, lease management, dead-lettering, outbox-based response routing.

Those mechanics are correct and necessary for one specific case: the **client↔server mailbox**. That's a genuine network boundary with the genuine concerns — message loss, server-side offline, version skew across independent client/server deployments. The framework's apparatus is load-bearing there, and this branch **keeps it**. The client retains `mailbox_egress` and `mailbox_ingress_cursors`; durable delivery semantics across the network boundary are unchanged.

For everything else — Tells between client subsystems inside the same Go binary — the framework's apparatus is real, but **additional to state durability rather than a substitute for it**:

- **Crash mid-Tell.** The framework persists the mailbox row before delivery so a recipient crash leaves the message recoverable. State durability already gives you this: the producer's state advance commits in one tx with the effect intent; the consumer derives pending work by scanning state on startup. **The state-layer protection comes for free with any FSM that has to recover across restarts; the framework's mailbox layer adds a second copy of the same guarantee.**
- **Recipient handler failure.** The framework Nacks, retries with backoff, dead-letters after max attempts. The effect table has the same machinery in its columns — `status`, lease, `attempts`, `max_attempts`, `dead`. Per-subsystem, but ~10 lines of SQL per table rather than a shared framework.
- **Codec versioning.** The framework supports senders and receivers on different code versions. Not applicable in-process — client subsystems are built and deployed together as one binary, so versioning doesn't earn its keep here.
- **DurableAsk response routing.** The framework routes responses back via the outbox publisher. In the SQL shape the "response" is just another effect row the requesting subsystem polls for. The outbox publisher is solving a routing problem the effect table doesn't have.

So for the in-process majority, **the framework's apparatus is additional to state durability, not a substitute for it**. Any FSM-bearing subsystem has to persist its state regardless — that's how it recovers across restarts at all. Once that's in place, every in-flight crash scenario the mailbox row protects against is *already* protected at the state layer. The framework's machinery for in-process Tells isn't replacing state durability; it's stacked on top of it, duplicating protection that's already at the state layer.

That's Rule 5 of the design doc, stated structurally: *"One transaction; producer owns it."* The framework didn't get this wrong by accident — it generalized from the cross-process case (where durable messages are *the* mechanism for crash recovery, because there's no shared state to fall back to) to the in-process case (where shared state is right there, and once it's durable the message doesn't need to be). **This branch is undoing the generalization, not the underlying apparatus.**

### What's kept vs dropped, expressed this way

| Concern | Branch's treatment |
|---|---|
| Cross-process boundary to the server | **Kept.** Client retains its egress and ingress-cursor halves of the mailbox plumbing, durable delivery, codec versioning, ack cursors. This is the boundary where the framework's design assumptions hold. |
| In-process Tell between two client subsystems in the same binary | **Replaced.** State advance in SQL + an effect row that the consumer's worker picks up. No durable mailbox, no codec, no lease machinery — because there's no boundary the message has to survive. |

The arithmetic is roughly: the framework's apparatus is correct, kept exactly where it's load-bearing, removed where it was paying its cost without earning it.

### This isn't an argument that the framework was wrong

A reasonable read of "we removed the framework" is *"the framework was wrong."* That's not the claim. The framework's distributed-system primitives are well-built and would be the right call in a system where in-process Tells could legitimately fail — across processes, across nodes, across machines, across deployment-version skew. None of those conditions hold for the client's in-process flows.

The closer phrasing is: *the framework's design assumptions hold for one boundary; this branch keeps the framework's apparatus exactly there, and matches the simpler in-process case with a simpler mechanism.* A bigger codebase (HA, multi-tenant, cross-cluster, independent client/server deployments) might earn the framework back. The current one has one process per client, one DB per process, no users, and the cross-process boundary is already isolated to one place (the server mailbox) — so the generalization is paying its cost without earning much.

### A note on what replaces it

Worth saying explicitly because "no framework" can read more aggressively than intended: the pattern this branch uses **isn't really "no framework"** — it's *framework via the tools the Go ecosystem has already converged on*. The persistence layer leans on `sqlc` (compile-time codegen from SQL schemas and queries, type-safe, no reflection). The "framework" for the SQL durability shape is:

- The shared effect-table column shape (§5.1 in the design doc) — applied as a convention per subsystem
- `sqlc`-generated queries for the standard claim / lease / ack patterns
- CI grep gates (`make sql-durability-check`) that enforce the discipline rules
- Operator runbook that documents the SQL queries per subsystem

In a meaningful sense this is the same *kind* of thing the durable-actor framework is — typed persistence with codegen, schema as contract, generic worker patterns. It just lands on the tools Go has chosen for that job (`sqlc` + discipline + grep gates) rather than building a separate runtime framework on top. The durable-actor framework's role would be similar but for *messaging*. The question this branch poses is whether messaging needs to be durable in the same generic way for the in-process case, given that the state already is.

---

## The TLV evolution cost

The single thing that motivated me to actually try this branch is the cost of evolving TLV-encoded payloads. It's worth stating concretely before the comparison table, because it's the dimension where the trade is least close.

Under the framework, persisted state lives in TLV payloads inside `mailbox_messages`, `outbox_messages`, `fsm_checkpoints`, etc. That has consequences:

- **Adding a field** to a persisted message requires a backwards-compatible codec, handling unknown TLV types on the read path, and often dual code paths until in-flight messages drain.
- **Removing a field** is effectively blocked until every in-flight message has been consumed, because old messages still carry the field.
- **Renaming** requires a parallel codec path that lives indefinitely.
- **Production introspection** requires a decoder — you can't `SELECT * FROM fsm_checkpoints WHERE state = 'X'` and read what you get back. The shape is implicit in Go code, not visible in the schema.
- **Cross-version debugging** requires knowing which codec version produced a given payload, which means versioning the payload itself.

Compare with `ALTER TABLE`: it handles every row atomically, the schema is the contract, every column is grep-able, every state is visible in `psql`. The change is recorded in the migration that anyone can read in a year.

The design doc's Rule 1 ("no column named `payload`") is rejecting exactly this — the cost shows up later as TLV-evolution debt that's invisible at the time the codec is written.

---

## How the two shapes compare

Both shapes solve the same underlying problem: persist FSM state across restarts, deliver side effects exactly-once-ish, let subsystems coordinate. They distribute complexity differently; neither is "simple" in any absolute sense.

| Dimension | Durable Actor Framework | Per-Subsystem SQL |
|---|---|---|
| **Conceptual surface** | One model: actors, mailboxes, codecs, delivery, restart, ask/tell, outbox. Learned once, reused everywhere. | One rule: state + effect commit in one tx. Each subsystem invents its own table layout within that rule. |
| **Framework code** | Thousands of lines of generic machinery, amortized across subsystems. | None. The "framework" is `<x>_effects` table conventions + code-review discipline. |
| **Persisted shape** | TLV payloads — implicit in Go code, versioned via codec registry, not directly readable in production. | Typed columns — explicit in migrations, readable with `SELECT`, version-controlled with the rest of the schema. |
| **Payload evolution** | Adding/removing fields requires backwards-compatible codecs and in-flight migration handling. Renames need parallel codec paths. | `ALTER TABLE` is atomic across all rows. The migration is the change record. |
| **Adding a subsystem** | Wire codecs, register handlers, design checkpoint payload shape. Durability machinery is inherited, but payload design and codec maintenance offset most of that saving. | Design tables, write worker, write queries. Cost is visible up front in migrations; payload evolution is `ALTER TABLE`. |
| **Operational surface** | "What's stuck?" needs mailbox introspection, TLV decoding, checkpoint reading. Framework knowledge required. | "What's stuck?" is a `SELECT`. Operator answers with SQL alone. |
| **Crash recovery** | Framework reloads checkpoints, replays outbox, reconciles cursors. | Each worker reads its effect table on startup and resumes. No reconciliation step. |
| **Concurrency / claiming** | Mailbox + delivery store handles ordering and at-least-once delivery. Uniform across subsystems. | Lease/owner/token columns per effect table. ~10 lines of SQL each, repeated per subsystem. |
| **Idempotency** | Actor message IDs + delivery cursors. Generic, framework-coupled. | Domain-stable keys. Survive framework redesigns. |
| **Cross-subsystem messaging** | Uniform Tell/Ask. Subsystems don't know each other's storage. | Explicit effect routing; receivers poll. More plumbing per integration. |
| **Onboarding** | Steep but flat. Learn the framework once; every subsystem reads similarly. | Shallow but bumpy. Each subsystem is self-contained, read on its own terms. |
| **Adding a state** | Edit FSM; framework persists transparently. | `ALTER TABLE` + CHECK constraint + tests. Visible in schema review. |
| **Test surface** | Framework has its own test suite; subsystem tests build on framework primitives. Crash tests need framework hooks. | No framework to test. Crash tests are "kill DB connection, restart worker." Deterministic. |
| **Generality** | Fixed cost; per-subsystem savings compound with N. | Cost scales linearly with N. No leverage from a 6th subsystem. |
| **Failure modes** | Subtle and framework-wide. A delivery-store bug affects every subsystem. | Subtle and subsystem-wide. A handler bug is bounded to that subsystem. |
| **MuSig2 / persist-before-send** | Awkward: the "send" is a framework concern, but "persist what you'll send" must happen before the framework sees it. | Natural: INSERT pub_nonce + sec_nonce + send effect in one tx; worker reads on send. |
| **Cross-cluster / HA** | Maps onto distributed mailboxes plausibly. | Requires real work; "one DB per process" assumption breaks. |

### Where durable actor pays off

- **N subsystems large and growing.** Fixed framework cost amortizes; new subsystems become cheap *at the durability boundary* (the payload-design cost remains).
- **Heterogeneous subsystems with similar delivery needs.** Uniform Tell/Ask is genuinely useful when messaging matters more than typed state.
- **Cross-process or cross-cluster.** Mailbox model maps onto distributed delivery; SQL-only does not.
- **Independently-deployed senders and receivers.** Codec versioning earns its keep when sender and receiver can be on different code versions in production.
- **Operator team that already knows the framework.** Generic tooling is reusable across all subsystems.

**Worth noting:** none of these conditions currently apply to the client. Subsystems are few (OOR, rounds, unroll, wallet boarding) and not growing; they aren't heterogeneous in their delivery needs; there's one process per client; sender and receiver are always the same binary; the only cross-process boundary is the server mailbox, already isolated.

### Where per-subsystem SQL pays off

- **Mature, stable state shapes that benefit from being readable.** Typed columns are an asset; TLV blobs hide things you'd want to query.
- **Single-process or single-DB deployment.** The "one tx" invariant is genuinely simple here.
- **Operations-heavy environment.** SQL queries on domain tables beat framework introspection for "what's stuck?", "show me sessions in state X".
- **MuSig2-style "persist-before-send".** Typed columns + one-tx commits handle this without contortion.
- **Schema reviewed in PR.** Every state, every effect, every constraint visible in migrations — same discipline that catches API bugs catches durability bugs.
- **Evolving subsystems.** `ALTER TABLE` handles atomic schema change; no in-flight payload reconciliation.

### Where each shape struggles

**Durable actor struggles when:**
- A persisted message shape needs to evolve frequently. TLV versioning debt accumulates.
- An operator needs to answer a domain question without engineer help. Payload introspection isn't a SELECT.
- A subsystem's correctness depends on the exact bytes of a typed value being persisted *before* an actor message is sent. The payload abstraction becomes a workaround.
- The framework's generic shape doesn't match a subsystem's natural shape; the subsystem ends up bent to fit, not the other way around.
- A failure is subtle and framework-wide; one delivery-store bug affects every subsystem.

**Per-subsystem SQL struggles when:**
- A new subsystem doesn't fit "one producer, one tx, switch on effect_type." Generic delivery would have saved table-design time.
- Discipline rules ("no payload BLOB", "no generic helper", "one tx per state advance") start to feel restrictive, and someone factors out shared machinery — quietly recreating the framework.
- The system grows to many subsystems and table sprawl outpaces operators' ability to hold the schema in their head.
- Cross-subsystem coordination grows past 2–3 producers per consumer; explicit effect-routing becomes noisier than uniform actor messaging.

### Complexity, plainly stated

The framework concentrates complexity in **one place** and lets each subsystem be light at the durability boundary. SQL distributes complexity across **many places** and keeps each piece narrow but more numerous. The total amount of complexity is roughly comparable — the framework hides some of it in payload design and codec maintenance, SQL surfaces it in migrations and worker code.

Imagine adding a 10th subsystem under each:
- **Framework**: design TLV payload, write codec, register message types, wire handlers, design checkpoint shape. Durability machinery is inherited; payload work is not.
- **SQL**: design 2–4 tables, write queries, write a worker, write migrations. Durability is your problem — but visible and contained.

---

## Retention is unresolved on both sides

One thing the comparison above doesn't cover, and worth flagging rather than papering over: **neither shape has finished retention policy today**.

Under the framework, most tables do clean up:
- `mailbox_messages` deleted on `Ack`
- `ask_results`, `processed_messages`, `dead_letters` have TTL- or age-based sweepers
- `fsm_checkpoints` deleted when the actor terminates normally

But `outbox_messages` has no GC wired up today — `DeleteOutboxMessage` and `CleanupCompletedOutbox` are noted as an explicit follow-up in the queries file, with retention policies to be added later. So completed outbox rows accumulate indefinitely in the current state. It hasn't been visible because there are no long-running deployments yet, but in production it would grow without bound. `fsm_checkpoints` is also fragile: it's only cleaned up on *normal* actor termination — crashed or replaced actors leave orphans.

Under per-subsystem SQL the same shape of problem exists in different form: rows in `<x>_effects` with `status IN ('done','dead')` need eventual cleanup, and the design doc doesn't prescribe a policy. Each effect table would presumably want its own `DELETE WHERE status IN ('done','dead') AND updated_at < retention` sweeper. This branch doesn't include those sweepers.

The honest trade is:
- **Framework**: missing GC affects all subsystems through one shared table. One fix, broad impact.
- **SQL**: missing GC is per subsystem. N small fixes, narrower impact each, but harder to overlook because each table is visible in code review.

Flagging it because it's a gap that exists on both sides, not a point of contrast.

---

## Observations I came out with

- **Any FSM-bearing subsystem has to persist its state regardless of which durability framework wraps it.** Once that's true — which it always is — the framework's mailbox-layer durability is *additional*, not equivalent. For in-process flows it solves crash scenarios that state durability already solves.
- **MuSig2 persist-before-send** is a one-INSERT pattern under SQL (`client_round_nonce_state.pub_nonce` and `sec_nonce` + a send effect row in the same tx). Historically the most awkward correctness case under the framework; under SQL it stops being a special case at all.
- The "what's stuck?" question collapses into a few SELECTs. Useful for ops and for code review.
- Crash recovery is "read domain tables on startup" — no separate checkpoint reconciliation step.
- Reviewing schema for a new subsystem is reviewing the subsystem's contract — same discipline that catches API bugs catches durability bugs.
- The framework's code volume is what gets deleted; the SQL shape's schema volume is what gets added. The client side is a net deletion overall — except in payload evolution, where SQL's `ALTER TABLE` is meaningfully cheaper than TLV codec maintenance over the project's lifetime.

---

## What I'm explicitly not claiming

The framework's design isn't wrong, and its generality would earn its keep in a system with cross-cluster delivery, dozens of heterogeneous subsystems, or independently-deployed senders and receivers. None of those conditions currently apply to the client. The honest claim is narrower: **in this codebase the framework's machinery is paying its cost without earning much**, and the branch is a working demonstration of what the alternative shape looks like. In a different codebase the trade would look different.

---

## A quick sense of scale

21 commits on this branch. 269 files touched. 4 subsystems migrated end-to-end: OOR client, rounds client, unroll, wallet boarding. 54 tables in the client schema. Net diff is ~14.4k lines smaller (framework removal dominates: ~22.5k insertions versus ~36.9k deletions). So the client side is a net deletion overall; the SQL machinery that's added back is significantly less than the framework code it replaces.

---

## Test status

Static gates, unit tests, integration tests (`itest`), and systest are green on this branch. `make sql-durability-check` enforces the framework-deletion grep gates as part of CI.

---

## Things I'd expect (and welcome) pushback on

A few places I think a reviewer might reasonably push, listed because I'd rather hear it than not:

- **Is per-subsystem retention the right answer?** I split retention into one sweeper per effect table. There's a real argument for keeping a shared sweeper pattern instead — single fix, broad impact. I went per-subsystem because each table has different liveness needs, but I'm not certain that's the right call.
- **Client schema mailbox set.** The client only has `mailbox_egress` and `mailbox_ingress_cursors`. I think this is intentional (the client doesn't host an inbox; it only sends, and tracks ack progress on what the server delivers) but worth a sanity check against the design assumptions.
- **CI grep gates are crude.** `make sql-durability-check` is just `rg` for forbidden names. It works, but anyone who wants something less brittle has a point.
- **Discipline rules aren't type-enforced.** The "one transaction, producer owns it" / "no payload BLOB" / "no generic helper" rules are enforced by code review and grep, not by the compiler. A reviewer who'd prefer stronger guardrails has a real argument; I just couldn't find a cheap way to express them as types.

---

## What I'd value feedback on

- Reading the per-subsystem code (`oor`, `unroll`, `wallet`, and the new tables in `db/sqlc/migrations`), is the domain shape clearer or muddier than under the framework?
- Are there durability cases that don't fit a "one producer, one tx, one effect type" mould, where this shape would be worse?
- If we wanted to keep the framework but borrow ideas — typed columns over payloads, domain-owned state for specific subsystems — where would the cheapest wins be? The TLV→column move alone might be worth doing inside the framework.
- Where does the operator runbook help or fall short for someone diagnosing a stuck session?


bhandras:
Closing this stale experimental branch.

This PR explored a per-subsystem SQL durability architecture. We did not select that rewrite as the product direction, and current durability work has evolved independently on `main`. Restacking this experiment would add review and maintenance cost without a current delivery goal.

No follow-up is planned from this PR.

PR #470, opened 2026-05-19
walletdk: Auto-join rounds on deposit and prefer cooperative exit

Roasbeef:
In this PR, we close the two round-join gaps that show up when a host
drives the daemon through `sdk/walletdk` rather than `darepocli`.
Wallet-shaped SDK hosts want "deposit funds here" and "exit this VTXO"
to be one-shot user interactions, but the daemon today stages those
intents and waits for an explicit `Board` / `LeaveVTXOs` RPC to push
them onto the round FSM. The default suits operator-driven hosts but
strands SDK callers a second RPC short of done.

To fix this without changing the default semantics that `darepocli`
and server hosts rely on, we add an opt-in "eager round join" mode
that lives in the wallet actor and surfaces up through `darepod.Config`
and `walletdk.Config`. When the flag is on, two sites change: a fresh
boarding UTXO confirming in a block triggers `handleBoard` inline, and
cooperative-leave intents are forwarded with `TriggerRegistration=true`
so the round FSM advances out of `PendingRoundAssembly` immediately.
Off remains the default everywhere.

We also flip `walletdk.Exit` from "always unilateral" to
"cooperative-first, unilateral fallback". `ExitRequest.Destination`
lets the caller name the on-chain address the leave output should land
on. When it's set, the SDK first tries `daemonrpc.LeaveVTXOs`, and any
failure (operator unreachable, queue rejected, deadline) falls through
transparently to the existing `Unroll` path so the user is never
blocked from getting their money back. The result type now reports
which path the daemon took and carries the original cooperative error
when the fallback fires, so the host can retry the cooperative attempt
later if it cares to.

## Why this lives in the wallet, not in the SDK

The boarding-confirmation event already arrives at the wallet actor,
and `handleBoard` already owns the persist-then-Tell ordering the
eager path needs. Wiring the auto-board hook one layer up (e.g. a
walletdk-side watcher polling `Subscribe`) would duplicate the
confirmation detection that the wallet already does, and would race
the wallet's own persist write. Putting the hook in
`handleBlockEpoch` keeps the existing invariants intact and gives us
restart-safety for free via the existing `PendingBoardRequest` rows.

The eager-board call coalesces per block (one `TriggerBoardMsg` per
epoch that surfaced any new UTXOs) rather than per UTXO, because
multiple confirmations in the same block share an assembling round
and `handleBoard` walks the full confirmed set on each call.

## Why `Exit` falls back on any error

`Exit` is an "I want out now" operation. Classifying which cooperative
errors are transient vs. terminal would either misdetect operator
outages as permanent (leaving the user stuck) or mistake a real client
error for a transient one. Unilateral `Unroll` always lands on a
wallet-derived script, so the fallback can't succeed into an
unexpected destination. The original cooperative error is preserved on
`ExitResult.CooperativeError` so hosts that want to be picky can
retry just the cooperative branch later.

## Test plan

- [x] `make build` and `make build tags=walletrpc,swapruntime`
- [x] `go test ./wallet/ ./darepod/ ./sdk/walletdk/`
- [x] `make fmt-changed` + `make lint-native` (0 issues)
- [x] `make commitmsg-lint range=origin/main..HEAD`
- [x] New `wallet/eager_round_join_test.go` covering both on/off cases for the auto-board path and the leave intent's `TriggerRegistration` field
- [ ] Manual SDK smoke: deposit → confirm → round-joined VTXO with `EagerRoundJoin=true` and without
- [ ] Manual `walletdk.Exit` smoke against a live operator (cooperative path) and against an unreachable operator (fallback path)

See each commit message for a detailed description w.r.t the
incremental changes.


Roasbeef:
@claude review this 

@codex review this

Roasbeef:
@claude review this 

@codex review this

PR #474, opened 2026-05-19
swapruntime: authenticate swap mailbox clients

sputn1ck:
## Summary

- rebase on main now that https://github.com/lightninglabs/darepo-client/pull/459 has merged
- export shared mailbox gateway auth helpers from serverconn
- provide a configurable server-side gateway mailbox auth interceptor and gateway-token client interceptor
- provide a reusable authenticated MailboxServiceClient wrapper for Send, Pull, and AckUpTo
- switch swapclientserver to the shared authenticated mailbox client wrapper

Used by:

- https://github.com/lightninglabs/darepo/pull/466
- https://github.com/lightninglabs/swapdk-server/pull/36

## Validation

- `go test ./serverconn ./darepod -count=1`
- `go test -tags swapruntime ./swapclientserver -count=1`
- `go test ./swaprpc -count=1`

Note: `make lint-local` still reports existing line-length issues in `cmd/darepocli/internal/gen-devrpc/main.go` on this base.

### DIFF
--- a/swapclientserver/service.go
+++ b/swapclientserver/service.go
@@ -28,6 +28,7 @@ import (
 	"github.com/lightninglabs/darepo-client/rpc/swapclientrpc"
 	sdkark "github.com/lightninglabs/darepo-client/sdk/ark"
 	"github.com/lightninglabs/darepo-client/sdk/swaps"
+	"github.com/lightninglabs/darepo-client/serverconn"
 	"github.com/lightningnetwork/lnd/invoices"
 	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/lightningnetwork/lnd/lntypes"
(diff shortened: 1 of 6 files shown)


sputn1ck:
@claude please review

swapclientserver/service.go
```diff
@@ -581,67 +603,65 @@ func (d *daemonAuthOnlyInvoiceCreator) CreateInvoiceWithKey(ctx context.Context,
...
 
-	case isLocalSwapServerAddr(addr) || cfg.ServerInsecure:
+	case cfg.ServerInsecure:
 		return []grpc.DialOption{
 			grpc.WithTransportCredentials(
 				insecure.NewCredentials(),
 			),
 		}, nil
 
 	default:
```
bhandras, line 633: codex: This drops the old loopback/plaintext default, but `DefaultConfig` still points swapruntime at `localhost:10030` with `ServerInsecure` left false. As a result a stock swapruntime daemon now builds TLS credentials here and the first call to the local dev swap server fails the handshake (`first record does not look like a TLS handshake`); REST has the same issue because `swapServerRESTBaseURL` now returns `https://localhost:10030` unless the user explicitly sets `swap.serverinsecure=true`. Either the local default needs to set `ServerInsecure: true`, or the loopback dev fallback needs to stay in this transport selection so the documented local/regtest default remains usable.

sputn1ck: Addressed in f2c20b4: restored the loopback/unix plaintext fallback when no TLS cert path is configured, kept explicit TLS cert paths authoritative, and added tests for local default, remote TLS default, and TLS-cert override behavior.

bhandras (approved):

PR #476, opened 2026-05-20
manual-testing-fixes-2: 4 bug fixes uncovered in the May 2026 walkthrough

ellemouton:
## Summary

Bug fixes uncovered during the post-`manual-testing-fixes` walkthrough
(see `MANUAL_TESTING.md` in the working tree for the full scenario log
and `BUGS_FOUND.md` for the issue catalogue). Each commit is atomic
and addresses one issue.

### bug-1 — lwwallet boarding preflight admits CPFP fee inputs that LND would later reject
**Before:** `RPCServer.Unroll` calls a fee-input preflight that
enumerates the wallet's spendable UTXOs. On the lwwallet backend the
underlying `btcwallet` credit-tracking pipeline removes spent
boarding outputs asynchronously, so a freshly-consumed boarding UTXO
stays credit-marked for hundreds of ms after the round commit. The
preflight observed the stale UTXO and admitted an unroll that the
actual CPFP broadcast then rejected with `cpfp fee input
unavailable`. On `lnd` the same scenario rejected cleanly because
LND removes the spent UTXO synchronously, so the asymmetry was
backend-specific and hid in CI.
**After (`darepod: filter boarding outputs from ListWalletUnspent`):**
`ListWalletUnspent` now snapshots the persisted boarding scripts
from `BoardingStore` and discards any UTXO whose pkScript matches.
The preflight only ever sees fee-input candidates. Same path on both
backends.

### balance-display correctness — `darepocli balance` reports unboarded UTXOs as spendable
**Before:** `walletrpc.Balance.confirmed_sat` is documented as "the
total spendable VTXO amount in satoshis", but `fetchBalance` was
returning `total_confirmed_sat` (vtxo + boarding). A user who
faucets a boarding address but has not yet run `ark board` would see
the boarding amount as spendable, even though every `send` verb
rejected it.
**After (`swapwallet: route boarding-confirmed into pending_in_sat in Balance`):**
`confirmed_sat` is now strictly `vtxo_balance_sat`;
`boarding_confirmed_sat + boarding_unconfirmed_sat` lands in
`pending_in_sat`. Matches the proto contract and the wallet's
runtime behaviour.

### bug-2 — `darepocli balance` hangs for ~7 minutes after a 200-block CSV catch-up
**Before:** The wallet actor ran the heavy per-block work
(`ListUnspent` + up-to-three 1 s retries, per-UTXO `processUtxo`,
boarding-sweep resume kick) inline inside the
`BlockEpochNotification` handler. A chain that advances rapidly
(`arktest mine 200` for a CSV-mature unilateral exit, or a fresh
sync after downtime) fires 200 epoch Tells into the actor mailbox in
seconds. Each handler takes ~2 s of serial actor-turn time, so every
other wallet Ask — `GetBoardingBalance`, `CreateBoardingAddress`,
`RegisterIntent`, `RefreshVTXOs`, `LeaveVTXOs`, `SendVTXOs`,
`SweepBoardingUTXOs`, … — hangs behind a ~7-minute mailbox drain.
**After (two commits):**
- `darepod: bypass wallet actor for GetBalance boarding read` —
  immediate stopgap: `GetBalance` now reads the boarding store
  directly from the RPC layer instead of issuing an Ask through the
  jammed actor mailbox. Reads are transactional, so a concurrent
  actor write yields a snapshot-consistent answer with at most one
  commit's worth of staleness — well within balance-display
  semantics. `darepocli balance` returns sub-second under burst.
- `wallet: coalesce block-epoch processing into a tick-driven loop` —
  proper fix that unblocks every other wallet Ask: `handleBlockEpoch`
  shrinks to an atomic-pointer store of the latest tip (microseconds,
  no actor work), and a single 1 s ticker drives the per-tip work
  once against the latest tip. A burst of N notifications resolves to
  N atomic writes plus one work pass. A `tickInflight` CAS caps
  pending-tick mailbox depth at 1 so a long retry pass cannot let
  `ticker.C` accumulate.

### bug-3 — received in-round VTXOs cannot be unilaterally exited
**Before:** Bob receives a VTXO via the in-round directed-send path
(`alice: SendVTXO(bob.pubkey, amount)` → round commits → operator
publishes `IncomingVTXOEvent` → `bob.IncomingVTXOHandler`
materializes the recipient VTXO). The push event carries
outpoint/value/script/round_id/commitment_txid/batch_expiry but not
the round commit-tree fragments or the created height. The handler
persisted a `Descriptor` with empty `Ancestry` and zero
`CreatedHeight`, and `validateProofDescriptorShape` on the
unilateral-exit unroll path rejected both — so `darepocli exit` on
any received VTXO immediately terminated the unroll actor with
`descriptor missing ancestry`. The cooperative spend paths (refresh,
leave, OOR forward) all worked because they don't need ancestry,
which is why this stayed hidden until end-to-end testing.
**After (two commits):**
- `vtxo: consolidate AncestryFromRPC into the vtxo package` — prep
  refactor that pulls the two near-identical `ancestryFromRPC`
  copies (one in `oor`, one in `darepod`) into a shared
  `vtxo.AncestryFromRPC` helper. Both the OOR receive path and the
  new in-round path now use the same validator. No behaviour change.
- `vtxo: fetch ancestry + created_height for incoming VTXOs` —
  `IncomingVTXOHandlerConfig` grows an `IncomingAncestryFetcher`
  hook; the `darepod` wiring resolves it via the daemon's indexer
  client with a per-script proof-of-control signer. The handler
  runs the fetch synchronously after building the descriptor and
  before `SaveVTXO`, so the persisted descriptor lands with full
  lineage from the first write. Fetch failures degrade gracefully —
  cooperative spend paths still work; only unilateral exit is
  blocked until backfill (matches the pre-bug-3 behaviour).
  Verified end-to-end via the companion server itest
  `TestUnilateralExitInRoundDerivedCompletion`.

### bug-4 — `darepocli send <addr> --onchain` strands the leave intent forever
**Before:** The everyday `send --onchain` verb is documented as a
one-shot, but `swapwallet.router.sendOnchain` only dispatched
`LeaveVTXOs` (which queues the leave intent in the round actor's
`PendingAssembly`) and returned. The caller had to follow up with
`darepocli ark rounds join` for the round to actually seal, otherwise
the source VTXO sat in `PendingForfeitState` forever and the operator
saw `leave_count=0` on the next round.
**After (`swapwallet: auto-join next round after onchain Send cooperative leave`):**
`sendOnchain` now calls `JoinNextRound` immediately after a
successful `LeaveVTXOs`. The raw `ark vtxos refresh` / `ark vtxos
leave` CLI still exposes `--no_join` for batching callers; the
top-level `send` verb has no batching contract and shouldn't. If the
implicit join fails the error wraps with the exact `ark rounds join`
recovery hint so a stranded intent is never silent.

## Server-side dependency

The companion server PR
([lightninglabs/darepo#469](https://github.com/lightninglabs/darepo/pull/469))
adds the `TestUnilateralExitInRoundDerivedCompletion` itest that
bug-3 needs to pass against, plus a server-side `batchwatcher` fix
for bug-5 (multi-leaf partial-exit operator-status reconciliation),
plus a hardening of `TestDirectedSendSelfSend` that the new bug-3
client behaviour exposed. That PR's submodule bump points at this
PR's tip and is amended in lockstep on every force-push here.

## Test plan

- [x] `make lint-native` clean
- [x] `make commitmsg-lint range="origin/main..HEAD"` clean
- [x] `make unit pkg=./swapwallet/... ./vtxo/... ./wallet/...
      ./darepod/... ./oor/...` green
- [x] End-to-end via the `arktest` harness on regtest:
  - bug-1: lwwallet unroll preflight rejects with
    `insufficient wallet UTXOs` (was admitting and failing
    downstream)
  - bug-2: `darepocli balance` returns sub-second after
    `arktest mine 200`
  - bug-3: in-round receive + refresh + exit ⇒ on-chain sweep
    credits the receiver's LND wallet (cases A and B from the
    walkthrough log)
  - bug-4: `darepocli send --onchain --amt N` lands without a
    follow-up `ark rounds join`


ellemouton:
@claude review this

darepod/incoming_ancestry_fetcher.go
```diff
@@ -0,0 +1,124 @@
...
+
+		if len(op.GetTxid()) != len(target.Hash) {
+			continue
+		}
+
+		var txid wire.OutPoint
+		copy(txid.Hash[:], op.GetTxid())
+		if txid.Hash != target.Hash {
+			continue
+		}
```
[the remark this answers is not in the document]
ellemouton: Done in the latest force-push —  now uses a typed `chainhash.Hash` local for the comparison.

darepod/rpc_server.go
```diff
@@ -467,6 +453,73 @@ func (r *RPCServer) GetBalance(ctx context.Context,
...
+func (r *RPCServer) fetchBoardingBalance(ctx context.Context,
+	resp *daemonrpc.GetBalanceResponse) {
+
+	dbStore := db.NewStore(
+		r.server.db.DB, r.server.db.Queries, r.server.db.Backend(),
+		r.server.subLogger(db.Subsystem),
+	)
+	boardingStore := dbStore.NewBoardingStore(
+		r.server.chainParams, r.server.clk,
+	)
```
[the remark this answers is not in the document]
ellemouton: Done — `fetchBoardingBalance` now uses `r.server.newBoardingStore()`.

darepod/wallet_testhooks.go
```diff
@@ -104,6 +138,76 @@ func (s *Server) ListWalletUnspent(ctx context.Context, minConfs,
...
+// because in-memory tracking on lwwallet / btcwbackend can drift
+// across restarts before the wallet actor re-imports scripts.
+func (s *Server) boardingScripts(ctx context.Context) (
+	map[string]struct{}, error) {
+
+	dbStore := db.NewStore(
+		s.db.DB, s.db.Queries, s.db.Backend(),
+		s.subLogger(db.Subsystem),
+	)
+	boardingStore := dbStore.NewBoardingStore(s.chainParams, s.clk)
```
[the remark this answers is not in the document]
ellemouton: Done — `boardingScripts` now uses `s.newBoardingStore()`.

ellemouton:
Thanks Claude — all three nits addressed in the latest force-push:

- `filterBoardingScripts`: now allocates with `make([]*wallet.Utxo, 0, len(utxos))` instead of `utxos[:0]`, with a comment explaining the footgun we avoided.
- `fetchBoardingBalance` comment: reworded to say "returns early on the first error" so it matches the actual behaviour.
- `darepod/CLAUDE.md` + `AGENTS.md`: the stale `ancestryFromRPC` entry now points readers at `vtxo.AncestryFromRPC` (the consolidated home from PR commit `vtxo: consolidate AncestryFromRPC into the vtxo package`).

ellemouton:
Both addressed in the latest force-push:

**High — `wallet/wallet.go` empty-scan retry budget regression.** Real bug: pre-coalescing each block-epoch ran its own retry window, so a 200-block burst gave 200 chances for a lagging backend to surface a UTXO; post-coalescing the burst yielded one tip advance with one in-tick retry window, so a backend that lagged past `listUnspentMaxRetries × listUnspentRetryDelay` would strand boarding detection until the next block arrived.

The fix carries the retry budget across tick boundaries. Two new atomic fields on `Ark` (`emptyScansAtTip`, `emptyScansAtTipHeight`) track consecutive empty scans at the same height. `handleProcessTipTick` now only stores `processedTipHeight` when either (a) the scan found new boarding UTXOs, or (b) `maxEmptyScansBeforeTipProcessed = 5` consecutive empty scans at this height have elapsed. The counter resets on tip advance so a new tip gets a fresh budget rather than inheriting the partial one accumulated at the previous tip. Steady-state cost is unchanged: once `processedTipHeight` matches the latest tip, ticks short-circuit. Worst-case detection latency is now `~5 × (in-tick retry budget + tick interval) ≈ 20 s` of retry effort per tip — well past the LND / neutrino / Esplora lag windows the original per-block retry was sized for.

`TestBlockEpochBurstCoalescesToOneTipPass` rewritten to pin the new contract: empty scans do NOT mark the tip processed until the budget exhausts; the COALESCING property (total `ListUnspent` calls stay well below the pre-fix `200×3` worst case) is still asserted. New focused test `TestTipTickEmptyScanRetryBudgetResetsOnTipAdvance` pins that the counter resets on each tip advance rather than carrying a partial budget forward.

**Medium — `darepod/incoming_ancestry_fetcher.go` single-page query.** Real bug: `ListVTXOsByScriptsTaproot(..., limit=0)` maps to the server default of 100 entries. A reused receive script with more than ~100 VTXOs would push the target outpoint past the first page; the fetcher would miss it, fall back to no-ancestry persist, and reintroduce the bug-3 unilateral-exit failure for that VTXO.

The fix mirrors `darepod.ResolveIncomingMetadataFromIndexer` line-for-line: bounded page size (`ancestryFetcherIndexPageSize = 128`, same as `incomingMetadataIndexPageSize` in the OOR resolver), max total scan (`oor.DefaultMaxVTXOMatches = 128`), cursor loop with `indexerlimits.ValidateVTXOsByScriptsCursor` on every hop, and a deterministic exit when the next cursor stops advancing or the cap is hit. The previous "response missing vtxo for outpoint" error is preserved at the exhausted-scan path.

ellemouton:
Partial-lag also addressed in the latest force-push.

The empty-scan retry budget now also covers the case the reviewer flagged where ListUnspent returns some but not all of a block's boarding outputs at scan time: the counter resets on `foundNew=true` instead of marking the tip processed, so the budget always allows `maxEmptyScansBeforeTipProcessed` more scans after the latest discovery. The tip is only marked processed once the set has been stable for N consecutive scans.

Concretely the counter semantics changed from "consecutive empty scans, any foundNew exits the budget" to "consecutive empty scans since the latest discovery; foundNew resets the counter but the budget keeps running". For backends that report all-or-nothing per poll (Esplora, where lwwallet sits) this is a no-op; for LND / neutrino (where credit-tracking is per-UTXO and a partial set is possible across a block boundary) we now match the pre-coalescing per-block retry coverage.

New test `TestTipTickPartialLagDoesNotMarkProcessedOnFirstFind` (using `newEagerBoardFixture` so processUtxo returns `true` on the first scan and `false` thereafter) pins the new contract: the first found-new tick does NOT advance `processedTipHeight`, the next N-1 empty ticks stay under budget, and the N-th finally marks the tip processed.

Server submodule bump amended to point at the new client tip.

darepod/rpc_server.go
```diff
@@ -514,6 +500,71 @@ func (r *RPCServer) GetBalance(ctx context.Context,
...
+		}
+
+		return total
+	}
+
+	confirmed, err := boardingStore.FetchBoardingIntentsByStatus(
+		ctx, wallet.BoardingStatusConfirmed,
+	)
+	if err != nil {
+		r.server.log.WarnS(
```
Roasbeef: Hmm on one hand nice to return the full balance even if portions fail, but also lean towards propagating the error here.

wallet/wallet.go
```diff
@@ -797,19 +973,73 @@ func (a *Ark) handleUnregisterNotifier(ctx context.Context,
...
+// complete. See handleBlockEpoch for why the work was hoisted off the
+// per-block path.
+func (a *Ark) handleProcessTipTick(ctx context.Context) fn.Result[WalletResp] {
+	// Clear the inflight latch on every exit path so the next
+	// ticker.C event can fire a fresh tick. Pair-bonded with the
+	// CompareAndSwap(false, true) gate in runTipTickLoop; together
+	// they cap pending-tick mailbox depth at 1.
+	defer a.tickInflight.Store(false)
+
+	tip := a.latestKnownTip.Load()
```
Roasbeef, line 937: I think we can reduce complexity a bit here. If the goal is just to make the block catch up and UTXO detection snappier, we can remove the loop for `ListUnspent`, we can just check once there. Then retries can live at a wallet/itest layer.

Roasbeef: As with this change, the concurrency store becomes more involved (extra goroutine, 2 new atomic variables, etc). 

If we want to catch that lag that can occur when we hear about things vs the wallet, then I don't think we need any new synchronization. Instead, a background goroutine (or just a timeout `time.Ticker` type interaction) can call `processUtxo`, then we make that map a `sync.Map`.

darepod/incoming_ancestry_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+// broken dependency in production.
+//
+// The fetched ancestry travels through the same vtxo.AncestryFromRPC
+// validator the OOR receive path uses; structural errors (missing or
+// over-cap paths) surface as fetch failures and the handler then
+// persists without ancestry rather than dropping the VTXO entirely.
+// This keeps cooperative spend paths (refresh, OOR forward, leave)
+// usable on the receiver even if the indexer is temporarily
+// unhealthy — only unilateral exit is degraded.
+func (s *Server) incomingAncestryFetcher() (vtxo.IncomingAncestryFetcher,
```
Roasbeef: Perhaps we should move this to another sub-system? To avoid the anti-pattern of "server struct is god with all the necessary attributes+methods).

Roasbeef: eg: `vtxo` package.

darepod/incoming_ancestry_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+	return func(ctx context.Context, outpoint wire.OutPoint,
+		pkScript []byte, clientKey keychain.KeyDescriptor) (
+		vtxo.IncomingVTXOExtras, error) {
+
+		// One scope per pkScript; the indexer signs each
+		// ScriptScope with the supplied signer so the
+		// proof-of-control attaches to the owner key for this
+		// specific receive script.
+		scope := indexer.TaprootScriptScope{
+			PkScript: append([]byte(nil), pkScript...),
```
Roasbeef: Append here as a short cut for copy?

darepod/incoming_ancestry_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+			if err := indexerlimits.
+				ValidateVTXOsByScriptsCursor(
+					nextCursor,
+				); err != nil {
+				return vtxo.IncomingVTXOExtras{}, fmt.Errorf(
+					"indexer next "+
+						"cursor: %w", err)
+			}
+
+			cursor = append(cursor[:0], nextCursor...)
```
Roasbeef: Could just re-assign instead?

darepod/incoming_ancestry_fetcher.go
```diff
@@ -0,0 +1,192 @@
...
+			)
+			if err != nil {
+				return vtxo.IncomingVTXOExtras{}, fmt.Errorf(
+					"indexer list vtxos by "+
+						"scripts: %w", err)
+			}
+
+			page := resp.GetVtxos()
+
+			if match := findVTXOByOutpoint(
```
Roasbeef: Breaking change on RPC layer, but could have it return this as a map to avoid this scan here.

Roasbeef (changes requested):

darepod/incoming_ancestry_fetcher.go
Roasbeef: Also could use unit tests (extracting can help with that)

ellemouton:
@claude review this again

ellemouton:
@Roasbeef @claude pushed addressing the three open items from the CHANGES_REQUESTED review:

**1. `wallet/wallet.go` retry budget complexity** — picked Option A ("check once there, retries can live at a wallet/itest layer"). Dropped the entire `emptyScansAtTip` / `emptyScansAtTipHeight` multi-tick retry. `handleProcessTipTick` is now a single-pass-per-tip: one `ListUnspent`, run `processUtxo` for each result, advance `processedTipHeight`. No new atomics, no inline retry budget. Lag coverage: the next block's tip advance re-runs the scan. Two now-obsolete tests deleted (`TestTipTickEmptyScanRetryBudgetResetsOnTipAdvance`, `TestTipTickPartialLagDoesNotMarkProcessedOnFirstFind`); `TestBlockEpochBurstCoalescesToOneTipPass` reshaped to pin the simpler contract (one tick services the burst → one `ListUnspent` → `processedTipHeight` advances; idle follow-up tick short-circuits).

**2. `darepod/rpc_server.go` error propagation** — all three balance components now propagate. VTXO fetch returns `codes.Internal` on error instead of warn-log + zero; `sumOnchainWalletConfirmed` returns `(amount, error)` and the first fetcher error short-circuits the sum, mirroring `fetchBoardingBalance`. `TestSumOnchainWalletConfirmed` rewritten to pin the new error-strict contract. Rationale documented inline: a zero-on-failure response is structurally indistinguishable from "no balance" and misleads any UI that trusts the body.

**3. `vtxo/incoming_ancestry.go` `InputIndices` copy** — `append([]uint32(nil), p.GetInputIndices()...)` → `slices.Clone(p.GetInputIndices())` for consistency with the resolver's `pkScript` clone.

Other comments triaged earlier this cycle:
- ✅ #4 (move to `vtxo`): heavy logic is in `vtxo/incoming_ancestry_resolver.go` with tests; only a 71-line wiring closure remains in darepod.
- ✅ #6 (cursor re-assign): `cursor = nextCursor` in resolver.
- ✅ #8 (unit tests): `vtxo/incoming_ancestry_resolver_test.go` (208 lines).
- ⏸ #7 (RPC return as map): Roasbeef tagged it himself as a breaking change / follow-up.

Local on the merged tip: `make lint-local` 0 issues, `make commitmsg-lint` clean, `go test` green across wallet/darepod/vtxo/swapwallet/oor.

darepod/rpc_server.go
ellemouton: Already addressed in the latest force-push — `fetchBoardingBalance` now propagates each `FetchBoardingIntentsByStatus` error and `GetBalance` short-circuits with `codes.Internal` instead of swallowing partial failures.

wallet/wallet.go
ellemouton: Picked Option A ("check once, retries can live at a wallet/itest layer") in the latest force-push. `handleProcessTipTick` now does a single `ListUnspent` per tip advance, with no inline retry budget; the per-block work has been hoisted off the actor mailbox onto a separate tick goroutine (`runTipTickLoop`). The atomics that remain (`latestKnownTip` + `tickInflight` CAS + `processedTipHeight`) are the minimum needed to coalesce a 200-block burst into a single `ListUnspent` and cap pending-tick mailbox depth at 1 — going further (e.g. dropping the actor seam entirely and reading UTXOs from the ticker goroutine via `sync.Map`) would be a bigger refactor than the bug-2 scope wants; happy to file a follow-up if you'd like that to happen here.

A backend whose UTXO reporting lags one tick interval surfaces the missing UTXO on the next chain advance (whichever tick processes the new tip re-runs the scan), so a separate retry surface isn't needed at the wallet layer. `TestBlockEpochBurstCoalescesToOneTipPass` pins the new contract.

darepod/incoming_ancestry_fetcher.go
ellemouton: Done — pulled the resolver into the `vtxo` package as `vtxo.ResolveIncomingAncestry` (`vtxo/incoming_ancestry_resolver.go`). The `darepod/incoming_ancestry_fetcher.go` file is now just the daemon-shaped wiring (~70 lines): build a per-script signer factory, hand the indexer-backed query callback to `vtxo.ResolveIncomingAncestry`, return the resulting `IncomingVTXOExtras`. Unit tests live with the resolver in `vtxo/incoming_ancestry_resolver_test.go` (`TestResolveIncomingAncestryMatchesPaginatedOutpoint`, `...RejectsOversizedNextCursor`, `...CapsScannedVTXOs`).

darepod/incoming_ancestry_fetcher.go
ellemouton: Done — `vtxo.ResolveIncomingAncestry` does `script := slices.Clone(pkScript)` once at function entry (`vtxo/incoming_ancestry_resolver.go:52`). The explicit `append([]byte(nil), pkScript...)` is gone from the daemon-side wiring.

darepod/incoming_ancestry_fetcher.go
ellemouton: Done — the loop now does `cursor = nextCursor` directly (`vtxo/incoming_ancestry_resolver.go:116`). The old `append(cursor[:0], nextCursor...)` is gone.

darepod/incoming_ancestry_fetcher.go
ellemouton: Went the other direction in commit 7dd08a7f — `ListVTXOsByScriptsResponse` is now `repeated VTXO vtxos` rather than a per-script map. Reasoning:

A per-script map keyed by hex-encoded pkScript pretends the most useful response index is always per-script, but every current call site queries with a single script, and the indexer's keyset cursor is keyed per-outpoint (not per-script). A map shape would discard that natural ordering and force any per-outpoint lookup (the unroll resolver here, the OOR metadata resolver) to flatten the map back out before iterating. The flat slice is the indexer's storage shape, the cursor's stride shape, and the wire shape — and lets callers build whatever index they want locally (per-script, per-outpoint, per-round) without another proto change.

In this resolver specifically, the scan is per-outpoint, so the map would not have avoided the scan — it would just have hidden one extra hop (`m[hex(script)].Vtxos`) before the same outpoint match runs. The current code uses `vtxo.FlattenListVTXOsByScriptsResponse(resp)` and matches by outpoint as before.

ellemouton:
@Roasbeef @claude pushed addressing the latest review.

**Force-push 7dd08a7f** is the long-term cleanup of the indexer response shape:

- `ListVTXOsByScriptsResponse` is now `repeated VTXO` (a flat slice) instead of the per-pkScript map. Each VTXO already carries its own pkScript and outpoint, so callers that need a per-script, per-outpoint, or per-round index build it locally; the wire format stays neutral.
- All three single-script call sites (`vtxo/incoming_ancestry_resolver.go`, `darepod/incoming_metadata.go`, `darepod/rpc_swap_lookup.go`) switch to the flat helper `vtxo.FlattenListVTXOsByScriptsResponse`; the per-script slicer is removed as dead.
- Test fixtures follow the same shape.

Disagreed with Roasbeef's suggestion of going to a per-outpoint map for the resolver-hot path — the map shape would have only hidden one extra hop before the same outpoint match runs (the keyset cursor is already per-outpoint, not per-script), and locking the wire format into one access pattern is what the flat slice avoids. Rationale on the thread.

**Force-push 586a2dc3** is doc-only cleanup of stale comments left by the previous Option A force-push (the multi-tick empty-scan budget that no longer exists): `wallet/wallet_test.go` (two test comments), `wallet/CLAUDE.md` + `wallet/AGENTS.md` (invariant at line 60), and `darepod/rpc_server.go` (the "up to three 1 s retries" justification on the boarding-balance bypass). No runtime behaviour changes.

I also went through Roasbeef's threads — all four `incoming_ancestry_fetcher.go` nits and the `wallet/wallet.go:938` complexity thread are now addressed in code (resolver moved to `vtxo/incoming_ancestry_resolver.go` with unit tests, single-call ListUnspent in `handleProcessTipTick`, `cursor = nextCursor` re-assignment, `slices.Clone` replacing the explicit `append([]byte(nil),...)`). Detail on each thread.

Roasbeef (approved):
LGTM 🥙

PR #488, opened 2026-05-20
walletcore: restore duplicate boarding imports

### LINKED ISSUE #487: Bug: ImportedAddrs not restored on daemon restart
Timeline

  Time: ~14:03
  Event: darepocli recv --onchain called — daemon creates boarding address,
  calls
    ImportTaprootScript → address added to ImportedAddrs in memory ✓
  ────────────────────────────────────────
  Time: ~14:05
  Event: Block 305255 processed, ListUnspent called, utxo_count=0 (faucet not
    sent yet) ✓
  ────────────────────────────────────────
  Time: ~14:05–14:49
  Event: IPv6 PMTU timeouts cause repeated daemon restarts during our debugging
  ────────────────────────────────────────
  Time: ~14:35
  Event: Faucet tx confirmed in block 305264
  ────────────────────────────────────────
  Time: Every restart
  Event: ImportedAddrs wiped (in-memory), startup recovery runs but silently
    fails to restore it — ListUnspent always returns 0
  ────────────────────────────────────────
  Time: 15:39
  Event: Debug log confirms the bug: Boarding address re-import skipped ... 
    already exists
  
  ---
  What the code intended vs. what happens
  
  The comment in wallet/wallet.go:384 says:

  ▎ "For in-memory backends (lwwallet), this restores tracked scripts lost on 
  ▎ restart. For persistent backends (LND), the script may already exist and the
  ▎  import will return a benign 'already exists' error which we log and skip 
  ▎ since the backend is already tracking the address."       

  The assumption embedded in this comment is: if btcwallet returns "already 
  exists", the backend is already tracking the address. That's true for LND
  (persistent). It's false for lwwallet — btcwallet persists the import to
  SQLite, but ImportedAddrs is in-memory only and gets wiped every restart.

  ---
  The call stack

  wallet/wallet.go: handleStartupRecovery()
    └─ store.ListAllBoardingAddresses()          → returns 1 address ✓
    └─ backend.ImportTaprootScript(tapscript)
         └─ walletcore/boarding_backend.go: ImportTaprootScript()
              └─ BtcWallet.ImportTaprootScript() → waddrmgr.ErrDuplicateAddress
              └─ returns error immediately        ← never reaches ImportedAddrs
  tracking
         └─ wallet.go logs "re-import skipped", continues
    └─ ImportedAddrs = {} (empty)

  ... later on block arrival ...

  wallet/wallet.go: handleBlockEpoch()
    └─ backend.ListUnspent()
         └─ lwwallet/boarding_backend.go: ListUnspent()
              └─ SnapshotAddrs() → {} (empty map)
              └─ loop never executes
              └─ returns [] (no UTXOs)
    └─ logs: utxo_count=0

  Debug log confirming the failure point:
  [DBG] ARKW: Boarding address re-import skipped
        address=tb1p8zyjdrhdylsh6xt73ux56gzslzxl0akharpcsh39xjpmjhl5n6dstjet27
        reason="import taproot script: address for script hash/key
                3889268eed27e17d197e8f0d4d2050f88df7f6d7e8c3885e253483b95ff49e9b
                already exists"

  ---
  Three potential fixes
  
  Fix 1 — Handle ErrDuplicateAddress in walletcore/boarding_backend.go (best 
  fix)

  When BtcWallet.ImportTaprootScript returns waddrmgr.ErrDuplicateAddress, look
  up the existing managed address from btcwallet and still add it to
  ImportedAddrs. The address is already in the btcwallet database — we just need
   to retrieve it.

  if errors.Is(err, waddrmgr.ErrDuplicateAddress) {
      // look up existing managed address, add to ImportedAddrs
  }

  Pros: fixes the root cause cleanly, one location. Cons: requires knowing how
  to look up a managed address from btcwallet by script (slightly non-trivial).

  ---
  Fix 2 — Add TrackAddress() to BoardingBackendBase, call it from wallet.go 
  recovery (second best)
  
  Separate the btcwallet import from the in-memory tracking. wallet.go's startup
   recovery already has addr.Address from the database — it can call
  TrackAddress(addr.Address) directly whether or not the btcwallet import
  succeeds.
  
  // wallet/wallet.go startup recovery:
  _, err := a.backend.ImportTaprootScript(ctx, addr.Tapscript)
  if err != nil {
      // log skipped
  }
  // Always track — ImportedAddrs needs it regardless of import result
  a.backend.TrackAddress(addr.Address)

  Pros: clean separation of concerns, doesn't depend on btcwallet API. Cons:
  adds a new exported method to the interface, two-line fix in two files.

  ---
  Fix 3 — Make ImportTaprootScript idempotent by not treating 
  ErrDuplicateAddress as an error (simplest)
  
  Just treat ErrDuplicateAddress as a success case: compute the address from the
   tapscript (or re-look it up), add to ImportedAddrs, and return normally.
  Callers already expect this to be a no-op for duplicates.

  Pros: one-file fix, no interface changes. Cons: requires btcwallet API to
  either look up existing address by script, or derive the address from the
  tapscript independently.
  
  ---
  The funds (2,000,000 sats, txid 90f68fe283..., block 305264) are sitting
  on-chain waiting. The daemon will detect them the moment the bug is fixed and
  a new block arrives.

sputn1ck:
## Summary

Fixes #487 by making btcwallet-backed taproot script imports idempotent for the duplicate-address case.

When lwwallet restarts, btcwallet still has the imported boarding script in its DB, but `walletcore.BoardingBackendBase.ImportTaprootScript` used to return on `ErrDuplicateAddress` before repopulating the in-memory `ImportedAddrs` filter. That left lwwallet polling Esplora with an empty address set, so funded boarding UTXOs were never found after restart.

This PR derives the P2TR address from the duplicate tapscript, restores it into `ImportedAddrs`, and returns success. A focused lwwallet regression test covers the restart-shaped path by clearing the in-memory filter and re-importing the same tapscript.

## Tests

```
go test ./lwwallet -run TestBoardingBackendDuplicateImportRestoresTracking -count=1
go test ./walletcore ./lwwallet ./btcwbackend ./wallet
make commitmsg-lint commit=HEAD
```

### DIFF
--- a/lwwallet/boarding_backend_test.go
+++ b/lwwallet/boarding_backend_test.go
@@ -28,30 +28,29 @@ func TestBoardingBackendDuplicateImportRestoresTracking(t *testing.T) {
 	}
 
 	tipHash := chaincfg.RegressionNetParams.GenesisHash.String()
-	esplora := httptest.NewServer(http.HandlerFunc(
-		func(w http.ResponseWriter, r *http.Request) {
-			switch r.URL.Path {
-			case "/blocks/tip/height":
-				_, err := w.Write([]byte("0"))
-				require.NoError(t, err)
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/blocks/tip/height":
+			_, err := w.Write([]byte("0"))
+			require.NoError(t, err)
 
-			case "/block-height/0":
-				_, err := w.Write([]byte(tipHash))
-				require.NoError(t, err)
+		case "/block-height/0":
+			_, err := w.Write([]byte(tipHash))
+			require.NoError(t, err)
 
-			case "/block/" + tipHash:
-				err := json.NewEncoder(w).Encode(esploraBlock{
-					ID:        tipHash,
-					Height:    0,
-					Timestamp: 1,
-				})
-				require.NoError(t, err)
+		case "/block/" + tipHash:
+			err := json.NewEncoder(w).Encode(esploraBlock{
+				ID:        tipHash,
+				Height:    0,
+				Timestamp: 1,
+			})
+			require.NoError(t, err)
 
-			default:
-				http.NotFound(w, r)
-			}
-		},
-	))
+		default:
+			http.NotFound(w, r)
+		}
+	}
+	esplora := httptest.NewServer(http.HandlerFunc(handler))
 	defer esplora.Close()
 
 	w, err := New(Config{
--- a/lwwallet/boarding_backend_test.go
+++ b/lwwallet/boarding_backend_test.go
@@ -1,0 +1,93 @@
+package lwwallet
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btclog/v2"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/stretchr/testify/require"
+)
+
+// TestBoardingBackendDuplicateImportRestoresTracking verifies restart recovery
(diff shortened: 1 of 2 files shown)


sputn1ck:
@claude review

PR #507, opened 2026-05-21
sdk/swaps: make per-swap mailbox receiver survive transport errors

### LINKED ISSUE #505: sdk/swaps: per-swap mailbox receiver has no reconnect — should reuse serverconn ingress-loop machinery
## Summary

`MailboxOutSwapEventReceiver` (`sdk/swaps/out_swap_mailbox.go`) does a
one-shot `edge.Pull` and surfaces any transport error as a terminal swap
failure. A single TCP RST against the mailbox/operator endpoint is enough to
kill an in-flight receive swap, even though the protocol is supposed to be
connectionless at the message layer. Meanwhile `serverconn` already has the
exact retry/backoff/cursor-checkpoint machinery we want; the SDK is
effectively a divergent reimplementation of the same loop, minus the
reliability features.

The mailbox-based protocol is designed to be connectionless: event delivery
should survive arbitrary TCP failures (RST, timeout, dial errors) as long as
the swap deadline / session lifetime allows. The SDK receiver currently
violates that contract.

## Production evidence (signet)

Receive swap
`77b3f54bf06de7f3be19abae1463939eccf941037d72acd11cec4a382df21fa3` (20 000
sats) created 2026-05-20 13:17:49 with a 1h deadline. Almost 4 hours later
the swap died with:

```
2026-05-20 17:02:19.180 [WRN] SWAP: receive swap 77b3f54b... stopped:
  wait for incoming vHTLC event: pull incoming vHTLC mailbox:
  rpc error: code = Unavailable desc = error reading from server:
  read tcp 10.1.63.109:51261->54.70.59.251:443: read: connection reset by peer
```

(`~/.darepod-signet/logs/signet/darepod.log:1971`)

For contrast, the SVRC ingress loop survived **eight separate identical TCP
RSTs / timeouts** against `16.148.235.208:443` over the same log window
(lines 1089, 1361, 1424, 1465, 1473, 1476, 1479, 1721, 1755, 1913, 2070).
Each one logs `Pull failed, retrying` and the loop heals.

`arkcli-signet swap list` still reports `state: SWAP_STATE_FAILED,
terminal_reason: ...read: connection reset by peer` for this hash — the
swap is permanently dead from a single transient network blip.

## Root cause

`MailboxOutSwapEventReceiver` has two wait methods, both with the same
shape:

- `WaitOutSwapHtlc`        — `sdk/swaps/out_swap_mailbox.go:75-94`
- `WaitIncomingVHTLC`      — `sdk/swaps/out_swap_mailbox.go:144-160`

```go
cursor := uint64(0)
for {
    resp, err := r.edge.Pull(ctx, &mailboxpb.PullRequest{...})
    if err != nil {
        return nil, fmt.Errorf("pull incoming vHTLC mailbox: %w", err)
    }
    ...
}
```

Any non-nil `err` (including `codes.Unavailable` from a TCP RST or
`io.EOF` from an HTTP/2 stream close) escapes the loop immediately. There
is no backoff, no reconnect, no cursor preservation across attempts, no
discrimination between transport-level and protocol-level errors. The
caller (`waitIncomingVHTLCNotification` at `sdk/swaps/out_swap.go:740-746`)
wraps the error and the receive FSM transitions to terminal failure.

## Reference shape that already works

`serverconn/ingress.go:30-130` (`ingressLoop`) does what we need:

```go
envelopes, nextCursor, err := a.pullBatch(ctx, state.PullCursor)
if err != nil {
    a.log.WarnS(ctx, "Pull failed, retrying", err,
        slog.Uint64("cursor", state.PullCursor),
    )
    a.sleepBackoff(ctx, &failCount)
    continue
}
```

with `retryDelay` (`ingress.go:462`) doing exponential backoff + jitter,
capped by `RetryMaxDelay`, and the cursor state checkpointed durably across
retries. This is exactly the loop shape every long-running mailbox consumer
needs, and it has already been battle-tested on the same endpoint in the
same daemon.

## Why duplicate? — proposal: refactor and reuse, don't duplicate

The user's question on review: *can the SDK just reuse `serverconn`
entirely?* Three options in increasing order of unification:

### A. Tactical fix — local retry loop inside the SDK receiver

Wrap each `edge.Pull` call in a retry-with-backoff inside
`MailboxOutSwapEventReceiver`. Preserve `cursor` across reconnect.
Distinguish `codes.Unavailable`, `io.EOF`, `io.ErrUnexpectedEOF`, and other
transport errors (retry) from real protocol errors (terminate). Bound only
by `ctx` / swap deadline.

Pros: smallest change, low risk.
Cons: duplicates the ingress-loop logic in a second place. Drift is
inevitable.

### B. Structural fix — extract a shared `MailboxPullLoop` helper

Extract a reusable helper that owns pull + backoff + cursor checkpoint from
`serverconn/ingress.go` and have both consumers use it:

- `serverconn.ServerConnectionActor.ingressLoop`
- `sdk/swaps.MailboxOutSwapEventReceiver.{WaitOutSwapHtlc,WaitIncomingVHTLC}`

Pros: single source of truth for mailbox reliability semantics.
Cons: helper must support both the actor-checkpoint path (durable
`AckState`) and the simpler SDK in-memory cursor path. Likely needs a
checkpoint-store interface so the SDK can plug in a memory or SQLite
implementation.

### C. Full unification — one ingress loop per identity, per-swap routes

Register the per-swap mailbox as a route on the existing `serverconn`
`EventRouter` (`serverconn/event_router.go:64`) so the daemon runs *one*
ingress loop on the identity-key mailbox and demultiplexes
`OutSwapHtlcEvent` / `InArkHtlcEvent` by payment hash to the receive
worker that's waiting for it. Eliminates the per-swap connection entirely.

Pros: matches the connectionless design intent literally; one TCP stream
to heal, not N; no per-swap mailbox churn; backpressure / cursor / ack
semantics already solved.
Cons: requires checking whether the swap server can multiplex out-swap
events onto the client's identity mailbox. Today the per-swap mailbox is
keyed by `OutSwapMailboxID(identityKey, paymentHash)`
(`sdk/swaps/out_swap_mailbox.go:141`), so this is a server-side coupling
to investigate. Construction site that would change:
`swapclientserver/service.go:369-373` (currently builds a per-receiver
`NewMailboxOutSwapEventReceiver(swapClients.mailbox, "")`).

## Recommendation

- **Minimum bar**: B (shared `MailboxPullLoop` helper). Anything less than
  B leaves us shipping two copies of the same loop, which is what the
  current bug is.
- **Destination**: C. The fact that the SDK keeps its own gRPC stub and
  cursor state is a layering accident; the swap event stream is just
  another route on the identity mailbox.

Acceptance criteria for whichever option lands:

- A single TCP RST or `codes.Unavailable` against the operator/mailbox
  endpoint must not move a receive swap to a terminal failure state.
- Reconnect must preserve the mailbox cursor so envelopes are neither
  missed nor double-delivered.
- The wait is bounded only by `ctx` cancellation and the swap deadline.
- Existing fakes in `sdk/swaps/out_swap_mailbox_test.go` should be
  extended with disconnect/reconnect coverage (loops that return
  `codes.Unavailable` N times then succeed).

## Code references

- One-shot pulls (the bug): `sdk/swaps/out_swap_mailbox.go:75-94`,
  `sdk/swaps/out_swap_mailbox.go:144-160`
- Caller that turns transport error into terminal swap failure:
  `sdk/swaps/out_swap.go:740-746`
- Reference retry loop: `serverconn/ingress.go:30-130`, with
  `sleepBackoff` and `retryDelay` at `serverconn/ingress.go:442-490`
- EventRouter that option C would extend:
  `serverconn/event_router.go:64-265`
- Construction site (one per-receiver mailbox today):
  `swapclientserver/service.go:369-373`

Roasbeef:
Closes #505.

In this PR, we make the SDK swap event receiver survive transient mailbox transport errors instead of treating them as terminal swap failures. The bug surfaced on signet: receive swap `77b3f54b...` (20 000 sats) died after a single `codes.Unavailable` TCP RST against the operator/mailbox endpoint, even though the `serverconn` ingress loop on the same daemon survived eight identical RSTs against the same endpoint over the same log window. The mailbox protocol is supposed to be connectionless at the message layer, so the SDK was clearly missing reliability semantics the daemon already had.

The fix is option B from the issue body: extract the retry+backoff+cursor primitives `serverconn/ingress.go` was already using inline, expose them as a small `serverconn/mailboxpull` package, and have `MailboxOutSwapEventReceiver` build on the same helper. Both consumers now back off identically when the shared mailbox endpoint flaps. The SDK's outer matching loop is unchanged: `cursor` is still a local `uint64` advanced from `resp.NextCursor` on success, and the helper does not mutate the `PullRequest`, so a reconnect resumes from the last cursor instead of rewinding to zero. The wait is bounded only by `ctx` cancellation and the swap deadline, which is what the connectionless-at-the-message-layer contract calls for.

Option C from the issue body (fold per-swap events onto the identity mailbox via the existing `EventRouter` and demultiplex by payment hash) is the right destination, but it needs `swapserver`-side coordination in a separate repo, so it's deferred to a follow-up. This PR closes the immediate bug without touching the per-swap mailbox layering.

## Helper shape

`mailboxpull.PullWithRetry` calls `Edge.Pull` and retries any non-`ctx` error after exponential backoff with jitter. It returns `ctx.Err` on cancellation rather than the underlying transport error, so callers further up the stack (the receive FSM, etc.) can tell "we're shutting down" from "endpoint is flapping." A nil logger is treated as `btclog.Disabled`. The zero-value `BackoffConfig` falls back to the package defaults (200 ms base, 30 s cap, matching `serverconn`), so callers cannot accidentally produce a tight retry loop by leaving the config blank.

`ServerConnectionActor.sleepBackoff` now delegates to `mailboxpull.Sleep` via a snapshot of the existing `RetryBaseDelay` / `RetryMaxDelay` knobs on `ConnectorConfig`. The outer `ingressLoop` body (ack → pull → dispatch → checkpoint) is untouched on purpose: replacing `pullBatch` with `PullWithRetry` would change how the dispatch/ack backoff counter interacts with pull failures, and that thread is worth pulling separately.

## Test coverage

The legacy `testOutSwapMailboxEdge` returned a single static `PullResponse` for every call and ignored the request cursor, so it couldn't tell a clean first-pull success from a disconnect-then-reconnect. We extend it with a `pullScript` of typed steps (each step is either an error or a response, consumed in order) and record every cursor seen on the wire, so the new cases drive the receiver deterministically without adding wall time to the suite.

Three new cases pin the contract the old loop violated. `RetriesOnUnavailable` scripts three `codes.Unavailable` errors followed by a successful envelope and asserts the receiver keeps pulling instead of failing on the first blip; this is the direct regression test for #505. `RespectsCtxCancel` scripts a permanently flapping endpoint, cancels the context, and asserts the receiver returns `context.Canceled` promptly rather than spinning forever. `PreservesCursorAcrossRetry` threads an empty batch that advances the cursor to 5, a transient `Unavailable`, and then a real event, and asserts the second and third pull requests both carry cursor 5: the previous code would have rewound to zero and potentially re-processed already-delivered envelopes.

## Test plan

- [x] `go test ./sdk/swaps/... ./serverconn/...` passes locally
- [x] new `serverconn/mailboxpull/pull_test.go` cases cover success, retry-then-success, ctx-cancel-mid-backoff, cursor pass-through
- [x] three new SDK cases (`TestMailboxOutSwapEventReceiverRetriesOnUnavailable`, `TestMailboxOutSwapEventReceiverRespectsCtxCancel`, `TestMailboxOutSwapEventReceiverPreservesCursorAcrossRetry`) pass
- [x] existing `TestMailboxOutSwapEventReceiverPullsAndAcks` / `TestMailboxOutSwapEventReceiverPullsInArkEvent` still pass unmodified
- [x] `make fmt-changed` and lint-local clean on the diff
- [x] `make commitmsg-lint range="origin/main..HEAD"` clean

### DIFF
--- a/serverconn/connector_test.go
+++ b/serverconn/connector_test.go
@@ -773,28 +773,7 @@ func TestIngress_PartialDispatch_NoDuplicateRedelivery(t *testing.T) {
 	)
 }
 
-// TestRetryDelay verifies the exponential backoff formula with jitter.
-func TestRetryDelay(t *testing.T) {
-	t.Parallel()
-
-	base := 100 * time.Millisecond
-	maxDelay := 5 * time.Second
-
-	// First attempt should be approximately base (within jitter range).
-	d := retryDelay(base, maxDelay, 1)
-	require.GreaterOrEqual(t, d, base/2)
-	require.LessOrEqual(t, d, base)
-
-	// High attempt should be capped at maxDelay.
-	d = retryDelay(base, maxDelay, 100)
-	require.LessOrEqual(t, d, maxDelay)
-	require.GreaterOrEqual(t, d, maxDelay/2)
-}
-
-// TestRetryDelay_DefaultsOnZero verifies that zero base/max get defaults.
-func TestRetryDelay_DefaultsOnZero(t *testing.T) {
-	t.Parallel()
-
-	d := retryDelay(0, 0, 1)
-	require.Greater(t, d, time.Duration(0))
-}
+// Backoff formula tests now live in serverconn/mailboxpull/pull_test.go
+// (TestRetryDelayClampsToMax, TestRetryDelayUsesDefaults) -- the formula
+// itself moved to that subpackage so the SDK pull loop and this actor's
+// ingress loop share one schedule.
--- a/sdk/swaps/out_swap_mailbox.go
+++ b/sdk/swaps/out_swap_mailbox.go
@@ -7,8 +7,10 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btclog/v2"
 	mailboxpb "github.com/lightninglabs/darepo-client/mailbox/pb"
 	"github.com/lightninglabs/darepo-client/serverconn"
+	"github.com/lightninglabs/darepo-client/serverconn/mailboxpull"
 	"github.com/lightninglabs/darepo-client/swaprpc"
 	"github.com/lightningnetwork/lnd/lntypes"
 )
@@ -37,20 +39,60 @@ type MailboxOutSwapEventReceiver struct {
 	mailboxID        string
 	pullWaitTimeout  time.Duration
 	pullMaxEnvelopes uint32
+
+	// pullBackoff controls the exponential backoff schedule applied when
+	// edge.Pull returns a transport error. The default schedule matches
+	// the serverconn ingress loop so the daemon's reconnect cadence is
+	// uniform across both consumers.
+	pullBackoff mailboxpull.BackoffConfig
+
+	// log receives one WARN per failed pull attempt. nil is treated as a
+	// no-op logger.
+	log btclog.Logger
+}
+
+// MailboxReceiverOption tweaks a MailboxOutSwapEventReceiver at construction
+// time. Defaults match the production wiring; tests use these to drive the
+// retry schedule deterministically.
+type MailboxReceiverOption func(*MailboxOutSwapEventReceiver)
+
+// WithMailboxPullBackoff overrides the exponential backoff schedule used when
+// edge.Pull returns a transport error.
+func WithMailboxPullBackoff(
+	cfg mailboxpull.BackoffConfig) MailboxReceiverOption {
+
+	return func(r *MailboxOutSwapEventReceiver) {
+		r.pullBackoff = cfg
+	}
+}
+
+// WithMailboxReceiverLog wires a structured logger so retry attempts surface
+// in daemon logs alongside the equivalent serverconn ingress messages.
+func WithMailboxReceiverLog(log btclog.Logger) MailboxReceiverOption {
+	return func(r *MailboxOutSwapEventReceiver) {
+		r.log = log
+	}
 }
 
 // NewMailboxOutSwapEventReceiver creates a mailbox-backed out-swap event
 // receiver. When mailboxID is empty, WaitOutSwapHtlc derives the mailbox from
 // the client identity key supplied by the caller.
 func NewMailboxOutSwapEventReceiver(edge mailboxpb.MailboxServiceClient,
-	mailboxID string) *MailboxOutSwapEventReceiver {
+	mailboxID string,
+	opts ...MailboxReceiverOption) *MailboxOutSwapEventReceiver {
 
-	return &MailboxOutSwapEventReceiver{
+	r := &MailboxOutSwapEventReceiver{
 		edge:             edge,
 		mailboxID:        mailboxID,
 		pullWaitTimeout:  5 * time.Second,
 		pullMaxEnvelopes: 1,
+		pullBackoff:      mailboxpull.DefaultBackoffConfig(),
 	}
+	for _, opt := range opts {
+		opt(r)
+	}
+
+	return r
 }
 
 // WaitOutSwapHtlc waits until the matching out-swap HTLC mailbox event is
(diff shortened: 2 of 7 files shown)


Roasbeef:
@claude review this 

@codex review this

serverconn/connector_test.go
```diff
@@ -773,28 +773,7 @@ func TestIngress_PartialDispatch_NoDuplicateRedelivery(t *testing.T) {
...
-}
-
-// TestRetryDelay_DefaultsOnZero verifies that zero base/max get defaults.
-func TestRetryDelay_DefaultsOnZero(t *testing.T) {
-	t.Parallel()
-
-	d := retryDelay(0, 0, 1)
-	require.Greater(t, d, time.Duration(0))
-}
+// Backoff formula tests now live in serverconn/mailboxpull/pull_test.go
```
Roasbeef, line 776: Remove comment

PR #508, opened 2026-05-21
multi: close walletdk visibility gaps for in-flight rounds + deposits

### LINKED ISSUE #500: daemonrpc: GetRound returns mostly-empty RoundInfo for live (actor-served) rounds
## Problem

`darepocli ark rounds get --round-id <id>` (and `ark rounds list`)
returns a `RoundInfo` with most non-state fields empty whenever the
round is served from the live actor path rather than the persisted
store. In practice this affects every round between intent
submission and the commitment tx confirming onchain — i.e. precisely
the window users want to watch.

## Reproduction

Boarding flow on signet against a fresh wallet, immediately after
the boarding deposit confirms (round joined, MuSig2 signing done,
commitment tx broadcast but not yet confirmed):

```
$ darepocli ark rounds get --round-id 019e4c0f-287f-77e7-b7b0-63009a55b478
{
  "round": {
    "round_id":          "019e4c0f-287f-77e7-b7b0-63009a55b478",
    "state":             "ROUND_STATE_INPUT_SIG_SENT",
    "is_temp":           false,
    "vtxos":             [],
    "commitment_txid":   "",
    "commitment_height": 0,
    "creation_time":     "0",
    "last_update_time":  "0",
    "input_outpoints":   [],
    "output_outpoints":  [],
    "failure_reason":    ""
  }
}
```

`commitment_txid` is empty even though the daemon already knows it
(it appears in the log):

```
[INF] ROND: temp(...) Received commitment transaction from server
  commitment_txid=4933176d99022f257ecb0af228db7f0eb89f129ec5c33d0e338cd39da1962ec9
```

## Root cause

`queryRoundStates` in `darepod/rpc_server.go:3077-3120` builds
`RoundInfo` with only four fields:

```go
rounds = append(rounds, &daemonrpc.RoundInfo{
    RoundId:       roundID,
    State:         clientStateToProto(info.State),
    IsTemp:        info.IsTemp,
    FailureReason: roundFailureReason(info.State),
})
```

The persisted-store path (`roundSummaryToProto`) populates many more
fields. `GetRound` (`darepod/rpc_operation_status.go:27-77`) checks
the live path first and returns early once a match is found, so the
extra fields never get filled in for in-flight rounds.

## Impact

Users currently have to grep the daemon log to find the commitment
txid they should be watching onchain — there is no other API that
exposes it once the round has been signed but before it has
confirmed. This is the window where users most want to observe
state, since it's the one period the operator has accepted but
nothing is final yet.

## Suggested direction

Populate the additional fields from `info` (the entries returned by
`round.GetClientStateResponse.States`) on the live path. At minimum
`CommitmentTxid` should be available — the round actor has the
commitment tx in hand by the time it transitions into
`ROUND_STATE_INPUT_SIG_SENT`. `CreationTime`, `LastUpdateTime`,
`InputOutpoints` and `OutputOutpoints` should also be derivable from
the actor's own state.

## Discovered via

Boarding test on signet against LND-backed darepod (
`wallet.type=lnd`). Documenting an inspection workflow for the
walletdk user guide.

### LINKED ISSUE #501: walletrpc: no way to inspect VTXOs the wallet is about to receive from an in-flight round
## Problem

`darepocli ark vtxos list` and `darepocli list --view vtxos` only
surface VTXOs whose commitment transaction has already confirmed.
There is no "pending VTXO" view for the window between a round being
signed (`ROUND_STATE_INPUT_SIG_SENT`) and the commitment tx
confirming onchain.

The `--status` filter on `ark vtxos list` accepts {`live`,
`pending_forfeit`, `forfeiting`, `forfeited`, `spent`,
`unilateral_exit`, `failed`, `spending`} — all of which are
post-creation lifecycle states. There is no value that means "output
of a round commitment tx I have signed but which has not yet
confirmed onchain."

## Reproduction

After a successful boarding intent → MuSig2 → input-sig-sent flow,
balance and VTXOs are zero even though the wallet has irrevocably
committed to receive a VTXO:

```
$ darepocli ark vtxos list
{ "vtxos": [] }

$ darepocli list --view vtxos
{ "vtxos": { "vtxos": [], "total": 0 } }

$ darepocli balance
{ "confirmed_sat": "0", "pending_in_sat": "0", "pending_out_sat": "0" }
```

The only signal that something is in flight is `ark rounds list`:

```
"state": "ROUND_STATE_INPUT_SIG_SENT"
```

…but that's coarse (no amount, no destination, no commitment txid —
see #500 for the related read-API gap on `RoundInfo`).

## Impact

Users in the boarding-deposit-confirmed → commitment-tx-confirmed
window have no way to know what VTXO they are about to receive, or
even confirm via the wallet that one is incoming. The `pending_in_sat`
balance bucket goes from zero (pre-deposit-confirmation) to zero
(post-deposit-confirmation, in-flight round) — there is no visible
transition until the commitment tx is mined, which can take many
minutes on signet.

## Suggested direction

Two options, either or both:

1. Add a `--status pending_round` (or similar name) to
   `ark vtxos list` that joins against in-flight rounds in the actor
   system, projecting each upcoming output as a synthetic VTXO entry
   with its expected `amount_sat`, the `commitment_txid` it will be
   created by, and the originating `round_id`. Could be paired with a
   note that these entries are not yet spendable.

2. Have `balance.pending_in_sat` reflect the value of upcoming-but-
   not-yet-created VTXOs as well. Currently `pending_in_sat` is only
   non-zero between deposit confirmation and round seal, which on a
   well-running operator can be too brief for users to observe.

Option 1 is structurally cleaner: it does not require redefining the
meaning of `pending_in_sat` and it keeps the VTXO list as the
authoritative view of upcoming + live + ended VTXOs.

## Related

Largely overlaps with #500 — fixing the empty `RoundInfo` fields on
the live path would at least let users derive the upcoming VTXO set
from `ark rounds get`. Option 1 above would be the canonical fix.

## Discovered via

Boarding test on signet against LND-backed darepod (
`wallet.type=lnd`). Documenting an inspection workflow for the
walletdk user guide.

### LINKED ISSUE #502: walletrpc: confirmed_sat includes confirmed boarding UTXOs, contradicting its own proto contract
## Problem

The `BalanceResponse.confirmed_sat` field is documented in the proto
as "total spendable VTXO amount" but currently also includes confirmed
on-chain boarding UTXOs that have **not** yet been swept into a VTXO.
Those funds are not spendable via the wallet's normal in-Ark flows —
the user has to call `ark board` (or wait for the eager-round-join
path) before they become a VTXO. So the field's name + docstring
diverge from the value it reports.

## Proto definition

`rpc/walletrpc/wallet.proto:403-414`:

```proto
message BalanceResponse {
    // confirmed_sat is the total spendable VTXO amount in satoshis.
    int64 confirmed_sat = 1;

    // pending_in_sat is the total in-flight inbound amount (boarding
    // plus receive operations).
    int64 pending_in_sat = 2;
    ...
}
```

By the proto's own contract, confirmed boarding UTXOs should be
counted under `pending_in_sat` (the docstring even calls out
"boarding") and only live VTXOs should appear under `confirmed_sat`.

## Reproduction

Signet, `eagerroundjoin=false`, with one live VTXO (99 745 sat from a
previous round) and one freshly confirmed boarding UTXO (100 000 sat,
not yet boarded):

```
$ darepocli list --view onchain
... { "txid": "bc10da58…", "kind": "boarding", "amount_sat": "100000",
      "status": "confirmed", "confirmation_height": 305423 } ...

$ darepocli ark vtxos list
{ "vtxos": [ { "outpoint": "203db6da…:0", "amount_sat": "99745" } ] }

$ darepocli balance
{
  "confirmed_sat":   "199745",   ← should be 99745 (live VTXO only)
  "pending_in_sat":  "0",         ← should be 100000 (boarding awaiting board)
  "pending_out_sat": "0"
}
```

`confirmed_sat` lumps the 100 000 boarding-confirmed sat together
with the 99 745 VTXO sat, even though the underlying `daemonrpc.
GetBalanceResponse` (`daemonrpc/daemon.proto:363-400`) already
exposes the two pools separately:

```
boarding_confirmed_sat   - confirmed boarding UTXOs
vtxo_balance_sat         - live VTXOs
total_confirmed_sat      - the sum above
```

The walletrpc layer appears to map `total_confirmed_sat` →
`confirmed_sat`, losing the distinction the proto field promises.

## Impact

Wallet-shape SDK callers (and any user reading `darepocli balance`)
see a number that overstates immediately-spendable Ark balance. With
`eagerroundjoin=false` (or any operator-driven host) this can be
indefinitely wrong until `ark board` runs.

## Suggested direction

Map walletrpc balance fields to match the proto docstrings:

- `confirmed_sat` ← `daemonrpc.vtxo_balance_sat`
- `pending_in_sat` ← `daemonrpc.boarding_confirmed_sat` +
  `daemonrpc.boarding_unconfirmed_sat` + any inbound swap rows
  already counted today

Or, alternatively, update the proto comment to admit the current
behavior — but the docstring is much more useful as the spec, since
the daemonrpc surface already has the granular fields available.

## Discovered via

walletdk guide testing on signet with LND-backed darepod, `eager
roundjoin=false`. Same session that surfaced #500 and #501.

### LINKED ISSUE #503: walletrpc: deposit activity row goes ENTRY_STATUS_COMPLETE on onchain confirm, not on boarding completion
## Problem

A `kind=DEPOSIT` row in `darepocli list` flips to
`status=ENTRY_STATUS_COMPLETE` the moment the boarding UTXO confirms
on-chain, even if the deposit has not yet been boarded into a VTXO.
Under `eagerroundjoin=false` (the operator-driven default) the
deposit can sit at "COMPLETE" indefinitely while still being
unspendable through the Ark wallet path — it only becomes
VTXO-spendable after a successful `ark board` round.

## Reproduction

Signet, `eagerroundjoin=false`, fresh wallet, one boarding deposit
of 100 000 sat sent + confirmed but not yet boarded:

```
$ darepocli list
{ "activity": { "entries": [
    {
      "id":     "bc10da5808a0224ec3c881c1403a73f8c154b5c31a1d5fcc786877c275963b72",
      "kind":   "ENTRY_KIND_DEPOSIT",
      "status": "ENTRY_STATUS_COMPLETE",            ← misleading
      "amount_sat":   "100000",
      "counterparty": "boarding"
    }
] } }

$ darepocli ark vtxos list
{ "vtxos": [] }                                     ← no VTXO yet

$ darepocli balance
{ "confirmed_sat": "199745",  "pending_in_sat": "0",
  "pending_out_sat": "0" }                          ← 100k is in confirmed_sat
                                                       (see also #502)
```

There is no follow-up status the user can wait for from the activity
view — a deposit that gets explicitly boarded later doesn't update
its row, because it's already "COMPLETE".

Compare to `eagerroundjoin=true` (the walletrpc-build default): under
that mode the same row only flips to COMPLETE after the round
commitment tx confirms, which is the "boarded, your VTXO is live"
moment. So the COMPLETE status string means two materially different
things depending on a server-side config bit.

## Root cause

`swapwallet/history.go:562-574`:

```go
func statusFromLedgerConfirmation(s string) walletrpc.EntryStatus {
    switch s {
    case "confirmed", "swept":
        return walletrpc.EntryStatus_ENTRY_STATUS_COMPLETE
    ...
    }
}
```

Any ledger row with `confirmation_status="confirmed"` is mapped to
COMPLETE regardless of whether the row represents a boarding
intermediate state or a finalised VTXO. The mapping doesn't consult
the boarding lifecycle (`BoardingStatus` in `wallet/interfaces.go`).

## Impact

Wallet-shape consumers using activity as their primary UI surface
will tell users their deposit is "complete" when in fact:

- It cannot be sent via OOR or in-swap yet.
- It does not appear in `ark vtxos list`.
- It needs an explicit `ark board` (eagerroundjoin=false) or another
  round tick (eagerroundjoin=true) to actually finish.

## Suggested direction

Distinguish boarding lifecycle states in the wallet-facing entry
status. Either:

1. Add an `ENTRY_STATUS_CONFIRMED` (or similar intermediate) state
   to walletrpc.EntryStatus, and only flip the DEPOSIT row to
   COMPLETE once the originating boarding UTXO has been adopted into
   a confirmed round commitment tx.

2. Keep the three-state enum but redefine the mapping for DEPOSIT
   rows so COMPLETE means "VTXO live" — i.e. the row stays PENDING
   from onchain confirmation through `ark board` and only transitions
   to COMPLETE on round commitment confirmation. This matches the
   intuitive read of "complete deposit".

Option 2 is less of a wire-protocol change and matches the
`eagerroundjoin=true` behaviour the codebase already exhibits.

## Related

- #500 — empty fields on live RoundInfo (next-step info also hard to
  find from the activity surface)
- #501 — no pending-VTXO view
- #502 — confirmed_sat lumps boarding UTXOs with live VTXOs

These four together describe the user-visible "what state are my
boarding funds in?" gap.

## Discovered via

walletdk guide testing on signet with LND-backed darepod, `eager
roundjoin=false`.

### LINKED ISSUE #504: walletrpc: list --view onchain 'round' rows have empty txid; commitment txid + outpoint only in free-text description
## Problem

`darepocli list --view onchain` returns entries with `kind="round"`
for VTXOs that were received via a boarding round, but the
structured `txid` field is empty and the actual commitment txid +
outpoint are only available by string-parsing the human-readable
`description` field.

This is the same shape of bug as #500 (live RoundInfo dropping
fields), in a different RPC surface: data exists, structured field
is empty, info is only available through unstructured text.

## Reproduction

After completing one boarding round (commitment tx confirmed,
VTXO live):

```
$ darepocli list --view onchain
{
  "onchain": {
    "txs": [
      {
        "txid":                "",                                    ← empty
        "kind":                "round",
        "amount_sat":          "99745",
        "fee_sat":             "0",
        "status":              "recorded",
        "confirmation_height": 0,
        "created_at_unix":     "1779394430",
        "description":         "VTXO received via round_boarding: 980c5911b4b2bc981c2b1ee7caa2a3d9e256f9bb1d84a0e595605463dab63d20:0"
      },
      ...
    ]
  }
}
```

The string `980c5911…:0` in `description` is the VTXO outpoint
(`<commitment_txid>:<vout>`), so the daemon has it — it just isn't
projected into the structured response.

For comparison, the sibling `kind=boarding` entries in the same
response do populate `txid` and `confirmation_height` correctly:

```
{ "txid": "bc10da58…", "kind": "boarding", "status": "confirmed",
  "confirmation_height": 305423, ... }
```

So the gap is specific to the `round` kind.

## Impact

- Programmatic consumers cannot link a round-kind entry to its
  commitment tx without text parsing.
- `confirmation_height` is 0 even though the commitment tx is
  confirmed; users get no signal of when the VTXO became live from
  this response.
- Discoverability of the commitment txid is the central UX problem
  flagged in #500 — same workaround (grep the daemon log)
  reappears here.

## Suggested direction

Populate `txid`, `confirmation_height`, and an `outpoint` (or
`vout`) field on `round`-kind onchain entries from the same
underlying record that the `description` string is built from.
Treat `description` as a *narrative* of structured data, not its
sole carrier.

## Related

- #500 (same anti-pattern in `ark rounds get`/`list` — fixing the
  live `RoundInfo` would likely give this surface the data it needs
  too).
- #503 (DEPOSIT row flips to COMPLETE on onchain confirm).
- #502 (confirmed_sat lumping).

## Discovered via

walletdk guide testing on signet with LND-backed darepod.

ellemouton:
## Summary

Five commits closing the user-visible visibility gaps surfaced during
walletdk guide testing on signet. Each commit is independently
reviewable and ships with a regression test that fails against the
pre-fix code.

- 46493f72 \`darepod: surface upcoming VTXOs and commitment txid for live rounds\` — closes #500
- 85a94fee \`darepod: add pending_round filter to ListVTXOs\` — closes #501
- 0c4f801a \`swapwallet: confirmed_sat should be VTXO-only\` — closes #502
- 241565ad \`swapwallet: keep DEPOSIT row PENDING after chain confirmation\` — closes #503
- 8d93568a \`ledger: stamp VTXO outpoint on vtxo_received chain fields\` — closes #504

The five issues together describe the same theme — \"what state are my
boarding funds in?\" — and were observed in one signet session with
\`eagerroundjoin=false\`. They sit in three independent layers
(\`daemonrpc\`, \`swapwallet\`/\`walletrpc\`, \`ledger\`) so they ship as one
stack rather than one tangled patch.

### #500 — live RoundInfo dropped commitment txid + upcoming VTXOs

\`queryRoundStates\` only populated \`RoundId\`, \`State\`, \`IsTemp\`, and
\`FailureReason\` for live actor-served rounds. The commitment txid was
visible in the daemon log but nowhere in the API response, and the
upcoming-VTXO set the wallet had just signed for never reached the
caller. Adds a state-aware \`liveRoundDetails\` helper that pulls the
commitment txid (from the \`TxID\` field, or by recomputing from the
embedded PSBT on MuSig2-phase states) and the per-owner VTXO amounts
(filtered to \`HasLocalOwner=true\` so outputs destined for other
clients don't inflate the local view).

### #501 — no pending-VTXO view between round-signed and commitment-confirmed

\`ark vtxos list\` only surfaced VTXOs whose commitment tx had
already confirmed; there was no enum value for \"signed but not yet
on-chain\". Adds \`VTXO_STATUS_PENDING_ROUND\` and routes \`ListVTXOs\`
with that filter to the round actor instead of the store, projecting
each in-flight round's owner-VTXOs as synthetic entries carrying
amount + round_id + commitment txid. Outpoint is left empty: the
precise leaf outpoint isn't finalised until the commitment tx
confirms. \`min_amount_sat\` still applies.

### #502 — \`walletrpc.confirmed_sat\` lumped boarding UTXOs with VTXOs

\`fetchBalance\` mapped \`daemonrpc.total_confirmed_sat → confirmed_sat\`,
which silently bundled confirmed-but-not-boarded boarding UTXOs into
the field the proto documents as \"total spendable VTXO amount\". Maps
\`vtxo_balance_sat → confirmed_sat\` and
\`boarding_confirmed_sat + boarding_unconfirmed_sat → pending_in_sat\`
so the proto contract is upheld.

### #503 — DEPOSIT activity row flipped to COMPLETE on chain confirm

\`statusFromLedgerConfirmation\` mapped every \`confirmation_status=\"confirmed\"\`
row to COMPLETE. For a DEPOSIT row backed by \`wallet_utxo_created\`
that's misleading — chain confirmation of the boarding UTXO is NOT
the boarded-into-a-VTXO signal. Under \`eagerroundjoin=false\` the row
can sit at COMPLETE indefinitely while being unspendable through any
in-Ark flow. Adds \`statusForLedgerRow(t, kind)\`; for DEPOSIT rows
backed by \`wallet_utxo_created\` the status is held at PENDING until a
follow-up \"boarded into a round\" ledger event lands (noted in the
helper's comment).

### #504 — round-kind onchain rows had empty txid + confirmation_height

\`handleVTXOReceived\` was writing the VTXO outpoint only into the
human-readable description and the idempotency key. The structured
\`ChainTxid\`/\`ChainVout\` columns sat empty, so \`list --view onchain\`
rendered round-kind entries with an empty txid that callers had to
recover by string-parsing the description. Stamps the outpoint hash
on \`ChainTxid\` and the index on \`ChainVout\` for every
\`vtxo_received\` entry.

## Test plan

- [x] \`make fmt-changed\` clean
- [x] \`make lint-changed-local\` clean across all 5 commits
- [x] \`go test ./darepod/\` passes — covers #500 and #501
- [x] \`go test -tags=\"walletrpc swapruntime\" ./swapwallet/\` passes — covers #502 and #503
- [x] \`go test ./ledger/\` passes — covers #504
- [x] Each regression test was verified to fail against the pre-fix code by stashing the impl and rerunning the targeted test
- [ ] CI green

## Notes for reviewer

- #500/#501 add a \`commitment_txid\` field and a new VTXO status enum to \`daemonrpc/daemon.proto\` — generated stubs regenerated via \`make rpc\` and bundled into the same commits per the proto-edit convention.
- #500's \`liveRoundDetails\` helper threads through \`PendingRoundAssembly\`, \`IntentSentState\`, \`QuoteReceivedState\`, \`RoundJoinedState\`, the MuSig2-phase states, \`InputSigSentState\`, and \`ConfirmedState\`. Anything else falls through to no-data, matching the prior contract.
- #503's fix is a half-step: DEPOSIT rows now stay PENDING after chain confirmation rather than reporting a misleading COMPLETE, but they cannot flip back to COMPLETE until a \"boarded into a round\" ledger event exists. That follow-up is intentionally out of scope here.

### DIFF
--- a/daemonrpc/daemon.pb.go
+++ b/daemonrpc/daemon.pb.go
@@ -4239,12 +4239,20 @@ type RoundInfo struct {
 	// is_temp indicates this round has a temporary key and hasn't
 	// been assigned a server round ID yet.
 	IsTemp bool `protobuf:"varint,3,opt,name=is_temp,json=isTemp,proto3" json:"is_temp,omitempty"`
-	// vtxos lists the VTXOs created by this round. Only populated
-	// for persisted rounds (input_sig_sent, confirmed). Empty for
-	// pending/in-memory rounds.
+	// vtxos lists the VTXOs the wallet expects to receive from this
+	// round. For rounds still in flight (any state before confirmed)
+	// only amount_sat is populated; the outpoint is empty because the
+	// precise leaf position in the VTXO tree is not finalised until
+	// the commitment transaction confirms. For confirmed rounds both
+	// outpoint ("txid:index") and amount_sat are populated. The
+	// commitment_txid field below carries the commitment tx hash
+	// independently once the round actor has received it from the
+	// operator, so callers do not need a populated outpoint to know
+	// which commitment they have signed.
 	Vtxos []*RoundVTXOInfo `protobuf:"bytes,4,rep,name=vtxos,proto3" json:"vtxos,omitempty"`
-	// commitment_txid is the round commitment transaction id when the round
-	// has been persisted.
+	// commitment_txid is the round commitment transaction id once the
+	// round actor has received it from the operator, or once the
+	// round has been persisted.
 	CommitmentTxid string `protobuf:"bytes,5,opt,name=commitment_txid,json=commitmentTxid,proto3" json:"commitment_txid,omitempty"`
 	// commitment_height is the block height that confirmed the commitment
 	// transaction, when known.
--- a/daemonrpc/daemon.pb.go
+++ b/daemonrpc/daemon.pb.go
@@ -121,6 +121,14 @@ const (
 	// VTXO_STATUS_SPENDING indicates the VTXO has been claimed for an
 	// out-of-round spend but the spend has not yet completed.
 	VTXOStatus_VTXO_STATUS_SPENDING VTXOStatus = 8
+	// VTXO_STATUS_PENDING_ROUND indicates a VTXO the wallet is about to
+	// receive from a round that has been signed but whose commitment
+	// transaction has not yet confirmed on-chain. Entries with this
+	// status are synthetic projections from the live round actor; they
+	// are not yet persisted, not yet spendable, and carry only the
+	// amount, originating round_id, and (once known) the commitment
+	// txid that will create them.
+	VTXOStatus_VTXO_STATUS_PENDING_ROUND VTXOStatus = 9
 )
 
 // Enum value maps for VTXOStatus.
@@ -135,6 +143,7 @@ var (
 		6: "VTXO_STATUS_UNILATERAL_EXIT",
 		7: "VTXO_STATUS_FAILED",
 		8: "VTXO_STATUS_SPENDING",
+		9: "VTXO_STATUS_PENDING_ROUND",
 	}
 	VTXOStatus_value = map[string]int32{
 		"VTXO_STATUS_UNSPECIFIED":     0,
@@ -146,6 +155,7 @@ var (
 		"VTXO_STATUS_UNILATERAL_EXIT": 6,
 		"VTXO_STATUS_FAILED":          7,
 		"VTXO_STATUS_SPENDING":        8,
+		"VTXO_STATUS_PENDING_ROUND":   9,
 	}
 )
 
@@ -6443,7 +6453,7 @@ const file_daemon_proto_rawDesc = "" +
 	"\x11WALLET_STATE_NONE\x10\x01\x12\x17\n" +
 	"\x13WALLET_STATE_LOCKED\x10\x02\x12\x16\n" +
 	"\x12WALLET_STATE_READY\x10\x03\x12\x18\n" +
-	"\x14WALLET_STATE_SYNCING\x10\x04*\x81\x02\n" +
+	"\x14WALLET_STATE_SYNCING\x10\x04*\xa0\x02\n" +
 	"\n" +
 	"VTXOStatus\x12\x1b\n" +
 	"\x17VTXO_STATUS_UNSPECIFIED\x10\x00\x12\x14\n" +
@@ -6454,7 +6464,8 @@ const file_daemon_proto_rawDesc = "" +
 	"\x11VTXO_STATUS_SPENT\x10\x05\x12\x1f\n" +
 	"\x1bVTXO_STATUS_UNILATERAL_EXIT\x10\x06\x12\x16\n" +
 	"\x12VTXO_STATUS_FAILED\x10\a\x12\x18\n" +
-	"\x14VTXO_STATUS_SPENDING\x10\b*\xf7\x03\n" +
+	"\x14VTXO_STATUS_SPENDING\x10\b\x12\x1d\n" +
+	"\x19VTXO_STATUS_PENDING_ROUND\x10\t*\xf7\x03\n" +
 	"\n" +
 	"RoundState\x12\x17\n" +
 	"\x13ROUND_STATE_UNKNOWN\x10\x00\x12\x14\n" +
--- a/daemonrpc/daemon.proto
+++ b/daemonrpc/daemon.proto
@@ -436,6 +436,15 @@ enum VTXOStatus {
     // VTXO_STATUS_SPENDING indicates the VTXO has been claimed for an
     // out-of-round spend but the spend has not yet completed.
     VTXO_STATUS_SPENDING = 8;
+
+    // VTXO_STATUS_PENDING_ROUND indicates a VTXO the wallet is about to
+    // receive from a round that has been signed but whose commitment
+    // transaction has not yet confirmed on-chain. Entries with this
+    // status are synthetic projections from the live round actor; they
+    // are not yet persisted, not yet spendable, and carry only the
+    // amount, originating round_id, and (once known) the commitment
+    // txid that will create them.
+    VTXO_STATUS_PENDING_ROUND = 9;
 }
 
 message VTXO {
--- a/daemonrpc/daemon.proto
+++ b/daemonrpc/daemon.proto
@@ -1115,13 +1115,21 @@ message RoundInfo {
     // been assigned a server round ID yet.
     bool is_temp = 3;
 
-    // vtxos lists the VTXOs created by this round. Only populated
-    // for persisted rounds (input_sig_sent, confirmed). Empty for
-    // pending/in-memory rounds.
+    // vtxos lists the VTXOs the wallet expects to receive from this
+    // round. For rounds still in flight (any state before confirmed)
+    // only amount_sat is populated; the outpoint is empty because the
+    // precise leaf position in the VTXO tree is not finalised until
+    // the commitment transaction confirms. For confirmed rounds both
+    // outpoint ("txid:index") and amount_sat are populated. The
+    // commitment_txid field below carries the commitment tx hash
+    // independently once the round actor has received it from the
+    // operator, so callers do not need a populated outpoint to know
+    // which commitment they have signed.
     repeated RoundVTXOInfo vtxos = 4;
 
-    // commitment_txid is the round commitment transaction id when the round
-    // has been persisted.
+    // commitment_txid is the round commitment transaction id once the
+    // round actor has received it from the operator, or once the
+    // round has been persisted.
     string commitment_txid = 5;
 
     // commitment_height is the block height that confirmed the commitment
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -659,12 +659,21 @@ func sumOnchainWalletConfirmed(ctx context.Context,
 }
 
 // ListVTXOs returns the set of VTXOs known to the wallet, optionally
-// filtered by status and minimum amount.
+// filtered by status and minimum amount. The VTXO_STATUS_PENDING_ROUND
+// filter is special: it bypasses the on-disk store and projects each
+// upcoming VTXO from the live round actor as a synthetic VTXO entry,
+// giving callers a way to see the outputs they have signed for but
+// which have not yet been created by an on-chain commitment.
 func (r *RPCServer) ListVTXOs(ctx context.Context,
 	req *daemonrpc.ListVTXOsRequest) (*daemonrpc.ListVTXOsResponse, error) {
 
 	if err := r.requireWalletReady(); err != nil {
 		return nil, err
+	}
+
+	if req.StatusFilter ==
+		daemonrpc.VTXOStatus_VTXO_STATUS_PENDING_ROUND {
+		return r.listPendingRoundVTXOs(ctx, req)
 	}
 
 	if r.server.vtxoStore == nil {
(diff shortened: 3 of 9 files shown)


ellemouton:
@claude review this

ellemouton:
Thanks for the review! Addressed findings 1–4 via fixup + autosquash; final 5-commit shape is unchanged (`fe1c0837` → `d4a3f2a4`).

### Finding 1 — proto comment inaccuracy (fixed)

`46493f72` → `fe1c0837`: rewrote the `RoundInfo.vtxos` comment to match the implementation — in-flight rounds carry empty outpoint, only `ConfirmedState` populates `txid:index`. Also added a cross-reference to `commitment_txid` so callers know where the txid lives during the in-flight window.

### Finding 2 — PSBT-phase duplication (fixed)

`46493f72` → `fe1c0837`: extracted `psbtPhaseDetails(state) (*psbt.Packet, []types.VTXORequest, bool)` helper. The 6 case bodies are now stated once inside the helper; `liveRoundDetails` collapses to a single call site after the typed-state cases. Kept the explicit type switch inside the helper rather than adding methods to the FSM state contract in the `round` package — your "type-assertion helper" suggestion exactly.

### Finding 3 — `ConfirmedState` double-reporting (fixed)

`85a94fee` → `07d0d6bb`: added an `info.State == ROUND_STATE_CONFIRMED` skip in `listPendingRoundVTXOs`, plus a new regression test (`TestListVTXOsPendingRoundSkipsConfirmedRounds`) verified to fail without the guard. Good catch — the actor's transient post-confirmation window is short but it's still a real double-count.

### Finding 4 — DEPOSIT→COMPLETE follow-up (fixed)

`241565ad` → `e7606431`: added an inline `// TODO(#503 follow-up):` in `statusForLedgerRow` documenting that the row is intentionally stuck at PENDING until a `boarding_fee_paid` (or equivalent "boarded into a round") ledger entry exists. Future reader now sees the gap explicitly rather than having to dig through PR history.

### Finding 5 — intent vs quote amount (not changed)

Left as-is per your "low risk and likely acceptable for a best-effort in-flight view" framing. Happy to revisit if you'd prefer the in-flight display to fall back to the quote amount once `QuoteReceivedState` carries it.

CI is green on the new tip.

Roasbeef (approved):
LGTM 🫒

PR #511, opened 2026-05-22
Expire unfunded receive swaps

### LINKED ISSUE #491: swaps: expire unpaid receive invoices and stop receive workers
## Problem

When the client creates a Lightning-to-Ark receive invoice, the invoice and swap-server route hint are created with a finite expiry. The receive session persists this deadline, but the active receive worker does not appear to stop listening when that deadline passes while still waiting for the server/mailbox HTLC event.

The current path is:

- `sdk/swaps/out_swap.go` sets `defaultReceiveExpirySeconds = 3600` and stores `s.deadline` when preparing the invoice.
- `swapclientserver` starts a daemon-owned receive worker for the payment hash and calls `session.Wait(rootCtx)`.
- In `ReceiveStateInvoiceCreated`, the FSM blocks in `waitForHTLCEvent` / `waitIncomingVHTLCNotification`.
- The mailbox receiver long-polls with only the daemon/root context, so an unpaid invoice can remain in `InvoiceCreated` after the invoice deadline.
- `ListPendingReceiveSwaps` filters only by terminal state, not by `deadline_unix`, so an expired unpaid receive can be resumed again on daemon restart.

Expiry is enforced later when waiting for a funded vHTLC via `waitForVHTLC(..., s.deadline, ...)`, but that does not help if no mailbox event arrives before the invoice expires.

## Expected behavior

Once an unpaid receive invoice reaches its persisted deadline, the client should stop waiting for matching swap events, transition the receive swap to `Expired`, release the daemon worker, and avoid resuming that row as pending on restart.

## Suggested fix

- Bound the `InvoiceCreated` mailbox wait by the receive session deadline.
- On deadline expiry, persist `ReceiveStateExpired` and return `ErrSwapExpired`.
- Ensure the daemon worker exits and publishes the updated summary.
- Consider making pending-list/resume logic robust against already-expired `InvoiceCreated` rows, either by excluding them from pending results or expiring them during the resume sweep.
- Add focused coverage for an unpaid receive invoice expiring before any mailbox event arrives, plus restart/resume behavior for that expired row.

## Relevant code

- `sdk/swaps/out_swap.go`: receive expiry setup, `waitForHTLCEvent`, `waitIncomingVHTLCNotification`, and `waitForVHTLC` deadline handling.
- `sdk/swaps/out_swap_mailbox.go`: mailbox long-poll loop uses the supplied context.
- `sdk/swaps/queries/swaps.sql`: `ListPendingReceiveSwaps` filters only terminal states.
- `swapclientserver/service.go`: receive workers call `ResumeReceiveViaLightning` then `session.Wait(rootCtx)`.

sputn1ck:
## Summary
- expire receive swaps that remain in invoice-created past the persisted invoice deadline
- add a 10s overdue-resume mailbox grace window so already-delivered funding messages can win over expiry
- keep wallet list/subscription status for SEND and RECV rows sourced from the swap FSM instead of wallet timeout overlays

Fixes #491

## Tests
- make unit pkg=sdk/swaps timeout=5m
- make unit pkg=swapwallet tags="swapruntime walletrpc" timeout=5m
- make unit pkg=swapclientserver tags="swapruntime" timeout=5m
- make fmt-changed-check base=origin/main
- git diff --check origin/main..HEAD

Note: make lint-local is currently blocked by unrelated line-length findings in cmd/darepocli/internal/gen-devrpc/main.go.

### DIFF
--- a/sdk/swaps/client.go
+++ b/sdk/swaps/client.go
@@ -441,7 +441,6 @@ type SwapClient struct {
 
 	waitPollInterval         time.Duration
 	overdueReceivePollWindow time.Duration
-	waitVHTLCTimeout         time.Duration
 	fundingResumeGracePeriod time.Duration
 	claimResumeGracePeriod   time.Duration
 	fundingExpiryBuffer      time.Duration
@@ -500,7 +499,6 @@ func NewSwapClientWithStore(server SwapServerConn, daemon DaemonConn,
 		log:                      log,
 		waitPollInterval:         2 * time.Second,
 		overdueReceivePollWindow: defaultOverdueReceiveMailboxPollWindow,
-		waitVHTLCTimeout:         60 * time.Second,
 		fundingResumeGracePeriod: defaultFundingResumeGracePeriod,
 		claimResumeGracePeriod:   defaultClaimResumeGracePeriod,
 		fundingExpiryBuffer:      defaultFundingExpiryBuffer,
--- a/sdk/swaps/client.go
+++ b/sdk/swaps/client.go
@@ -440,6 +440,7 @@ type SwapClient struct {
 	log        btclog.Logger
 
 	waitPollInterval         time.Duration
+	overdueReceivePollWindow time.Duration
 	waitVHTLCTimeout         time.Duration
 	fundingResumeGracePeriod time.Duration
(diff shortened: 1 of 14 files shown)


sputn1ck:
@claude review

PR #513, opened 2026-05-22
btcwbackend: Retry list unspent while syncing

sputn1ck:
## Summary
- return a retryable ListUnspent error while btcwallet is still chain-syncing/recovering
- avoid treating a pre-recovery empty UTXO result as authoritative during restart
- let the wallet actor's existing retry path rescan the same tip once btcwallet catches up

## Test Plan
- go test ./btcwbackend
- GOPATH=$(go env GOPATH) make fmt-changed-check
- env SYSTEST_PARALLEL=1 make systest backend=btcwallet case=TestBoardingRestartBeforeConfirmation (run from darepo root)


sputn1ck:
@claude please review

PR #498, opened 2026-05-21
walletrpc: Add wallet activity metadata

bhandras:
## Summary

- add typed wallet activity metadata for original requests and lifecycle progress while preserving existing field numbers such as `counterparty` and `progress.payment_hash`
- carry durable swap invoices through swap summaries so send/receive rows can surface invoices, payment hashes, and receive memos later
- derive pending boarding activity and boarding fees from existing wallet, ledger, and fee-accounting data without a DB migration
- hide internal swap OOR execution legs from the friendly activity feed using structured funding/claim session ids rather than amount-only matching
- add `WalletInspectionService.InspectActivity` for technical swap, VTXO, and ledger drill-downs, including structured history output indexes where available
- rename the everyday CLI surface to `darepocli activity`; activity defaults to table, inspection defaults to expanded, and JSON is emitted only with `--format json`
- render human-readable table/expanded timestamps with seconds and millisecond digits
- keep `cmd/darepocli/darepoclicommands/CLAUDE.md` and `AGENTS.md` in sync with the renamed command surface

Closes #497.

## Design Shape

The PR splits wallet history into two layers:

- `WalletService.List` remains the friendly activity feed. It exposes normalized request/progress metadata, hides reliably paired internal OOR swap legs, and keeps compact defaults suitable for UIs and CLI tables.
- `WalletInspectionService.InspectActivity` is the explicit technical drill-down. It correlates one activity id with swap summary data, vHTLC output details, VTXO movement rows, and local ledger rows.

Boarding visibility is derived from existing state instead of adding a migration. Zero-conf UTXOs paying to known boarding scripts become a pending boarding activity row, confirmed deposits come from the ledger, and displayed boarding fees are derived from fee-accounting rows.

The inspection path now carries transaction output indexes through the daemon history RPC. OOR receive outputs are correlated from structured txid/output-index fields rather than parsing ledger description strings.

## CLI Defaults

- `darepocli activity` defaults to `--format table`.
- `darepocli activity inspect <id>` defaults to `--format expanded`.
- JSON output is opt-in with `--format json`.
- Human-readable table and expanded timestamps render as `YYYY-MM-DD HH:MM:SS.mmm`.
- The table intentionally omits request payloads so invoices do not dominate the friendly feed; expanded and inspection views retain full identifiers.

## Review Notes

[...]

PR #518, opened 2026-05-22
swapwallet: hide structured OOR internals

bhandras:
Follow-up to #498.

## Motivation

The friendly wallet activity feed should not show ledger rows that are implementation details of OOR-backed swap execution. The previous follow-up only tightened malformed OOR send session handling, but testing showed the correct model also needs two related constraints:

- OOR receive rows should be correlated through structured OOR artifact bindings instead of description parsing.
- OOR receive rows paired with an OOR send in the same session are wallet-local change/materialization rows and should be hidden from the activity list, while remaining visible in inspection.

## Changes

- Classify structured OOR receive ledger rows as `oor` when they are transfer-in rows with an OOR created-output binding.
- Enrich structured receive rows with the OOR binding session id for swap/change correlation.
- Synthesize OOR receive rows from `oor_vtxo_bindings` and `vtxos` for older local rows that predate structured ledger outpoints.
- Require OOR receive correlation to consume rows already classified as `oor`.
- Hide paired OOR receive rows from activity when an OOR send in the same session proves they are wallet-local change/materialization.
- Keep OOR send hiding anchored to structured swap metadata, including the exact external payment amount.
- Require OOR send session ids to be exactly `chainhash.HashSize` before normalization.
- Keep inspection output able to show the hidden ledger rows and their roles.

## Validation

```sh
make fmt-changed-check base=origin/main
make sqlc-check
go test ./db ./darepod ./cmd/darepocli/darepoclicommands -tags 'walletrpc swapruntime'
go test ./swapwallet -tags 'walletrpc swapruntime'
```

### DIFF
--- a/db/ledger_store_test.go
+++ b/db/ledger_store_test.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"database/sql"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


bhandras:
@claude this follow-up PR addresses the remaining local review nit from #498.

Mapping against the review comments:

- `oorSendSessionID` now checks `len(row.GetSessionId()) != chainhash.HashSize` before normalizing the raw session bytes. This addresses the remaining minor nit that the helper previously only rejected empty session ids and relied on `chainhash.NewHash` for malformed non-32-byte input.
- The `hiddenOORLedgerRows` concern from the prior review was the same correlation path through `internalOORSends`/`oorSendSessionID`, so the earlier byte-order fix plus this explicit size guard covers both history and inspection callers.
- Added `TestOORSendSessionIDRequiresHashSizedSession` to reject a short session id and verify a valid raw session id normalizes to the production display string.

Validation:

```sh
make fmt-changed-check base=origin/main
go test ./swapwallet -tags 'walletrpc swapruntime'
```

Please review whether this closes the remaining review feedback.

bhandras:
@claude please re-review this updated version.

Context for the review:

- The branch is still one commit on top of `main`, but the scope is now broader than the first tiny guard patch.
- We are intentionally not adding description-string compatibility for old local rows. The correctness contract is structured data only.
- Please focus on whether the layering is right:
  - SQL history query classifies structured OOR receive rows as `oor` only when the transfer-in row has no round id and has persisted outpoint fields.
  - The wallet activity merger consumes only rows already typed as `oor` for OOR receive/change correlation.
  - Paired OOR receive rows are hidden as wallet-local change/materialization when a same-session send exists.
  - OOR send rows themselves are hidden only when swap metadata proves the external payment amount.
  - Inspection still exposes the hidden ledger rows and their roles.
- Please also check the edge cases covered by the tests: malformed send session ids, structured OOR receives without session_id, unstructured/malformed receives staying out of OOR classification, same-amount unmatched sessions, ordinary OOR send with change, and pay/receive swap internals.

Validation run locally:

```sh
make sqlc-check
go test ./swapwallet -tags 'walletrpc swapruntime'
go test ./db
go test ./darepod
```

PR #524, opened 2026-05-22
lwwallet: gap-fill skipped heights on next TipBlock event

### LINKED ISSUE #369: [security][high] Esplora failures can make lwwallet permanently skip blocks
Imported from Codex security findings CSV `codex-security-findings-2026-05-12T19-29-42.462Z.csv`.

Triage status: imported scanner finding, not yet validated by a human.

## Scanner Metadata

- Finding URL: https://chatgpt.com/codex/cloud/security/findings/f6a9f02c55148191ba7c595820fe18d2
- Repository: `lightninglabs/darepo-client`
- Severity: `high`
- Scanner status: `new`
- Detected at: `2026-05-12T15:41:45.966090Z`
- Commit: `5dba8352d747821c0b0a37da3aaeb4dd22ff268b`
- Committed at: `2026-05-05 16:11:33 -0500`
- Author email: `laolu32@gmail.com`
- Has patch: `false`
- Configured scan ID: `user-SFHXEfRY6GFhQZ5usE2FvSk3:github-1095286530`

## Relevant Paths

- lwwallet/tip_poller.go
- lwwallet/esplora_chain.go

### Original Title

Esplora failures can make lwwallet permanently skip blocks

### Original Description

This commit changes lwwallet block processing from per-consumer polling to a shared TipPoller that emits each height once. TipPoller updates its cached tip and broadcasts the TipBlock event, but subscriber processing is not part of the retry decision. EsploraChainService.processTipEvent fetches the raw block when wallet addresses are watched; if that raw-block fetch fails, it logs and returns before sending FilteredBlockConnected or BlockConnected to btcwallet. Since TipPoller has already advanced past that height, the chain service will not retry the skipped block on the next poll. When a later block arrives, btcwallet can be advanced using later notifications while transactions in the failed block were never scanned. Compared to the previous EsploraChainService poll loop, this is a regression: the old loop owned its bestBlock height and would leave it unchanged on a block fetch failure, causing the same height to be retried on the next poll.

## Original CSV Row

```text
finding_url: https://chatgpt.com/codex/cloud/security/findings/f6a9f02c55148191ba7c595820fe18d2
repository: lightninglabs/darepo-client
repository_url: https://github.com/lightninglabs/darepo-client
title: Esplora failures can make lwwallet permanently skip blocks
description: This commit changes lwwallet block processing from per-consumer polling to a shared TipPoller that emits each height once. TipPoller updates its cached tip and broadcasts the TipBlock event, but subscriber processing is not part of the retry decision. EsploraChainService.processTipEvent fetches the raw block when wallet addresses are watched; if that raw-block fetch fails, it logs and returns before sending FilteredBlockConnected or BlockConnected to btcwallet. Since TipPoller has already advanced past that height, the chain service will not retry the skipped block on the next poll. When a later block arrives, btcwallet can be advanced using later notifications while transactions in the failed block were never scanned. Compared to the previous EsploraChainService poll loop, this is a regression: the old loop owned its bestBlock height and would leave it unchanged on a block fetch failure, causing the same height to be retried on the next poll.
severity: high
status: new
detected_at: 2026-05-12T15:41:45.966090Z
committed_at: 2026-05-05 16:11:33 -0500
author_email: laolu32@gmail.com
assignee_name: 
assignee_email: 
has_patch: false
configured_scan_id: user-SFHXEfRY6GFhQZ5usE2FvSk3:github-1095286530
commit_hash: 5dba8352d747821c0b0a37da3aaeb4dd22ff268b
relevant_paths: lwwallet/tip_poller.go | lwwallet/esplora_chain.go
resolution_reason: 
```

ellemouton:
## Summary

Fix the regression reported in #369: a transient `/block/:hash/raw`
failure inside `EsploraChainService.processTipEvent` can permanently
strand a height between btcwallet's view and the chain tip.

When the shared `TipPoller` was introduced, it became the single source
of truth for the chain tip and only emits each height once. The
previous per-component poll loop owned its own `bestBlock` cursor and
would naturally retry the same height on the next tick after a fetch
failure. That invariant was lost on the subscriber side: a `GetRawBlock`
failure inside `processTipEvent` returned without sending
`FilteredBlockConnected` or `BlockConnected`, and TipPoller never
re-emits the same `TipBlock` event. btcwallet's sync height eventually
advanced past the skipped block via a later `BlockConnected` without
ever scanning it for relevant transactions.

This PR restores the cursor on the subscriber side. `EsploraChainService`
now treats `s.bestBlock.Height` as its canonical delivery cursor and
walks any gap between it and the latest `TipBlock` event before
processing the live event.

## Changes

- `lwwallet: add regression test for transient block-fetch tip skip` — the reproducer (gated on `t.Skip` so this commit alone does not introduce a failing test).
- `lwwallet: gap-fill skipped heights on next TipBlock event` — the fix and the corresponding un-skip.

Mechanics of the fix:

- `processTipEvent` snapshots `s.bestBlock.Height` and the watched
  pkScript set, drops duplicates (`event.Height <= lastDelivered`),
  and walks `(lastDelivered, event.Height-1]` via `fillGap` before
  processing the live event.
- `fillGap` re-fetches hash + header for each intermediate height
  (both LRU-cached in `EsploraClient`, so the warm path costs no extra
  network round trips) and delegates per-height emission to
  `deliverHeight`. Work is bounded by `defaultMaxGapFillPerTipEvent`
  (256, documented at the const) so a deep outage doesn't hold the
  `handleTipEvents` goroutine for longer than one poll interval.
- `deliverHeight` extracts the `GetRawBlock` -> filter -> emit pair
  and advances `s.bestBlock` only on full success. On any failure
  (raw-block fetch error, shutdown mid-send) `s.bestBlock` stays at
  the last successfully delivered height, so the next `TipBlock` event
  re-walks the unfinished stretch.

Same watched-scripts snapshot is passed to gap-fill and the live event
so every height processed in one invocation sees the same watch set.

## Test plan

- [x] `go test ./lwwallet/ -count=1` — passes.
- [x] `go test ./lwwallet/ -race -count=1` — passes (~20s).
- [x] `make fmt-changed` + `make lint-changed-local` — clean (0 issues).
- [x] Reproduction confirmed on the pre-fix branch: the new test fails with `observed=[102 103]` (height 101 is missed).
- [x] Reproduction is fixed by the second commit: the test passes with `observed=[101 102 103]`.

Closes #369.

### DIFF
--- a/lwwallet/esplora_chain.go
+++ b/lwwallet/esplora_chain.go
@@ -57,25 +57,55 @@ type EsploraChainService struct {
 	// whose methods do not accept a context parameter.
 	runCtx context.Context //nolint:containedctx
 
+	// maxGapFillPerTipEvent caps the number of missed heights that a
+	// single processTipEvent invocation will walk before yielding
+	// back to the handleTipEvents loop. Initialized from
+	// defaultMaxGapFillPerTipEvent and overridable via the
+	// WithMaxGapFillPerTipEvent functional option (tests use this to
+	// exercise the cap branch without revealing 256+ heights).
+	maxGapFillPerTipEvent int32
+
 	quit     chan struct{}
 	stopOnce sync.Once
 	wg       sync.WaitGroup
+}
+
+// EsploraChainServiceOption configures an EsploraChainService at
+// construction time. Options are applied in order, so a later option
+// overrides an earlier one for the same field.
+type EsploraChainServiceOption func(*EsploraChainService)
+
+// WithMaxGapFillPerTipEvent overrides the per-TipBlock gap-fill cap.
+// Intended for tests that need to exercise the bounded-walk branch
+// of fillGap; production callers should leave the default in place.
+func WithMaxGapFillPerTipEvent(n int32) EsploraChainServiceOption {
+	return func(s *EsploraChainService) {
+		s.maxGapFillPerTipEvent = n
+	}
 }
 
 // NewEsploraChainService creates a new chain.Interface backed by the
 // Esplora REST API. The provided TipPoller drives new-block
 // detection; the caller is responsible for starting and stopping it.
 func NewEsploraChainService(esplora *EsploraClient, tipPoller *TipPoller,
-	logger btclog.Logger) *EsploraChainService {
+	logger btclog.Logger,
+	opts ...EsploraChainServiceOption) *EsploraChainService {
 
-	return &EsploraChainService{
-		esplora:       esplora,
-		tipPoller:     tipPoller,
-		log:           logger,
-		notifications: make(chan interface{}, 100),
-		watchedAddrs:  make(map[string]btcutil.Address),
-		quit:          make(chan struct{}),
+	s := &EsploraChainService{
+		esplora:               esplora,
+		tipPoller:             tipPoller,
+		log:                   logger,
+		notifications:         make(chan interface{}, 100),
+		watchedAddrs:          make(map[string]btcutil.Address),
+		maxGapFillPerTipEvent: defaultMaxGapFillPerTipEvent,
+		quit:                  make(chan struct{}),
 	}
+
+	for _, opt := range opts {
+		opt(s)
+	}
+
+	return s
 }
 
 // Start seeds the initial chain tip from the configured TipPoller
@@ -683,8 +713,10 @@ func (s *EsploraChainService) handleTipEvents(ctx context.Context,
 // height, so the cap is essentially never hit. 256 is well above any
 // realistic burst of consecutive previously-failed events while still
 // bounding each invocation's wall-clock to well under one second of
-// Esplora HTTP under typical mempool.space latency.
-const defaultMaxGapFillPerTipEvent = 256
+// Esplora HTTP under typical mempool.space latency. Overridable via
+// WithMaxGapFillPerTipEvent for tests that need to exercise the
+// per-event cap branch without producing 256+ heights of traffic.
+const defaultMaxGapFillPerTipEvent int32 = 256
 
 // processTipEvent applies one TipBlock to btcwallet's notification
 // channel. The chain service owns its own delivery cursor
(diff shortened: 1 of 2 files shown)


ellemouton:
@claude review this

ellemouton:
Thanks @claude and @gemini-code-assist for the reviews. Pushed a squashed update (force-pushed onto the same two commits) addressing the substantive items:

**Addressed**
- @claude #1 — Inaccurate "LRU-cached" claim in `fillGap`'s doc comment. Reworded to make clear the block header is served warm from `blockHeaderCache`, but the height→hash lookup is a live `/block-height/:h` request (`EsploraClient` deliberately does not cache mutable height→hash mappings).
- @claude #3 — Test now uses `require.Equal` instead of `require.ElementsMatch` so the in-order delivery contract is asserted, not just presence. Added a comment explaining why ordering matters (btcwallet's sync pipeline relies on contiguous, in-order `BlockConnected` delivery).
- @claude #5 nit — Live `deliverHeight` call in `processTipEvent` now reads `_ = s.deliverHeight(...)` with a one-line rationale, making the dropped return value explicit.
- @gemini ×2 — Added `<-ctx.Done()` cases alongside `<-s.quit` in both notification sends inside `deliverHeight`. Goroutine now has a second exit so a full notifications buffer cannot wedge shutdown if the lifecycle context is cancelled without going through `Stop()`.

**Skipped (with rationale)**
- @claude #2 (double `FilteredBlockConnected` on shutdown mid-send) — pre-existing design issue, not introduced by this PR. Better tracked as a separate issue than expanded in a comment here.
- @claude #4 (deep-gap case drops live event for the invocation) — deliberate and already documented at the const and in the PR description; the alternative (delivering out-of-sequence while the middle is still missing) would break btcwallet's contiguous-sync assumption.
- @claude godoc nit — the `defaultMaxGapFillPerTipEvent` const comment already begins with the identifier name, which is the form `go doc` picks up.

Local checks: `make fmt-changed-check` clean, `make lint-changed-local` 0 issues, `go test ./lwwallet/ -race` passes. CI re-running.

Roasbeef (approved):
LGTM added on some additional tests 

🛼

PR #525, opened 2026-05-22
multi: rebuild refresh output template against current operator key (#520)

### LINKED ISSUE #520: client: refresh/join round can fail after operator key rotation
## Problem

A normal client refresh/join-round path can fail after an Ark operator key rotation. This is separate from the swap-specific stale operator-key issue in #514 and the swap quote binding race in #519.

Observed log from a manual refresh followed by `JoinNextRound`:

```text
2026-05-22 16:14:58.888 [INF] ROND: temp(c3bb96e6) State transition from_state=Idle to_state=PendingRoundAssembly
2026-05-22 16:14:58.888 [INF] ROND: Registered intent package forfeits=1 vtxos=1 leaves=0
2026-05-22 16:14:58.888 [INF] ARKW: Registered refresh intent package forfeits=1 vtxos=1 errors=0
2026-05-22 16:14:58.888 [INF] DRPD: VTXOs queued for refresh queued_count=1 error_count=0
2026-05-22 16:14:58.888 [INF] ROND: temp(c3bb96e6) Registration requested, preparing to join round boarding_intent_count=0 vtxo_intent_count=1
2026-05-22 16:14:58.889 [INF] ROND: temp(c3bb96e6) Intent balance check passed total_input="0.00001000 BTC" total_output="0.00001000 BTC" estimated_operator_fee="0 BTC"
2026-05-22 16:14:58.889 [INF] ROND: temp(c3bb96e6) Sending JoinRoundRequest to server boarding_requests=0 vtxo_requests=1 forfeit_requests=1 leave_requests=0
2026-05-22 16:14:58.902 [INF] ROND: temp(c3bb96e6) Building join round auth boarding_intent_count=0 vtxo_request_count=1 forfeit_request_count=1 leave_request_count=0
2026-05-22 16:14:58.904 [INF] ROND: temp(c3bb96e6) Built join round auth proof_input_count=1 message_len=288 signature_len=342 valid_from_block=305525 valid_until_block=305669
2026-05-22 16:14:58.904 [INF] ROND: temp(c3bb96e6) State transition from_state=PendingRoundAssembly to_state=IntentSent
2026-05-22 16:14:58.908 [INF] DRPD: JoinNextRound accepted
2026-05-22 16:14:59.256 [INF] ROND: temp(c3bb96e6) State transition from_state=IntentSent to_state="ClientFailed: join request invalid: operator key does not match: got 024f08048a96d562fc6271aef61dfd0be72953ee60764692547279293a941b3c8a, want 023c444b1d06e4608309dcd3947914f6b489c9582776bc0f07c91290825b4d7283"
```

The client successfully queues the refresh intent and emits a `JoinRoundRequest`, but the operator rejects it because the request contains VTXO/forfeit material tied to an old operator key.

## Likely cause

Normal VTXO descriptors and refresh requests persist/carry the operator key that was used when the VTXO was created. After an operator key rotation, refresh/join-round logic may reuse that stored key when building the refresh output, forfeit proof, or join auth request, while the operator validates against its current key.

Relevant client-side areas:

- `darepod/rpc_server.go`: `RefreshVTXOs` loads operator terms and builds `wallet.RefreshVTXOsRequest`.
- `wallet/wallet.go`: `handleRefreshVTXOs` registers refresh intents and logs `Registered refresh intent package`.
- `round/actor.go`: `buildVTXORequestFromRefresh` and `handleRefreshVTXORequest` build the round VTXO request from the refresh intent.
- `round/join_auth.go`: join auth material includes VTXO operator keys and is sent with `JoinRoundRequest`.
- `db/vtxo_store.go` / `db/round_store.go`: VTXO/round descriptors persist `operator_pubkey` / `operator_key`.

## Desired behavior

A client with spendable VTXOs should have a deterministic recovery path after an operator key rotation. At minimum:

- Refresh should not silently construct an invalid `JoinRoundRequest` that the operator rejects with `operator key does not match`.
- The client should either refresh/rebuild the new output side with the current operator key while preserving the historical key needed to spend the old input, or fail early with an actionable rotation-specific error before queueing a doomed round intent.
- Any old-key material that is still required to spend the existing VTXO must remain tied to the VTXO being spent; only newly-created outputs should use the active operator key.

## Acceptance criteria

- Add a regression test that simulates a VTXO created with operator key `K1`, then rotates the live operator key to `K2`, then attempts `RefreshVTXOs` + `JoinNextRound`.
- The test should prove the client no longer emits a join request that fails with `operator key does not match`.
- New VTXO outputs created by refresh use the current operator key `K2`.
- Historical spend/forfeit material for the old VTXO still uses the key committed to that VTXO (`K1`) where required by the original script.
- The user-facing state/logging should make the recovery behavior clear if automatic refresh across rotation is not possible.

ellemouton:
## Summary

Fixes #520. After an operator key rotation, the client's refresh path
silently constructed a `JoinRoundRequest` whose new VTXO output template
still carried the operator's pre-rotation key (K1), and the server
rejected the round with `ErrOperatorKeyMismatch`. This PR rebuilds the
new output's standard policy template against the operator's current
long-term key (K2) while leaving spend-time material for the old VTXO
untouched (it still has to commit to K1, because the on-chain output's
taproot tree was built against K1).

The cause sat in two places, both of which cloned the descriptor's stored
`PolicyTemplate` bytes verbatim into the new output:

- `vtxo/actor.go` — auto-refresh emission on expiry (`processOutbox` /
  `ForfeitRequest`).
- `wallet/wallet.go` — explicit `RefreshVTXOs` RPC handler
  (`handleRefreshVTXOs`).

Both now consult a `func() *btcec.PublicKey` provider wired in by
`darepod` from the cached `OperatorTerms` snapshot, and rebuild the new
output's template against the returned key. When the provider is unset
or the descriptor's stored policy is not the standard shape (vHTLC,
custom), the code falls back to the descriptor's stored bytes so legacy
paths and non-standard policies keep their current behavior.

## What's in the diff

- **`vtxo/policy.go`, `wallet/policy.go`** — new `RefreshOutputTemplate`
  helper on each descriptor type. Standard shape only; non-standard
  shapes return `ErrRefreshOperatorKeyUnsupported` so callers can fall
  back or surface rotation-specific UX explicitly.
- **`vtxo/actor.go`, `vtxo/manager.go`** — new `CurrentOperatorKey func()`
  field on `VTXOActorConfig` and `ManagerConfig` (manager forwards to
  each spawned actor). The auto-refresh emission goes through a new
  `refreshOutputTemplate` helper that prefers the rebuilt template and
  falls back to the stored bytes on non-standard shapes.
- **`wallet/wallet.go`** — new `WithCurrentOperatorKey` `ArkOption` and a
  parallel `refreshOutputTemplate` helper used by `handleRefreshVTXOs`.
- **`darepod/server.go`** — small `currentOperatorPubKey()` accessor that
  reads the cached `OperatorTerms`, wired into both `wallet.NewArk` and
  the `vtxo.ManagerConfig` at construction time.
- **`vtxo/refresh_operator_rotation_test.go`** — regression test with two
  subtests:
  - **rebuilds against rotated key when provider wired** — pins the fix:
    a `VTXOActor` configured with a `CurrentOperatorKey` returning K2
    emits a `RefreshVTXORequest` whose `PolicyTemplate` decodes to K2,
    not the K1 baked into the descriptor.
  - **falls back to stored template when provider unset** — pins the
    legacy path so harness tests and non-standard policy holders keep
    working unchanged.

## Design notes for review

1. **Forfeit and unilateral-exit material is intentionally untouched.**
   Only the new output's template is rewritten. The forfeit witness,
   control block, and unilateral exit script for the *input* VTXO still
   commit to K1 because that is what the on-chain output's taproot tree
   was built against. This matches the acceptance criteria in #520.

2. **Non-standard policies keep the legacy fallback.** vHTLC and other
   custom shapes still ship the stored bytes verbatim. A user holding a
   vHTLC VTXO under K1 across a rotation will see the same failure as
   before — out of scope for #520, but worth a follow-up if it matters.

3. **A nil provider is treated as 'not wired yet'.** The actor and wallet
   helpers fall back to `EffectivePolicyTemplate` so harnesses that don't
   mock `OperatorTerms` keep their current behavior.

## Test plan

- [x] `go test ./vtxo/ ./wallet/ ./darepod/ -count=1` — green
  (full unit suites for the three touched packages)
- [x] New regression test `TestRefreshEmissionUsesCurrentOperatorKey`
  passes both subtests
- [ ] Verify CI is green
- [ ] Recommended follow-up before bumping the submodule pointer in
  `darepo`: re-run the existing `TestRefreshIntegrationSingleVTXOLifecycle`
  itest against this branch to confirm the seal-time fee handshake and
  forfeit signing still work end-to-end through the new emission path
  (this PR's tests don't drive a real round).

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -600,6 +600,31 @@ func (s *Server) storeOperatorTerms(terms *types.OperatorTerms) {
 	s.operatorTerms.Store(terms)
 }
 
+// fetchCurrentOperatorPubKey issues a fresh GetInfo round-trip to the
+// operator and returns the operator's current long-term public key. The
+// daemon-startup OperatorTerms cache is also refreshed as a side effect so
+// other readers see the same snapshot. Used to plumb a live operator-key
+// lookup into the wallet and VTXO subsystems so refresh emissions build
+// the NEW VTXO output's policy template against the operator's join-time
+// key — VTXOs commit to their operator key for life, so the new output's
+// key is chosen at join time and stays stable on that VTXO forever.
+func (s *Server) fetchCurrentOperatorPubKey(ctx context.Context) (
+	*btcec.PublicKey, error) {
+
+	terms, err := s.fetchOperatorTerms(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("fetch operator terms: %w", err)
+	}
+
+	// Refresh the cache so unrelated readers (e.g. GetInfo) reflect the
+	// snapshot the refresh path used. The cache was previously only
+	// hydrated at daemon startup, which is what made it the wrong source
+	// of truth in the first place.
+	s.storeOperatorTerms(terms)
+
+	return terms.PubKey, nil
+}
+
 // isServerConnected returns the latest mailbox-ingress connectivity signal
 // reported by the daemon runtime.
 func (s *Server) isServerConnected() bool {
@@ -3231,6 +3256,7 @@ func (s *Server) initWalletActor(ctx context.Context,
 		),
 		wallet.WithClock(s.clk),
 		wallet.WithEagerRoundJoin(s.cfg.EagerRoundJoin),
+		wallet.WithFetchOperatorKey(s.fetchCurrentOperatorPubKey),
 	)
 	walletKey := actor.NewServiceKey[
 		wallet.WalletMsg, wallet.WalletResp,
@@ -3436,6 +3462,7 @@ func (s *Server) initVTXOManager(ctx context.Context,
 		LedgerSink:       fn.Some(ledger.NewSink(s.actorSystem)),
 		ChainResolver:    chainResolver,
 		RefreshFeeQuoter: s.autoRefreshFeeQuoter(),
+		FetchOperatorKey: s.fetchCurrentOperatorPubKey,
 		TerminalVTXOObserver: func(ctx context.Context,
 			outpoint wire.OutPoint) error {
 
--- a/vtxo/actor.go
+++ b/vtxo/actor.go
@@ -2,9 +2,11 @@ package vtxo
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/wire"
@@ -89,6 +91,22 @@ type VTXOActorConfig struct {
 	// emits RefreshVTXORequest with OperatorFee=0, which is fine:
 	// the seal-time quote is still the source of truth.
 	RefreshFeeQuoter RefreshFeeQuoter
+
+	// FetchOperatorKey, when set, returns the operator's current
+	// long-term public key by issuing a fresh GetInfo round-trip to
+	// the operator at the moment of an auto-refresh emission. The
+	// fetched key is used to build the NEW VTXO output's policy
+	// template; the input VTXO's stored operator key is intentionally
+	// not reused for the new output because VTXOs commit to their
+	// operator key for their entire lifetime, and the new output is a
+	// fresh VTXO whose operator key is chosen at join time.
+	//
+	// A nil callback causes refreshOutputTemplate to fall back to the
+	// descriptor's stored bytes (harness paths and pre-fix behavior).
+	// A non-nil callback that errors propagates the error so the
+	// refresh fails loudly rather than silently emitting against a
+	// stale key; the next expiry tick will retry.
+	FetchOperatorKey func(context.Context) (*btcec.PublicKey, error)
 }
 
 // VTXOActor manages the lifecycle of a single VTXO. It processes events using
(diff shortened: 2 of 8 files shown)


ellemouton:
@claude review this

ellemouton:
Reworked per review feedback. The previous iteration read the operator key from the daemon-startup `OperatorTerms` cache, which is exactly the source darepo#335 calls out as the wrong one. This version resolves the operator key at **refresh-join time** via a fresh `GetInfo` round-trip, so the new VTXO output commits to whatever key the operator declares at that moment — and then stays bound to it for that VTXO's life (consistent with the per-VTXO-stability model).

Key changes vs the previous push:

- Callback signature is now `func(context.Context) (*btcec.PublicKey, error)` (fetch, not cache read). On the darepod side it calls `fetchOperatorTerms(ctx)` and also refreshes the cache via `storeOperatorTerms`.
- Wallet's `handleRefreshVTXOs` hoists the fetch to a single call at the top of the handler so one `GetInfo` covers the whole batch instead of N per-outpoint round-trips.
- VTXO actor's auto-refresh fetches per emission. Fetch errors propagate so the refresh fails rather than silently emitting against a stale key; auto-refresh retries on the next expiry tick.
- Added 4-subtest coverage (vtxo) including error-propagation and nil-key paths, plus a parallel wallet-side test for `composeRefreshTemplate`.
- Fixed the `wallet/policy.go` error-wrap asymmetry that the earlier review caught (both decode failure modes now wrap with `ErrRefreshOperatorKeyUnsupported`).

@claude review this

ellemouton:
Round 3 — all of @claude's actionable items + the two CI failures addressed:

- **Commit msg L4 length**: rewrapped the body so every line ≤72.
- **`ll` 80-col**: aliased the callback type in the test (`fetchOperatorKeyFn`) so `newRefreshTestActor`'s signature fits.
- **`VTXOActorConfig.FetchOperatorKey` doc**: fixed — nil callback falls back, non-nil-that-errors propagates. The comment now matches the wallet-side parallel.
- **`processOutbox` log level**: downgraded `ErrorS` → `WarnS` with a brief comment citing the CLAUDE.md rule (`error` is for internal bugs, not external triggers like an unreachable operator). Stale message updated to "Failed to build refresh output template" since the helper can now fail for fetch reasons too.
- **`composeRefreshTemplate` godoc identifier break**: rephrased so the `refreshOutputTemplate` reference stays on one line.
- **`WarnS` formatting nit** in `handleRefreshVTXOs`: normalized to multi-line style for consistency with adjacent calls.

ellemouton:
@claude re-review

ellemouton:
Round 4 review addressed:

- **Issue-number citations stripped from code comments** (8 places across vtxo/actor.go, vtxo/manager.go, vtxo/policy.go, wallet/wallet.go, wallet/policy.go, darepod/server.go, and the two test files). The surrounding explanatory prose stands on its own — the design principle ("VTXOs commit to their operator key for life, the new output's key is chosen at join time") was already spelled out around the citations. Per CLAUDE.md, that context belongs in the PR description, not in the code where it rots once the issue closes.
- **"per CLAUDE.md's logging policy" self-reference removed** from the WarnS comment in processOutbox; the reasoning ("external trigger, not internal bug") stands on its own.

Roasbeef (approved):
LGTM 🏏

PR #536, opened 2026-05-27
darepod: fix vhtlc preimg registry startup

hieblmi:
This PR attempts to fix:
```
2026-05-27 08:10:01.533 [DBG] ACTR: Transaction begin failed during shutdown err="context canceled"                                                                                                                                                                                                                                                   
2026-05-27 08:10:01.533 [DBG] ACTR: Transaction begin failed during shutdown err="context canceled"                                                                                                                                                                                                                                                   
2026-05-27 08:10:01.533 [INF] CBKD: Stopping LND chain backend                                                                                                                                                                                                                                                                                        
2026-05-27 08:10:01.533 [INF] CBKD: LND chain backend stopped                                                                                                                                                                                                                                                                                         
Error: register recovery preimage resolver: rpc error: code = Unavailable desc = vhtlc recovery preimage registry not initialized
```

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -363,14 +363,16 @@ type Server struct {
 }
 
 // NewServer allocates a Server from a validated Config. The server is
-// inert until RunUntilShutdown is called. The walletReady channel is
-// initialized here so RPC handlers can select on it immediately.
+// inert until RunUntilShutdown is called. The walletReady channel and
+// recovery preimage registry are initialized here so RPC handlers can use
+// them immediately.
 func NewServer(cfg *Config) (*Server, error) {
 	return &Server{
-		cfg:         cfg,
-		clk:         clock.NewDefaultClock(),
-		walletReady: make(chan struct{}),
-		daemonReady: make(chan struct{}),
+		cfg:            cfg,
+		clk:            clock.NewDefaultClock(),
+		walletReady:    make(chan struct{}),
+		daemonReady:    make(chan struct{}),
+		vhtlcPreimages: &unrollpolicy.PreimageResolverRegistry{},
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


litbot-9000 (changes requested):
The code change itself looks correct — `PreimageResolverRegistry` is safe to zero-initialize (just a `sync.RWMutex` + nil resolver), and moving it to `NewServer` ensures the pointer is non-nil before `RegisterVHTLCRecoveryPreimageResolver` can be called. The `initUnrollSubsystem` side correctly reads from `s.vhtlcPreimages` instead of allocating a new one.

**Nit:** the commit message and PR title say `derepod` — should be `darepod`.

bhandras (approved):

litbot-9000 (approved):
Code review: the fix is correct.

Moving `PreimageResolverRegistry` allocation from `initUnrollSubsystem` to `NewServer` eliminates the startup race where the swap runtime's `RegisterVHTLCRecoveryPreimageResolver` call arrives before the unroll subsystem is wired. The registry is a concurrency-safe indirection point (mutex-guarded, starts with nil resolver) so creating it early is safe — the concrete resolver is installed later via `SetResolver` as before.

The pointer sharing is preserved: `initUnrollSubsystem` reads `s.vhtlcPreimages` into its local `preimages` var and passes it to the `ExitSpendPolicyResolver`, so the swap runtime's `SetResolver` call and the unroll registry's `ResolvePreimage` call both operate on the same object.

Existing test `TestNewSwapClientServiceRequiresRecoveryPreimageRegistry` still covers the nil-server path (it creates `NewRPCServer(nil)` which hits `r.server == nil` before reaching the registry check).

CI is green. LGTM.

PR #545, opened 2026-05-27
wallet: keep adopted boarding balance pending

### LINKED ISSUE #542: Balance "disappears" while boarding utxo is swept
1. I generate an onchain boarding address
2. I send funds to it

While this transaction is unconfirmed, I can see it as a pending balance in `darepocli balance`:

```
{
  "confirmed_sat": "0",
  "pending_in_sat": "100000",
  "pending_out_sat": "0"
}
```

Next, this utxo is swept by the service provider, and my balance shows 0:

```
{
  "confirmed_sat": "0",
  "pending_in_sat": "0",
  "pending_out_sat": "0"
}
```

Only once this transaction is confirmed does the balance show up:

**EXPECTED**:

While the funding utxo is swept by the provider, the pending balance should still be shown.

```
{
  "confirmed_sat": "99745",
  "pending_in_sat": "0",
  "pending_out_sat": "0"
}
```

bhandras:
## Summary

Fixes #542 by keeping boarded funds visible while they are adopted into a round but before the resulting VTXO is live.

When a boarding UTXO is accepted into a persisted round checkpoint, the intent moves to `BoardingStatusAdopted`. During that window the original on-chain UTXO can disappear from wallet `ListUnspent`, while the new VTXO is not spendable yet because the commitment has not confirmed. The balance projection previously skipped this adopted state, so wallet balance could temporarily drop to zero.

This adds explicit adopted-boarding accounting to `GetBalance` and maps it into wallet `pending_in_sat`, while keeping `confirmed_sat` VTXO-only.

## Validation

- `make rpc`
- `make fmt-changed`
- `make unit pkg=./wallet`
- `make unit-swapruntime pkg=./swapwallet`
- `make unit pkg=./darepod`
- `make unit pkg=./daemonrpc`
- `go run scripts/verify-schema-registry/main.go`
- `git diff --check`
- `make lint-changed-local`

### DIFF
--- a/daemonrpc/daemon.pb.go
+++ b/daemonrpc/daemon.pb.go
@@ -1388,8 +1388,10 @@ type GetBalanceResponse struct {
 	// vtxo_balance_sat is the total balance held in live (spendable)
 	// VTXOs.
 	VtxoBalanceSat int64 `protobuf:"varint,3,opt,name=vtxo_balance_sat,json=vtxoBalanceSat,proto3" json:"vtxo_balance_sat,omitempty"`
-	// total_confirmed_sat is the sum of all confirmed balances
-	// (boarding_confirmed_sat + vtxo_balance_sat).
+	// total_confirmed_sat is the sum of boarding_confirmed_sat +
+	// vtxo_balance_sat. boarding_adopted_sat is intentionally excluded
+	// because adopted funds are no longer confirmed boarding UTXOs and
+	// the resulting VTXOs are not yet live.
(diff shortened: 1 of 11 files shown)


bhandras:
@claude pls review

bhandras:
Addressed the review nit and folded it into the existing commit.

- Updated the `total_confirmed_sat` proto comment to explicitly state that `boarding_adopted_sat` is intentionally excluded.
- Regenerated RPC code with `make rpc`.
- Re-ran `make unit pkg=./daemonrpc` and `git diff --check`.

@claude pls review

Issue #543, opened 2026-05-27
stale mailbox state meeting a fresh local database


hieblmi:
Seeing the following after a fresh `darepod` install. Looks like my previous install left some stale server state.

wallet config
```
wallet.type=lnd                                                                
lnd.host=127.0.0.1:11009                                                       
lnd.tlspath=/home/bob/.lnd-signet/tls.cert                                     
lnd.macaroonpath=/home/bob/.lnd-signet/data/chain/bitcoin/signet/admin.macaroon
```

```
2026-05-27 12:22:41.541 [INF] WRPC: Running unified swapwallet resume sweep                                                                                                
2026-05-27 12:22:41.541 [INF] SVRC: Ingress loop starting mailbox_id=0349408eebe95246fb4e10835a6ed8b4a161ac8726b122014b84cccc0a754f6c32                                    
2026-05-27 12:22:41.541 [INF] DRPD: Daemon ready                                                                                                                           
2026-05-27 12:22:41.728 [DBG] SVRC: Pulled envelopes count=14 cursor=0 next_cursor=215383                                                                                  
2026-05-27 12:22:41.729 [WRN] SVRC: Dispatch failed err="adapt arkrpc.IndexerService/ListVTXOsByScripts event: unexpected incoming metadata correlation id: \"00aa8bfb11f09
881bbd2c7eb5df8735b\"" committed_to=0                                                                                                                                      
2026-05-27 12:22:42.082 [DBG] SVRC: Pulled envelopes count=14 cursor=0 next_cursor=215383                                                                                  
2026-05-27 12:22:42.082 [WRN] SVRC: Dispatch failed err="adapt arkrpc.IndexerService/ListVTXOsByScripts event: unexpected incoming metadata correlation id: \"00aa8bfb11f09
881bbd2c7eb5df8735b\"" committed_to=0                                                                                                                                      
```

Why this happens after a fresh install:

  - Your local darepod DB/checkpoint was reset, so mailbox ingress starts from cursor 0.
  - Your mailbox identity is still probably the same, especially in lnd mode because it is derived from the same lnd wallet key.
  - The remote operator/mailbox still has old response envelopes for that mailbox.
  - The new daemon has no in-memory waiter or durable session for those old responses.
  - The first old ListVTXOsByScripts response blocks dispatch, so you see committed_to=0.

  This is not a wallet/funds problem, but it is unhealthy operationally: ingress may keep retrying the same stale envelope and fail to process newer mailbox messages behind it.

[...]

PR #547, opened 2026-05-27
Tolerate stale vHTLC recovery state

bhandras:
## Summary

Fixes the stale vHTLC recovery follow-up from #543 by handling two stale local-state paths after a daemon database reset.

First, canceling a missing vHTLC recovery row is now treated as an idempotent success. If old swap metadata still has a `recovery_id` but the daemon's durable recovery table no longer has that row, cooperative settlement already won on that path, so cancellation can safely behave like the row was already gone.

Second, receive-side funding polls now stop when the authoritative indexer says the accepted vHTLC script is not registered for the current principal. That condition is persistent for the current identity: either script derivation disagrees, or stale local swap metadata survived while the server-side script registration did not. Retrying the same query only repeats the same stale-state log, so the receive FSM records a terminal failure with a clear reason.

## Validation

- `make fmt-changed`
- `make unit pkg=./darepod case=TestCancelVHTLCRecoveryMissingIsIdempotent`
- `make unit pkg=./sdk/swaps case=TestCancelVHTLCRecoveryIgnoresNotFound`
- `make unit pkg=./sdk/swaps case=TestWaitForVHTLCFailsOnUnregisteredScript`
- `make unit pkg=./darepod`
- `make unit pkg=./sdk/swaps`
- `make lint-changed-local`
- `git diff --check`
- `make commitmsg-lint range="HEAD~1..HEAD"`

### DIFF
--- a/darepod/rpc_vhtlc_recovery.go
+++ b/darepod/rpc_vhtlc_recovery.go
@@ -141,6 +141,10 @@ func (r *RPCServer) CancelVHTLCRecovery(ctx context.Context,
 		ctx, req.RecoveryId, req.Reason, cooperativeTxid,
 	)
 	if err != nil {
+		if errors.Is(err, db.ErrVHTLCRecoveryJobNotFound) {
+			return &daemonrpc.CancelVHTLCRecoveryResponse{}, nil
+		}
+
 		return nil, recoveryErrorToStatus(err)
 	}
 
--- a/darepod/rpc_vhtlc_recovery_test.go
+++ b/darepod/rpc_vhtlc_recovery_test.go
@@ -2,9 +2,14 @@ package darepod
 
 import (
 	"context"
+	"errors"
 	"testing"
 
+	"github.com/btcsuite/btcd/wire"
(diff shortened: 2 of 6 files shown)


bhandras:
@claude review

bhandras:
@claude re-review please.

Context since your last pass: this PR now also addresses the remaining `script not registered for principal` stale-state loop from the #543 follow-up logs.

The new bit is in the receive-side funding poll path: if the authoritative indexer says the accepted vHTLC script is no longer registered for the current principal, we classify that as terminal for this stale receive attempt instead of retrying forever. The reasoning is that the same query cannot become valid under the current identity without new registration/state; in the local-reset case, the local swap metadata survived but the server-side script registration did not.

Still intentionally narrow:
- generic indexer errors continue to be logged and retried;
- recovery escalation is still strict on missing recovery rows;
- only recovery cancellation is idempotent for missing rows.

New validation added: `TestWaitForVHTLCFailsOnUnregisteredScript`, plus the full `sdk/swaps` and `darepod` unit packages and `make lint-changed-local`.

PR #550, opened 2026-05-27
swaps: support immediate failed in-swap refunds

bhandras:
## Problem

When an Ark-to-Lightning pay-side in-swap funds its vHTLC and the swap server later hits a terminal Lightning payment failure, the payer's Ark funds stay locked until the timeout refund path matures. That is safe, but unnecessarily slow: the server already knows the Lightning payment failed and no preimage was learned.

## Proposed design

This PR adds the client/daemon pieces for a cooperative immediate refund:

- add daemon RPCs to prepare a deterministic custom-input OOR and to sign one prepared custom input;
- carry externally produced tapscript signatures through custom OOR inputs and durable OOR snapshots;
- assemble custom input witnesses in script-key order with externally supplied signatures;
- add the swap RPC client surface for in-swap refund authorization;
- have pay sessions ask the swap server to sign the exact prepared cooperative refund spend after safe terminal failure, then submit the signed custom OOR before `refund_locktime`.

The timeout `RefundWithoutReceiver` path remains the fallback when the server has not safely failed yet, is unavailable, or is an older implementation.

## Tests

- `make fmt-changed`
- `make lint-changed-local`
- `make unit pkg=./sdk/swaps timeout=5m`

### DIFF
--- a/cmd/darepocli/darepoclicommands/devrpc/registry_generated.go
+++ b/cmd/darepocli/darepoclicommands/devrpc/registry_generated.go
@@ -127,6 +127,20 @@ func generatedRegistry() []serviceSpec {
 					Comments: "SendOOR initiates an out-of-round transfer directly between the\nclient and operator, without waiting for a round.",
 				},
 				{
+					Name:     "PrepareOOR",
+					Aliases:  []string{"prepare-oor"},
+					Input:    "daemonrpc.PrepareOORRequest",
+					Output:   "daemonrpc.PrepareOORResponse",
+					Comments: "PrepareOOR builds the deterministic OOR package without submitting it.\nThis lets callers collect signatures from external custom-input\nparticipants before calling SendOOR with the same parameters.",
+				},
+				{
+					Name:     "SignOORCustomInput",
+					Aliases:  []string{"sign-oor-custom-input"},
+					Input:    "daemonrpc.SignOORCustomInputRequest",
+					Output:   "daemonrpc.SignOORCustomInputResponse",
+					Comments: "SignOORCustomInput signs one prepared custom OOR checkpoint input with\nthe daemon identity key.",
+				},
+				{
 					Name:     "RefreshVTXOs",
 					Aliases:  []string{"refresh-vtxos"},
 					Input:    "daemonrpc.RefreshVTXOsRequest",
--- a/daemonrpc/daemon.pb.go
+++ b/daemonrpc/daemon.pb.go
@@ -2952,9 +2952,14 @@ type CustomOORInput struct {
 	// amount_sat is the VTXO value in satoshis.
 	AmountSat int64 `protobuf:"varint,4,opt,name=amount_sat,json=amountSat,proto3" json:"amount_sat,omitempty"`
 	// pk_script is the VTXO taproot output script.
-	PkScript      []byte `protobuf:"bytes,5,opt,name=pk_script,json=pkScript,proto3" json:"pk_script,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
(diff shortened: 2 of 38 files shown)


bhandras:
@claude please review when you get a chance. This is the client/daemon side of safe immediate in-swap refunds; server companion PR: https://github.com/lightninglabs/swapdk-server/pull/55

bhandras:
@claude please review the refreshed client stack.

Key context: this adds the client-side refund authorization plumbing for failed in-swaps, including custom OOR preparation/signing and an automatic cooperative refund path. The latest push also fixes the custom OOR checkpoint owner leaf selection so the vHTLC `Refund` leaf is used for the actual custom input spend, while the checkpoint output falls back to the normal local-wallet/operator owner leaf when the policy leaf would require an extra participant signature.

Local validation run:
- `make fmt-changed-check`
- `make lint-changed-local`
- `make unit`
- targeted darepod tests for the custom owner-leaf fallback

bhandras:
codex: Re-reviewed the latest pushed head `bb2a982d`.

The cooperative refund retry issue is fixed: once the daemon accepts the custom-input refund OOR, the session keeps the `refundSessionID` in memory even if the following swap DB persist fails, and the focused regression test verifies the retry path does not submit a second refund. The old-server `Unimplemented` behavior and refunded wallet status assertions are also covered.

GitHub CI is fully green for this head. No blocking findings remain from my review; the broader restart/concurrency/TLV/witness-ordering cases are good follow-up hardening rather than blockers.

bhandras:
claude: re-reviewed at `bb2a982d`. All three fixes from the last pass land cleanly.

**Retry-bug fix** — the split is correctly shaped: `persistRefundSessionID` sets in-memory state first, then persists *without* rollback; `initiateRefund` short-circuits on `state == PayStateRefundInitiated`. The next FSM tick re-enters `tryCooperativeRefund`, hits the `if s.refundSessionID != "" { ... }` short-circuit, and the repeat `initiateRefund` no-ops. `TestPaySessionCooperativeRefundKeepsAcceptedSessionOnPersistFailure` pins `sendCustomCalls == 1` across both ticks — the regression-protection is real.

One residual edge worth a one-line comment near `persistRefundSessionID`: the in-memory promotion only protects **same-process** retries. On crash-during-persist, the reloaded session has empty `refundSessionID` and the next tick would re-submit, relying on daemon-side OOR idempotency. Worth documenting the layered contract so future readers don't read this as airtight against crashes.

**`codes.Unimplemented` fallback** — added to the silently-ignored set at `in_swap.go:1130-1133`; existing test extended to a loop over `{Unavailable, Unimplemented}` asserting `sendCustomCalls == 0` for both. Closes the old-server-noise gap.

**Pending REFUNDED → PENDING** — assertion added to `TestStatusFromSwapState`.

Agreed on the deferral calls. The same-pubkey external-sig overwrite is defense-in-depth only — exploitation would require forging Schnorr under the client key, which is the security model itself.

**Follow-ups to track** (all darepo-side hardening, not blockers):

1. TLV round-trip for `ExternalSignatures` at the three boundaries (`maxExternalSignatures=8`, script ≤520, sig ≤65) plus trailing-byte and corrupted-pubkey rejection. Pairs with extending `TestTransferInputSnapshotRoundTrip` to set the field non-empty.
2. Daemon-level `PrepareOOR` / `SignOORCustomInput` suites in `darepod/`: happy path, `SessionID` determinism, single-input enforcement, error-code mapping (`ErrCustomCheckpointInputSigning` → `Internal`, other → `InvalidArgument`), `BuildCustomTransferInputs` rejection for unowned VTXO.
3. Positive witness-ordering test for `assembleCustomFinalWitness` — 2-key (and ideally 3-key) leaf, assert the final stack is in reverse-script-key order. The current test only covers the rejection branch; a forward-order revert would not be caught.

LGTM to merge.

PR #556, opened 2026-05-27
darepod: expand ~ in config paths during Validate

bitromortac:
Related to https://github.com/lightninglabs/darepo-client/issues/553. Tilde-prefixed paths set via the config file or CLI flags (e.g. `datadir: ~/.darepod`) were only expanded inside NetworkDir() and LogDir(); every other DataDir consumer read the literal string. As a result, e.g. the swap subserver default `filepath.Join(DataDir, "swaps.db")` produced a directory named "~" in the daemon CWD.

Centralize expansion in a new Config.expandPaths() called at the top of Validate so downstream code can treat path fields as absolute. Covers DataDir, LogDirPath, Lnd TLS/macaroon, wallet password and btcwallet sources, and the swap DB / TLS cert paths. NetworkDir and LogDir drop their per-call expandTilde now that DataDir is normalized upfront.

Users will probably need to migrate from the `~` folder to the intended folder.

### DIFF
--- a/darepod/config.go
+++ b/darepod/config.go
@@ -719,6 +719,10 @@ func DefaultConfig() *Config {
 // Validate checks the config for internal consistency and returns an error
 // if any required fields are missing or invalid.
 func (c *Config) Validate() error {
+	if err := c.expandPaths(); err != nil {
+		return err
+	}
+
 	switch c.Network {
 	case "mainnet", "testnet", "regtest", "simnet", "signet":
 	default:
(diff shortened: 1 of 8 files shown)


darepod/config.go
```diff
@@ -954,28 +958,62 @@ func networkToChainParams(network string) (*chaincfg.Params, error) {
...
+			&c.Wallet.BtcwalletDataDir, &c.Wallet.BtcwBlockSource,
+			&c.Wallet.BtcwFilterSource,
+		)
+	}
+	if c.Swap != nil {
+		fields = append(
+			fields, &c.Swap.DatabaseFileName,
+			&c.Swap.ServerTLSCertPath,
+		)
 	}
```
bhandras, line 1018: `Server.TLSCertPath` should be included here too. Unlike `RPC.TLSCertPath`/`RPC.TLSKeyPath`, this field is live today: both the gRPC and REST operator clients read it directly with `os.ReadFile`. So `server.tlscertpath=~/ark/tls.cert` still resolves against a literal `~` directory after this change, even though the rest of the path normalization now happens in `Validate`.

darepod/config.go
```diff
@@ -954,28 +958,62 @@ func networkToChainParams(network string) (*chaincfg.Params, error) {
...
+// /home/user/.darepod/data/regtest). DataDir is expected to be
+// tilde-expanded already; Validate normalizes it via expandPaths.
 func (c *Config) NetworkDir() (string, error) {
-	dataDir, err := expandTilde(c.DataDir)
-	if err != nil {
-		return "", err
-	}
-
-	return filepath.Join(dataDir, "data", c.Network), nil
+	return filepath.Join(c.DataDir, "data", c.Network), nil
```
bhandras: This makes `NetworkDir` depend on `Validate` having already run. The standalone daemon and `walletdk.Start` do validate first, but `sdk/ark.StartEmbedded` currently clones the supplied `darepod.Config` and starts `NewServer` without calling `Validate`. A caller passing `darepod.DefaultConfig()` or any `DataDir` beginning with `~` would now get a literal `~/.darepod/data/<network>` path; before this PR, `NetworkDir` expanded that on demand. I think either `NetworkDir`/`LogDir` should keep doing the expansion defensively, or `StartEmbedded` should validate before starting.

bhandras (commented):
I left two path-normalization comments. The main things are: one live TLS cert path is still skipped, and moving expansion out of `NetworkDir`/`LogDir` creates a new dependency on callers having run `Validate` first.

darepod/config.go
bitromortac: Fixed by calling `Validate` in `StartEmbedded` to centralize the tilde exansion and to add more defensive code paths.

darepod/config.go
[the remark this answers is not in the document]
bitromortac: Added them to not forget `c.RPC.TLSCertPath` and `c.RPC.TLSKeyPath` later.

darepod/config.go
```diff
@@ -954,28 +958,74 @@ func networkToChainParams(network string) (*chaincfg.Params, error) {
...
+			&c.Swap.ServerTLSCertPath,
+		)
+	}
+
+	for _, p := range fields {
+		expanded, err := expandTilde(*p)
+		if err != nil {
+			return fmt.Errorf("expand path %q: %w", *p, err)
+		}
+		*p = expanded
```
ellemouton, line 1025: note that this writes through to the original (pre copy) value of the config passed to `StartEmbedded` which i think it should not. way to fix is to do a deep copy in `cloneDaemonConfig` instead of a shallow copy

ellemouton (approved):
thanks for this! 

left one comment & i think one of bhandras's comments still need to be resovled

darepod/config.go
bitromortac: Added `Server.TLSCertPath`

darepod/config.go
bitromortac: Ah thanks, added clones for user configuration (but not for instantiated resources on the cfg)

bhandras (approved):
Re-reviewed the latest head and checked the comment threads. The earlier path-normalization issues look addressed: `Server.TLSCertPath` is included, `StartEmbedded` validates before startup, and the embedded config clone now avoids writing path expansion/default mutations back into the caller's config.

Validated locally:
- `make unit pkg=./darepod`
- `make unit pkg=./sdk/ark`
- `make unit pkg=./cmd/darepod`
- `make lint-changed-local`
- `git diff --check origin/main...HEAD`

PR #559, opened 2026-05-28
btcwbackend: reorg-aware chain notifier forwarding (C0b)

ellemouton:
Task 1b for the reorg-safety epic (darepo#454) — Neutrino-backed sibling of
#461 (lwwallet/Esplora).

neutrino's chainntnfs adapter already drives the full
`Confirmed / NegativeConf / Done` (and `Spend / Reorg / Done`) lifecycle
natively from its compact-block-filter scanner, but btcwbackend's
`RegisterConf` and `RegisterSpend` previously only read the first
`Confirmed` / `Spend` event and let the rest fall on the floor. The result:
any downstream consumer subscribed to a `chainsource.ConfRegistration.Reorged`
or `.Done` channel silently saw nothing on the neutrino backend, defeating
the whole reorg-safety substrate (#422).

This PR mirrors what `chainbackends/lnd.go` already does for the LND
backend:

- Persistent select loop that forwards `Confirmed`/`Spend`,
  `NegativeConf`/`Reorg`, and `Done` to the matching chainsource channel.
- Closed-channel arms `nil`'d out so the select doesn't busy-loop after
  one upstream sub-channel tears down.
- LIFO defer order: `event.Cancel` → `notifyCtx` cancel → close
  downstream chans, so the upstream notifier stops writing before the
  outgoing channels close.
- Field type widened from `*neutrinonotify.NeutrinoNotifier` to the
  `chainntnfs.ChainNotifier` interface so a stub notifier can drive the
  lifecycle in tests without standing up a real neutrino chain service.

The reorg depth carried on `NegativeConf` is intentionally dropped — same
choice the LND backend makes — so both backends present the same wire
shape to chainsource consumers.

## Stack

#422 (chainsource substrate) → **this PR**. Sibling of #461 (lwwallet).

## Test plan

- [x] `go vet ./...`
- [x] `make unit pkg=btcwbackend`
  (`TestRegisterConfForwardsReorgAndDone`,
  `TestRegisterSpendForwardsReorgAndDone`,
  plus pre-existing `TestSubmitPackage*`)
- [ ] CI integration tests

### DIFF
--- a/btcwbackend/chain_backend.go
+++ b/btcwbackend/chain_backend.go
@@ -38,8 +38,13 @@ type ChainBackend struct {
 	neutrinoCS *neutrino.ChainService
 
 	// notifier provides chain notification services backed by
-	// neutrino's compact block filter scanning.
-	notifier *neutrinonotify.NeutrinoNotifier
+	// neutrino's compact block filter scanning. Stored under the
+	// chainntnfs interface (concrete type is
+	// *neutrinonotify.NeutrinoNotifier in production) so reorg-aware
+	// forwarder tests can substitute a stub that drives the full
+	// Confirmed / NegativeConf / Done lifecycle without spinning up a
+	// neutrino chain service.
+	notifier chainntnfs.ChainNotifier
 
 	// feeEstimator provides fee estimation from a web API since
 	// neutrino has no mempool visibility.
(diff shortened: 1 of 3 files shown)


ellemouton:

PR #589, opened 2026-05-28
darepod: report only spendable VTXOs in GetBalance vtxo_balance_sat

### LINKED ISSUE #584: darepod GetBalance: vtxo_balance_sat overstates spendable balance (includes Spending + PendingForfeit VTXOs)
## Summary

`darepod`'s `GetBalance` RPC reports `vtxo_balance_sat` that **includes
non-spendable VTXOs** (status `Spending` and `PendingForfeit`), overstating the
truly spendable Ark-side balance. Downstream consumers that treat
`vtxo_balance_sat` as spendable liquidity — notably the swapdk-server
auto-boarder — then make wrong decisions and can stop replenishing liquidity
entirely.

## Root cause

`darepod/rpc_server.go` `GetBalance` computes the VTXO balance by summing
`ListLiveVTXOs`:

```go
liveVTXOs, err := r.server.vtxoStore.ListLiveVTXOs(ctx)
...
resp.VtxoBalanceSat = int64(vtxo.SumBalance(liveVTXOs))
```

But `ListLiveVTXOs` is not a "spendable" query — it is the **actor-recovery**
query. From `db/sqlc/queries/vtxo.sql`:

```sql
-- name: ListLiveVTXOs :many
-- ListLiveVTXOs returns all VTXOs that are not in a terminal state.
-- Terminal states are: Forfeited (3), Spent (4), UnilateralExit (5), Failed (6).
-- Non-terminal states: Live (0), PendingForfeit (1), Forfeiting (2), Spending (7).
-- This is used during startup to recover active VTXO actors.
SELECT * FROM vtxos
WHERE (status < 3 OR status = 7) AND spent = FALSE
ORDER BY creation_time DESC;
```

So it deliberately returns Live(0), PendingForfeit(1), Forfeiting(2) **and**
Spending(7) — everything non-terminal, because those actors must be recovered on
startup. Reusing it for a balance figure means `vtxo_balance_sat` counts VTXOs
that cannot currently fund anything.

## Observed impact on signet

A single 4,834,745-sat VTXO was pinned in `Spending` (a stuck OOR checkpoint) and
a 5,000-sat VTXO in `PendingForfeit`. Spendable balance was **0**, yet:

```
darepod GetBalance.vtxo_balance_sat = 4,839,745   (= 4,834,745 + 5,000)
```

The swapdk-server auto-boarder consumes this as `LiveVTXOSat` (its comment:
"currently spendable Ark-side VTXO balance", `autoboarder/policy.go:78`). With
`live_vtxo_sat=4,839,745` plus a pending 3M board, `effective_ark_liquidity_sat`
read ~10.8M — above the 5M low-water mark — so the auto-boarder boarded
`amount_sat=0` on every 5-min tick and **never refilled actual spendable
liquidity** (which was 0). Out-swap funding stayed starved.

```
[INF] Auto-boarder decision action=board_confirmed
  reason=below_low_water_with_confirmed_boarding amount_sat=0
  effective_ark_liquidity_sat=10839121 live_vtxo_sat=4839745
  boarding_confirmed_sat=2999688 active_reservation_sat=2999688 ...
```

## Suggested fix

`GetBalance` should report **only spendable** VTXOs in `vtxo_balance_sat`:
status `Live(0)` with `spent = FALSE`. Concretely, add a dedicated query
(`ListSpendableVTXOs` / `SumSpendableVTXOBalance` filtering `status = 0 AND
spent = FALSE`) and use it in `GetBalance`, leaving `ListLiveVTXOs` for its
actor-recovery purpose. (Optionally expose a separate "non-terminal/total" figure
if a caller genuinely wants it.)

Defensively, swapdk-server could also stop equating `vtxo_balance_sat` with
spendable liquidity, but the authoritative fix is to make the RPC field mean what
its name and consumers assume.

Roasbeef:
In this PR, we fix #584: `darepod`'s `GetBalance` reports a `vtxo_balance_sat`
that overstates the spendable Ark-side balance, because it sums VTXOs that
can't actually fund anything.

The handler summed `ListLiveVTXOs`, but that query backs actor recovery, not
spendability. By design it returns every non-terminal VTXO, i.e. `Live(0)`,
`PendingForfeit(1)`, `Forfeiting(2)`, and `Spending(7)`, since those actors all
have to be respawned on startup. Summing that set for a balance figure means a
VTXO pinned in `Spending`, or one parked in `PendingForfeit`, gets counted as
spendable when it isn't.

That mismeasure bites downstream. The swapdk-server auto-boarder reads
`vtxo_balance_sat` as spendable liquidity (`LiveVTXOSat`). On signet a single
4,834,745 sat VTXO stuck in `Spending` plus a 5,000 sat `PendingForfeit` made
the field read ~4.84M while real spendable balance was 0. With a pending 3M
board on top, effective liquidity looked like ~10.8M, above the 5M low-water
mark, so the auto-boarder boarded `amount_sat=0` on every tick and never
refilled the actual spendable pool. Out-swap funding stayed starved.

## The fix

We add `vtxo.SumSpendableBalance`, which filters a descriptor set to `Live` via
the existing `FilterDescriptors` and then sums it. `GetBalance` now uses that
over the recovered set. `ListLiveVTXOs` is left alone for its recovery callers
(queued-outpoint synthesis, recovery snapshots, manager startup), and
`SumBalance` stays for callers that want the total over an already-filtered set.

We do this as a Go-side filter over the set `GetBalance` already fetches, rather
than a dedicated `status = 0 AND spent = FALSE` SQL query. The reason is purely
mechanical: sqlc codegen here is Docker-based and Docker isn't available in this
environment, so a new query can't be regenerated and CI's `sqlc-check` would
reject hand-written generated code. The Go filter makes the field mean what its
name and consumers assume without touching the generated layer. A scalar DB sum
stays a clean follow-up if we'd rather not load the descriptors at all.

## Tests

`vtxo/filter_test.go` covers the helper directly. `TestSumSpendableBalance`
rebuilds the issue's exact mixed-status set (two `Live`, plus `PendingForfeit`,
`Forfeiting`, and a 4.8M `Spending`) and asserts only the `Live` amounts sum,
while `SumBalance` over the same set still totals everything (that contrast is
why `GetBalance` must not use it). `TestSumSpendableBalanceAllNonSpendable`
covers the all-stuck case reporting zero, and `TestSumSpendableBalanceEmpty`
guards empty input.

## Test plan

- [x] `go build ./vtxo/ ./darepod/`
- [x] `go vet ./vtxo/`
- [x] `go test ./vtxo/ -count=1`

### DIFF
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -502,19 +502,25 @@ func (r *RPCServer) GetBalance(ctx context.Context,
 			"balance: %v", err)
 	}
 
-	// Fetch VTXO balance by summing all live VTXOs using the
-	// package-level SumBalance helper. Propagate query errors rather
+	// Fetch the spendable VTXO balance. Propagate query errors rather
 	// than silently zeroing — a returned zero is functionally
 	// indistinguishable from "no VTXOs" and would mislead a UI that
 	// trusts the response, which is the same rationale that makes
 	// fetchBoardingBalance error-strict above.
 	if r.server.vtxoStore != nil {
+		// ListLiveVTXOs returns every non-terminal VTXO (Live,
+		// PendingForfeit, Forfeiting, Spending) because it backs
+		// actor recovery, not spendability. Sum only the spendable
+		// (Live) subset: counting PendingForfeit/Spending here would
+		// overstate vtxo_balance_sat, which consumers such as the
+		// auto-boarder read as spendable liquidity and would then
+		// stop replenishing.
 		liveVTXOs, err := r.server.vtxoStore.ListLiveVTXOs(ctx)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "fetch vtxo "+
 				"balance: %v", err)
 		}
-		resp.VtxoBalanceSat = int64(vtxo.SumBalance(liveVTXOs))
+		resp.VtxoBalanceSat = int64(vtxo.SumSpendableBalance(liveVTXOs))
 	}
 
 	// Fetch the confirmed balance of the backing on-chain wallet so
--- a/vtxo/filter.go
+++ b/vtxo/filter.go
@@ -54,3 +54,19 @@ func SumBalance(descs []*Descriptor) btcutil.Amount {
 
 	return total
 }
+
+// SumSpendableBalance computes the total balance across only the
+// spendable subset of descriptors. Only Live VTXOs can fund a spend;
+// the other non-terminal states (PendingForfeit, Forfeiting, Spending)
+// are still "live" for actor-recovery purposes but cannot back a spend,
+// so they are excluded here. Callers reporting a spendable balance must
+// use this rather than SumBalance over a recovery-oriented VTXO set,
+// otherwise the figure overstates spendable liquidity.
+func SumSpendableBalance(descs []*Descriptor) btcutil.Amount {
+	spendable := FilterDescriptors(descs, FilterOptions{
+		Status:    VTXOStatusLive,
+		StatusSet: true,
+	})
+
+	return SumBalance(spendable)
+}
--- a/vtxo/filter_test.go
+++ b/vtxo/filter_test.go
@@ -1,0 +1,81 @@
+package vtxo
+
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/stretchr/testify/require"
+)
+
+// TestSumSpendableBalance verifies that SumSpendableBalance counts only
+// Live VTXOs, excluding the other non-terminal states that ListLiveVTXOs
+// returns for actor recovery (PendingForfeit, Forfeiting, Spending). This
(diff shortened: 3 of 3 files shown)


ellemouton (approved):

Roasbeef:
## Codex 5.5 automated review

VERDICT: APPROVE

The change makes `vtxo_balance_sat` match the existing proto semantics by summing only `VTXOStatusLive` descriptors, while keeping `ListLiveVTXOs` available for actor recovery. I checked the status model, the generated `ListLiveVTXOs` query, and the `GetBalance` totals; the new predicate excludes `PendingForfeit`, `Forfeiting`, and `Spending` while preserving the documented `total_confirmed_sat = boarding_confirmed_sat + vtxo_balance_sat` relationship.

Blocking: none.

Major: none.

Minor: none.

Nit: none.

Verification: `git diff --check` passed. I attempted `go test ./vtxo -count=1`, but this review worktree does not have the Go toolchain on PATH (`go: command not found`).

PR #593, opened 2026-05-28
vtxo+darepod: reconcile orphaned spending reservations on startup

### LINKED ISSUE #587: vtxo: VTXO permanently stuck in Spending status when reserving spend dies without release (orphaned reservation, no resume reconciliation)
## Summary

A VTXO reserved for a spend (`SpendingState`, DB `status = 7`) can be left
**permanently stuck** across restarts when the spend that reserved it (an OOR
send or a round registration) terminates without emitting a `SpendReleasedEvent`.
On resume the VTXO actor keeps it in `SpendingState` on the assumption that a
resumable OOR session will eventually release or complete the claim — but if no
such session exists, nothing ever releases it. The VTXO is then unspendable until
batch expiry forces an on-chain unilateral exit.

This is the root blocker behind the signet liquidity starvation, and it compounds
with #584 (the stuck `Spending` VTXO is still counted in `vtxo_balance_sat`, so
the auto-boarder thinks liquidity is fine and never refills).

## Evidence (signet, 2026-05-28)

The `0b02c92e…800a:1` VTXO (4,834,745 sat) is stuck:

- `ark vtxos list` → `status: VTXO_STATUS_SPENDING`, `spent_by_txid: ""` — no
  actual on-chain spend exists.
- It **survives a full pod restart**: on the fresh pod the VTXO actor reloads
  straight into `SpendingState` and only ever receives `BlockEpochEvent`s, never
  a release.
- `ark oor list` shows the only OOR session referencing it is the **incoming,
  completed** session that *created* it:
  ```
  session_id: 0b02c92e…800a
  direction:  OOR_SESSION_DIRECTION_INCOMING
  status:     OOR_SESSION_STATUS_COMPLETED
  created_outpoints: [ 0b02c92e…800a:1 ]
  ```
  There is **no outgoing / resumable** OOR session to "resume and later release
  the claim." So the reservation is orphaned.

The reservation most likely originated from a failed OOR vHTLC-funding attempt
(`SendOOR` reserving the VTXO, then failing) or a failed auto-boarder round
registration (round failed "quote pass closed with no survivors", see #585), in
either case without a persisted, resumable spend to roll it back.

## Root cause (code)

`vtxo/transitions.go`:

- `handleSpendReserve` (LiveState, ~L81) moves the VTXO to `SpendingState` and
  persists `VTXOStatusSpending`.
- `SpendingState.ProcessEvent` (~L745) only leaves `SpendingState` via:
  - `SpendCompletedEvent` → `Spent`,
  - `SpendReleasedEvent` → `Live` (sent by `manager.rollbackSpend`, ~L712),
  - `BlockEpochEvent` at **critical expiry** → `UnilateralExit`,
  - `ForceUnrollEvent` → `UnilateralExit`.
- On resume:
  ```go
  case *ResumeVTXOEvent:
      // On resume, stay in SpendingState. The OOR session will
      // resume and later release or complete the claim.
      return &VTXOStateTransition{ NextState: s }, nil
  ```

The `ResumeVTXOEvent` handler assumes a resumable spend session always exists. If
the reserving spend died without `rollbackSpend` (crash, abandoned funding
attempt, failed round registration), no `SpendReleasedEvent` is ever sent and the
VTXO is stranded in `SpendingState`. There is **no startup reconciliation** that
cross-checks persisted `Spending` VTXOs against actually-live spend sessions and
releases orphans.

## Impact

- The VTXO is unspendable indefinitely → out-swap funding (OOR) reports
  `insufficient funds` despite a 4.8M VTXO sitting there.
- Via #584 it is still summed into `vtxo_balance_sat`, so the swapdk-server
  auto-boarder reads liquidity as above low-water and boards `amount_sat=0`,
  never refilling spendable liquidity.
- The only automatic escape is batch expiry (`batch_expiry 307246`, ~6 days out),
  which forces an on-chain unilateral exit — wasteful, and currently blocked by
  the CPFP zero-output unroll bug (#580).

## Suggested fix

1. **Resume-time reconciliation:** when resuming a VTXO in `SpendingState`,
   cross-check it against live/resumable spend references (OOR session store +
   in-flight round registrations). If nothing references the VTXO, emit
   `SpendReleasedEvent` to return it to `Live` instead of trusting an implicit
   resume.
2. **Rollback robustness:** ensure every spend-initiation path (OOR send, round
   registration) guarantees a `rollbackSpend` / `SpendReleasedEvent` on failure
   (deferred/reconciled), so a `Spending` reservation cannot outlive its spend.
3. **Operator remediation:** add an RPC/CLI to release an orphaned `Spending`
   reservation back to `Live` without forcing an on-chain unilateral exit.

Roasbeef:
In this PR, we fix #587: a VTXO reserved for an OOR spend can get stuck in
`SpendingState` (DB `status = 7`) forever. The reservation is meant to be
short-lived: the spend either completes the VTXO or emits `SpendReleasedEvent`
to free it. But if the spend dies without releasing (a crash, an abandoned
vHTLC-funding attempt), the reservation is orphaned. On restart the VTXO actor
reloads straight into `SpendingState` and only ever sees `BlockEpochEvent`s, so
nothing releases it, and the VTXO is unspendable until batch expiry forces an
on-chain unilateral exit. On signet this was the root of the out-swap liquidity
starvation: a 4,834,745 sat VTXO pinned in `Spending`, never coming back.

## Approach (option A): startup reconciliation against live OOR sessions

A VTXO only ever enters `SpendingState` through one path: the OOR send flow,
which reserves the selected VTXOs as inputs (`SelectAndReserveSpendRequest` ->
`SpendReserveEvent`). Round registration reserves via the *forfeit* path
(`PendingForfeit`/`Forfeiting`), a separate lifecycle, so it's out of scope
here. That means the complete set of legitimate `Spending` holders after a
restart is exactly the set of input outpoints held by live, resumable outgoing
OOR sessions. Anything in `Spending` that isn't in that set is an orphan.

So at startup, once the VTXO manager and the OOR subsystem have both recovered,
we run one reconcile pass:

- The server asks the OOR actor for the input outpoints of its live (pending)
  outgoing sessions via `ListSessionsRequest`, which the actor already supports.
  The OOR actor processes its prepended restart message before this Ask (the
  mailbox is FIFO), so its in-memory session set is fully restored from the
  durable checkpoint by the time we read it.
- The server hands that claimed set to the VTXO manager via
  `ReconcileSpendingReservationsRequest`. The manager reads the authoritative
  current `Spending` set from the store (not the Start-time snapshot, so a spend
  that resolved during recovery isn't mistaken for an orphan) and releases every
  Spending VTXO not in the claimed set, reusing the same `SpendReleasedEvent`
  path as a normal release.

Safety: if the OOR query fails we abort rather than release, since an incomplete
claimed set would free a VTXO a live spend still holds. The pass is non-fatal to
startup; a failure just leaves the VTXOs reserved until the next restart, same
as today.

The claimed set is gathered in the server and passed into the manager rather
than queried from inside the VTXO actor. The manager has no handle on the OOR
session store, and keeping the query in the server avoids a `vtxo -> oor` import
cycle.

## Why not option B (durable reservation index)

The more thorough fix is to make the reservation itself durable: persist
`(owner_id, input_outpoint)` atomically with the `-> Spending` transition,
delete it on release/complete, and on startup release any Spending VTXO whose
owner record is absent or resolves to a terminal/nonexistent owner. That closes
the gap deterministically instead of leaning on startup ordering, and it
generalizes cleanly to the forfeit path too.

We didn't take that here because it costs a schema migration plus write-path
hooks at every reserve/release/complete site, and option A already recovers the
stuck VTXOs using state we already persist (the OOR durable checkpoint plus the
VTXO status). Option B is the natural follow-up if we want crash-proof
reservation accounting across both the spend and forfeit paths. The keys are: a
`spending_reservations(owner_kind, owner_id, outpoint)` table, atomic
insert/delete on the status transitions, and a startup sweep that releases rows
with no live owner.

## Operator escape-hatch (pending)

The issue also asks for an operator CLI/RPC to force-release an orphaned
`Spending` reservation without waiting for the automatic pass. That needs a new
gRPC method, and proto generation in this repo is Docker-based (`make rpc` ->
`scripts/gen_protos_docker.sh`), which isn't available in my environment, so I
can't generate or verify the stubs here. With the automatic reconciliation in
place a restart already unsticks orphans, so this is a convenience rather than
the critical path. I'll fold it in once proto regen is available, or it can land
as a fast follow-up.

## Tests

`vtxo/reconcile_test.go` covers the manager handler.
`TestReconcileSpendingReservationsReleasesOrphans` builds three reserved VTXOs,
marks one as still claimed by a live session, and asserts the other two are
released back to Live while the claimed one is untouched.
`TestReconcileSpendingReservationsNoOrphans` confirms a fully-claimed set
releases nothing, and `TestReconcileSpendingReservationsEmptyClaimedReleasesAll`
covers the no-live-sessions case. `TestOrphanedReservations` exercises the pure
set-difference helper.

## Test plan

- [x] `go build ./vtxo/... ./darepod/... ./oor/... ./wallet/... ./round/...`
- [x] `go vet ./vtxo/ ./darepod/`
- [x] `go test ./vtxo/ -count=1`
- [ ] operator escape-hatch CLI/RPC (blocked on proto regen, see above)

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -1965,7 +1965,98 @@ func (s *Server) startWalletDependentActors(ctx context.Context,
 		return err
 	}
 
+	// -------------------------------------------------------
+	// 15. Reconcile orphaned spending reservations.
+	// -------------------------------------------------------
+	// With the VTXO manager and OOR subsystem both recovered, release any
+	// VTXO left stranded in SpendingState by a spend that died without
+	// emitting SpendReleasedEvent (see issue #587). Non-fatal: a failure
+	// here just leaves those VTXOs reserved until the next restart.
+	if err := s.reconcileOrphanedSpendingReservations(
+		ctx, vtxoManagerRef,
+	); err != nil {
+		s.log.WarnS(ctx, "Failed to reconcile orphaned spending "+
+			"reservations", err)
+	}
+
 	s.log.InfoS(ctx, "Wallet-dependent actors started")
+
+	return nil
+}
+
+// reconcileOrphanedSpendingReservations releases VTXOs stranded in
+// SpendingState that no live OOR session still claims. A VTXO enters
+// SpendingState only via the OOR send path, which reserves the selected
+// VTXOs as inputs; if that spend dies without emitting SpendReleasedEvent,
+// the reservation outlives its owner and the VTXO is unspendable until batch
+// expiry forces an on-chain unilateral exit.
+//
+// We build the claimed set from the OOR actor's live (pending) outgoing
+// sessions and hand it to the VTXO manager, which releases any Spending VTXO
+// not in that set. If the OOR set can't be determined we abort rather than
+// release, since an incomplete claimed set would free VTXOs a live spend
+// legitimately holds. The OOR actor processes its restart message before this
+// query (the mailbox is FIFO and the restart is prepended), so its in-memory
+// session set is fully restored by the time we read it.
+//
+// Forfeit-state reservations (PendingForfeit/Forfeiting) from round
+// registration are a separate lifecycle and are intentionally out of scope:
+// only the OOR SpendingState path is reconciled here.
+func (s *Server) reconcileOrphanedSpendingReservations(ctx context.Context,
+	vtxoManagerRef actor.ActorRef[vtxo.ManagerMsg, vtxo.ManagerResp]) error {
+
+	if s.actorSystem == nil {
+		return nil
+	}
+
+	// Gather the input outpoints still held by live outgoing OOR sessions.
+	oorRef := oor.NewServiceKey().Ref(s.actorSystem)
+	listResult := oorRef.Ask(ctx, &oor.ListSessionsRequest{
+		Direction:   oor.SessionDirectionOutgoing,
+		PendingOnly: true,
+	}).Await(ctx)
+
+	actorResp, err := listResult.Unpack()
+	if err != nil {
+		return fmt.Errorf("query live OOR sessions: %w", err)
+	}
+
+	listResp, ok := actorResp.(*oor.ListSessionsResponse)
+	if !ok {
+		return fmt.Errorf("unexpected OOR list response type: %T",
+			actorResp)
+	}
+
+	var claimed []wire.OutPoint
+	for _, summary := range listResp.Sessions {
+		claimed = append(claimed, summary.InputOutpoints...)
+	}
+
+	// Hand the claimed set to the manager, which releases any Spending VTXO
+	// not present in it.
+	reconcileResult := vtxoManagerRef.Ask(ctx,
+		&vtxo.ReconcileSpendingReservationsRequest{
+			ClaimedInputs: claimed,
+		},
+	).Await(ctx)
+
+	resp, err := reconcileResult.Unpack()
+	if err != nil {
+		return fmt.Errorf("reconcile spending reservations: %w", err)
+	}
+
+	reconcileResp, ok := resp.(*vtxo.ReconcileSpendingReservationsResponse)
+	if !ok {
+		return fmt.Errorf("unexpected reconcile response type: %T", resp)
+	}
+
+	if reconcileResp.ReleasedCount > 0 {
+		s.log.InfoS(ctx, "Released orphaned spending reservations",
+			slog.Int("released", reconcileResp.ReleasedCount),
+			slog.Int("scanned", reconcileResp.ScannedCount),
+			slog.Int("claimed", reconcileResp.ClaimedCount),
+		)
+	}
 
 	return nil
 }
--- a/vtxo/manager.go
+++ b/vtxo/manager.go
@@ -274,6 +274,9 @@ func (m *Manager) Receive(ctx context.Context,
 	case *CompleteSpendRequest:
 		return m.handleCompleteSpend(ctx, req)
 
+	case *ReconcileSpendingReservationsRequest:
+		return m.handleReconcileSpendingReservations(ctx, req)
+
 	case *ReserveForfeitRequest:
 		return m.handleReserveForfeit(ctx, req)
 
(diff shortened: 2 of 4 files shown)


Roasbeef:
## Codex 5.5 automated review

VERDICT: COMMENT (draft — design sound, 3 observations for final pass)

This is well-reasoned startup reconciliation. The key design choices are all correct:
fail-close on OOR query error (abort rather than release with an incomplete claimed
set), fresh DB read inside the manager (avoids startup-snapshot divergence), non-fatal
outer wrapper (a failed reconcile leaves VTXOs stuck but doesn't crash the daemon), and
conservative skip when no actor is present for an outpoint. The OOR-before-reconcile
ordering argument (restart message prepended → FIFO → fully restored before `ListSessionsRequest`
arrives) is sound given the actor model, though it's worth a comment pointing at the
OOR actor's startup code for future readers.

Test coverage is good for the core paths. Three things worth addressing before marking
ready for review:

---

**Minor — store error path untested in `reconcile_test.go`**

`handleReconcileSpendingReservations` returns `fn.Err` if `ListVTXOsByStatus` fails, but
no test exercises that branch. The outer caller would log a warning and continue; that's
correct, but the behavior should be pinned. Add a test that mocks the store to return an
error and verifies the outer call returns that error and nothing is released.

---

**Minor — partial-success result lost on error**

When some VTXO actor releases succeed and others fail, `handleReconcileSpendingReservations`
returns `fn.Err` — so the caller never sees `ReleasedCount`. The server-level log entry
is gated on `err != nil` (WarnS) and never logs the partial count. A reader watching
logs can't tell whether 0 or N-1 reservations were cleared before the failure.

Consider either (a) returning the partial response alongside the error (pattern used
elsewhere in the manager), or (b) logging the partial `released` count before returning
the error in `reconcileOrphanedSpendingReservations`.

---

**Minor — "no actor for orphan" edge is silent and untested**

`vtxo/manager.go`, the `if !ok { continue }` path at the actor-map lookup:
if a VTXO is in `SpendingState` in the DB but has no live actor (actor map diverged
from DB state during startup), it is silently skipped, leaving that VTXO stuck reserved.
The comment says "already cleaned up to terminal state" but doesn't prove that invariant
holds after every crash scenario. A test (or at minimum an `InfoS` log line naming the
outpoint) would make this observable. Given that the operator-RPC escape hatch (#598)
handles the stuck case, this is acceptable in a draft, but name it explicitly.

---

No blocking issues. The wiring into `startWalletDependentActors` step 15 is live (not
dead code), which I mention since the PR is still marked draft — double-check the draft
flag is intentional pending the follow-up index PR (#597) rather than indicating the
reconcile path itself isn't ready.

Roasbeef:
Going with option B: https://github.com/lightninglabs/darepo-client/pull/597

PR #597, opened 2026-05-28
vtxo+db+oor: durable spending-reservation index for orphan recovery (#587 option B)

### LINKED ISSUE #587: vtxo: VTXO permanently stuck in Spending status when reserving spend dies without release (orphaned reservation, no resume reconciliation)

Roasbeef:
In this PR, we fix #587 with a durable spending-reservation index. This is the
"option B" alternative to #593, which solves the same bug by querying live OOR
sessions at startup; the two are mutually exclusive ways to recover the same
orphaned reservations, so we can pick one.

The bug: a VTXO reserved for an OOR send sits in `SpendingState` (DB status 7).
If the spend dies without releasing it (a crash, an abandoned vHTLC-funding
attempt), the reservation is orphaned. On restart the VTXO actor reloads into
`SpendingState` and nothing ever releases it, so the VTXO is unspendable until
batch expiry forces an on-chain unilateral exit. On signet this was the root of
the out-swap liquidity starvation: a 4.8M sat VTXO pinned in `Spending`.

## The index

We add a `spending_reservations` table: one row per reserved outpoint, recording
the owner (an outgoing OOR session) and a timestamp. The lifecycle is what makes
recovery deterministic:

- The row is written in `handleStartTransfer` right after `persistCheckpoint`, so
  it exists IFF the owning session was durably checkpointed. A crash between the
  checkpoint and the row write can only leave a row-less (orphaned) reservation,
  never a live session with a dangling row.
- The row is deleted when the VTXO leaves `SpendingState`, in the VTXO manager's
  `handleReleaseSpend` and `handleCompleteSpend`.
- At startup the manager sweeps: it lists Spending VTXOs and releases every one
  with no reservation row. A checkpointed session is always restored and resumed
  by the OOR actor on restart, so it's never an orphan; only a pre-checkpoint
  death leaves a row-less Spending VTXO, which is exactly what we release.

The decisive property over #593: the sweep needs no live OOR query and is
order-independent of OOR recovery. The rows are durable, so the manager can sweep
the moment it finishes recovering actors, in `Start`. It also generalizes: a
forfeit-path owner (round registration, the `PendingForfeit`/`Forfeiting`
states) could record its own `owner_kind` rows later and reuse the same sweep,
which the live-query approach can't do as cleanly.

Safety: if the sweep can't read the reservation list it releases nothing, since
acting on a partial view could free a VTXO an in-flight spend still owns. The
recording side is best-effort and never fails a transfer; the sweep is the
backstop.

## Layout

- `db/sqlc/migrations/000017_spending_reservations.*` plus
  `db/sqlc/queries/spending_reservations.sql`, regenerated through sqlc (pinned
  1.29.0; the regen diff is limited to the new table and queries).
- `db/spending_reservation_store.go`: the persistence store, following the same
  transaction-executor pattern as the VTXO store.
- `vtxo`: a narrow `SpendingReservationStore` interface (so the package doesn't
  import db or oor), the delete hooks, and the startup sweep.
- `oor`: a minimal `ReservationStore` interface and the post-checkpoint write.
- `darepod`: constructs the store and injects it into both the OOR actor and the
  VTXO manager.

The interfaces live in `vtxo` and `oor` rather than a shared package, so neither
imports the db type and we avoid a `db -> oor` cycle. The structural match is
pinned with a compile-time assertion at the wiring site.

## Tests

- `vtxo/manager_reservation_test.go`: the sweep releases only row-less orphans
  and leaves reserved VTXOs in `SpendingState`; it aborts without releasing
  anything when the reservation list can't be read; a nil store disables it; and
  release/complete delete the reservation row.
- `db/spending_reservation_store_test.go`: store CRUD against a real SQLite DB
  (upsert, idempotent re-upsert, list, delete, no-op delete), which also
  exercises the new migration.

## Test plan

- [x] `go build ./vtxo/... ./darepod/... ./oor/... ./db/...`
- [x] `go vet ./vtxo/ ./oor/ ./db/ ./darepod/`
- [x] `go test ./vtxo/ ./db/ -count=1`
- [x] sqlc regenerated with the pinned 1.29.0; diff limited to the new table/queries


Roasbeef:
## Codex 5.5 automated review

VERDICT: REQUEST_CHANGES

The durable index is the right shape for orphan recovery, but the current
write ordering does not actually make the row authoritative. A checkpointed
OOR session can be persisted without a reservation row, and the VTXO manager
will then release its inputs on the next startup before the OOR actor is even
registered.

Blocking

- `oor/actor.go:756`: `persistCheckpoint` commits the outgoing session, but
  `recordReservations` is called afterwards and explicitly ignores all write
  failures at `oor/actor.go:792`. That violates the "row exists iff
  checkpointed" invariant the sweep relies on. If the process crashes between
  lines 756 and 767, or the upsert fails because the request context is
  canceled after the checkpoint commit, the DB contains a live checkpointed OOR
  session with no reservation row. On restart, `vtxo/manager.go:250` runs the
  sweep during VTXO manager startup, and `darepod/server.go:1964` starts the
  OOR actor only later, so `vtxo/manager.go:794` sees no row and
  `vtxo/manager.go:828` releases an input that the recovered OOR session still
  owns. Please make the checkpoint and reservation rows atomic in one durable
  transaction, or make the sweep consult a source that includes checkpointed OOR
  sessions before releasing. At minimum, reservation write failure after a
  checkpoint cannot be best-effort while missing rows are treated as orphan
  proof.

Major

- `vtxo/manager.go:885` and `vtxo/manager.go:939`: reservation deletion is
  also best-effort, but a stale row is not harmless. If delete fails after a
  release, the VTXO returns to LiveState with an old reservation row still
  present; if a later outgoing send reserves the same outpoint and dies before
  checkpointing, the startup sweep will see the stale row and refuse to release
  the true orphan. Please either delete the row in the same durable transition
  that moves the VTXO out of SpendingState, or otherwise make stale rows
  distinguishable by owner/session liveness so they cannot mask future orphans.

Minor

- No additional minor findings.

Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
Pushed fixes for every finding from the automated passes (gemini, codex, and both Claude reviews). Force-pushed onto `daedalus/587-option-b-reservation-index` — still three commits, with the fixes folded into the originals and the commit bodies updated to match the new behavior.

**Critical — checkpoint/reservation atomicity.** One clarification on the mechanism: the upsert is *not* a second SQLite writer racing the actor tx. The durable actor wraps each message in `TxAwareDeliveryStore.ExecTx`, which attaches the tx to the context via `actor.WithTx` (`db/actordelivery/store_impl.go:1400`), and `TransactionExecutor.ExecTx` joins that tx via `actor.TxFromContext` (`db/interfaces.go:247`), ignoring its own `WriteTxOption` — exactly how the ledger store joins. So `UpsertReservation` already runs in the same transaction as `SaveCheckpoint`. The only real bug was swallowing the write error. `recordReservations` now returns it and `handleStartTransfer` propagates, so the rows and the checkpoint commit or roll back together and the row-exists-IFF-checkpointed invariant holds by construction.

**Major — stale rows masking orphans.** Deletion moved out of the manager's best-effort call into the VTXO actor's FSM transition: a `ReleaseSpendReservation` flag rides the status update and the store deletes the row in the same transaction as the status change (`UpdateVTXOStatusReleasingReservation`), on all three SpendingState exits (complete / release / critical-expiry). The "stale row is harmless" comment is gone.

**Major — upgrade backfill.** Migration 17 now backfills a reservation row for every VTXO already in `SpendingState`, so a node upgrading mid-send doesn't have its in-flight inputs released by the first post-upgrade sweep.

**Minor — hash decode + CHECK constraints.** Reads now decode via `chainhash.NewHash` (errors on a short/corrupt blob instead of zero-padding); `outpoint_hash` and `owner_id` carry `CHECK (length(...) = 32)`.

**Tests.** Added the checkpoint-succeeds / reservation-write-fails coverage (`recordReservations` error propagation), a db-level atomic status-change-plus-row-delete test, and tightened the SpendingState transition tests to assert the release flag.

Each of the three commits compiles on its own (verified per-commit during the rebase) and the `vtxo` / `oor` / `db` / `unroll` / `darepod` unit suites pass.

Roasbeef:
@claude review this

PR #598, opened 2026-05-28
daemonrpc+darepod: ReleaseSpendingVTXO operator RPC + CLI

Roasbeef:
In this PR, we add an operator escape-hatch to force-release a VTXO stuck in
`SpendingState` back to Live. It's the manual companion to the automatic #587
recovery (#593 / #597): when a spend reservation is orphaned (the owning OOR send
died without releasing it), the VTXO is unspendable until batch expiry, and an
operator can recover it on demand.

## The RPC

`ReleaseSpendingVTXO(outpoint)` on the daemon service. The handler routes through
the VTXO manager: it parses the outpoint and, to keep the hatch from being
misused on a healthy VTXO, refuses the call unless the VTXO is currently in
`SpendingState`. It then asks the manager to release it through the same
`SpendReleasedEvent` path a normal release uses, and reports whether the release
happened.

## CLI

`darepocli ark vtxos release-spending <outpoint>` gives operators a direct verb.
Because the RPC also lands in the regenerated dev registry, it's reachable
through the generated `dev` command too (alias `release-spending-vtxo`), but the
dedicated verb is the ergonomic path.

## Notes

The proto stubs were regenerated with the repo's pinned toolchain (protoc
3.21.12 plus the go.mod-pinned plugins); the diff is limited to the daemonrpc
generated files and the dev RPC registry. The hand-rolled REST client gains the
matching method to stay in step with the `DaemonServiceClient` interface it
implements.

## Test plan

- [x] `go build ./daemonrpc/... ./darepod/... ./cmd/darepocli/... ./rpc/restclient/...`
- [x] `go vet ./daemonrpc/ ./darepod/ ./cmd/darepocli/darepoclicommands/...`
- [x] `go test ./daemonrpc/ ./darepod/ ./cmd/darepocli/darepoclicommands/... -count=1`
- [x] proto regen scoped to daemonrpc + dev registry (pinned protoc 3.21.12)

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_vtxos.go
+++ b/cmd/darepocli/darepoclicommands/cmd_vtxos.go
@@ -24,9 +24,55 @@ func newVTXOsCmd() *cobra.Command {
 
 	cmd.AddCommand(
 		newVTXOsListCmd(), newVTXOsRefreshCmd(), newVTXOsLeaveCmd(),
+		newVTXOsReleaseSpendingCmd(),
 	)
 
 	return cmd
+}
+
+// newVTXOsReleaseSpendingCmd creates the vtxos release-spending
+// subcommand. It is an operator escape-hatch that force-releases a VTXO
+// stuck in spending state (an orphaned spend reservation) back to Live;
+// the daemon rejects the call if the VTXO is not actually spending.
+func newVTXOsReleaseSpendingCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "release-spending [outpoint]",
+		Short: "Force-release a VTXO stuck in spending state",
+		Long: "Operator escape-hatch for a VTXO whose spend " +
+			"reservation was orphaned (the owning spend died " +
+			"without releasing it). Releases the outpoint " +
+			"(txid:index) back to Live. The daemon rejects the " +
+			"call unless the VTXO is currently in spending " +
+			"state.\n\n" +
+			"Example:\n" +
+			"  darepocli ark vtxos release-spending TXID:VOUT",
+		Args: cobra.ExactArgs(1),
+		RunE: vtxosReleaseSpending,
+	}
+}
+
+// vtxosReleaseSpending executes the ReleaseSpendingVTXO RPC.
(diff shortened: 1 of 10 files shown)


Roasbeef:
## Codex 5.5 automated review
VERDICT: REQUEST_CHANGES

The new RPC takes the right high-level path by routing the release through the VTXO manager and actor transition instead of mutating VTXO state directly. However, the safety precondition currently fails open if the durable VTXO lookup is unavailable or returns an error, which is not acceptable for an operator escape hatch that can make funds spendable again.

**Blocking**

- `darepod/rpc_server.go:3184`: `GetVTXO` errors are ignored unless they are `nil`, and a nil `vtxoStore` also skips the guard entirely. `VTXOPersistenceStore.GetVTXO` returns an error for both "not found" and real storage failures, so a typo or database/read failure falls through to `ReleaseSpendRequest`; if the runtime actor exists and is in `SpendingState`, this releases the reservation without the durable-state check that the PR description relies on. Please fail closed here: require `vtxoStore` to be initialized, return a clear `FailedPrecondition`/`NotFound` for missing or non-spending VTXOs, and return `Internal`/`Unavailable` for lookup errors before sending anything to the manager. Add handler tests for lookup error, missing VTXO, non-spending VTXO, and happy path so this cannot regress.

**Major**

- `darepod/rpc_server.go:3184`: the current not-found handling branch is effectively unreachable because `GetVTXO` documents and implements "returns error if not found"; `desc == nil && err == nil` is not how a missing row is represented. This means malformed-but-valid outpoints are reported through the manager as an internal `release spend: ... no actor for outpoint ...` error rather than the advertised clean rejection before touching the actor boundary. Once the lookup is made fail-closed, split not-found from other store errors and return a caller-facing precondition/not-found status before calling `vtxoMgrRef.Ask`.

**Minor**

- No dedicated tests were added for the new RPC/CLI behavior. Given this is a money-affecting manual recovery path, please cover the server rejection paths above plus CLI arg validation/output plumbing.

litbot-9000:

Roasbeef:
Bug fixed, no longer needed.

Issue #538, opened 2026-05-27
recv: paying `recv --offchain --amt 5000` results in `script not registered for principal`


hieblmi:
Note: I use `lnd` as wallet
```
wallet.type=lnd                                                                
lnd.host=127.0.0.1:11009                                                       
lnd.tlspath=/home/bob/.lnd-signet/tls.cert                                     
lnd.macaroonpath=/home/bob/.lnd-signet/data/chain/bitcoin/signet/admin.macaroon
wallet.pollinterval=30s                                                        
```

My ark client seems borked after i did
```
da recv --offchain --amt 5000
```
Then paying the invoice
```
slncli payinvoice lntbs50u1p4pd0ek
```
`da activity` shows:
```
2026-05-27 09:54:49.000  RECV     PENDING   5000    0    settling   d3a6f959e11622a5a5f7f604b795ac72ad0ce78190070671cfe89b7d318b74c8 
```
Results in a error loop at the end of the log.
A restart doesn't fix it.

```
2026-05-27 09:49:10.907 [INF] DRPD: Starting darepod version=0.0.1-alpha commit=b8655c8 network=signet wallet_type=lnd
2026-05-27 09:49:10.907 [INF] DRPD: Connecting to lnd host=127.0.0.1:11009
2026-05-27 09:49:10.983 [INF] DRPD: Connected to lnd alias=haggis pubkey=0349408eebe95246fb4e10835a6ed8b4a161ac8726b122014b84cccc0a754f6c32
2026-05-27 09:49:10.983 [INF] DRPD: Actor system initialized
2026-05-27 09:49:10.983 [DBG] ACTR: Actor registered with system actor_id=timeout service_key=timeout
2026-05-27 09:49:10.983 [INF] CBKD: Starting LND chain backend
2026-05-27 09:49:10.983 [INF] CBKD: LND chain backend started successfully
2026-05-27 09:49:10.983 [INF] DRPD: Chain backend started type=lnd
2026-05-27 09:49:10.983 [DBG] ACTR: Actor registered with system actor_id=chain-source service_key=chainsource
2026-05-27 09:49:10.983 [INF] DRPD: Chain source actor registered
2026-05-27 09:49:10.983 [INF] DABS: Opening SQLite database db_file=/home/bob/.darepod/data/signet/arkd.db max_conns=25 conn_max_lifetime=10m0s
2026-05-27 09:49:10.983 [DBG] DABS: SQLite connection pool configured
[...]

Issue #67, opened 2026-01-08
db: evaluate alternatives to TLV encoding for complex type serialization


Roasbeef:
The round persistence layer uses custom TLV encoding for serializing complex types (`tree.Tree`, `proof.TxProof`) to database blobs. Consider whether an alternative encoding format would provide benefits.

## Current Approach
- Custom TLV encoding using lnd's `tlv` package
- Implemented in `db/tree_codec.go`
- Property-based tests verify round-trip correctness

### Protocol Buffers (protobuf)
- **Pros**: Mature tooling, schema evolution, compact, fast, already used in project for RPC
- **Cons**: Requires `.proto` files and code generation, adds build complexity

### MessagePack (msgpack)
- **Pros**: Schemaless, JSON-like but binary, simple API
- **Cons**: Less type safety, schema evolution less explicit

### CBOR (RFC 8949)
- **Pros**: IETF standardized, similar to msgpack, good tooling
- **Cons**: Less common in Go ecosystem

### FlatBuffers / Cap'n Proto
- **Pros**: Zero-copy access, very fast reads
- **Cons**: More complex, overkill for blob storage

### Keep TLV
- **Pros**: Consistent with lnd patterns, no new dependencies, already working
- **Cons**: Custom code to maintain, less discoverable format

## Evaluation Criteria
1. Serialization/deserialization performance
2. Encoded size
3. Schema evolution story
4. Code maintenance burden
5. Debuggability (can we inspect stored blobs?)
6. Consistency with rest of codebase

Roasbeef:
Closing — the design question is effectively settled. TLV is the committed encoding: it is implemented (db/tree_codec.go), hardened against untrusted blobs (commit a9a32fb7 added depth/child-count recursion bounds), consolidated into lib/types (commit 17f21f3e), and used across ~48 files. This open-ended "evaluate alternatives" ticket is moot; if a migration to another format is ever desired it should be a fresh, concrete proposal.

Issue #59, opened 2026-01-08
round: support fan-in boarding (multiple inputs → single VTXO)


Roasbeef:
## Summary

The current client round FSM assumes a 1:1 mapping between boarding inputs and VTXOs. We should support "fan-in" where multiple boarding UTXOs can be combined to create a single, larger VTXO.

## Motivation

From PR #34 review discussion with @ellemouton:

> in my model, a boarding request is not really linked to the requested vtxos. instead, the user submits a JoinRequest as a whole which contains a set of: inputs (boarding requests and forfeits) and outputs (leave requests and vtxos). This makes things very flexible in that you can combine any number of boarding utxos with a set of vtxo descriptors (could have one boarding input funding multiple vtxos or multiple boarding inputs funding a single vtxo for example).

Use cases:
- User has multiple small confirmed boarding UTXOs and wants to consolidate into one VTXO
- Combining a boarding UTXO with a forfeit to create a larger single VTXO
- More flexible fund management without on-chain consolidation

## Current Behavior

`BoardingIntent` currently couples a single boarding input to its VTXO expectations:

```go
type BoardingIntent struct {
    Address     BoardingAddress
    Outpoint    wire.OutPoint
    ChainInfo   BoardingChainInfo
    VTXOs       []VTXOExpectation  // Currently assumes these map 1:1
}
```

## Proposed Changes

1. Decouple boarding inputs from VTXO outputs in the `JoinRoundRequest`
2. Allow specifying arbitrary input/output combinations in a single request
3. Update FSM state tracking to handle N inputs → M outputs mapping
4. Ensure validation correctly verifies total input value ≥ total output value + fees

## Related

- PR #34 review thread: https://github.com/lightninglabs/darepo-client/pull/34#discussion_r2584544839
- This flexibility also enables future "swap requests" (forfeits → VTXOs) without FSM changes

ellemouton:
addressed in #84

Issue #55, opened 2026-01-06
chainsource: BlockEpochActor iterator can panic on closed channel


bhandras:
### Summary
`chainsource.BlockEpochActor` can panic with `send on closed channel` in iterator mode, because the receiver-side iterator/cancel path closes `epochChan` while `monitorBlocks` may still be sending.

This appears to be a pre-existing issue (not introduced by PR #50), but it is made more visible while reviewing the new systest framework changes.

### Details
In iterator mode, the iterator cleanup and `cancelFunc` both call `close(a.epochChan)` (guarded by `closeOnce`). Meanwhile, `monitorBlocks` sends into the same channel:

- `BlockEpochActor.monitorBlocks`: `case a.epochChan <- blockEpoch:`
- iterator cleanup / cancel: `close(a.epochChan)`

Even with a `select` on `<-a.ctx.Done()`, sending to a closed channel can still race and panic.

### Impact
- Potential intermittent panics in iterator mode under cancellation/cleanup timing.
- Violates the usual Go ownership rule: the sender should close the data channel.

### Suggested Fix
- Do not close `epochChan` from the iterator/cancel path.
- Use context cancellation to signal shutdown.
- Have the sender (`monitorBlocks`) be the only goroutine that closes `epochChan` (e.g., `defer close(epochChan)` on exit).

### Notes
While reviewing PR #50, I also noticed multiple actors (Conf/Spend/BlockEpoch) create internal contexts from `context.Background()` and rely on `Stop()` methods, but the actor system uses `actor.Stoppable.OnStop(ctx)` for cleanup. If those behaviors are stopped via the actor system without `OnStop` implemented, their goroutines may outlive the actor.

PR #623, opened 2026-05-29
docs+swaps: swap system guide and offchain receive race regression tests

Roasbeef:
## Summary

Two related, low-risk additions that came out of digging into #538:

1. **`docs/swap_system.md`** — an end-to-end guide to the Lightning↔Ark swap
   system: the vHTLC taproot tree, the receive (out-swap) and pay (in-swap)
   flows, same-Ark p2p settlement detection, the swap-server RPC surface, and
   how the proof-gated indexer authorizes script-scope queries. Mermaid sequence
   and flow diagrams for both directions. Linked from `docs/index.md`. Written
   to document how the system works today, not tied to any one incident.

2. **Regression tests for the offchain receive funding race** in
   `sdk/swaps/out_swap_test.go`. On a receive the swap server funds the vHTLC,
   so the receiver has no local copy and depends on the operator's indexer.
   Between accepting the HTLC event and the indexer materializing a queryable
   row, the proof-gated poll is rejected with `script not registered for
   principal`. PR #550 already made that non-terminal; these tests pin the
   behavior so it cannot regress:
   - `TestWaitForVHTLCRetriesThroughIndexerRowLag` — retries through the
     pre-funded window and completes once the row lands.
   - `TestWaitForVHTLCExpiresOnPersistentUnregisteredScript` — a persistent
     rejection still terminates via the refund-locktime guard (`ErrSwapExpired`),
     never looping forever or reporting false funding.
   - `TestReceiveSessionWaitForFundingSurvivesIndexerRowLag` — the end-to-end
     receive session reaches `VHTLCFunded` despite an initial burst of
     rejections.

   Adds a sequenced `FindLiveVTXOByPkScript` hook to the test daemon.

## Context for #538

The reported loop has two distinct causes that surface the identical indexer
error: (a) the transient pre-funded window (a race; fixed by the non-terminal
poll in #550 and exercised here), and (b) a `lib/arkscript` client/server skew
that derives a different pkScript so the row never appears (a deployment issue;
the persistent-failure test confirms it expires cleanly rather than hanging).
Follow-ups tracked separately: server-side `vhtlc_outpoint` in
`OutSwapHtlcEvent` to detect skew immediately, and clearer indexer error
semantics in arkd.

## Validation

- `go test ./sdk/swaps/` (full package)
- `make fmt-changed`
- `make lint-changed-local` — 0 issues

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,6 +15,7 @@ into specific topics below.
 | [fee_ledger.md](fee_ledger.md) | Client-side double-entry fee ledger: chart of accounts, per-flow walkthroughs, emission sites, replay safety |
 | [fee-change-model.md](fee-change-model.md) | Seal-time fee handshake (#270): change-output designation rules, 11-scenario catalogue, proto contract, CLI mapping |
 | [sdk_layered_architecture.md](sdk_layered_architecture.md) | SDK layering rationale: `sdk/ark` facade, remote vs. embedded modes, `sdk/swaps` future direction |
+| [swap_system.md](swap_system.md) | End-to-end swap walkthrough: vHTLC tree, receive (out-swap) and pay (in-swap) flows, same-Ark p2p detection, swap-server RPCs, and proof-gated indexer authorization, with mermaid diagrams |
 | [walletdk_integration.md](walletdk_integration.md) | Basic `walletdk` integration flow, startup/config examples, swap accounting, and wrapper guidance |
 | [walletdkrpc_build.md](walletdkrpc_build.md) | How to build and install the daemon and CLI with the wallet RPC subserver enabled (`walletdkrpc` + `swapruntime` tags) |
 | [swap_background_execution.md](swap_background_execution.md) | Optional build-tagged daemon executor for background swap progress and daemon-backed CLI control |
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,7 +15,7 @@ into specific topics below.
 | [fee_ledger.md](fee_ledger.md) | Client-side double-entry fee ledger: chart of accounts, per-flow walkthroughs, emission sites, replay safety |
 | [fee-change-model.md](fee-change-model.md) | Seal-time fee handshake (#270): change-output designation rules, 11-scenario catalogue, proto contract, CLI mapping |
 | [sdk_layered_architecture.md](sdk_layered_architecture.md) | SDK layering rationale: `sdk/ark` facade, remote vs. embedded modes, `sdk/swaps` future direction |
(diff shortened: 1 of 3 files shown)


Roasbeef:
## Follow-up: documented the off-chain cancellation / recovery story

Expanded `docs/swap_system.md` to cover how a swap is cancelled or timed
out **off-chain** without going to the base layer — the part the original
draft listed (the six leaves) but never explained. No code changes; docs only.

### What was added / corrected

**1. The collaborative-vs-exit split (§1).** Fixed the leaf list to match
`lib/arkscript/vhtlc.go` (it used `Checksig` where the policy compiles
`Multisig`) and explained the spine of the whole cancellation story: the six
leaves are three pairs (claim / refund / refund-without-receiver), each with a
*collaborative* member (carries the operator key → settles off-chain via OOR)
and a *unilateral exit* member (operator-free, CSV-gated → on-chain). Cited the
`ValidatePolicy` invariants: the operator can always *help* settle but can never
*seize* a vHTLC alone.

**2. Fixed the on-chain/off-chain framing (§4.4, §9 table).** The CLTV refund
locktime gates an **off-chain OOR spend** of the RefundWithoutReceiver leaf — it
is *not* itself an on-chain event. The old text implied the timeout refund meant
going on-chain.

**3. New §5 — the recovery ladder.** The "additional exit paths" turned out not
to be new tapscript leaves (still exactly six, here and in swapdk-server) but an
operational layer in `sdk/swaps/recovery.go`:
- **Arm before you risk** — a dormant recovery row is stored the instant a vHTLC
  is funded, before committing to the risky half of the flow.
- **Receive side** — claim leaf 1 cooperatively before the refund locktime (the
  server can reclaim via leaf 3 after it); escalate to unilateral claim (leaf 4)
  only under deadline pressure.
- **Pay side** — the three-rung refund ladder: immediate cooperative Refund
  (leaf 2, `AuthorizeInSwapRefund`) → CLTV-gated cooperative RefundWithoutReceiver
  (leaf 3, still off-chain) → unilateral exit (leaf 6) only if cooperation breaks.
- **Escalation policy** — `AutoEscalate` off by default; otherwise gated by a
  block-margin deadline check and a one-hour grace period, with a frozen fee cap.
- **State machine + terminal mapping** — ARMED → UNROLL_STARTED →
  COMPLETED/FAILED/CANCELLED, and how each outcome lands the session in
  Refunded / Completed / NeedsIntervention / Failed.
- **Server mirror** — swapdk-server arms its own symmetric recovery (#43), the
  conditions under which it co-signs a cooperative refund (#55), and fast-fail
  paths for AMP (#67) and out-swap liquidity exhaustion (#59).

**4. In-Ark (§6) + housekeeping.** Noted that same-Ark swaps reuse the identical
six-leaf vHTLC, so the ladder applies unchanged; renumbered the trailing sections
and fixed all shifted cross-references; updated `docs/index.md`.

Written in the same classic-style register as the original pass.

PR #631, opened 2026-05-29
swapwallet: show unboarded deposits as boarding

### LINKED ISSUE #629: UX: Show confirmed but unboarded deposits as "boarding"
Currently, once you deposit funds, they show up as `PENDING` while they are still unconfirmed, and as `COMPLETE` once they are confirmed.

In my case, the funds are not being swept, but they show up as `COMPLETE`.

I believe there should be something in between, e.g. `BOARDING` to indicate progress.

bhandras:
## Summary

- report chain-confirmed wallet UTXO deposits as `boarding` until the matching boarding input is part of a confirmed round
- map that intermediate ledger status to a pending wallet activity entry with `phase_label="boarding"`
- keep legacy wallet UTXO rows without full outpoint metadata on the old confirmed path
- keep full-metadata `wallet_utxo_created` rows without a matching boarding intent confirmed, so sweep returns and future non-deposit wallet UTXOs do not remain pending forever

Fixes #629.

## Testing

- `go test ./db -run 'TestLedgerStoreTransactionHistoryWalletUTXOCreated'`
- `go test -tags="walletdkrpc swapruntime nolog" ./swapwallet -run 'TestHistoryDeposit|TestHistoryPendingFilterIncludesPendingBoarding|TestHistorySurfacesPendingBoardingBalance'`
- `go test ./db`
- `go test -tags="walletdkrpc swapruntime nolog" ./swapwallet`
- `make lint-changed-local`
- `make fmt-changed-check`

### DIFF
--- a/db/ledger_store_test.go
+++ b/db/ledger_store_test.go
@@ -546,6 +546,146 @@ func TestLedgerStoreTransactionHistoryWalletUTXOCreatedChainFields(
 	require.Equal(t, depositAmount-vtxoAmount, row.FeeSat)
 }
 
+// TestLedgerStoreTransactionHistoryWalletUTXOCreatedBoardingStatus confirms
+// a chain-confirmed deposit is not reported as complete until its boarding
+// outpoint has moved through a confirmed round and produced a spendable VTXO.
+func TestLedgerStoreTransactionHistoryWalletUTXOCreatedBoardingStatus(
+	t *testing.T) {
+
+	t.Parallel()
+
+	ctx := t.Context()
+	store, db := newLedgerStoreAndDBForTest(t)
+
+	outpointHash := testHash32(0x52)
+	const (
+		outpointIndex      = uint32(1)
+		confirmationHeight = int32(304_091)
+		depositAmount      = int64(75_000)
+		createdAt          = int64(1_700_000_601)
+	)
+	pkScript := testBytes(34, 0x09)
+
+	require.NoError(
+		t,
+		store.InsertLedgerEntry(
+			ctx, ledger.LedgerEntry{
+				DebitAccount:   ledger.AccountWalletBalance,
+				CreditAccount:  ledger.AccountOpeningBalance,
+				AmountSat:      depositAmount,
+				EventType:      ledger.EventWalletUTXOCreated,
+				Description:    "boarding deposit",
+				CreatedAt:      createdAt,
+				IdempotencyKey: testBytes(36, 0x61),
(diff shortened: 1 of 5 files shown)


bhandras:
Addressed the non-deposit `wallet_utxo_created` case.

The history SQL now only falls through to the intermediate `boarding` status when the row has full chain metadata and a matching `boarding_intents` row. Full-metadata wallet UTXO rows without a matching boarding intent, such as boarding sweep returns, remain `confirmed` so they do not show as pending forever.

I also added a regression test that inserts a chain-bound `wallet_utxo_created` row without a boarding intent and asserts it stays confirmed.

bhandras:
claude: Re-reviewed the amended commit (`6a32af55`). The blocking concern from the previous pass is fully addressed.

## What changed

**Fix applied** — the status `CASE` now gates `'boarding'` on the existence of a boarding intent:

```sql
WHEN le.event_type = 'wallet_utxo_created'
     AND NOT EXISTS (
         SELECT 1 FROM boarding_intents AS status_bi
         WHERE status_bi.outpoint_hash = le.chain_txid
           AND status_bi.outpoint_index = le.chain_vout
     )
THEN 'confirmed'
```

Branch order is now: confirmed-round → `confirmed`; legacy NULL-outpoint → `confirmed`; **no boarding intent → `confirmed`**; else → `boarding`. This keeps boarding-sweep-returns / change / future non-deposit `wallet_utxo_created` rows at `COMPLETE` instead of stranding them at `PENDING` forever, while genuine deposits (which always have a `boarding_intents` row) still read `boarding` until their round confirms.

**Test gap closed** — `TestLedgerStoreTransactionHistoryWalletUTXOCreatedNonBoardingStatus` inserts a `wallet_utxo_created` row with full chain metadata and no boarding intent, asserting it stays `confirmed` — exactly the case the earlier logic got wrong.

## Verification

- `.sql` ↔ generated `.sql.go` status block are byte-identical (in sync).
- The `NOT EXISTS` correlated subquery is cross-DB safe (Postgres + SQLite) and short-circuits correctly on NULL outpoints.
- CI is fully green: unit (`unit-cover`), system tests on both postgres and sqlite, lint, static checks, and all cross-compile targets.

## Minor note (non-blocking)

`JOIN rounds AS r ON r.round_id = rbi.round_id` is an INNER join, relying on every `round_boarding_intents.round_id` having a `rounds` row. That's the FK invariant, so fine — just flagging it.

LGTM.

bhandras:
Tested this manually as well.

PR #639, opened 2026-05-29
darepod: add opt-in pprof debug HTTP server

### LINKED ISSUE #635: Add pprof flags for darepod
Add disabled-by-default pprof config/flags to darepod, on a dedicated HTTP server/mux. Suggested options: `pprof.listen`, optionally `pprof.blockprofilerate` and `pprof.mutexprofilefraction`. Empty listen address disables it. When enabled, expose standard endpoints like `/debug/pprof/goroutine?debug=2`, `/debug/pprof/profile`, `/debug/pprof/heap`, `/debug/pprof/block`, `/debug/pprof/mutex`, and `/debug/pprof/trace`.

bhandras:
## Summary

Adds disabled-by-default pprof support to `darepod` (issue #635).

- New `pprof` config group, served on a **dedicated, private `http.ServeMux`** so the `net/http/pprof` endpoints are never attached to the gRPC or gateway listeners.
- `pprof.listen` — listen address; **empty (default) disables pprof entirely**.
- `pprof.blockprofilerate` — `runtime.SetBlockProfileRate` value, applied at startup when `> 0` (default `0`).
- `pprof.mutexprofilefraction` — `runtime.SetMutexProfileFraction` value, applied at startup when `> 0` (default `0`).
- Started in the daemon lifecycle only when a listen address is set, with a bounded graceful shutdown mirroring the other listeners. `http.ErrServerClosed` is treated as a clean shutdown. `WriteTimeout` is unset so streaming CPU/trace profiles (`?seconds=N`) are not cut off.

### Endpoints (when enabled)
`/debug/pprof/` (index → goroutine, heap, block, mutex, allocs, threadcreate), `/debug/pprof/cmdline`, `/debug/pprof/profile`, `/debug/pprof/symbol`, `/debug/pprof/trace`.

### Security
pprof exposes sensitive runtime/debug data and is opt-in. Operators must bind `pprof.listen` to a loopback (e.g. `127.0.0.1:6060`) or otherwise firewalled address. Documented in `sample-darepod.conf` and config GoDoc.

## Testing
- `make build`
- `go test ./darepod/ -run TestNewPprofMux` (new unit tests: endpoint registration + private-mux confinement)
- `make fmt-changed`
- sample-conf check (`check-sample-darepod-conf`) — passes
- `make commitmsg-lint` — passes
- `darepod --help` lists the three `--pprof.*` flags

Verify manually:
```
darepod --pprof.listen=127.0.0.1:6060
curl http://127.0.0.1:6060/debug/pprof/
curl -o cpu.prof 'http://127.0.0.1:6060/debug/pprof/profile?seconds=30'
```

Draft pending full CI (unit/itest/systest).

Closes #635.

### DIFF
--- a/cmd/darepod/main.go
+++ b/cmd/darepod/main.go
@@ -194,6 +194,8 @@ func newRootCmd() *cobra.Command {
 
 	registerSwapRuntimeFlags(f, cfg)
 
+	registerPprofFlags(f, cfg)
+
 	// Safety flag for mainnet operation.
 	f.Bool(
 		"allow-mainnet", cfg.AllowMainnet, "allow the daemon to "+
@@ -332,6 +334,32 @@ func registerSwapRuntimeFlags(f *pflag.FlagSet, cfg *darepod.Config) {
 		cfg.Swap.VHTLCRecovery.MaxFeeRateSatPerKW, "maximum fee "+
 			"rate in sat/kw for swapruntime vHTLC recovery "+
 			"exit spends",
+	)
+}
+
+// registerPprofFlags registers the optional pprof debug-server flags. pprof is
+// disabled by default and only starts when pprof.listen is set to a non-empty
+// address. The flag names match the mapstructure config keys, so viper binds
+// them without explicit aliases.
+func registerPprofFlags(f *pflag.FlagSet, cfg *darepod.Config) {
+	f.String(
+		"pprof.listen", cfg.Pprof.ListenAddr, "address for the "+
+			"pprof debug HTTP server (e.g. 127.0.0.1:6060); "+
+			"empty disables pprof. Exposes sensitive runtime "+
+			"data, so bind to a loopback or firewalled address "+
+			"only",
+	)
+	f.Int(
+		"pprof.blockprofilerate", cfg.Pprof.BlockProfileRate, "value"+
+			" passed to runtime.SetBlockProfileRate at startup "+
+			"when greater than zero; zero leaves block "+
+			"profiling off",
+	)
+	f.Int(
+		"pprof.mutexprofilefraction", cfg.Pprof.MutexProfileFraction,
+		"value passed to runtime.SetMutexProfileFraction at "+
+			"startup when greater than zero; zero leaves mutex "+
+			"profiling off",
 	)
 }
 
--- a/darepod/config.go
+++ b/darepod/config.go
@@ -258,6 +258,12 @@ type Config struct {
 	// build tag (both the standalone cmd/darepod binary and the
 	// sdk/walletdk embedded path).
 	EagerRoundJoin bool `mapstructure:"eagerroundjoin"`
+
+	// Pprof configures the optional net/http/pprof debug server. It is
+	// disabled by default and must be explicitly opted into via a
+	// non-empty listen address. A value type so a zero-value Config can
+	// never carry a nil pprof config into the start path.
+	Pprof PprofConfig `mapstructure:"pprof"`
 }
 
 // RPCServiceRegistrar registers one optional daemon gRPC subserver on the
@@ -602,6 +608,33 @@ func DefaultGatewayConfig() *GatewayConfig {
 		Enabled:    true,
 		ListenAddr: DefaultRPCGatewayHost,
 	}
+}
+
+// PprofConfig configures the optional net/http/pprof debug server.
+//
+// SECURITY: pprof exposes sensitive runtime and debug data — goroutine
+// stacks, heap and CPU profiles, the command line, and the symbol table —
+// any of which can leak internal state or enable denial-of-service. It is
+// opt-in (disabled by default) and is served on its own private HTTP mux
+// that is never attached to the daemon's gRPC/gateway listeners. Operators
+// who enable it should bind ListenAddr to a loopback or firewalled address
+// (e.g. "127.0.0.1:6060") and never expose it to untrusted networks.
+type PprofConfig struct {
+	// ListenAddr is the network address the pprof HTTP server binds to.
+	// An empty value (the default) disables pprof entirely. A non-empty
+	// value such as "127.0.0.1:6060" starts a dedicated HTTP server
+	// exposing the standard net/http/pprof endpoints.
+	ListenAddr string `mapstructure:"listen"`
+
+	// BlockProfileRate, when greater than zero, is passed to
+	// runtime.SetBlockProfileRate at startup to enable blocking-profile
+	// collection. Zero (the default) leaves block profiling disabled.
+	BlockProfileRate int `mapstructure:"blockprofilerate"`
+
+	// MutexProfileFraction, when greater than zero, is passed to
+	// runtime.SetMutexProfileFraction at startup to enable mutex-contention
+	// profiling. Zero (the default) leaves mutex profiling disabled.
+	MutexProfileFraction int `mapstructure:"mutexprofilefraction"`
 }
 
 // WalletConfig selects and configures the wallet backend.
--- a/darepod/logging.go
+++ b/darepod/logging.go
@@ -39,6 +39,7 @@ var allSubsystems = []string{
 	db.Subsystem,
 	SwapSubsystem,
 	WalletRPCSubsystem,
+	PprofSubsystem,
 	"TXCF",
 	"UNRL",
 	VHTLCRecoverySubsystem,
@@ -59,6 +60,11 @@ const (
 	// VHTLCRecoverySubsystem is the subsystem tag used for vHTLC
 	// on-chain recovery coordination logs.
 	VHTLCRecoverySubsystem = "VREC"
+
+	// PprofSubsystem is the subsystem tag used for the optional pprof
+	// debug server so its logs can be level-tuned independently of the
+	// main daemon logs.
+	PprofSubsystem = "PPRF"
 )
 
 // SetupLoggersWithShutdownFn registers all subsystem loggers using a plain
--- a/darepod/pprof.go
+++ b/darepod/pprof.go
@@ -1,0 +1,184 @@
+package darepod
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net"
+	"net/http"
+	"net/http/pprof"
+	"runtime"
+	"sync"
+	"time"
+
+	"github.com/btcsuite/btclog/v2"
+)
+
+const (
+	// defaultPprofReadHeaderTimeout bounds request-line/header reads so a
+	// slow client cannot hold the local debug listener open indefinitely.
+	defaultPprofReadHeaderTimeout = 5 * time.Second
+
+	// defaultPprofReadTimeout bounds the full request read. It is larger
+	// than the header timeout because /debug/pprof/symbol accepts a POST
+	// body (a list of program counters) that must be read fully.
+	defaultPprofReadTimeout = 30 * time.Second
+
+	// CPU and trace profiles intentionally hold the response open for the
+	// caller-requested sampling window, so no fixed write deadline applies.
+	defaultPprofWriteTimeout = 0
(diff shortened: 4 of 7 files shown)


bhandras:
claude: thanks for the review — addressed all points and folded into the commit (force-pushed):

1. **Subtest slashes** — subtest names are now sanitized (`strings.ReplaceAll(TrimPrefix(path,"/"), "/", "_")`), so `-run` filtering is unambiguous.
2. **Dedicated pprof subsystem** — added `PprofSubsystem = "PPRF"`, registered it in `allSubsystems`, and the server now logs via `s.subLogger(PprofSubsystem)` so pprof logs are independently level-tunable.
3. **Unnecessary nil-receiver guards** — dropped the `p == nil` / `p.cfg == nil` checks in `Start`/`Stop`/`Addr` (impossible per `DefaultConfig`/`newPprofServer`); kept the real-state checks (`ListenAddr == ""`, `httpSrv == nil`, `listener == nil`).
4. **`0b` section label** — dropped the number; it's now a plain stanza header consistent with the surrounding blocks.

Minor:
- `defaultPprofWriteTimeout` now has an explicit `time.Duration` type.
- Added a `Start`/`Stop` lifecycle test (binds `127.0.0.1:0`, serves, then asserts the listener is released) plus a disabled-`Start` no-op test.
- `Stop` now resets the block/mutex profiling rates it enabled (same point raised on the arkd PR) so their overhead doesn't outlive the server.
- `sample-darepod.conf` now notes the server is plain HTTP (no TLS) and unauthenticated.

The `Stop` context-detach `//nolint:contextcheck` is kept as-is — it matches `gateway_server.go` and the caller ctx is already cancelled at shutdown.

bhandras:
claude: round 2 addressed (force-pushed):

1. **Latent nil deref** — changed `Config.Pprof` from `*PprofConfig` to a **value type** `PprofConfig`, so a zero-value `Config{}` can never carry a nil into the start path. This also keeps the round-1 request to drop the impossible nil-receiver guards (a value config can't be nil).
2. **`ErrorS` for an external trigger** — switched the serve-error log to `WarnS` per the convention (error level is for internal bugs only).
3. **`wg.Wait()` vs shutdown timeout** — added a comment: `Shutdown` closes the listener before draining, so `Serve` has already returned `ErrServerClosed` and `wg.Wait()` only joins the (already-exiting) serve goroutine.
5. **profile/trace coverage** — added `TestNewPprofMuxRegistersProfileAndTrace`, which asserts both routes resolve via `mux.Handler` *without* triggering an actual capture.
7. **PR description** — fixed the endpoint paths to the full `/debug/pprof/...` form.

Holding on two (with reasons):
- **(4) rate reset on `Stop`** — keeping it: an earlier review on the sibling arkd PR asked for exactly this (sampling overhead must not outlive the server), and `Start`/`Stop` are symmetric. Restart re-applies rates on the next `Start`. Now applied only *after* the listener binds so a bind failure leaves globals untouched.
- **(6) `defaultPprofWriteTimeout time.Duration = 0`** — keeping the named constant: round 1 explicitly asked for this type annotation, and the name + comment document intent at the `http.Server` field. Reverting would just re-open the round-1 note.

bhandras:
claude: round 3 addressed (force-pushed):

1. **Weak profile/trace assertion** — good catch; `mux.Handler` returns the `/debug/pprof/` prefix pattern as a fallback, so `NotEmpty` passed even without the explicit routes. The test now asserts the matched pattern **equals** the exact path (`/debug/pprof/profile` / `/debug/pprof/trace`), so removing either explicit registration fails it.
2. **Mid-word `"runti"+"me"` split** — reworded the flag descriptions to start with "value passed to …" so the formatter wraps at word boundaries; no more mid-word break.
3. **`Stop` context discard** — moved the rationale into the `Stop` GoDoc so it's visible from the signature.

Holding (with reasons):
- **(4) typed-zero const** — keeping it; round 1 explicitly asked for the `time.Duration` annotation, so this is a flip from that. The name + comment document intent at the field.
- **(5) Start/Stop race** — the daemon lifecycle calls these sequentially and this mirrors `gatewayServer`; noted for future test evolution, not changing here.

bhandras:
claude: round 4 addressed (force-pushed); CI is green.

1. **Raw `net.Addr` in the listening log** — fixed: now `slog.String("addr", listener.Addr().String())`, matching the gRPC listener log style.
3. **Profiling-rate paths untested** — added `TestPprofServerAppliesProfilingRates`, which starts/stops with `BlockProfileRate`/`MutexProfileFraction` set so the `Start` apply and `Stop` reset paths are exercised.

Holding:
2. **No mutex on `listener`/`httpSrv`** — declining: the daemon drives `Start`/`Stop`/`Addr` sequentially from the startup path and this mirrors `gatewayServer`. The tests don't call `Addr()` concurrently with `Start`, so the race detector has nothing to flag. A mutex would be dead weight given the single-goroutine lifecycle.

bhandras:
claude: round 5 addressed (force-pushed):

2. **`Stop` defer passed the cancelled daemon ctx** — fixed via your option (b): `pprofServer.Stop` now uses the caller-provided context directly (no internal bounded ctx), and the defer builds a `context.WithTimeout(context.Background(), DefaultShutdownTimeout)` with `//nolint:contextcheck`, matching `gateway.Stop`.
3. **`Addr()` misleading after `Stop`** — fixed: `Stop` now clears `p.listener`, so `Addr()` returns nil once stopped; added an assertion for it.

Nits:
- **30s `ReadTimeout`** — kept, with a comment: `/debug/pprof/symbol` accepts a POST body (a list of program counters) that must be read in full, so the full-request read timeout is intentionally larger than `ReadHeaderTimeout`.
- **`t.Cleanup` reset** — added to `TestPprofServerAppliesProfilingRates` so the process-global rates are reset even on a panic.

Declining (1) **`allSubsystems` ordering**: the premise that the list is alphabetical doesn't hold — the existing tail is `SWAP, WRPC, TXCF, UNRL, VREC` (`WRPC` already precedes `TXCF`). The list is grouped logically (daemon core → backends → feature subsystems), so appending the new `PPRF` feature subsystem at the end is consistent. Happy to reorder if you'd prefer a specific grouping.

bhandras:
claude: current head is `97c67cbb`. I cleaned up the actionable review items and folded them into the single PR commit.

Changed in the folded head:

- Added an inline reason to the `//nolint:contextcheck` in `darepod/server.go`.
- `PprofServer.Start` now logs `DebugS` when pprof is disabled.
- Runtime profiling rates are only applied after the listener bind succeeds, so a bind failure cannot leave process-global profiling enabled.
- `Stop` now clears both `listener` and `httpSrv`, making `Addr` inactive after stop and making repeated `Stop` calls harmless.
- Tests now exercise `Stop` idempotency and keep explicit loop-var capture for the endpoint table loops.
- Removed the redundant zero-value `Pprof` initializer from `DefaultConfig`.
- Moved `PPRF` before `TXCF` in `allSubsystems`.

Not changing in this pass:

- No mutex around `Start`/`Stop`/`Addr`: the daemon lifecycle calls these sequentially, matching the current server lifecycle pattern, and the tests do not exercise concurrent lifecycle calls.
- No extra change for the captured run context in the serve goroutine: it is only used for logging metadata and has no functional effect on shutdown or request handling.

Revalidated locally on the folded head:

- `make fmt-changed`
- `go test ./darepod -run 'TestNewPprofMux|TestPprofServer' -count=1`
- `go test ./cmd/darepod ./darepod -count=1`
- `make lint-changed-local`
- `make fmt-changed-check`

Please re-review current head only; several older comments are stale after the folded changes.

bhandras:
> claude: current head is `97c67cbb`. I cleaned up the actionable review items and folded them into the single PR commit.
> 
> Changed in the folded head:
> 
> * Added an inline reason to the `//nolint:contextcheck` in `darepod/server.go`.
> * `PprofServer.Start` now logs `DebugS` when pprof is disabled.
> * Runtime profiling rates are only applied after the listener bind succeeds, so a bind failure cannot leave process-global profiling enabled.
> * `Stop` now clears both `listener` and `httpSrv`, making `Addr` inactive after stop and making repeated `Stop` calls harmless.
> * Tests now exercise `Stop` idempotency and keep explicit loop-var capture for the endpoint table loops.
> * Removed the redundant zero-value `Pprof` initializer from `DefaultConfig`.
> * Moved `PPRF` before `TXCF` in `allSubsystems`.
> 
> Not changing in this pass:
> 
> * No mutex around `Start`/`Stop`/`Addr`: the daemon lifecycle calls these sequentially, matching the current server lifecycle pattern, and the tests do not exercise concurrent lifecycle calls.
> * No extra change for the captured run context in the serve goroutine: it is only used for logging metadata and has no functional effect on shutdown or request handling.
> 
> Revalidated locally on the folded head:
> 
> * `make fmt-changed`
> * `go test ./darepod -run 'TestNewPprofMux|TestPprofServer' -count=1`
> * `go test ./cmd/darepod ./darepod -count=1`
> * `make lint-changed-local`
> * `make fmt-changed-check`
> * `make commitmsg-lint range=origin/main..HEAD`
> 
> Please re-review current head only; several older comments are stale after the folded changes.

@claude review

bhandras:
claude: current head is `47c5a16f`. I folded the remaining small review items into the single PR commit.

Changed in this head:

- Decoupled the pprof serve goroutine's log context from the daemon run context using `context.WithoutCancel(ctx)`, so unexpected listener errors are not logged with an already-cancelled context during shutdown.
- Added an inline note at the `http.Server` literal explaining why `WriteTimeout` intentionally uses the no-deadline default for streaming CPU/trace profiles.
- Added a short comment before the profiling-rate reset conditions explaining they mirror the `Start` conditions.
- Removed the explicit `path := path` loop-variable shadowing in the pprof tests. The module is Go 1.25, so each range iteration already has its own variable; the older Gemini comment requesting the shadow is stale for this repo's Go version.

Revalidated locally:

- `make fmt-changed`
- `go test ./darepod -run 'TestNewPprofMux|TestPprofServer' -count=1`
- `go test ./cmd/darepod ./darepod -count=1`
- `make lint-changed-local`
- `make fmt-changed-check`

Please re-review current head only.

bhandras:
claude: current head is `c5a5c1f8`. I folded the latest small review items.

Changed in this head:

- Documented that the daemon owns `pprofServer` lifecycle and calls `Start`/`Stop` sequentially.
- Added a comment in `TestNewPprofMuxServesStandardEndpoints` explaining why `/debug/pprof/profile` and `/debug/pprof/trace` are omitted from the live-serving endpoint list and covered by pattern lookup instead.
- Changed `defaultPprofWriteTimeout` to `time.Duration(0)`.
- Tightened the `WriteTimeout` inline comment to explicitly say `0` means no deadline for CPU/trace streaming.

Revalidated locally:

- `make fmt-changed`
- `go test ./darepod -run 'TestNewPprofMux|TestPprofServer' -count=1`
- `go test ./cmd/darepod ./darepod -count=1`
- `make lint-changed-local`
- `make fmt-changed-check`

Please re-review current head only.

bhandras:
claude: current head is `f20c49ae`. I folded the latest test/constant polish.

Changed in this head:

- Simplified the after-Stop request assertion while still closing a response body if one is unexpectedly returned, satisfying `bodyclose` without keeping the earlier confusing require/fail pattern.
- Added comments for `defaultPprofReadHeaderTimeout` and `defaultPprofIdleTimeout`.
- Changed `defaultPprofWriteTimeout` back to plain `0`.
- Expanded the profiling-rate reset comment to call out the single-lifecycle invariant.

Revalidated locally:

- `make fmt-changed`
- `go test ./darepod -run 'TestNewPprofMux|TestPprofServer' -count=1`
- `go test ./cmd/darepod ./darepod -count=1`
- `make lint-changed-local`
- `make fmt-changed-check`

Please re-review current head only.

PR #640, opened 2026-05-29
oor: bound incoming metadata retries with backoff and give-up

Roasbeef:
## Motivation

On signet the `swapdk-ark-client` was issuing `arkrpc.IndexerService/ListVTXOsByScripts` queries at ~77/sec, nonstop, each with a distinct msg_id. The backing mailbox held only a handful of rows at any instant — the queries were answered and acked instantly, so it was a live hot loop, not a backlog.

Root cause: the OOR incoming-receive phase-2 metadata resolution has no convergence guard. When materialization reports the VTXO is not yet in the indexer, it returns a *retryable* error with a flat **1s** delay (`oor/transitions.go` `defaultRetryDelay`) and **no attempt cap**. A session whose VTXO never lands re-issues the query once per second forever; many such sessions stuck at once sum to the observed rate. (Upstream those stuck sessions came from a swapd in-swap retry storm — fixed separately in swapdk-server.)

## Changes

- **`oor`** — `handleReceiveOutboxError` now routes `ReceiveNotified` retries through bounded **exponential backoff** (1s base, 5m cap) keyed off a new `MetadataAttempts` counter, and **gives up terminally** (→ `Failed`) after `maxMetadataRetries` (20) instead of scheduling another query. The counter rides on the `ReceiveNotified` state and is persisted in the incoming snapshot (new TLV record type 19) so the bound survives restarts — the resume path clears *volatile* retry metadata, so it must live in persisted FSM state.
- **`serverconn`** — demote the per-pull "Pulled envelopes" log from debug to trace. The ingress success path has no delay, so a continuously-fed mailbox spins the loop at RTT speed and floods debug logs with one line per single-envelope pull.

## Tests

New `oor` tests cover the backoff progression + cap, the terminal give-up, the backoff helper bounds, and the snapshot round-trip of the attempt counter. Full `oor` package passes; `make lint-changed-local` reports 0 issues.

Behavior change to call out: an incoming receive session that cannot resolve its metadata now fails terminally after ~an hour of attempts rather than retrying indefinitely. The terminal `Failed` reason carries the underlying error.

### DIFF
--- a/oor/actor.go
+++ b/oor/actor.go
@@ -2432,6 +2432,26 @@ func (b *oorDurableBehavior) resumeOutboxForHandle(handle *sessionHandle,
 		return OutboxForState(outgoingState)
 
 	case sessionKindIncoming:
+		// An incoming session mid-backoff on metadata resolution must
+		// resume by re-scheduling the retry rather than firing the
+		// query immediately. Otherwise a restart during one of the
+		// capped backoff windows resets the wait to zero, and repeated
+		// restarts could burn through maxMetadataRetries far faster
+		// than the intended backoff schedule. The persisted attempt
+		// count reproduces the same deterministic delay.
+		if notified, ok := state.(*ReceiveNotified); ok &&
+			notified.MetadataAttempts > 0 {
+			return []OutboxEvent{
+				&ScheduleRetryRequest{
+					After: metadataRetryBackoff(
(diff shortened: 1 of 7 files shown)


Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #641, opened 2026-05-29
Add passkey-derived wallets to the walletdk web demo

jamaljsr:
## Summary

The walletdk web demo only supports password unlock today. This adds a **passkey wallet** as a separate choice at create time; password wallets remain available and work exactly as before. A passkey wallet derives its seed deterministically from a platform passkey's WebAuthn PRF output, so there's no password to choose or remember. Because a synced passkey produces the same PRF output on every device that has it, the same wallet re-derives anywhere. A passkey wallet is portable across your devices rather than tied to the browser it was created in.

The recovery phrase shown at create is the backstop: it's how you recover if the passkey is lost or you're on a device outside the passkey's sync ecosystem.

Based on #431

## Demo

https://github.com/user-attachments/assets/13868a33-eb03-4406-8ecd-1f5c8c6db17c

## How it works

A passkey wallet is rooted entirely in the passkey's PRF secret:

1. **Create.** Register a platform passkey with the PRF extension (eval input pinned to `SHA-256("walletdk-passkey:v1")`). The raw PRF output is handed to the Go SDK, which runs HKDF-SHA256 with domain-separated labels to derive (a) 16 bytes of entropy and (b) a local DB password. The entropy becomes a deterministic aezeed cipher seed (pinned version and birthday) and the wallet is initialized from it. The recovery phrase is shown for backup.
2. **Open.** Assert the passkey, get the PRF output, and call the Go SDK's `OpenWalletFromPasskey`, which branches on wallet state: if no local wallet exists it re-derives the seed and imports; if the wallet is locked it unlocks with the derived DB password; if it's already unlocked it's a no-op. One entry point covers first create, local unlock, and fresh-device import.
3. **Cross-device.** A synced passkey (iCloud Keychain or Google Password Manager, within one ecosystem) gives the same PRF output on the second device, so it re-derives the identical wallet. Across ecosystems, you recover with the phrase.

All derivation lives in Go (shared by the WASM bridge and any future native binding); the web layer only runs the WebAuthn ceremony and hands over the raw PRF bytes.

## What gets stored

A passkey wallet stores **no secret in `localStorage`**: the seed is re-derived from the passkey each time. localStorage holds only non-secret markers per data dir:

- `walletdk:wallet-kind:{dataDir}` stores `passkey` or `password`, so the unlock screen shows the right path.
- `walletdk:passkey-cred:{dataDir}` stores the credential id, so a return unlock can scope the assertion to the right passkey (no OS chooser).

The encrypted seed lives in OPFS SQLite, and its DB-encryption password is also derived from the passkey rather than chosen by the user.

## Changes

- **Go SDK (`sdk/walletdk`):** adds `OpenWalletFromPasskey`, which derives entropy and a DB password from the PRF output (HKDF), builds a deterministic aezeed, and imports or unlocks based on wallet state. No daemon or proto changes.
- **WASM bridge:** adds an `openWalletFromPasskey` entrypoint that decodes the PRF hex and calls the SDK.
- **`walletdk-wasm-web`:** a ceremony-only `passkey.ts` that runs the WebAuthn PRF register and assert and returns the raw output as hex (a discoverable assertion so a synced passkey works on a fresh device).
- **`walletdk-react`:** a `usePasskeyWallet` hook.
- **Demo app:** create picks a **passkey wallet** (derived, no password, portable) or a **password wallet** (classic); onboarding split into focused screens; passkey unlock scoped to the stored credential; passkeys named with a timestamp so multiple test credentials stay distinguishable; and a "Clear wallet data" action (localStorage and OPFS) behind a confirm modal.

## Testing

- [ ] Create a passkey wallet on a PRF-capable browser; confirm no password is requested and the recovery phrase is shown.
- [ ] Reload and unlock with the passkey (scoped, no OS chooser); confirm the same wallet and address.
- [ ] On a second device or browser with the same synced passkey, "Unlock with passkey" re-derives and imports the same wallet.
- [ ] Recover a wallet from its phrase on a device without the passkey.
- [ ] Password-wallet create and unlock still work; restore-from-phrase still works.
- [ ] "Clear wallet data" clears localStorage and OPFS and returns to fresh onboarding (theme preserved).
- [ ] Graceful behavior on browsers without PRF support.


jamaljsr:
This work has been extracted out of this repo.

PR #646, opened 2026-05-29
rpc: Add SendOnChain for exact-amount onchain sends (fixes #634)

### LINKED ISSUE #634: darepocli: onchain send transfers more sats to destination than --amt requested
## Summary

`darepocli send <addr> --onchain --amt <amt>` can transfer significantly
more sats to the destination address than the user-specified `--amt`.

The CLI prints a `note:` line about "whole-VTXO sweep semantics" and
returns `actual_amount_sat` in the response body, but the underlying
behaviour is that the user asked to send `--amt` and the destination
receives substantially more.

Either:

1. The exact `--amt` should be sent to the destination (any excess
   should go back to the user as change), or
2. The CLI should make it very clear up front — before the transaction
   is broadcast — that more funds than `--amt` are about to leave the
   user's wallet to the destination, and require confirmation.

## Example

This is the same transaction that prompted #615.

Command:

```bash
darepocli --no-tls send tb1qge3qhgnzn8337wm3540fm0ng4lm47wlw3t0z8u --onchain --amt 25000
```

Output:

```
note: actual_amount_sat=904745 exceeds --amt=25000 due to whole-VTXO sweep semantics
{
  "entry": {
    "id": "16064dcc8f5f21e4393086c643bdb2f708a83b3ad4fa5866c57eeb55134f3259:0",
    "kind": "ENTRY_KIND_EXIT",
    "status": "ENTRY_STATUS_PENDING",
    "amount_sat": "-25000",
    "fee_sat": "0",
    "counterparty": "tb1qge3qhgnzn8337wm3540fm0ng4lm4",
    ...
    "request": {
      "onchain_address": {
        "address": "tb1qge3qhgnzn8337wm3540fm0ng4lm47wlw3t0z8u"
      }
    },
    "progress": {
      "phase": "WALLET_ENTRY_PHASE_REQUEST_CREATED",
      ...
    }
  },
  "actual_amount_sat": "904745"
}
```

Resulting tx:
https://mempool.space/signet/tx/3c3b93c4d7f5d605280de7ab98043e739e0b2ea2649226d5a4b18509df08394e

The first 4 outputs all go to my address
`tb1qge3qhgnzn8337wm3540fm0ng4lm47wlw3t0z8u`:

```
tb1qge3qhgnzn8337wm3540fm0ng4lm47wlw3t0z8u    0.00002356 sBTC
tb1qge3qhgnzn8337wm3540fm0ng4lm47wlw3t0z8u    0.00005000 sBTC
tb1qge3qhgnzn8337wm3540fm0ng4lm47wlw3t0z8u    0.00005000 sBTC
tb1qge3qhgnzn8337wm3540fm0ng4lm47wlw3t0z8u    0.00889745 sBTC
```

User asked for `--amt 25000`, destination received `904,745` sats — over
36× the requested amount.

## Why it matters

A user who runs `--amt 25000` reasonably expects ~25,000 sats to be
sent. Silently sending 36× more, with the only signal being a `note:`
line in the response body, is a footgun — especially in scripted or
automated flows where the response body may not be inspected before
the tx is broadcast.

## Expected

See the two options in the Summary. Option 1 (exact amount + change) is
the principle-of-least-surprise behaviour for an `--amt` flag. If
whole-VTXO sweep semantics are intentional and unavoidable in some
cases, Option 2 should at minimum gate the behaviour behind explicit
user confirmation rather than a single `note:` line.

## Related

- #615 — `wallet: cooperative leave / onchain send sweeps multiple
  small VTXOs instead of selecting minimally` (same flow, different
  angle — that issue is about how many VTXOs get swept; this one is
  about the destination amount itself diverging from `--amt`).

## Environment

- Client: darepocli built from `main` at commit
  [`32b33fb5d908afd599a3efaecbd887a206273bb1`](https://github.com/lightninglabs/darepo-client/commit/32b33fb5d908afd599a3efaecbd887a206273bb1)
  (the `main` HEAD at the time #615 was opened).
- Network: signet

ellemouton:
## Summary

- Adds a new \`SendOnChain\` daemonrpc RPC that gives \`darepocli send --onchain --amt N\` exact-amount semantics: the destination receives EXACTLY N sats, and any residual returns to the caller as a change VTXO.
- The handler picks live VTXOs via the existing manager admission gate, emits a mixed intent (1 forfeit set + 1 fixed leave output + 1 IsChange=true VTXO), and registers it atomically.
- Rewires \`walletrpc.Send\` (the wallet-shaped multiplexed send verb) to dispatch onchain through \`SendOnChain\` instead of the old whole-VTXO \`LeaveVTXOs\` + \`JoinNextRound\` path. The \`ark vtxos leave\` raw CLI keeps using \`LeaveVTXOs\` unchanged for advanced batching.
- Drops the \"v1 whole-VTXO sweep semantics\" warning from the CLI; the overpay it warned about can no longer happen.

Fixes lightninglabs/darepo-client#634.

## Background

Before: \`darepocli send --onchain --amt 25000\` silently swept whole VTXOs to the destination because the underlying \`LeaveVTXOs\` primitive has no change output. Viktor reported sending 25,000 sats and the destination receiving ~905,000 sats — a 36× overpay. The CLI emitted a \`note:\` to stderr after the fact, but the round was already in flight.

After: the daemon emits one fixed leave output of exactly \`amount_sat\` and a change VTXO. The seal-time fee handshake (#270) stamps the operator fee onto the change VTXO via \`IsChange=true\`, so the destination still lands the exact requested amount.

## Commit order

The 7 commits split along natural seams so each one builds and tests cleanly:

1. \`rpc: Add SendOnChain RPC definition\` — proto + grpc-gateway route.
2. \`rpc: Regenerate stubs for SendOnChain\` — \`make rpc\`.
3. \`restclient: Implement SendOnChain over the REST gateway\` — handwritten REST stub.
4. \`wallet: Add SendOnChain handler for exact-amount onchain sends\` — wallet actor handler + actor message types.
5. \`darepod: Wire SendOnChain RPC handler\` — RPC server method with the split-Ask pattern (mirrors \`LeaveVTXOs + JoinNextRound\`).
6. \`swapwallet: Route onchain send through SendOnChain\` — router rewrite + test updates.
7. \`darepocli: Drop whole-VTXO overpay warning from send\` — CLI text cleanup.

## Test plan

- [x] \`go test -tags \"walletrpc swapruntime\" ./...\` — all client unit tests pass.
- [x] swapwallet router tests updated to assert the new exact-amount invariants.
- [x] End-to-end coverage in the parent darepo repo: new \`TestSendOnChainExactAmount\` itest boards alice with 120k sats, sends \`amount_sat=30000\`, confirms the round, asserts bitcoind destination receives exactly 30000 sats AND alice retains a change VTXO ≈ 90k − operator fee.
- [x] Manual reproduction via \`arktest\` (\`darepocli --no-tls send <addr> --onchain --amt 25000\`): bitcoind \`getreceivedbyaddress\` returns exactly 0.00025000 and alice's wallet shows a change VTXO of ~872k sats.

## Companion PR

The parent darepo repo needs (a) a submodule bump to pick up these commits and (b) an unrelated rounds-actor fix that arktest's reproduction exposed. Will link here once opened.

### DIFF
--- a/daemonrpc/daemon.proto
+++ b/daemonrpc/daemon.proto
@@ -104,6 +104,17 @@ service DaemonService {
     // lands as an on-chain output at the specified destination.
     rpc LeaveVTXOs (LeaveVTXOsRequest) returns (LeaveVTXOsResponse);
 
+    // SendOnChain is the wallet-shaped onchain payment primitive: it
+    // picks live VTXOs to cover the requested amount, forfeits them
+    // in the next round, produces one on-chain output of the exact
+    // requested amount at the supplied destination, and returns any
+    // residual to the caller as a change VTXO. The atomic
+    // forfeit + leave + change-VTXO intent registers immediately
+    // (TriggerRegistration=true) so the caller does not need to also
+    // issue JoinNextRound. sweep_all skips selection and drains every
+    // live VTXO to the destination (no change VTXO produced).
+    rpc SendOnChain (SendOnChainRequest) returns (SendOnChainResponse);
+
     // Board triggers the client to join the next round with any
     // confirmed boarding UTXOs. This sends IntentRequested to
     // the round FSM, which emits a JoinRoundRequest to the server.
@@ -961,6 +972,63 @@ message LeaveVTXOsResponse {
 
     // status is the result: "queued" on success, "preview" on dry_run.
     string status = 2;
+}
+
+message SendOnChainRequest {
+    // destination is the on-chain recipient. Same shape (and standardness
+    // gating) as LeaveVTXOsRequest.default_destination.
+    LeaveDestination destination = 1;
+
+    // amount picks the send mode. Exactly one variant must be set.
+    oneof amount {
+        // amount_sat is the exact number of sats to send on-chain to
+        // destination. The daemon selects live VTXOs whose summed value
+        // covers amount_sat plus an operator-fee headroom plus dust,
+        // forfeits them, produces one on-chain output of exactly
+        // amount_sat, and returns the residual to the caller as a
+        // change VTXO. Must be > 0 and ≤ MaxSatoshi.
+        int64 amount_sat = 2;
+
+        // sweep_all, when true, drains every live VTXO into a single
+        // on-chain output at destination. No change VTXO is produced;
+        // the actual on-chain amount equals Σ(live VTXOs) − operator
+        // fee and is echoed back as actual_amount_sat.
+        bool sweep_all = 3;
+    }
+
+    // dry_run validates inputs and returns the planned selection
+    // without submitting the round intent.
+    bool dry_run = 5;
+}
+
+message SendOnChainResponse {
+    // actual_amount_sat is the exact value of the on-chain output
+    // that will land at destination. For amount_sat mode this echoes
+    // the requested amount. For sweep_all it is Σ(selected VTXOs) −
+    // operator fee.
+    int64 actual_amount_sat = 1;
+
+    // fee_sat is the operator fee deducted from the input set for
+    // this send. Under the #270 seal-time fee handshake this is the
+    // server-stamped figure, not a pre-flight estimate. Zero in
+    // dry_run mode.
+    int64 fee_sat = 2;
+
+    // selected_outpoints lists the VTXOs the daemon picked to
+    // forfeit for this send, in the order they were chosen.
+    repeated string selected_outpoints = 3;
+
+    // change_outpoint is the outpoint of the change VTXO produced
+    // by an amount_sat send. Empty in dry_run, in sweep_all mode,
+    // or when no change is needed (input set exactly covers
+    // amount_sat + fee + dust). The outpoint is populated once the
+    // round seals and the new VTXO is persisted; in fast-path
+    // responses it may be empty pending the round.
+    string change_outpoint = 4;
+
+    // status is "submitted" on a successful intent registration,
+    // "preview" when dry_run is set.
+    string status = 5;
 }
 
 message BoardRequest {
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -1507,6 +1507,189 @@ func (r *RPCServer) LeaveVTXOs(ctx context.Context,
 	}, nil
 }
 
+// SendOnChain plans and submits an atomic onchain payment from VTXOs.
+// The RPC enforces the wallet-API invariants on the request shape
+// (exactly one of amount_sat / sweep_all set, destination present,
+// destination resolves to a standard pkScript) and then dispatches to
+// the wallet actor's handleSendOnChain. For sweep_all the RPC layer
+// also enumerates the live VTXO set via vtxoStore so the wallet does
+// not have to bypass the manager's admission gate to drain the wallet.
+func (r *RPCServer) SendOnChain(ctx context.Context,
+	req *daemonrpc.SendOnChainRequest) (*daemonrpc.SendOnChainResponse,
+	error) {
+
+	// Pure-argument validation first so a malformed request surfaces
+	// InvalidArgument before any wallet-state check.
+	if req.GetDestination() == nil {
+		return nil, status.Errorf(codes.InvalidArgument,
+			"destination is required")
+	}
+
+	pkScript, err := r.resolveLeaveDestination(req.GetDestination())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument,
+			"destination: %v", err)
+	}
+
+	var (
+		targetAmount btcutil.Amount
+		sweepAll     bool
+	)
+	switch amt := req.Amount.(type) {
+	case *daemonrpc.SendOnChainRequest_AmountSat:
+		if amt.AmountSat <= 0 ||
+			amt.AmountSat > int64(btcutil.MaxSatoshi) {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"amount_sat must be between 1 and %d",
+				int64(btcutil.MaxSatoshi))
+		}
+		targetAmount = btcutil.Amount(amt.AmountSat)
+
+	case *daemonrpc.SendOnChainRequest_SweepAll:
+		if !amt.SweepAll {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"sweep_all must be true when selected")
+		}
+		sweepAll = true
+
+	default:
+		return nil, status.Errorf(codes.InvalidArgument, "amount "+
+			"mode is required (amount_sat or sweep_all)")
+	}
+
+	// Fetch operator terms for the change VTXO descriptor + the
+	// coin-selection headroom hint. Under #270 the binding fee is
+	// stamped server-side; MinOperatorFee is advisory only.
+	terms, err := r.server.fetchOperatorTerms(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "fetch operator "+
+			"terms: %v", err)
+	}
+
+	// For sweep_all, enumerate live VTXOs up front. We do this at
+	// the RPC layer (not in the wallet actor) so the dry_run path
+	// can echo the planned outpoint set without reserving anything,
+	// and so the wallet does not need its own live-set listing seam.
+	var sweepOutpoints []wire.OutPoint
+	if sweepAll {
+		if r.server.vtxoStore == nil {
+			return nil, status.Errorf(codes.Internal, "vtxo "+
+				"store not initialized")
+		}
+
+		liveVTXOs, err := r.server.vtxoStore.ListLiveVTXOs(ctx)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "list live "+
+				"VTXOs: %v", err)
+		}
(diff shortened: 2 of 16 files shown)


daemonrpc/daemon.proto
```diff
@@ -104,6 +104,17 @@ service DaemonService {
...
+    // SendOnChain is the wallet-shaped onchain payment primitive: it
+    // picks live VTXOs to cover the requested amount, forfeits them
+    // in the next round, produces one on-chain output of the exact
+    // requested amount at the supplied destination, and returns any
+    // residual to the caller as a change VTXO. The atomic
+    // forfeit + leave + change-VTXO intent registers immediately
+    // (TriggerRegistration=true) so the caller does not need to also
+    // issue JoinNextRound. sweep_all skips selection and drains every
+    // live VTXO to the destination (no change VTXO produced).
+    rpc SendOnChain (SendOnChainRequest) returns (SendOnChainResponse);
```
Roasbeef, line 116: Perhaps this should go solely in the walletrpc server? As this is the `ark` server which is lower level.

wallet/messages.go
```diff
@@ -718,3 +718,119 @@ func (m *SendVTXOsResponse) MessageType() string {
...
+	// SweepAll drains every live VTXO in SweepOutpoints to
+	// DestinationPkScript. No change VTXO is produced; the single
+	// leave output absorbs the residual under the #270 fee handshake.
+	SweepAll bool
+
+	// SweepOutpoints carries the live-VTXO outpoint set enumerated
+	// by the RPC server for SweepAll mode. Empty in bounded mode.
+	// The wallet does not enumerate live VTXOs itself; the RPC
+	// layer's vtxoStore listing is the single source of truth.
+	SweepOutpoints []wire.OutPoint
```
Roasbeef, line 754: Can collapse this wit hthe above, as if `SweepOutpoints` if `nil/None`, then we know it isn't the sweep all case.

wallet/messages.go
```diff
@@ -718,3 +718,119 @@ func (m *SendVTXOsResponse) MessageType() string {
...
+	// coin-selection headroom in bounded mode.
+	DustLimit btcutil.Amount
+
+	// OperatorKey is the operator's pubkey for the change-VTXO
+	// policy template. Required in bounded mode; unused in SweepAll.
+	OperatorKey *btcec.PublicKey
+
+	// VTXOExitDelay is the CSV delay for the change VTXO's exit
+	// path. Required in bounded mode; unused in SweepAll.
+	VTXOExitDelay uint32
```
Roasbeef, line 774: Why is this needed at all? Given the VTXOs will be looked up?

wallet/messages.go
```diff
@@ -718,3 +718,119 @@ func (m *SendVTXOsResponse) MessageType() string {
...
+// walletMsgSealed implements the sealed WalletMsg interface.
+func (m *SendOnChainRequest) walletMsgSealed() {}
+
+// SendOnChainResponse carries the outcome of a SendOnChainRequest.
+type SendOnChainResponse struct {
+	actor.BaseMessage
+
+	// Status is "submitted" for a successfully registered intent or
+	// "preview" for a dry-run.
+	Status string
```
Roasbeef: Can make a proper enum, with a `String()` method.

wallet/messages.go
```diff
@@ -718,3 +718,119 @@ func (m *SendVTXOsResponse) MessageType() string {
...
+	// is reduced by the server's seal-time operator fee.
+	ActualAmountSat btcutil.Amount
+
+	// SelectedOutpoints is the set of VTXOs forfeited for this send,
+	// in selection order.
+	SelectedOutpoints []wire.OutPoint
+
+	// TotalSelected is the sum of all SelectedOutpoints' VTXO
+	// amounts.
+	TotalSelected btcutil.Amount
```
Roasbeef, line 849: Given the disjoint fields in this req/resp, could consider reaching for two sub-strcuts composed with an `fn.Either`. Mainly a taste thing, as IMO it makes the code easier to scan (disjoin paths clear) but adds some boilerplate when you need to access/set the fields.

darepod/rpc_server.go
```diff
@@ -1507,6 +1507,196 @@ func (r *RPCServer) LeaveVTXOs(ctx context.Context,
...
+			len(sendResp.SelectedOutpoints)),
+		slog.Int64("change_amount",
+			int64(sendResp.ChangeAmount)),
+	)
+
+	return &daemonrpc.SendOnChainResponse{
+		ActualAmountSat:   int64(sendResp.ActualAmountSat),
+		SelectedOutpoints: outpointStrs,
+		Status:            sendResp.Status,
+
```
Roasbeef, line 1682: Would be nice if this gave a `txid`....tho we can't know that until the round is actually created. Something to think about re a general status/activity endpoint.

Roasbeef: Eg: should `activity/List` now attempt to link the: send request, round creation, and final round publish?

daemonrpc/daemon.proto
```diff
@@ -963,6 +974,69 @@ message LeaveVTXOsResponse {
...
+        // the actual on-chain amount equals Σ(live VTXOs) − operator
+        // fee and is echoed back as actual_amount_sat.
+        bool sweep_all = 3;
+    }
+
+    // max_fee_sat is an upper bound on the operator fee the caller is
+    // willing to pay. Zero means "use the daemon default" and is the
+    // common case. The cap is enforced when the server-stamped
+    // seal-time quote arrives; an overshoot rejects the round join.
+    int64 max_fee_sat = 4;
```
Roasbeef: This doesn't seem to be used anywhere

darepod/rpc_server.go
```diff
@@ -1507,6 +1507,196 @@ func (r *RPCServer) LeaveVTXOs(ctx context.Context,
...
+	// Trigger round registration on a fresh Ask boundary so the
+	// IntentRequested → outbound JoinRoundRequest publish → operator
+	// admission lookup chain runs under TriggerRoundRegistration's
+	// detached context. This mirrors the JoinNextRound RPC pattern:
+	// SendOnChain is a one-shot, so the wallet host must not be
+	// asked to also call JoinNextRound separately. Skip for dry_run
+	// since the wallet handler released the reservation rather than
+	// committing the intent.
+	if !req.DryRun {
+		if err := r.server.TriggerRoundRegistration(ctx); err != nil {
```
Roasbeef, line 1654: I wonder if we should make this reliable. Right now if the daemon dies before a new round starts, then this send intent is lost. 

We added something similar for new boarding attempts, if they user says they want to board, we'll remember that and on restart reply it if needed. 

One actor that we never made fully durable was the round actor, as it was sort of an open question if it should actually be. If it was, we could just try on the fact that `TriggerRoundRegistration` returning non nil would mean we alwys join a around.

daemonrpc/daemon.proto
Roasbeef: Got through the rest of the diff, can go either way now. Nice to have it still on the ark level IMO as it helps to take care of a lot of boilerplate.

ellemouton:
Thanks for the review @Roasbeef — pushed a round of fixups (force-pushed, autosquashed into the relevant commits). Per-comment:

**Actioned:**
- **`max_fee_sat` unused** — removed from `SendOnChainRequest` proto + the wallet message + all plumbing. It was threaded through four layers but never actually read (the cap isn't enforced anywhere under #270; the server is the fee authority at seal time). Will re-add when there's a real quote-handshake client that enforces it.
- **`Status` → enum** — now a typed `SendOnChainStatus` (Submitted/Preview) with a `String()` method; the RPC layer projects it onto the proto string field.
- **Collapse `SweepAll` bool** — dropped the redundant bool from the wallet message. Sweep-all is now implied by a non-empty `SweepOutpoints` (exposed via `IsSweepAll()`), since the RPC layer only populates it in sweep mode.

**Kept (with rationale):**
- **`VTXOExitDelay`** — this is the CSV exit delay for the *new change VTXO* output, sourced from `OperatorTerms`, not derivable from the input VTXOs being forfeited. It mirrors how `SendVTXOsRequest` carries `VTXOExitDelay`/`OperatorKey` for its self-change output. Happy to revisit if you'd rather the wallet fetch terms itself, but that'd diverge from the directed-send path.
- **`fn.Either` for the disjoint req/resp fields** — left as-is for now since you flagged it as taste; can do it in a follow-up if you feel strongly.
- **proto location (ark vs walletrpc)** — kept on the ark/daemonrpc level per your follow-up (\"nice to have it on the ark level... takes care of a lot of boilerplate\").

**Follow-ups (not in this PR):**
- **txid in the response / `activity/List` linking the send→round→publish chain** — agreed these want a general status/activity story; can't surface a txid until the round is created. Worth its own issue.
- **Durable send intent (restart safety)** — good point that a daemon crash before the next round drops the intent, same gap the durable Board replay closed. This ties into the open question of making the round actor fully durable. I'd prefer to tackle that as a dedicated follow-up rather than expand this PR's scope — filing an issue to track it.

ellemouton:
Durability follow-up filed: lightninglabs/darepo-client#660

Roasbeef (approved):
LGTM 💃🏾

Issue #661, opened 2026-06-02
ledger: in-round multi-receive can be silently dropped (vtxo_balance undercount)


Roasbeef:
## Summary

Two distinct in-round transfer receives for the same client in the **same
round** can be silently de-duplicated into one ledger row, undercounting
`vtxo_balance`. The per-VTXO `idempotency_key` does **not** prevent this, despite
comments/docs claiming it does.

This is **pre-existing** (the index and handler dedup logic predate the durable
Read/Commit conversion; introduced around commit `ae62dd22b`). Filing to track a
fix separately from that work.

## Root cause

`idx_client_ledger_idempotent_round` is defined on:

```sql
-- db/sqlc/migrations/000006_fee_accounting.up.sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_client_ledger_idempotent_round
    ON ledger_entries(round_id, event_type, debit_account, credit_account)
    WHERE round_id IS NOT NULL;
```

It does **not** include `idempotency_key`. Two `VTXOReceivedMsg{SourceRoundTransfer}`
for the same client in one round both produce the identical round-index key
`(round_id, vtxo_received, vtxo_balance, transfers_in)`. They differ only in the
per-outpoint `idempotency_key`.

`InsertClientLedgerEntry` uses an **untargeted** `ON CONFLICT DO NOTHING`, which
fires on *any* unique-index violation. So the second leg collides on the round
index and is silently dropped — even though its `idempotency_key` is distinct.
The separate `idx_client_ledger_idempotent_key` index does not save it because
the round-index violation triggers first.

## Reachability

A client that receives two separate VTXOs via in-round transfer in a single
round emits two `VTXOReceivedMsg{SourceRoundTransfer}` with the same `round_id`
(producer: `round/actor.go`). The handler comment itself enumerates same-round
multi-receive scenarios (three-way directed send, multi-leg refresh, a round
with both a boarding intent and a received transfer), so the author anticipated
multiple same-round receives — but the chosen mitigation (per-VTXO key) does not
actually prevent the collision.

## Misleading docs to correct

- `ledger/handlers.go:199-204` — comment claims the per-VTXO key means
  "multiple owned receives in the same round ... don't collide on
  idx_client_ledger_idempotent_round." This is false given the index columns.
- `ledger/CLAUDE.md` — repeats the same claim.

## Remediation options

1. Use a **targeted** `ON CONFLICT (idempotency_key, event_type, debit_account,
[...]

PR #662, opened 2026-06-02
actor: add durable Read/Commit execution path (+ ledger)

Roasbeef:
In this PR, we add a new "Read/Commit" execution path to the durable actor
framework so a behavior can do its slow side-effect IO (signing RPCs, calls
out to the server, polling, etc) *without* holding the SQLite writer the whole
time, while still getting an atomic, exactly-once commit at the end. This is
the framework-level fix for the OOR/swap concurrency stalls tracked in
lightninglabs/darepo#313: today `DurableActor` wraps the entire `Receive` in
one `ExecTx`, so if a behavior does any IO inside that tx the single WAL writer
is held across the network round-trip and every other actor (plus the ingress
dispatch write) stalls behind it.

The design follows section 4 of the framework sketch. A behavior opts in by
implementing `TxBehavior[M, R, S]` and is driven through an `Exec[S]` handle
with two phases. `Read` runs a short read-only snapshot (no writer lock); the
behavior then does its IO with no tx held; finally `Commit` runs one short
writer tx. The trick that keeps this exactly-once is a lease fence: inside
`Commit` we ack the mailbox row first (`DELETE ... WHERE id = ? AND lease_token
= ?`), and if that deletes zero rows the lease was reclaimed while we were off
doing IO, so we abort with `ErrLeaseLost` and the whole tx rolls back, applying
nothing. On the happy path the behavior's own writes and the dedup mark fold
into that same tx, so the state advance and the message consumption commit
together. Any duplicate effect a stale consumer might have already emitted is
deduped downstream by the receiver's `ON CONFLICT (id) DO NOTHING`.

changes.

## Framework

The store type `S` is a free type parameter injected via a `StoreFactory`, so
`DurableActor` itself stays `[M, R]` and the typed store is handed to the
behavior's `Commit` closure (it joins the tx via `actor.TxFromContext`). The
config's `Behavior` becomes an `fn.Either` of the classic `ActorBehavior`
(Left, the unchanged safe default) and the new `BoundTxBehavior` (Right).
`NewDurableActor` now returns an `fn.Result` and fails closed at construction:
a `TxBehavior` paired with a non-tx-aware store gets `ErrTxBehaviorNeedsTxStore`,
and a missing or zero-value behavior gets `ErrNoBehavior` instead of panicking
at first dispatch. `NewClassicBehavior`, `NewTxBehaviorEither`, and
`DefaultDurableTxActorConfig` keep the either-wrapping boilerplate out of
callers.

Property tests with rapid pin the two invariants that matter: the lease fence
(consume exactly when the token matches, else `ErrLeaseLost` with zero side
effects) and atomic grouping (any number of effects commit all-or-nothing
alongside consumption and the dedup mark).

## Ledger

We convert the client ledger actor as the first adopter. Each accounting
handler now books its ledger legs inside one fenced `Commit` via a typed
`ledgerTx` store, so multi-leg events like `ExitCost` commit every leg
atomically with the mailbox ack. We also fix the wiring: the ledger used to
register an in-memory actor via `RegisterWithSystem`, so producer `Tell`s were
handled synchronously and the durable mailbox went unused. We now register the
durable ref with the receptionist (the same shape OOR and serverconn already
use) and stop the actor explicitly during shutdown, after the producers drain
but before `db.Close`, so the Commit/fence path is actually exercised and
best-effort accounting Tells aren't dropped in the shutdown window. An
end-to-end test drives a real `ExitCostMsg` through the durable mailbox and
asserts two-leg atomicity, idempotent replay, and rollback-on-reject.

## Follow-ups

The OOR effect actors and serverconn egress are the next adopters (section 4.5
of the sketch) and aren't in this PR.

While reviewing we also turned up a pre-existing ledger dedup bug that's
orthogonal to this work: two in-round transfer receives in the same round can
collide on the `idx_client_ledger_idempotent_round` index (which doesn't
include the per-outpoint key) and the untargeted `ON CONFLICT DO NOTHING`
silently drops the second. Tracked separately in #661.

## Test plan

- [x] `make lint` (Docker) clean
- [x] `baselib/actor`, `ledger`, `internal/actortest`, `serverconn`, `oor`, `unroll`, `db/actordelivery` unit tests pass
- [x] rapid property tests for the lease fence + atomic grouping
- [x] end-to-end ledger test over a real SQLite-backed durable mailbox

### DIFF
--- a/baselib/actor/durable_actor.go
+++ b/baselib/actor/durable_actor.go
@@ -43,9 +43,22 @@ type DurableActorConfig[M TLVMessage, R any] struct {
 	// When unset, the runtime falls back to btclog.Disabled.
 	Log fn.Option[btclog.Logger]
 
-	// Behavior defines how the actor responds to messages.
-	// The runtime handles ack/nack automatically based on the result.
-	Behavior ActorBehavior[M, R]
+	// Behavior defines how the actor responds to messages. It is one of two
+	// shapes:
+	//
+	//   - Left: a classic ActorBehavior. The runtime wraps the whole
+	//     Receive in one transaction (processInTransaction) when the store
+	//     is tx-aware, else runs the short-slice path. This is the safe
+	//     default.
+	//
+	//   - Right: a BoundTxBehavior built via NewTxBehavior. The runtime
+	//     drives the behavior through the Read/Commit execution path
+	//     (processWithExec) so it can do side-effect IO outside the writer
+	//     transaction. This requires a TxAwareDeliveryStore.
+	//
+	// DefaultDurableActorConfig populates the Left case from its behavior
+	// argument; tx-aware actors overwrite this field with fn.NewRight.
+	Behavior fn.Either[ActorBehavior[M, R], BoundTxBehavior[M, R]]
 
 	// Store is the persistence layer for mailbox operations. If the store
 	// implements TxAwareDeliveryStore, message processing will be wrapped
@@ -99,6 +112,30 @@ type DurableActorConfig[M TLVMessage, R any] struct {
 	DeduplicationTTL time.Duration
 }
 
+// NewClassicBehavior wraps a classic ActorBehavior as the Left case of the
+// DurableActorConfig.Behavior either. DefaultDurableActorConfig applies it for
+// you; use it directly only when assigning Behavior on a hand-built config.
+func NewClassicBehavior[M TLVMessage, R any](
+	behavior ActorBehavior[M, R],
+) fn.Either[ActorBehavior[M, R], BoundTxBehavior[M, R]] {
+
+	return fn.NewLeft[ActorBehavior[M, R], BoundTxBehavior[M, R]](behavior)
+}
+
+// NewTxBehaviorEither binds a TxBehavior and its StoreFactory and wraps the
+// result as the Right case of the DurableActorConfig.Behavior either.
+// DefaultDurableTxActorConfig applies it for you; use it directly only when
+// assigning Behavior on a hand-built config.
+func NewTxBehaviorEither[M TLVMessage, R any, S any](
+	behavior TxBehavior[M, R, S],
+	factory StoreFactory[S],
+) fn.Either[ActorBehavior[M, R], BoundTxBehavior[M, R]] {
+
+	return fn.NewRight[ActorBehavior[M, R]](
+		NewTxBehavior(behavior, factory),
+	)
+}
+
 // DefaultDurableActorConfig returns a config with sensible defaults.
 func DefaultDurableActorConfig[M TLVMessage, R any](
 	id string,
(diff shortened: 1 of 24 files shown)


Roasbeef:
@claude review this

Roasbeef:
Pushed an update addressing the review bots plus a maintainer-requested change. Folded into the existing three commits (no new commits), so it's a force-update.

**Read/Commit framework (`baselib/actor`)**
- DurableAsk on the exec path is no longer silently consumed (gemini high / codex P1). On this path the message is acked inside the behavior's own `Commit`, so the framework can't fold the crash-safe callback response into the consuming tx (it has no result at Commit time). Rather than half-implement it with an out-of-tx write that can drop the response after consume, `processWithExec` now rejects a `DurableAsk` up front: it writes an error `AskResponse` to the outbox (so the caller gets a definitive error instead of hanging) and consumes the message under the lease fence, via the new `ErrDurableAskUnsupported`. Full support lands with the OOR effect-actor adopter. There are no DurableAsk users on the exec path today (the ledger is Tell-only).
- The lease heartbeat now stops via `sync.Once` right after the behavior returns, instead of at function end (gemini medium). On the committed path the row is already acked/deleted, so a late tick no longer tries to extend a gone lease and log a spurious warning.

**Ledger handlers now hold the writer only for the inserts**
The bigger one: as written, each handler ran entirely inside `ax.Commit`, so validation, account selection, description formatting, and idempotency-key derivation all held the SQLite writer -- effectively the old whole-Receive-in-one-tx shape. Each handler now owns its `Exec` handle: it validates and builds the entries first (no lock), then calls `ax.Commit` with a closure containing only the `InsertLedgerEntry` / `InsertUTXOAuditEntry` writes. Multi-leg `ExitCost` still commits both legs in one fenced tx. A future IO-bearing handler uses the same shape (read -> IO -> commit). Validation failures return before any Commit and never open a writer tx.

`make lint` clean; ledger, baselib/actor, and the e2e actortest suites pass.

Roasbeef:
@claude review this

Roasbeef:
Rebased this PR onto `mailbox-formal` (#436) and retargeted the base, so it's now stacked on the durable-actor mailbox P model.

Added a commit extending that model to cover this PR's Read/Commit design: `p-models: model the Read/Commit lease-fenced consume step`.

- New `eDurableMailboxCommit` op on `DurableMailboxSpec`: the behavior effect + dedup mark + lease-fenced ack as one atomic unit. Its ack always validates the lease token (mirroring `DELETE ... WHERE id AND lease_token`); a `fenced` flag decides whether the effect is bound to that token check.
- New `LeaseFencedCommitAppliesEffectAtMostOnce` monitor: a row's effect is applied at most once even when its lease expires mid-IO and the row is reclaimed and reprocessed.
- `tcMailboxReadCommitFence` (green): the fenced design holds — the stale consumer's Commit is an `ErrLeaseLost` no-op, effect applied exactly once.
- `tcMailboxUnfencedCommitCounterexample` (negative, no in-machine assertion): the unfenced profile double-applies under reclaim and is caught solely by the monitor — proving it catches the bug the fence prevents.
- Trace bridge: a new `commit` op replays the fence against the real `db/actordelivery` SQLite store (`ExecTx{ fenced AckMessage → MarkProcessed }`, rolling back with `ErrLeaseLost` on a zero-row ack), driven by the new `mailbox_read_commit_fenced_exactly_once.json` trace.

`./p-models/scripts/check.sh` is green: 3 green P cases (incl. the new one) find 0 bugs, 3 negatives (incl. the new one) each find their expected bug, and the Go bridge replays all traces.

Roasbeef:
@claude review this

Issue #509, opened 2026-05-21
walletrpc: add PrepareSend for interactive send confirmation


sputn1ck:
## Problem

The high-level wallet send path currently dispatches funds in one step:

- `darepocli send <invoice-or-onchain-address>` builds a `walletrpc.SendRequest`.
- `walletrpc.WalletService.Send` immediately starts the underlying payment flow.
- Invoice sends go through the daemon-owned swap runtime (`StartPay`).
- Onchain sends go through cooperative leave semantics.

That is awkward for interactive wallet UX. A CLI or UI should be able to show the user what will be paid before committing funds, especially for invoice sends where the wallet may settle either:

- in-Ark / same-Ark, without paying an external Lightning route, or
- over Lightning, with expected routing/swap fees.

The current API can only return `actual_amount_sat` after dispatch. That is too late for a human confirmation prompt.

This is related to, but separate from, #497. That issue is about durable activity-feed metadata after a wallet operation exists. This issue is about a pre-dispatch quote/intent flow that lets clients ask: "what would this send do?" before funds move.

## Proposed API shape

Split the wallet send flow into preparation and execution semantics:

```proto
service WalletService {
    rpc PrepareSend(PrepareSendRequest) returns (PrepareSendResponse);
    rpc Send(SendRequest) returns (SendResponse);
}
```

`PrepareSend` should validate the destination, parse payment metadata, ask the daemon/swap subsystem for the best available route/quote, and return a short-lived send intent. It must not dispatch funds.

`Send` should consume that prepared intent rather than re-parsing and dispatching directly from raw user input.

Sketch:

```proto
message PrepareSendRequest {
    oneof destination {
        string invoice = 1;
        string onchain_address = 2;
    }

    uint64 amt_sat = 3;
    uint64 max_fee_sat = 4;
    string note = 5;
    bool sweep_all = 6;
}

message PrepareSendResponse {
    string send_intent_id = 1;

    int64 amount_sat = 2;
    int64 expected_fee_sat = 3;
    int64 expected_total_outflow_sat = 4;

    SendRail rail = 5;
    string rail_label = 6;

    string destination_summary = 7;
    string invoice_description = 8;
    string payment_hash = 9;

    int64 expires_at_unix = 10;
    WalletEntry preview_entry = 11;
}

enum SendRail {
    SEND_RAIL_UNSPECIFIED = 0;
    SEND_RAIL_IN_ARK = 1;
    SEND_RAIL_LIGHTNING = 2;
[...]

PR #663, opened 2026-06-02
txconfirm: keep never-broadcast checkpoints retrying with operator escalation

Roasbeef:
## Summary

Addresses [issue #509](https://github.com/lightninglabs/darepo/issues/509) on the client side. A fraud-response checkpoint whose initial broadcast reached no mempool (CPFP child sign **and** direct-parent fallback both failing, surfaced as `ErrCPFPFeeInputUnavailable`) was advanced straight to `AwaitingConfirmation` and then retried forever with no escalation — a tracked tx silently "awaiting confirmation" that was never actually sent. On a fund-risk path this is dangerous: the checkpoint must confirm before the client's unilateral-exit CSV matures.

This PR:

1. **Keeps never-broadcast txs in `Broadcasting`** (item 2). The `Broadcasting` state now carries progress (last attempt height + a consecutive-failure counter) and self-loops on re-attempt. A total broadcast failure stays in `Broadcasting` and is re-attempted every `FeeBumpIntervalBlocks` from the block observer, sharing the broadcast-outcome handling with the initial submit. `ErrParentAlreadyBroadcast` still advances to `AwaitingConfirmation` (a live parent genuinely exists on another path).
2. **Escalates on repeated deterministic failure, never auto-fails** (item 3). After `BroadcastFailureAlertThreshold` (default 5) consecutive failures, the actor emits a distinct operator-visible `WARN`. It never transitions to terminal `Failed` — a tx that must land keeps retrying.
3. **Pins the CPFP-child finalize contract** (item 1). Investigation showed item 1's stated root cause does **not** reproduce on current `main`: `signCPFPChild` already pre-finalizes the canonical P2A anchor (empty witness) for *every* anchor-bearing TRUC parent, so LND `FinalizePsbt` only signs the wallet fee input. The original report predates this code. Rather than re-fix a non-bug, a regression test builds a checkpoint-shaped parent (pre-signed taproot key-spend input, unlike a round commitment tx) and asserts the child's anchor input is pre-finalized empty, the fee input is left for the wallet, the package is submitted, and the parent's signed input is preserved byte-for-byte.

## Test plan

- [x] `make unit pkg=./txconfirm/...` — all green, including new FSM, actor, and regression tests
- [x] `go test -race ./txconfirm/...` — race-clean
- [x] `go vet ./txconfirm/...`, `gofmt`, gopls diagnostics clean
- [ ] CI lint (local `custom-gcl` blocked by a `GOROOT==GOPATH` env issue in the dev sandbox)

## Notes

- New `Config.BroadcastFailureAlertThreshold` is additive (zero falls back to the default); no caller changes required.
- Server-side observability for the #508-amplified fund-risk window lands in the parent `darepo` repo.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,7 @@ require (
 	golang.org/x/exp v0.0.0-20251113190631-e25ba8c21ef6
 	golang.org/x/sync v0.20.0
 	golang.org/x/term v0.41.0
+	golang.org/x/time v0.11.0
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260414002931-afd174a4e478
 	google.golang.org/grpc v1.80.0
 	google.golang.org/protobuf v1.36.11
@@ -206,7 +207,6 @@ require (
 	golang.org/x/oauth2 v0.36.0 // indirect
 	golang.org/x/sys v0.42.0 // indirect
 	golang.org/x/text v0.36.0 // indirect
-	golang.org/x/time v0.11.0 // indirect
 	golang.org/x/tools v0.43.0 // indirect
 	google.golang.org/genproto v0.0.0-20240213162025-012b6fc9bca9 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20260414002931-afd174a4e478 // indirect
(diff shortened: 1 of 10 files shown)


Roasbeef:
@claude review this

PR #667, opened 2026-06-02
swapclientserver: return pending pay duplicates

sputn1ck:
## Summary
- decode StartPay invoices locally before creating a new pay swap
- return the existing pending swap summary when the payment hash is already in flight
- re-arm the pay worker for duplicate pending starts

## Testing
- go test -tags="dev nolog swapruntime" -timeout=5m ./swapclientserver
- make commitmsg-lint commit=HEAD
- git diff --check

### DIFF
--- a/sdk/swaps/in_swap.go
+++ b/sdk/swaps/in_swap.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btclog/v2"
+	"github.com/lightninglabs/darepo-client/daemonrpc"
 	"github.com/lightninglabs/darepo-client/lib/arkscript"
 	loopfsm "github.com/lightninglabs/loop/fsm"
 	"github.com/lightningnetwork/lnd/lntypes"
@@ -1056,6 +1057,14 @@ func (s *paySession) completeRefund(ctx context.Context) error {
 		return nil
 	}
 
+	sessionHandled, err := s.reconcilePayRefundSession(ctx)
+	if err != nil {
+		return newRetryableActionError(err)
+	}
+	if sessionHandled {
+		return nil
+	}
+
 	funded, err := s.observeRefundableVHTLC(ctx)
 	if err != nil {
 		return newRetryableActionError(
@@ -1065,10 +1074,6 @@ func (s *paySession) completeRefund(ctx context.Context) error {
 	}
 	if !funded {
 		return waitForFixedPoll(ctx, s.client.waitPollInterval)
-	}
-
-	if s.refundSessionID != "" {
-		return s.markRefundAccepted(ctx)
 	}
 
 	height, err := s.client.daemon.BlockHeight(ctx)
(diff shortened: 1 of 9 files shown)


sputn1ck:
@claude review

bhandras (commented):
Flagging one refund-state concern inline.

sdk/swaps/in_swap.go
```diff
@@ -1068,7 +1068,14 @@ func (s *paySession) completeRefund(ctx context.Context) error {
...
 	if s.refundSessionID != "" {
-		return s.markRefundAccepted(ctx)
+		s.client.log.DebugS(
+			ctx,
+			"Pay swap refund accepted; waiting for refund output",
+			btclog.Hex("hash", s.cfg.PaymentHash[:]),
+			slog.String("refund_session_id", s.refundSessionID),
+		)
+
+		return waitForFixedPoll(ctx, s.client.waitPollInterval)
```
bhandras: Is it safe to remove the accepted-refund terminal path here? Previously `markRefundAccepted` treated the accepted custom-input OOR as the durable refund point, transitioned the pay swap to `Refunded`, and cancelled the armed VHTLC recovery. Now a persisted `refundSessionID` only causes polling until the refund output or spent vHTLC is indexed. If that visibility never appears, `Wait()` can remain stuck in `RefundInitiated` even though the daemon accepted the refund. Can we preserve an accepted-session completion path, or add another durable reconciliation path that does not depend solely on refund-output indexing?

sputn1ck: Addressed in 215f1f52.

I kept the stricter retry boundary (an accepted refund send alone does not immediately make the pay swap terminal), but added daemon OOR-session reconciliation so we are not solely dependent on refund-output indexing:

- `refundSessionID` now calls `GetOORSession` during `completeRefund`.
- `COMPLETED` marks the pay swap `Refunded` and cancels the armed VHTLC recovery with a dedicated reason.
- `PENDING`/unknown keeps polling and keeps the duplicate `StartPay` path blocked on the existing pending swap.
- `FAILED` clears the stale refund session id and lets the same pending swap continue into refund/recovery logic rather than allowing an invoice retry.

I extended `TestPaySessionCooperativeRefundWaitsForOutputBeforeTerminal` to cover both halves: no indexed output + no completed OOR session stays `RefundInitiated`, then daemon `OOR_SESSION_STATUS_COMPLETED` transitions to `Refunded`.

Validation run locally:
- `go test -tags="dev test_sqlite nolog" -timeout=5m ./sdk/swaps -run 'TestPaySessionCooperativeRefund(UsesLocalFundingMetadata|WaitsForOutputBeforeTerminal|sBeforeTimeout)|TestPaySessionCooperativeRefundKeepsAcceptedSessionOnPersistFailure|TestPaySessionResumeFundingGraceEventuallyRetries|TestPaySessionResumeAfterAcceptedRefund'`
- `go test -tags="dev test_sqlite nolog" -timeout=5m ./sdk/ark ./sdk/swaps -run '^$'`
- `make lint-local`
- `make fmt-changed-check base=$(git merge-base origin/main HEAD)`

bhandras (approved):
LGTM 🎉

PR #658, opened 2026-06-01
vtxo: Recover VTXO to live on a failed no-broadcast unroll (#602)

ellemouton:
## Summary

Fixes #602.

Fixes https://github.com/lightninglabs/darepo-client/issues/690 

A unilateral exit (`darepocli exit`, or auto-triggered at critical expiry)
that **failed without broadcasting anything** still removed the VTXO from
`ark vtxos list` and the wallet balance — while the Ark server still
considered it live. The VTXO actor was reaped on the **user's intent to
exit** rather than on a terminal **on-chain** event, so a no-broadcast
failure permanently stranded the coin.

This PR makes the exit *observed and reversible*: the VTXO actor stays
alive through the exit and is either **rolled back to live** (clean
failure, nothing on-chain) or **retired to spent** (exit confirmed
on-chain).

---

## Root cause

`UnilateralExitState` was a **terminal** FSM state. Entering it emitted a
`VTXOTerminatedNotification`, so the manager deleted the actor *on intent*,
before anything hit the chain. There was no path back, and no feedback edge
from the unroll job to the VTXO manager.

```
 BEFORE
 ──────
  darepocli exit ─▶ VTXO actor ──▶ UnilateralExit (TERMINAL)
                       │             + VTXOTerminatedNotification
                       │                     │
                       │                     ▼
                       │               manager reaps actor  ─▶ ✗ dropped from
                       │ ExpiringNotification                    live set / balance
                       ▼
                  unroll registry ─▶ child ─▶ txconfirm
                                      │
                                      ▼ proof tx rejected: "min relay fee not met"
                                  Failed (nothing broadcast)
                                      │
                                      ▼ notifies its own registry ONLY
                                  ✗ no edge back to VTXO  ─▶ STRANDED forever
```

---

## The fix

`UnilateralExitState` is now **non-terminal**. The actor survives to observe
the exit and the manager applies the unroll job's terminal on-chain outcome.

```mermaid
stateDiagram-v2
    [*] --> Live
    Live --> UnilateralExit: ForceUnroll / critical expiry<br/>(emit ExpiringNotification; no terminated)
    UnilateralExit --> Live: ExitFailedEvent<br/>(clean failure, no on-chain footprint)
    UnilateralExit --> Spent: ExitConfirmedEvent<br/>(swept + confirmed on-chain)
    UnilateralExit --> UnilateralExit: block epoch / duplicate unroll<br/>(observe, self-loop)
    Spent --> [*]
    note right of UnilateralExit
[...]

Issue #606, opened 2026-05-28
oor: 'ark send oor' accepts sub-dust amounts; operator co-signs an unspendable dust VTXO


jtobin:
## Environment

- `darepo-client`: `cd035b84` (also reproducible on current main `32b33fb5`)
- Network: signet
- Wallet backend: `lnd`
- Ark server: `arkd-signet.testnet.lightningcluster.com:443`

## Summary

`darepocli ark send oor --amount N` with `N < dust_limit` is accepted end-to-end by both the client and the operator. The recipient ends up with a live sub-dust VTXO (1 sat, 999 sat — below the publicly advertised `dust_limit: 1000`). The operator co-signs without rejecting. This creates wallet dust that cannot be cooperatively swept (a leave producing sub-dust outputs would be rejected) and consumes forfeit slots.

## Expected Behavior

Reject any OOR send whose recipient amount is below `server_info.dust_limit`. The operator publishes the dust limit explicitly; honoring it on the client side is the only way to prevent the wallet from creating unspendable dust VTXOs that the user has paid round/forfeit fees on.

The CLI should reject before submitting; the daemon RPC should also reject if a caller goes direct; the operator should reject any sub-dust send that slips past. Error text should name the field and the limit, e.g. `amount 999 below operator dust_limit 1000`.

## Actual Behavior

With `server_info.dust_limit = "1000"` (visible from `getinfo`), both `--amount 1` and `--amount 999` sends are accepted:

```
NODE=b darepocli ark send oor \
  --pubkey d0cc46ed223c79550e848fd0ca6584cf9c154a9948e6ea17ce051496ae57fb18 \
  --amount 1
{ "status": "submitted",
  "session_id": "624c91e2...4923" }

NODE=b darepocli ark send oor \
  --pubkey d0cc46ed223c79550e848fd0ca6584cf9c154a9948e6ea17ce051496ae57fb18 \
  --amount 999
{ "status": "submitted",
  "session_id": "233940f8...44a7" }
```

Both completed end-to-end. After the 1-sat send, node a held a brand-new live VTXO of `amount_sat: 1` (below the operator's stated dust limit), and node b held a 24,999-sat change VTXO. After the 999-sat send, node a held a `amount_sat: 999` live VTXO, node b held a 24,000-sat change. Daemon logs show clean state transitions (`Idle → AwaitingArkSignatures → … → Completed`) and `change_sat=24999`, `change_sat=23001` — no warnings, no rejection at any layer.

[...]

Issue #602, opened 2026-05-28
vtxo: failed unilateral exit silently removes VTXO from wallet's live set despite no on-chain spend


jtobin:
## Environment

- `darepo-client`: `dcbacc0a` (also reproducible on current main `32b33fb5`)
- Network: signet
- Wallet backend: `lnd`
- Ark server: `arkd-signet.testnet.lightningcluster.com:443`

## Summary

A `darepocli exit --outpoint ...` that progresses to `EXIT_JOB_STATUS_FAILED` **without broadcasting anything to the mempool** still removes the underlying VTXO from `ark vtxos list`. The Ark server side still considers the VTXO live; the local wallet has forgotten about it. This is a soundness gap: the VTXO actor is reaped on the strength of the user's intent rather than on the strength of an on-chain commit.

## Expected Behavior

A failed unroll job — especially one whose state machine never advances past `Broadcasting → Failed` (no on-chain footprint) — must restore the VTXO to `VTXO_STATUS_LIVE` so the wallet's view of its coins matches what the operator still considers live.

## Actual Behavior

Initiated a unilateral exit on a 1-sat VTXO on signet node a:

```
# before
"outpoint": "624c91e2…:0", "amount_sat": "1", "status": "VTXO_STATUS_LIVE"

# after FAILED unroll, with no broadcast: outpoint absent from
# `ark vtxos list`. balance dropped 5000 → 4999.
```

Daemon log confirms the VTXO actor was reaped on the **intent** to unroll, not on an on-chain commit:

```
[INF] VTXO: Force-unroll accepted by VTXO actor
   outpoint=624c91e2…:0 reason="manual RPC request"
   new_state=*vtxo.UnilateralExitState
[INF] VTXO: VTXO actor terminated outpoint=624c91e2…:0
   final_state=UnilateralExit reason="manual RPC request"
```

The proof tx then failed to broadcast (`min relay fee not met` on the 1-sat VTXO; sub-dust acceptance covered in a separate issue). Job marked `FAILED`. No rollback restored the VTXO to LIVE.

Related to #587 (VTXOs stuck in `Spending` with no resume reconciliation) — same root-cause family: local VTXO state changes are gated on transient intent rather than terminal on-chain events.

## Reproduction

1. Have any small live VTXO on hand. A sub-dust input (e.g. 1 sat) makes this deterministic because the proof tx can't pay min relay fee.
2. `darepocli exit --outpoint <txid:vout>`
[...]

Issue #450, opened 2026-05-15
`unroll` results in "unilateral exit job not found"


hieblmi:
```
da unroll --outpoint abe9beddc595be85cb4af9be7ce2c96222bb31343a746e866fa93afa1d9256a5:0
```
results in
```
2026-05-15 09:37:27.775 [DBG] VTXO: VTXO actor received event event_type=*vtxo.ForceUnrollEvent outpoint=abe9beddc595be85cb4af9be7ce2c96222bb31343a746e866fa93afa1d9256a5:0 current_state=*vtxo.LiveState                                                                                                                                             
2026-05-15 09:37:27.775 [DBG] VTXO: VTXO FSM transition completed next_state=*vtxo.UnilateralExitState outbox_len=3                                                                                                                                                                                                                                   
2026-05-15 09:37:27.776 [DBG] VTXO: Processing VTXOStatusUpdate outpoint=abe9beddc595be85cb4af9be7ce2c96222bb31343a746e866fa93afa1d9256a5:0 new_status=unilateral_exit has_forfeit_tx=false is_forfeiting_with_tx=false                                                                                                                               
2026-05-15 09:37:27.783 [DBG] VTXO: Unsubscribed from block epochs vtxo=abe9beddc595be85cb4af9be7ce2c96222bb31343a746e866fa93afa1d9256a5:0                                                                                                                                                                                                            
2026-05-15 09:37:27.783 [INF] VTXO: Force-unroll accepted by VTXO actor outpoint=abe9beddc595be85cb4af9be7ce2c96222bb31343a746e866fa93afa1d9256a5:0 reason="manual RPC request" new_state=*vtxo.UnilateralExitState                                                                                                                                   
2026-05-15 09:37:27.783 [INF] VTXO: VTXO actor terminated outpoint=abe9beddc595be85cb4af9be7ce2c96222bb31343a746e866fa93afa1d9256a5:0 final_state=UnilateralExit reason="manual RPC request"                                                                                                                                                          
2026-05-15 09:37:27.783 [DBG] ACTR: Actor stopped and removed from system actor_id=epoch.vtxo.abe9beddc595be85cb4af9be7ce2c96222bb31343a746e866fa93afa1d9256a5:0                                                                                                                                                                                      
2026-05-15 09:37:27.784 [WRN] DABS: Transaction body failed err="unilateral exit job not found"                                                                                                                                                                                                                                                       
[...]

PR #671, opened 2026-06-02
darepod: Gate unilateral exit on economic feasibility

### LINKED ISSUE #608: unroll: 'exit --outpoint' accepts sub-dust VTXOs whose proof tx cannot meet min relay fee
## Environment

- `darepo-client`: `dcbacc0a` (also reproducible on current main `32b33fb5`)
- Network: signet
- Wallet backend: `lnd`
- Ark server: `arkd-signet.testnet.lightningcluster.com:443`

## Summary

`darepocli exit --outpoint <txid:vout>` accepts a VTXO whose `amount_sat` is below the dust limit + min-relay-fee headroom needed to broadcast the proof tx. The exit is admitted, the unroll FSM proceeds to `Broadcasting`, LND rejects the proof tx for `min relay fee not met`, the job marks `FAILED`, and — per #602 — the VTXO is silently removed from the local wallet's view. The whole cascade starts at the missing pre-flight here.

## Expected Behavior

Refuse unilateral exit when `vtxo.amount_sat < (dust_limit + estimated_proof_tx_fee_at_current_feerate)`. Surface a clean client-side error before admitting the job. Suggested message: `VTXO amount X sat insufficient to broadcast proof tx at current feerate (need at least Y sat); use cooperative leave instead`.

## Actual Behavior

Exit dispatched against a 1-sat VTXO on signet node a:

```
darepocli exit --outpoint 624c91e206e96d82d9bd2159817a394bb65ce39836b0ef580417f9db7fc4a923:0
{
  "created": true,
  "actor_id": "unroll-624c91e206e96d82d9bd2159817a394bb65ce39836b0ef580417f9db7fc4a923:0"
}
```

Exit code 0. The unroll FSM admitted the job, transitioned the VTXO to `*vtxo.UnilateralExitState`, and dispatched the proof tx. LND rejected:

```
[ERR] LNDC: PublishTransaction failed
   txid=ee770706083ad4cf50c469242eeeb44a810f639c8a7a675bcbee986dbe6ea35a
   error="min relay fee not met"
```

Job marked `EXIT_JOB_STATUS_FAILED` shortly after the watchdog timeout. `exit status` reports an opaque `txconfirm returned failed state` (filed separately as our #27, summary deferred).

Cascade:
- Pre-flight gap (this issue): nothing checked whether the 1-sat input could pay min relay fee.
- LND rebroadcast loop (filed separately as our #30, summary deferred): even after `FAILED`, LND keeps retrying the same doomed tx every ~60s.
- VTXO disappears from local view (#602 / our #28): the wallet's view of its own funds diverges from the operator's.

## Reproduction

```sh
# Get a sub-dust VTXO. The deterministic way is to OOR-send 1 sat to yourself
# (which itself is accepted — see our #20).
darepocli ark oor receive --label rxr
darepocli ark send oor --pubkey <hex> --amount 1
# wait for completion, then identify the 1-sat outpoint via vtxos list.

darepocli exit --outpoint <txid:vout>
# Accepted. Watch FAILED in ~30s.
darepocli exit status --outpoint <txid:vout>
darepocli ark vtxos list   # outpoint is GONE
```

## Notes

Discovered during a swapdk testing sprint Batch C unilateral-exit pass on the smallest available VTXO. Companion to #602 (cascade outcome) and #20 / a separate sub-dust-OOR-send filing (how the sub-dust VTXO was created in the first place).

ellemouton:
## Issue

Fixes #608.

`darepocli exit --outpoint …` admitted a unilateral exit after only
checking that the wallet held one confirmed UTXO per ancestry path. It
never looked at the target VTXO's own value, so a sub-dust VTXO was
admitted, transitioned to `UnilateralExitState`, and then failed
downstream with `min relay fee not met` when the proof/sweep tx could
not be broadcast — stranding the VTXO in an exit state the user can't
easily back out of.

## Background

A holistic fee check was proposed in #450 (`checkUnrollFeeBalance` /
`unrollFeeShortfall`, deriving `numRecoveryTxs` from the ancestry) but
was never merged — it was superseded by the coarse UTXO-count heuristic
`preflightUnrollFeeInputs` that ships today, which compares only against
the *wallet's* fee budget and never the VTXO's own value vs. dust. This
PR resurrects that idea and extends it with the dust / net-recovered and
economical checks it never had.

## Approach

Driving an exit to completion draws fees from two distinct purses:

- **Wallet-funded** — a CPFP child per recovery transaction, paid from a
  confirmed on-chain wallet UTXO. Never touches the VTXO value.
- **VTXO-funded** — the final sweep pays its own fee out of the VTXO
  value, so the amount that lands back in the wallet is
  `vtxo.Amount - sweepFee`.

A new pure, IO-free model (`unroll/feasibility.go`) folds both into one
verdict via `AssessExitFeasibility`, with four ordered checks:

1. **`ExitSweepBelowDust`** (impossible) — `net = vtxo - sweepFee` falls
   below the **network** P2TR dust limit, so the sweep tx can't relay.
   This is the #608 case and subsumes the plain dust check. The floor is
   network policy, **not** an operator-advertised value, so the verdict
   does not depend on the operator being reachable — a unilateral exit is
   precisely the path used when it isn't.
2. **`ExitUneconomical`** — total recovery cost (CPFP + sweep) exceeds a
   configurable fraction of the VTXO value. Default 100%, i.e. block only
   outright loss-making exits; merely-expensive-but-net-positive
   recoveries are left to the user.
3. **`ExitWalletUnderfunded`** — wallet can't cover the CPFP fees.
4. **`ExitWalletTooFewInputs`** — too few distinct fee inputs (one per
   ancestry path).

`RecoveryTxCount(desc)` derives the recovery-transaction and ancestry-
path counts from the descriptor (tree-path tx count + OOR `ChainDepth`).

`darepod`'s `preflightUnrollFeeInputs` is replaced by
`preflightUnrollFeasibility`, which gathers the inputs and maps each
infeasible verdict to a `FailedPrecondition` error naming the concrete
figures and pointing at cooperative leave.

All cost-model sizes, the dust floor, and the economical threshold are
configurable with documented defaults.

## Relationship to other work

Complementary to #658 (recover a VTXO after a failed no-broadcast unroll
— the #602 cascade): this PR is the admission gate that stops the doomed
exit from being admitted in the first place; #658 is the safety net if
one slips through.

## Reproduction note

The issue's original recipe (OOR-send 1 sat to self) is now blocked by
#606 (`validateOORRecipientDust`), so a full itest would need a sub-dust
VTXO seeded another way. The logic is covered by unit tests instead:
table-driven coverage of the pure model (`unroll/feasibility_test.go`)
and the verdict→gRPC mapping (`TestUnrollInfeasibleError`).

## Test Plan

- `make unit pkg=./unroll timeout=5m`
- `make unit pkg=./darepod case=TestUnrollInfeasibleError timeout=5m`

### DIFF
--- a/darepod/rpc_server.go
+++ b/darepod/rpc_server.go
@@ -3349,23 +3349,15 @@ func (r *RPCServer) Unroll(ctx context.Context, req *daemonrpc.UnrollRequest) (
 		}
 	}
 
-	// Pre-flight the wallet's CPFP fee budget against the VTXO's
-	// ancestry shape. Each independent ancestry path (cross-
-	// commitment multi-input OOR VTXOs may have several; a round-
-	// direct VTXO has exactly one) corresponds to a distinct CPFP
-	// parent that the unroll needs to fee-bump on chain. TRUC RBF
-	// lets a single parent reuse its own fee input across bumps,
-	// but distinct parents each require their own confirmed wallet
-	// UTXO at child-construction time. If the wallet has fewer
-	// usable UTXOs than ancestry paths, the unroll will admit, the
-	// VTXO actor will transition to UnilateralExitState, but the
-	// chain layer will then thrash forever on
-	// `cpfp fee input unavailable: no confirmed wallet UTXOs
-	// available`. That leaves the VTXO in an exit state the user
-	// can't easily back out of. Fail closed here instead: refuse
-	// admission with InvalidArgument so the caller can fund the
-	// wallet and retry without contaminating state.
-	if err := r.preflightUnrollFeeInputs(
+	// Pre-flight the whole-picture feasibility of the exit before
+	// committing the VTXO to UnilateralExitState. This refuses an exit
+	// that can never succeed (the swept output would be dust and the
+	// sweep tx unrelayable — darepo-client #608), that is economically
+	// irrational (it burns more in fees than the coin is worth), or
+	// that the wallet cannot fund (insufficient balance or too few
+	// distinct CPFP fee inputs). Failing closed here keeps the VTXO out
+	// of an exit state the user can't easily back out of.
+	if err := r.preflightUnrollFeasibility(
 		admissionCtx, desc,
 	); err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


ellemouton:
@claude review this

ellemouton:
Rebased on latest `main` and addressed the review feedback:

- **`RecoveryTxCount` floor of 1 per ancestry path** (Gemini + @claude): each ancestry entry now contributes at least one transaction, so a malformed fragment (nil `TreePath` and zero `TreeDepth`) can't zero out the CPFP-fee estimate and slip a wallet-underfunded exit through. Added a `TestRecoveryTxCount` subtest pinning the floor, plus one covering the previously-untested `TreePath != nil` branch.
- **Error message wording**: `ExitSweepBelowDust` fires only on a strict `net < dust`, so the message now says "below the N sat dust limit" instead of "at or below" (a net exactly at the limit is relayable and passes).

The CI failures on the prior run were: Lint + Static Checks (stale formatting/line-length from a late edit — fixed, `make lint-changed-local` now 0 issues) and the postgres unit-test job, which died with *"the self-hosted runner lost communication with the server"* (infra flake — the sqlite/race/cover jobs ran the same tests and passed; this change is pure Go with no DB).

ellemouton:
merging as this is quite simple

PR #676, opened 2026-06-03
chainbackends: add Esplora package submitter for lnd/neutrino unroll broadcast

ellemouton:
## Update: scope the startup warning to lnd (addresses #678)

Folded in a follow-up commit (`cmd/darepod: Scope no-package-relay warning to lnd backend`) that narrows the missing-submitter startup warning. It previously fired for both `lnd` and `btcwallet`; the `btcwallet` (neutrino SPV) backend relays the zero-fee parent + CPFP child over P2P 1p1c with no submitter, so warning it was a false alarm. The warning now only fires for the `lnd` backend (the one config that can actually strand an exit when backed by bitcoind), and `warnIfNoPackagePath` was refactored into a unit-tested `packageRelayWarning` helper.

### DIFF
--- a/chainbackends/esplorarpc/CLAUDE.md
+++ b/chainbackends/esplorarpc/CLAUDE.md
@@ -1,0 +1,59 @@
+# esplorarpc
+
+## Purpose
+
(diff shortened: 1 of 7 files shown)


litbot-9000:

PR #682, opened 2026-06-03
swapwallet: hide exact-input funding rows

### LINKED ISSUE #669: Refunded payment shows up as negative balance
I recently attempted an offchain send with `darepo-cli send --offchain`. The payment ultimately failed, and I was refunded.

In the activity log, this looks like this:

```
LAST UPDATE              KIND     STATUS    AMOUNT  FEE  PHASE             ID                                                                NOTE
2026-06-02 19:19:48.000  SEND     PENDING   -210    0    payment_detected  ledger-14                                                         
2026-06-02 19:19:48.000  SEND     PENDING   -210    0    payment_detected  ledger-13                                                         
2026-06-02 19:19:47.000  SEND     COMPLETE  -210    0    refunded          a5a1087b0d340d83f7dcc726a313ba99465ab48d711dbfe9bf38192d0185d7c0  
2026-06-02 19:19:47.000  SEND     COMPLETE  -210    0    refunded          92b3887b45ed1828913b19fe41678755f4169d9950640404d951448e1b3b22ef  
2026-06-02 19:19:47.000  SEND     PENDING   -42000  0    payment_detected  ledger-12                                                         
2026-06-02 19:19:46.000  SEND     COMPLETE  -42000  0    refunded          f57cb0e8f3806481ea901a34d70400a925678748ee119ad6ecdd26acde0cc69a
```

My questions about this:

- As the outgoing payments failed, should they still show up as `pending`?
- Should the `complete` refund be shown with a negative sign?
- Should the refund be marked as `send`?

In my expectation:
- An offchain payment should time out fast
- A refund is a received payment, not a sent payment
- The balance should be returned back to you, either by setting the value of the outgoing payment to zero, or by "sending back" an equivalent amount. E.g. `SEND FAILED -100` & `RECV COMPLETE +100`

bhandras:
## Summary

Fixes #669 by hiding pay-swap funding OOR ledger rows when the selected input exactly matches the vHTLC amount and therefore produces no wallet change row.

The issue screenshot already has the correct user-facing swap rows: the payment attempts are represented by `SEND ... refunded` entries. We do not need to add a separate refund/receive activity row. The confusing part was the extra `ledger-*` rows leaking through as pending sends.

## Root Cause

The existing activity-feed filter hid pay-swap funding OOR rows when the same OOR session also had a wallet receive/change row. That works for the usual case:

- OOR send spends a larger VTXO.
- OOR receive returns wallet change.
- The net outgoing delta equals the swap amount.
- The internal funding row is hidden.

But exact-input funding has no wallet change row:

- OOR send amount equals the swap amount.
- No paired receive/change row exists.
- The old filter skipped the session because `received == 0`.
- The internal funding row leaked into activity as `ledger-N` / `payment_detected`.

## Fix

When there is no receive/change row, still hide the OOR send row if structured swap metadata proves that:

- the row's OOR session matches a pay swap funding session, and
- the row amount matches that swap amount.

Ordinary user OOR sends remain visible because this path still requires the swap funding session correlation; it does not hide by amount alone.

## Validation

- Added a regression test that first reproduced the leaked shape: one refunded swap row plus a pending `ledger-*` send for the same amount.
- Verified the activity feed now shows only the refunded swap row.

Commands run locally:

```bash
make fmt-changed-check base=origin/main
make commitmsg-lint commit=HEAD
go test -tags='dev walletdkrpc swapruntime nolog' ./swapwallet -count=1
```

### DIFF
--- a/swapwallet/history.go
+++ b/swapwallet/history.go
@@ -568,10 +568,9 @@ type swapOORCorrelations struct {
 }
 
 // swapOORCorrelationsFromSwaps indexes swap session ids by the internal OOR leg
-// they explain. Pay swaps hide their funding input only when the same funding
-// session returns change and the delta equals the swap amount. Receive swaps
-// hide their claim input only when the same claim session materializes a wallet
-// VTXO.
+// they explain. Pay swaps hide their funding input when the session plus net
+// outgoing amount matches the swap amount. Receive swaps hide their claim input
+// only when the same claim session materializes a wallet VTXO.
 func swapOORCorrelationsFromSwaps(
 	swaps []*swapclientrpc.SwapSummary) swapOORCorrelations {
 
@@ -667,8 +666,13 @@ func projectOORLedgerActivity(rows []*daemonrpc.TransactionHistoryEntry,
 			continue
 		}
 
+		amounts := correlations.payAmountsByFundingSession[session]
 		received := receivedBySession[session]
 		if received == 0 {
+			if consumePayFundingAmount(amounts, row.GetAmountSat()) {
+				projection.hidden[row.GetEntryId()] = struct{}{}
+			}
+
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


bhandras:
@claude review

Issue #605, opened 2026-05-28
swapwallet: concurrent recv calls serialize end-to-end (~8x latency under 5-way burst)


jtobin:
## Environment

- `darepo-client`: `cd035b84` (also reproducible on current main `32b33fb5`)
- Network: signet
- Wallet backend: `lnd`
- Ark server: `arkd-signet.testnet.lightningcluster.com:443`
- Swap server: `swapd-signet.testnet.lightningcluster.com:443`

## Summary

`darepocli recv --amt N --memo M` (the offchain swap-in invoice-creation path) serializes end-to-end across concurrent callers. Five concurrent calls each take ~8.3× the single-call baseline, and concurrent throughput is **lower than serial throughput**. A lock somewhere in the invoice mint (LND `addinvoice` + swap-row insert + key allocation) is held across the whole path.

## Expected Behavior

N concurrent `recv` calls should finish in roughly `max(latency_i)` wall-time, or at worst match serial throughput. At minimum, the LND `addinvoice` step shouldn't be globally serialized.

## Actual Behavior

Measured on signet node a with a swapdk-sandbox harness (`harness.sh darepocli recv ...`).

**Baseline (single call):**
```text
single recv elapsed=0.552 s
```

**Five calls concurrent (`&`, `wait`):**
```text
recv-1 amt=5100 elapsed=4.596 s
recv-2 amt=5200 elapsed=4.587 s
recv-3 amt=5300 elapsed=4.586 s
recv-4 amt=5400 elapsed=4.592 s
recv-5 amt=5500 elapsed=4.580 s
TOTAL_WALL=4.608 s
```

Each call ~8.3× baseline. All 5 returned within ~16 ms of each other — strongly suggests a shared mutex / actor inbox releasing them together.

**Same five calls in series:**
```text
serial-1..5 elapsed=0.52..0.57 s
TOTAL_SERIAL_WALL=2.836 s
```

So concurrent throughput (5 / 4.6 s ≈ **1.09 recv/s**) is worse than serial throughput (5 / 2.8 s ≈ **1.79 recv/s**).

**Functional correctness OK:**
- 5 distinct payment_hashes (no collision).
- All 5 visible in `darepocli activity --pending`, phase `waiting_for_payment`, correct amounts (5100, 5200, 5300, 5400, 5500).
- All 5 visible in `darepocli swap list`, `state=SWAP_STATE_INVOICE_CREATED`, `pending: true`, matching invoices.
- No timeouts, no errors.

[...]

PR #693, opened 2026-06-05
oor: per-session durable actors for out-of-round transfers

Roasbeef:
In this PR, we replace the single global OOR actor with one durable actor
per OOR session. The old design funneled every out-of-round transfer on the
daemon through a single actor turn, so concurrent receives serialized behind
each other and the whole subsystem moved at the speed of its slowest
session. Each session now gets its own durable actor with its own mailbox,
spawned and reaped by a registry coordinator, with session state snapshotted
through a bridge so a crash at any point resumes exactly where the protocol
left off.

The shape of the change: a new `oor_session_registry` table and store, the
per-session actor and its snapshot bridge, the registry coordinator that
owns spawn/lookup/reap, the incoming receive flow rewired onto the
per-session actor, and the daemon cutover. With the cutover in place we
delete the legacy global actor and the signing-effect actor outright. The
registry also gains a durable inbound mailbox, server pushes route straight
to live session actors, and admission asks settle from the child via
promise handoff rather than bouncing through an extra turn. Systest gains a
concurrent incoming-receive test that exercises N sessions making progress
at once.

Two lifecycle bugs that surfaced while stress testing the new model are
fixed here as well rather than in a follow-up, since both live in code this
PR introduces: materialized-VTXO notifications now deliver on a
daemon-owned context instead of a (cancelable) turn context, and deferred
self-transfer hints are parked and redriven on terminal reap instead of
looping through redelivery.

On the high-contention stress shape (2 clients + operator in one process,
concurrency 16, sqlite), this refactor on its own takes the system from
0.51/s at ~60% success to ~1/s at ~90%. The rest of the gap to ~14/s at
100% comes from the optimization train stacked on top of this PR:

1. this PR: the per-session refactor itself
2. #726: db storage knobs (commit durability, fullfsync)
3. #727: durable actor extensions (leaseless single-worker consume, outbox fold)
4. #728: VTXO listing levers
5. #729: db transaction observability
6. #730: serverconn transactional ingress fold

changes. The server-side counterpart of this PR is darepo#538.


Roasbeef:
@claude review this

levmi:
**Triage linkage check** — heads-up that the `#605` reference here looks like a mislink. #605 is about end-to-end serialization of concurrent **`recv`** (swap-in invoice mint) calls, whereas this PR refactors the **OOR** (out-of-round transfer) subsystem and doesn't touch the `recv`/swap invoice path or its systest.

Recommend **not** treating this as a fix for #605 (i.e. no `Closes #605`) — keep #605 open and tracked under the actual `recv`-serialization work. Happy to be corrected if #605's scope was meant to include the OOR path.

litbot-9000:

Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #694, opened 2026-06-05
serverconn: send egress events through a lock-free worker pool

Roasbeef:
In this PR, we make the server connection's egress sender concurrent.
Today every outbound event the round and out-of-round actors send to the
operator goes through a single durable actor that holds the SQLite writer
across the whole gRPC `Edge.Send`, so sends serialize behind one another
and behind every other writer in the daemon. We move egress onto the
durable actor's Read/Commit path so the send runs with no writer held,
then run it as a pool of competing-consumer workers so independent sends
actually proceed in parallel.

This is the egress half of the work; a companion PR ports the
out-of-round system onto the same pooled, lock-free path.

## The writer held across the wire

The egress actor ran on the classic durable-actor path. Because the
delivery store is transaction-aware, the runtime wrapped the entire
`Receive` in one write transaction (`processInTransaction`), and
`Receive` is where the outbound `Edge.Send` happens. So the global
SQLite writer lock was held for the full network round-trip to the
operator's mailbox. SQLite has a single writer, so that one in-flight
send blocked every other writer in the system, and egress itself could
only ever have one message in flight.

The egress handlers don't write any domain state. They build an envelope
and send it. That makes them a clean fit for the Read/Commit
(`TxBehavior`) execution path added in #688: the slow side effect (the
send) runs outside any transaction, and a short, lease-fenced `Commit`
folds in only the framework's ack and dedup mark. After the conversion
the writer is free during the send and is taken only for the
sub-millisecond bookkeeping at the end.

## A worker pool over one mailbox

Freeing the writer is necessary but not sufficient: a single drain loop
still sends one message at a time. So we add a `NumWorkers` knob to
`DurableActor`. The durable mailbox already leases work via
`LeaseNextMailboxMessage`, keyed only by mailbox id, which is textbook
competing-consumers, so N worker loops over the same mailbox each lease
distinct messages and the lease fence guarantees no message is processed
twice. A supervisor goroutine joins the workers and runs teardown once,
so the actor's `done` / WaitGroup / `Stoppable` semantics are unchanged
regardless of worker count. At `NumWorkers == 1` it's the old single
loop, so every other durable actor is untouched.

`serverconn` opts in via `ConnectorConfig.EgressWorkers` (default four).
Per-session ordering survives the fan-out for free: a
`SendClientEventRequest` already carries its inner message's
`CorrelationKey` (e.g. `round/<id>`, `oor/<session>`), and the same
lease query enforces per-correlation-key FIFO, so same-session events
stay strictly ordered across workers while distinct sessions send in
parallel.

Egress stays at-least-once. The send was never atomic with the ack, even
on the old path, so a crash or a lost lease between a successful send and
its `Commit` redelivers and re-sends; the server absorbs the duplicate
via the stable `MsgId` / `IdempotencyKey`. The low-latency unary facade
is unchanged (it already calls `Edge.Send` directly), and the single
ingress puller is untouched.

## Showing it's faster

The benchmark drives a durable actor whose handler sleeps to model a
network send (20ms), against a sqlite store configured exactly like
production (WAL, `busy_timeout`, `_txlock=immediate`):

```
BenchmarkEgressClassicW1-16        400   29434782 ns/op   33.97 msgs/sec
BenchmarkEgressClassicW4-16        400   29706981 ns/op   33.66 msgs/sec
BenchmarkEgressReadCommitW1-16     400   29651921 ns/op   33.72 msgs/sec
BenchmarkEgressReadCommitW4-16     400   10878844 ns/op   91.92 msgs/sec
```

The story is in the W4 rows. On the classic path four workers buy
nothing: the writer is held across each send, so they serialize anyway
(33.66 vs 33.97). On the Read/Commit path four workers send concurrently
and throughput roughly triples (91.92). It's ~2.7x rather than 4x because
the per-commit fsync under `synchronous=full` is the remaining floor,
which is the honest production cost.

## Testing

The new tests run under the race detector. There's a property-based test
(rapid) that every message is processed exactly once across random worker
and message counts, a test that a single worker stays strictly
sequential, and a composition test against the real sqlite store proving
same-key FIFO holds while cross-key work interleaves under four workers.
See each commit message for the per-change detail.


bhandras (approved):
LGTM, love it!

bhandras:
Looks like the failures are due to self-hosted runner instability.

Roasbeef:
@claude review this

PR #696, opened 2026-06-05
darepod: support multi-recipient SendOOR

bhandras:
## Summary

This extends the existing `SendOOR` RPC instead of adding a second OOR send method. The OOR actor already accepts a recipient slice, so the daemon now exposes that shape directly at the RPC boundary.

`SendOORRequest` now has one canonical `recipients` field. The daemon validates each recipient, selects wallet inputs for the aggregate amount, and submits one OOR transfer with all requested outputs. The response returns `recipient_outpoints` in request order so callers can map each requested recipient to the VTXO it received.

The SDK still keeps single-recipient convenience helpers for existing swap code. Those helpers now build a one-element `recipients` list and read the first returned outpoint.

Custom-input OOR sends stay single-recipient for now. That path carries specialized signing material, so widening it should be a separate protocol change with its own tests.

## Testing

- `make unit pkg=./darepod`
- `make unit pkg=./sdk/ark`
- `make unit pkg=./rpc/oorpb`
- `go test ./cmd/darepocli/darepoclicommands/devrpc ./cmd/darepocli/darepoclicommands`
- `go test -tags "systest" -v ./systest/... -run TestSendOORMultipleRecipientsEndToEnd -timeout 10m`
- `make lint-changed-local`

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_ark_send.go
+++ b/cmd/darepocli/darepoclicommands/cmd_ark_send.go
@@ -210,7 +210,7 @@ func sendOOR(cmd *cobra.Command, _ []string) error {
 			return err
 		}
 
-		req.Recipient = recipient
+		req.Recipients = []*daemonrpc.Output{recipient}
 		req.DryRun = dryRun
 		req.IdempotencyKey = idempotencyKey
 
--- a/cmd/darepocli/darepoclicommands/cmd_mcp.go
+++ b/cmd/darepocli/darepoclicommands/cmd_mcp.go
@@ -395,7 +395,7 @@ func registerMCPSendTools(s *mcp.Server, client daemonrpc.DaemonServiceClient) {
 
 		resp, err := client.SendOOR(
 			ctx, &daemonrpc.SendOORRequest{
-				Recipient:      recipient,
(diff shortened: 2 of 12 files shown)


bhandras:
@claude review

bhandras:
@claude re-review please.

Context since your last review:

- Normalized SendOOR recipient amount validation so the single-recipient positive-amount error now uses the same indexed recipient wording as the other recipient validation errors.
- Updated the `recipient_outpoints` proto documentation to state that individual entries may be empty when one recipient outpoint cannot be resolved, and updated the matching generated Go comment. Full `make rpc` is Docker-backed in this repo and was blocked locally because Docker was unavailable, but this doc-only generated delta is aligned.
- Added an inline comment explaining why the request-recipient copy must remain change-free before appending the change output: response outpoint resolution maps caller-requested recipients into the finalized Ark transaction.
- Folded the fixes into the implementation commit; both PR commits are signed.

Local validation after folding:

```text
go test ./darepod -run 'TestSendOOR|TestAppendOORChangeRecipient' -count=1\nmake lint-changed-local\nmake commitmsg-lint range="origin/main..HEAD"\n```\n\nNew head: `b33bf42b`. GitHub CI has restarted on the folded branch.

Issue #672, opened 2026-06-02
unroll: externally-driven PSBT broadcast mode for unilateral exits


Roasbeef:
## Motivation

Today an unroll (unilateral exit) is fully self-driving. The durable
per-target actor in `unroll/` assembles the recovery proof, broadcasts each
proof node in dependency order, waits for CSV maturity on the confirmed
target, then builds + signs the timeout-path sweep and broadcasts it. All
broadcast and fee-bumping goes through the shared `txconfirm` actor, which
does BIP-431/TRUC anchor CPFP using the daemon's _own_ wallet UTXOs via
`CPFPBroadcaster`.

That's the right default, but it bakes in two assumptions that don't always
hold: that the daemon's hot wallet is the right place to source fee inputs,
and that the daemon should own broadcast + package submission. For some
deployments we want an external program to drive the costly part of the
exit. The external program may have a dedicated fee wallet, a coin-selection
policy of its own, an HSM in the signing path, or a relationship with a
miner / package-relay endpoint that the daemon doesn't have. It wants to
anchor the exit down with its own inputs and post the package itself.

In this issue, we propose a lower-level, PSBT-based RPC surface on
`DaemonService` that lets an external program drive the signing, anchoring,
and broadcast of an unroll, while the daemon keeps doing what it's uniquely
good at: assembling the proof, planning the dependency order, tracking
confirmations, advancing the CSV clock, and surfacing the next transaction
that's ready to go on-chain. I.e. the daemon stays the brain, the external
program becomes the hands.

## How unroll works today

A quick map of the moving parts, since the proposal slots into the existing
machine rather than replacing it.

The unit of work is a `recovery.Proof` (`lib/recovery/proof.go`): an
immutable DAG of `Node`s for one target outpoint. Each node carries a
`wire.MsgTx` and a `NodeKind`, one of `Tree`, `Checkpoint`, or `Ark`. These
are the "branch" transactions: the round-tree path, the OOR checkpoint
lineage, and the ark txs that spend checkpoint outputs down toward the
target VTXO. Proof nodes are operator-pre-signed; the actor never signs
them, it only broadcasts. The proof already exposes everything an external
broadcaster needs to fee-bump: `Node.AnchorOutpoint()` returns the BIP-431
ephemeral-anchor outpoint, and `parents`/`children`/`layers` give the
dependency structure.

The durable FSM (`unroll/fsm_types.go`, `unroll/fsm_logic.go`) walks:

```
[...]

PR #697, opened 2026-06-05
walletdk: add exit plan and wallet sweep APIs

sputn1ck:
## Summary

This adds two wallet-facing APIs for unilateral exits:

- `GetExitPlan` tells a caller whether the backing wallet has the confirmed fee UTXOs needed to start an exit. If it does not, the response gives the caller a fresh backing-wallet Bitcoin address and a recommended funding amount.
- `SweepWallet` previews or broadcasts a sweep of leftover confirmed backing-wallet funds to a caller-supplied Bitcoin address.

`Deposit` still returns boarding addresses. The new funding address is only for fee UTXOs that txconfirm can use while unrolling a VTXO.

part of #672.

## CLI walkthrough

I started arktest with Alice's LND wallet unfunded:

```sh
./arktest start \
  --client alice \
  --client-lnd-funds=0 \
  --datadir /tmp/arktest-exitplan-state \
  --artifacts-dir /tmp/arktest-exitplan-artifacts
```

After boarding and confirming one VTXO, Alice had this live outpoint:

```text
ebf1204d8c35e4a7428f43fab5ad36668a474297732779c3b3a0455e446306b1:0
```

Before funding the backing wallet, `GetExitPlan` showed the caller exactly what was missing:

```json
{
  "funding_address": "bcrt1pkccaqzxsecm6ywmcd69qqxvh60h9y5965ggwguhzst9vjlys2ndqmv6vlu",
  "required_confirmations": 1,
  "fee_rate_sat_per_vbyte": "25",
  "required_fee_utxo_count": 1,
  "usable_fee_utxo_count": 0,
  "recommended_utxo_amount_sat": "10000",
  "recommended_total_funding_sat": "10000",
  "funding_shortfall_sat": "10000",
  "can_start": false
}
```

Funding that address with the recommended amount made the exit admissible:

```sh
./arktest --datadir /tmp/arktest-exitplan-state faucet \
  bcrt1pkccaqzxsecm6ywmcd69qqxvh60h9y5965ggwguhzst9vjlys2ndqmv6vlu \
  10000
```

The next plan showed the backing wallet had one usable fee UTXO:

```json
{
  "required_fee_utxo_count": 1,
  "usable_fee_utxo_count": 1,
  "funding_shortfall_sat": "0",
  "can_start": true
}
```

The exit RPC then accepted the job:

```json
{
  "created": true,
  "actor_id": "unroll-ebf1204d8c35e4a7428f43fab5ad36668a474297732779c3b3a0455e446306b1:0"
}
```

Finally, `SweepWallet` found and swept the leftover backing-wallet change:

```json
{
  "total_input_sat": "3100",
  "estimated_fee_sat": "2475",
  "net_amount_sat": "625",
  "fee_rate_sat_per_vbyte": "25",
  "can_broadcast": true,
  "txid": "4c2bcca7d2b7d383295c675698fb22fc84260e346eae7e4738290ea8e2f344a0"
}
```

After mining one block, bitcoind reported the external destination received `0.00000625` BTC.

## Testing

```sh
make rpc
make fmt-changed
make lint-changed-local
go test -tags='walletdkrpc swapruntime' ./darepod ./swapwallet ./sdk/walletdk
```


sputn1ck:
@claude review

sputn1ck:
Addressed the review feedback in `20bc3074`:

- `GetExitPlan` no longer mints a new address on every preview. It caches one backing-wallet funding address per VTXO outpoint for the daemon lifetime and reuses it while callers poll the plan.
- The LND backing-wallet address helper now requests an external P2TR address (`change=false`) for this funding path. I verified the current helper was using `WalletKit.NextAddr`; LND routes that to wallet `NewAddress`, and btcwallet/lwwallet also call `NewAddress`, so repeated calls do advance the address chain rather than returning the same latest unused address.
- `GetExitPlan` now returns `FailedPrecondition` when the VTXO descriptor has no unilateral-exit ancestry, so it cannot report `can_start=true` for a malformed/unexit-able VTXO.
- The devrpc generator now emits a blank `walletdkrpc` import and the generated registry was updated.
- `SweepWallet` rejects negative explicit fee rates and caps positive explicit rates to the daemon unroll fee cap.
- The dust case now reports `sweep amount is dust after fees`.
- The wallet service/proto comments now describe the cached/conditional funding address behavior.

Validation:

```sh
make rpc
go run ./cmd/darepocli/internal/gen-devrpc
make fmt-changed
go test -tags='walletdkrpc swapruntime' ./darepod ./swapwallet ./sdk/walletdk ./cmd/darepocli/darepoclicommands/devrpc
make lint-changed-local
```

sputn1ck:
@claude rereview

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+
+// ExitPlanRequest asks the daemon to preview unilateral-exit funding
+// readiness for a wallet-facing caller.
+type ExitPlanRequest struct {
+	Outpoint   string
+	ConfTarget uint32
+}
+
+// ExitPlanResponse describes the current unroll CPFP fee-input readiness.
+type ExitPlanResponse struct {
```
Roasbeef: I think we should move more of this into the existing sub-systems. Then that way the shim on the rpc/daemon side is relatively small, and we can expose it more easily using the existing SDK/RPC cut outs.

Roasbeef: For example, we can put this within the existing well defined `unroll` package. Would be easier to unit test as well. Systest too, etc.

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+	RequiredFeeUTXOCount       uint32
+	UsableFeeUTXOCount         uint32
+	RecommendedUTXOAmountSat   int64
+	RecommendedTotalFundingSat int64
+	FundingShortfallSat        int64
+	CanStart                   bool
+	ExitJobFound               bool
+	ExitStatus                 daemonrpc.UnrollJobStatus
+	SweepTxid                  string
+	LastError                  string
```
Roasbeef: Why no an actual `error`?

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+	FeeRateSatPerVByte         int64
+	RequiredFeeUTXOCount       uint32
+	UsableFeeUTXOCount         uint32
+	RecommendedUTXOAmountSat   int64
+	RecommendedTotalFundingSat int64
+	FundingShortfallSat        int64
+	CanStart                   bool
+	ExitJobFound               bool
+	ExitStatus                 daemonrpc.UnrollJobStatus
+	SweepTxid                  string
```
Roasbeef: Same here, `chainhash.Hash`.

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+// ExitPlanRequest asks the daemon to preview unilateral-exit funding
+// readiness for a wallet-facing caller.
+type ExitPlanRequest struct {
+	Outpoint   string
+	ConfTarget uint32
+}
+
+// ExitPlanResponse describes the current unroll CPFP fee-input readiness.
+type ExitPlanResponse struct {
+	FundingAddress             string
```
Roasbeef, line 34: `txscrpit.PkScript`?

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+		ctx, outpoint.String(), usable < required,
+	)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "exit plan funding "+
+			"address: %v", err)
+	}
+
+	plan := &ExitPlanResponse{
+		FundingAddress:             fundingAddress,
+		RequiredConfirmations:      1,
```
Roasbeef: Shouldn't hard code a single conf

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+	plan.ExitStatus = statusResp.GetStatus()
+	plan.SweepTxid = statusResp.GetSweepTxid()
+	plan.LastError = statusResp.GetLastError()
+
+	return plan, nil
+}
+
+// SweepWallet previews or broadcasts a normal backing-wallet sweep that
+// excludes boarding outputs.
+func (r *RPCServer) SweepWallet(ctx context.Context, req *SweepWalletRequest) (
```
Roasbeef, line 349: Unclear what the use of this is, we already have `txconfirm` that handles this whole fee estimation, tx creation, broadcast loop, etc.

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+
+	return feeRate
+}
+
+func (s *Server) exitPlanFundingAddress(ctx context.Context, outpoint string,
+	create bool) (string, error) {
+
+	s.exitPlanFundingAddressesMu.Lock()
+	defer s.exitPlanFundingAddressesMu.Unlock()
```
Roasbeef: If this was rolled into the existing `unroll` pkg, wouldn't need this mutex or this state that's actually stored in the server itself (eg: exposing to sdk variants would need to adopt this entire surface area).

darepod/rpc_exit_plan.go
```diff
@@ -0,0 +1,631 @@
...
+		PreviousOutPoint: wire.OutPoint{},
+		Sequence:         wire.MaxTxInSequenceNum - 2,
+	})
+	tx.AddTxOut(&wire.TxOut{
+		Value:    int64(txconfirm.DustLimit),
+		PkScript: make([]byte, 34),
+	})
+	tx.AddTxOut(&wire.TxOut{
+		Value:    0,
+		PkScript: []byte{txscript.OP_TRUE},
```
Roasbeef: Why not the actual anchor script? I think this is even non standard in the current form?

rpc/walletdkrpc/wallet.proto
```diff
@@ -66,6 +66,16 @@ service WalletService {
...
 
+    // GetExitPlan previews unilateral-exit readiness for one VTXO. The
+    // response includes CPFP fee input requirements and, when funding is
+    // needed, a backing-wallet address callers can fund before forced unroll.
+    rpc GetExitPlan (GetExitPlanRequest) returns (GetExitPlanResponse);
+
+    // SweepWallet previews or broadcasts a normal backing-wallet sweep to a
+    // caller-supplied Bitcoin address. This sweeps wallet-managed funds left
+    // after CPFP/change/unroll proceeds and does not sweep boarding outputs.
+    rpc SweepWallet (SweepWalletRequest) returns (SweepWalletResponse);
```
Roasbeef, line 77: Wasn't sure what this did at first from the name, don't have a good alternative suggestion here yet though

Roasbeef (dismissed):
In terms of architecture, i think this can be better integrated into the existing sub systems and abstractions of the demon. As an example, for the sweep process, we already have the wallet actor, and the entire `txconfirm` package. Same for the unroll process as well. AFAICT, it also isn't locking any inputs that it might be using to sweep, which can conflict with existing `unroll` instances that may already be running.

darepod/rpc_exit_plan.go
sputn1ck: Moved the pure exit funding planner into `unroll` (`PlanExitFunding`, `ExitFundingSnapshot`, and the related tests). The daemon RPC now just loads the VTXO descriptor, wallet snapshot, and fee estimate, then projects the `unroll` plan into the wallet RPC response.

darepod/rpc_exit_plan.go
sputn1ck: Done. The daemon-side response now carries an `error`; it is converted to the protobuf string only at the wallet RPC boundary.

darepod/rpc_exit_plan.go
sputn1ck: Done. The daemon-side txid fields now use `*chainhash.Hash`; conversion to string only happens when projecting into walletdkrpc.

darepod/rpc_exit_plan.go
sputn1ck: Done. The sweep destination is parsed into `txscript.PkScript` internally and converted back to script bytes only where the tx builder/estimator needs bytes.

darepod/rpc_exit_plan.go
sputn1ck: Done. The literal one-conf value moved to `unroll.RequiredFeeInputConfirmations`, and both `GetExitPlan` and the wallet snapshot use that shared constant.

darepod/rpc_exit_plan.go
sputn1ck: Clarified this path by keeping it scoped to a normal backing-wallet sweep, not txconfirm CPFP package management. It now locks all selected wallet inputs before signing/broadcasting, so it cannot race active `txconfirm`/unroll fee-input leases.

darepod/rpc_exit_plan.go
sputn1ck: Done. The address reuse policy moved into `unroll.ExitFundingAddressBook`; the server just owns an instance and asks it for an address.

darepod/rpc_exit_plan.go
sputn1ck: Done. The synthetic parent tx/package-fee estimate was removed entirely. `GetExitPlan` now uses the existing unroll feasibility model, so there is no fake/non-standard anchor script in this path.

rpc/walletdkrpc/wallet.proto
sputn1ck: Kept the API name for now since this is the wallet-facing preview before `Exit`/forced unroll, and there was not a clearer replacement suggested. The RPC doc now describes that it previews unilateral-exit readiness and returns backing-wallet funding requirements.

darepod/rpc_exit_plan.go
Roasbeef: Can remove this commit now?  (very first one)

darepod/rpc_exit_plan.go
```diff
@@ -28,6 +29,13 @@ const (
 	walletSweepDustFailureReason = "sweep amount is dust after fees"
 )
 
+var walletSweepLockID = func() walletcore.LockID {
```
Roasbeef: Can just declare it as a var?

darepod/rpc_exit_plan.go
```diff
@@ -506,7 +464,49 @@ func walletSweepPreview(utxos []*wallet.Utxo, destScript []byte,
 	return resp
 }
 
-func estimateWalletSweepVSize(utxos []*wallet.Utxo, destScript []byte) int {
+func lockWalletSweepInputs(ctx context.Context, leaser walletcore.OutputLeaser,
```
Roasbeef: Still think this should go through the wallet and txconfirm pipeline.

unroll/exit_plan.go
```diff
@@ -0,0 +1,165 @@
...
+
+// ExitFundingPlan is the user-facing funding projection derived from the
+// unroll feasibility model.
+type ExitFundingPlan struct {
+	Feasibility ExitFeasibility
+
+	RequiredConfirmations      uint32
+	RecommendedUTXOAmountSat   btcutil.Amount
+	RecommendedTotalFundingSat btcutil.Amount
+	FundingShortfallSat        btcutil.Amount
```
Roasbeef, line 37: Missing `godoc` above

unroll/exit_plan.go
```diff
@@ -0,0 +1,165 @@
...
+// RecommendedExitFeeInputAmount derives the per-funding-output suggestion from
+// the full unroll feasibility cost breakdown.
+func RecommendedExitFeeInputAmount(verdict ExitFeasibility) btcutil.Amount {
+	if verdict.RequiredWalletInputs <= 0 {
+		return 0
+	}
+
+	perInputFee := ceilDivAmount(
+		verdict.CPFPFeeTotalSat,
+		btcutil.Amount(verdict.RequiredWalletInputs),
```
Roasbeef, line 84: Does this factor in the fan out? What really matters is that if we have `N` vtxos to sweep and we have and each needs a _total fee_ of `M`, then we have `M*N + f` in UTXOs, where `f` is the fee for the fanout.. With the fan out, doesn't matter the size of them just that we have enough funds.

unroll/exit_plan.go
```diff
@@ -0,0 +1,165 @@
...
+func ceilDivAmount(a, b btcutil.Amount) btcutil.Amount {
+	if b <= 0 {
+		return 0
+	}
+
+	return (a + b - 1) / b
+}
+
+// ExitFundingAddressBook reuses a funding address for each target outpoint so
+// polling a plan does not advance the backing wallet address index.
```
Roasbeef, line 135: > polling a plan does not advance the backing wallet address index.

Seems if it's just a plan, then it can not hit the backend at all?

Roasbeef: Or perhaps I misunderstand here

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+	// when a child sees exactly one usable wallet UTXO. Without this, the
+	// first concurrent child can lease that UTXO before sibling demand is
+	// visible, causing later children to fail instead of sharing a fanout.
+	feeInputFanoutMinPreemptiveOutputs = 3
+)
+
+// FeeInputFanoutCoordinator serializes backing-wallet fanout for all child
+// unroll actors owned by a registry.
+type FeeInputFanoutCoordinator struct {
+	mu sync.Mutex
```
Roasbeef: If the unroll child actors are communicating with this, then we can make it an actor and avoid this mutex. Another thing to factor in is that the unroll child actors are all durable, so they'll write requests, retry responses, etc. Still getting through the diff, but if we make this durable, then they'll be able to simplify some downstream logic, as if they need UTXOs for an unroll, they send a message to this, then just wait until the resp comes back.

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+func (c *FeeInputFanoutCoordinator) PlanFeeInputs(ctx context.Context,
+	ready []unrollplan.TxFrontier) (FeeInputPlan, error) {
+
+	return c.PlanFeeInputsForChild(ctx, "", ready)
+}
+
+// PlanFeeInputsForChild records one child actor's current ready-frontier demand
+// and returns the aggregate registry-level fanout plan.
+func (c *FeeInputFanoutCoordinator) PlanFeeInputsForChild(ctx context.Context,
+	childID string, ready []unrollplan.TxFrontier) (FeeInputPlan, error) {
```
Roasbeef: What's the string ref? Should make a new type/alias

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+	required := len(ready)
+	plan := FeeInputPlan{
+		RequiredFeeInputsNow: required,
+	}
+	if c == nil || required == 0 {
+		return plan, nil
+	}
+
+	c.mu.Lock()
+	if c.wallet == nil {
```
Roasbeef: Is this even an acceptable configuration? No wallet configured at all. If not, then we should check this as init time and return an error.

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+			if !timer.Stop() {
+				select {
+				case <-timer.C:
+				default:
+				}
+			}
+
+			return plan, ctx.Err()
+
+		case <-timer.C:
```
Roasbeef: Re above, this can use the time out actor instead here. Then it just gets a message when the time has elapsed. Otherwise, we'll just end up blocking here inline, when we could be processing other requests.

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+		plan.FanoutOutputsNeeded = 0
+	}
+
+	return plan, nil
+}
+
+// fanoutOutputConfirmed reports whether the backing wallet now sees at least
+// one confirmed output created by the pending fanout transaction.
+func fanoutOutputConfirmed(utxos []*walletcore.Utxo, txid chainhash.Hash) bool {
+	for _, utxo := range utxos {
```
Roasbeef: Can make a set then just check contains (`fn.Set`). Would save ~5 lines

unroll/fee_input_fanout.go
Roasbeef: Getting through this a bit more, what if we just moved this responsibility into the `txconfirm` layer? Its job is to make sure transactions confirm. It'll do CPFP, RBF, etc to make sure that happens. That way we also centralize some of the UTXO management a bit. Today the only thing that actually sends transactions at the core layer is that sub system.

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+			"nil")
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if pending := c.pending; pending.IsSome() {
+		value := pending.UnsafeFromSome()
+
+		return value.txid, append([]byte(nil), value.pkScript...), nil
```
Roasbeef: Why append nil to the pk script then return that? Short hand for copy?

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+
+	tx, watchScript, err := buildFeeInputFanoutTx(
+		ctx, c.wallet, plan.FanoutOutputsNeeded,
+		plan.RecommendedFanoutOutputAmountSat, feeRate,
+	)
+	if err != nil {
+		return chainhash.Hash{}, nil, err
+	}
+
+	resp, err := c.chainSource.Ask(
```
Roasbeef: Why direct here vs txconfirm?

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+	selected, fee, changeScript, err := selectFeeInputFanoutInputs(
+		ctx, signer, utxos, outputs, outputAmount, feeRate,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	tx := wire.NewMsgTx(feeInputFanoutTxVersion)
+	for _, utxo := range selected {
+		tx.AddTxIn(&wire.TxIn{
```
Roasbeef: Do we risk accidental double spends here? `txconfirm` uses the PSBT API which handles locking under the hood, if we move this duty further there then we can re-use all that protection.

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+
+		selected = append(selected, utxo)
+		feeNoChange := btcutil.Amount(
+			estimateFeeInputFanoutVSize(
+				selected, outputs, nil,
+			),
+		) * btcutil.Amount(feeRate)
+		neededNoChange := btcutil.Amount(outputs)*outputAmount +
+			feeNoChange
+		total := selectedTotal(selected)
```
Roasbeef: New diff in this pkg can use some comments and also a bit of re folding here to make things easier to scan, eg: 
```
feeNoChange := btcutil.Amount(
			estimateFeeInputFanoutVSize(
				selected, outputs, nil,
			),
		) * btcutil.Amount(feeRate)
```

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+		if total == neededNoChange ||
+			total-neededNoChange < txconfirm.DustLimit {
+
+			if total >= neededNoChange {
+				return selected, feeNoChange, changeScript, nil
+			}
+		}
+
+		feeWithChange := btcutil.Amount(
```
Roasbeef: We def have a coin selection loop elsewhere. No reason to add yet another instance.

unroll/fee_input_fanout.go
```diff
@@ -0,0 +1,674 @@
...
+	return total
+}
+
+// estimateFeeInputFanoutVSize estimates the fanout transaction vsize.
+func estimateFeeInputFanoutVSize(inputs []*wallet.Utxo, outputs int,
+	changeScript []byte) int64 {
+
+	var est input.TxWeightEstimator
+	if len(inputs) == 0 {
+		est.AddTaprootKeySpendInput(txscript.SigHashDefault)
```
Roasbeef: Why zero input assumes keyspend?

unroll/fee_input_fanout.go
Roasbeef: If we just use PSBT funding, then that's all taken care of for us.

unroll/fsm_logic.go
```diff
@@ -310,6 +311,51 @@ func deriveStateTransition(_ context.Context, job *JobState, env *Environment,
...
+			feePlanner := env.FeeInputPlanner
+			if feePlanner != nil {
+				feePlan, err := feePlanner.PlanFeeInputs(
+					ctx, ready,
+				)
+				if err != nil {
+					return nil, err
+				}
+
+				shortfall := feePlan.FanoutFundingShortfallSat
```
Roasbeef: Can move into a new function here to save the indentation.

unroll/fsm_logic.go
```diff
@@ -310,6 +311,51 @@ func deriveStateTransition(_ context.Context, job *JobState, env *Environment,
...
+				if shortfall > 0 {
+					failed := job.Copy()
+					failed.FailReason =
+						feeInputFanoutShortfallReason(
+							shortfall,
+						)
+
+					return transitionWithOutbox(
+						&Failed{
```
Roasbeef: Hmm I don't think this should permanently fail the unroll. Instead, seems like this is the scenario where we need to log/alert/notify the user.

unroll/fsm_types.go
```diff
@@ -247,6 +267,21 @@ type EnsureReadyTransactions struct {
 // outboxEventSealed marks EnsureReadyTransactions as an outbox event.
 func (o *EnsureReadyTransactions) outboxEventSealed() {}
 
+// RequestFeeInputFanout asks the actor boundary to create or wait for a
+// backing-wallet fanout transaction before submitting the ready proof frontier.
+type RequestFeeInputFanout struct {
```
Roasbeef: Flow wise, why should the unroll actor get this message to fan out? Instead during the unroll process, it may realize that it needs to fan out, then execute that via w/e means. This is a reactive model. 

Or is the idea that we go with a proactive model where we fan out ahead of time?

Roasbeef: Ah nvm this is an outbox message, makes sense!

unroll/fsm_logic.go
```diff
@@ -310,6 +311,51 @@ func deriveStateTransition(_ context.Context, job *JobState, env *Environment,
...
+				if feePlan.NeedsFanout() {
+					outbox = append(
+						outbox, &RequestFeeInputFanout{
+							Txids: readyTxids,
+							Plan:  feePlan,
+						},
+					)
+
+					return transitionWithOutbox(
+						&AwaitingFeeInputFanout{
```
Roasbeef: So if we do it at this level (instead of say, `txconfirm`), how do we make sure that we don't run into race conditions here? Consider we fan out two inputs for vtxos, but then fees rise, and one vtxo ends up taking both inputs to cover for them.

Roasbeef:
## Centralizing fee-input/UTXO management in `txconfirm`

Spent some time on the fanout coordinator and I think the fee-input
provisioning wants to live one layer down, inside `txconfirm`, rather than as
a registry-scoped coordinator above the unroll FSM. Sketching it here.

### The gap today

The race I keep coming back to isn't a double-spend. `txconfirm` already
prevents two CPFP parents from grabbing the same wallet UTXO: invariant #3 in
[`doc.go`](../blob/kon/exit-plan-wallet-sweep/txconfirm/doc.go) plus
`reserveFeeInput` reserve each input to a parent txid, lease it under
`txconfirmLockID`, and hold it across blocks until eviction.

The real gap is **ownership across the fanout boundary**. The coordinator's
`buildFeeInputFanoutTx` selects inputs and broadcasts a fanout, but never
leases those inputs and never reserves the resulting outputs to anything. Once
the N outputs confirm, they're just generic confirmed UTXOs in the shared
wallet pool. The actual consumption happens later, inside `txconfirm`, when
each unroll parent's CPFP child calls `SelectFeeInput` and reserves. But
`txconfirm` has no idea output X was minted *for* VTXO A's exit, so VTXO B's
child can legitimately reserve it. No double-spend, but A is starved and
re-fanouts.

So we have two independent selection systems sharing one wallet pool with no
shared ledger. The `demandByID` map, the 500ms
`feeInputFanoutDemandCoalesceDelay`, and `needsPreemptiveSingleUTXOFanout` are
all reconstructing, from outside, aggregate state the single `txconfirm` actor
already holds natively. That's the tell that it's in the wrong place.

### Why `txconfirm` is the right home

`txconfirm` is the one actor that *both* owns the authoritative reservation
ledger (`parentStates`) *and* does the real fee-input consumption. It's a
single actor that serializes all access to the non-concurrency-safe
`CPFPBroadcaster`, so it sees the whole tracked-tx map at once. Fold the
provisioning in and there's one ledger from mint to spend.

The hook is already there: a missing fee input surfaces as
`ErrCPFPFeeInputUnavailable`, which is *transient*. The FSM self-loops in
`Broadcasting` and retries every fee-bump interval, never failing. That is
already "the force-close just waits as normal." So the unroll side needs
nothing new: it submits its ready proof tx and waits. We delete
`AwaitingFeeInputFanout`, `RequestFeeInputFanout`, `FeeInputsAvailableMsg`, the
`FeeInputPlanner` interface, and the whole coordinator. The unroll FSM goes
back to "submit -> wait for confirm," which is the shape it had before this PR.

### What it looks like

A new internal step in the broadcaster, `ensureFeeInputSupply`. When selection
finds no usable unreserved UTXO for parent P:

1. `deficit = (# tracked parents needing a distinct fee input) - (# usable
   unreserved confirmed UTXOs)`. The single actor sees the full tracked map,
   so this is exact and synchronous. No coalesce timer, no demand map, no
   preemptive-single-UTXO heuristic.
2. If `deficit > 0` and no fanout is pending: build a fanout splitting a large
   UTXO into `deficit` (+ buffer) outputs, **lease its inputs under
   `txconfirmLockID`**, broadcast, and record a pending-fanout marker whose
   **predicted output outpoints are pre-reserved round-robin to the blocked
   parents** in the same `parentStates` ledger. That pre-reservation is the
   pinning the current design lacks.
3. Return `ErrCPFPFeeInputUnavailable` so each parent keeps self-looping.
4. Give the fanout its own conf watch (a lightweight tracked entry works). On
   confirmation, promote those reservations from predicted to spendable; P's
   next bump tick finds its *own* reserved outpoint and proceeds. No sibling
   could have taken it.

Because the fanout's inputs are leased and its outputs are reserved in the same
ledger that governs CPFP consumption, input selection for the fanout itself is
also race-safe against in-flight CPFP children, and it reuses the existing
`FinalizePsbt` flow.

### Caveats worth deciding up front

- **Restart window (the one real gap).** `txconfirm` is in-memory today, so a
  crash mid-fanout loses the pre-reservation ownership. After restart unroll
  re-issues, `txconfirm` recomputes demand, and the confirmed fanout outputs
  look generic again, so a re-race is possible. It's bounded and non-fund-losing
  (same reissue window unroll already tolerates), and strictly no worse than the
  current coordinator, which has zero durable ownership either. Durable
  ownership means persisting the reservation ledger, which is a much bigger
  lift. I'd punt it and document the window.
- **Charter creep.** Today `txconfirm` only spends wallet funds via CPFP
  *children*. A fanout is a standalone wallet tx it broadcasts on its own. I'd
  argue it's still subsystem-neutral ("manufacture the fee inputs needed to
  confirm a package") and it already owns fee-input selection + lease
  coordination, so it fits, but it's a deliberate scope call.
- **Pool widens from registry to global.** The coordinator aggregates demand
  across one unroll registry; `txconfirm` is shared across all subsystems.
  Folding in makes the ledger global, which is better for the race. Sizing
  already derives per-parent from `EstimatePackageFee`, so it generalizes.
- **DoS/fee-leak bound.** Keep the single-pending-fanout guard and the existing
  `MaxFeeRateSatPerVByte` cap so a flapping estimator can't spam fanout txs.
  Idempotency comes free from the pending marker.

Net: we trade a bolt-on coordinator with a hand-rolled demand model and a
timing hack for a few hundred lines *inside* the actor that already owns the
ledger, and the unroll FSM gets simpler instead of more complex. The only thing
we give up versus the current approach is nothing, since neither has durable
cross-restart ownership.

Not blocking this PR necessarily, but if we're going to centralize UTXO
management for fee bumps/CPFP/fanout, `txconfirm` is where the ledger already
lives.

darepod/rpc_exit_plan.go
sputn1ck: Done. `walletSweepLockID` is now a package var instead of a const-style helper.

unroll/exit_plan.go
sputn1ck: The estimate can be computed locally, but this endpoint also returns a real funding address that the caller can show immediately. The address book keeps that address stable for repeated polling so `GetExitPlan` does not keep advancing the backing wallet address index.

sputn1ck:
Addressed in `17e9ff21`.

- Removed the unroll-owned fanout coordinator, FSM state/messages, planner hook, and obsolete unroll fanout tests.
- Moved fee-input fanout into `txconfirm`. The broadcaster now reacts to `ErrCPFPFeeInputUnavailable`, aggregates active tracked anchor parents, and returns the same transient error so the existing `Broadcasting` retry loop keeps the exit alive instead of terminally failing unroll.
- Normal fanout construction uses wallet PSBT funding through the `txconfirm.Wallet` adapter and leases fanout inputs under `txconfirmLockID`.
- Predicted fanout output outpoints are reserved to blocked parents in `parentStates` and promoted after confirmation, so sibling parents cannot steal outputs minted for a blocked parent.
- If total wallet balance is insufficient, `txconfirm` keeps returning `ErrCPFPFeeInputUnavailable` and the existing retry/alert path remains in effect.
- Restart ownership is still intentionally in-memory for v1; after restart, reissued parents recompute demand and continue without fund loss.

Validation run:
- `go test ./txconfirm ./unroll -count=1`
- `go test -tags="walletdkrpc swapruntime" ./darepod ./swapwallet ./sdk/walletdk ./cmd/darepocli/darepoclicommands/devrpc -count=1`
- `make fmt-changed`
- `make lint-changed-local`
- parent repo focused exit-plan/concurrent-fanout itests
- parent repo `make systest`

sputn1ck:
@claude review, with roasbeefs latest comments in mind

sputn1ck:
Addressed Claude's latest txconfirm fanout review in a5eacd7a. Summary:

- Pending fee-input fanouts now store the funded tx plus last broadcast height. While parents remain blocked, txconfirm rebroadcasts the same fanout at the normal fee-bump interval.
- If that rebroadcast returns a non-ignorable failure, txconfirm releases the predicted output reservations/input leases, clears the pending marker, and lets the same retry build a fresh fanout at the current capped fee rate.
- Evict now prunes the parent from pending fanout assignments. If the last assignment is removed, the pending marker is cleared and predicted reservations are released.
- Fanout promotion now does a direct parentStates lookup and skips missing parents, so confirming a fanout cannot recreate state for an evicted parent.
- The reserved-input fanout fallback is documented and covered by a regression test for the intentional case where the only usable value is a large already-reserved fee input that must be split for active parents.
- Minor cleanups are included: removed the dead zero-input estimator branch, dropped the unused LND change index assignment, and documented btcwallet/lwwallet PSBT lock lifetime behavior.

SweepWallet's broader move through wallet/txconfirm is still intentionally deferred. This PR keeps SweepWallet as a separate backing-wallet sweep, with the safety property that selected inputs are locked before signing/broadcast.

Validation run:

- client: make fmt-changed
- client: go test ./txconfirm -count=1
- client: go test ./txconfirm ./unroll -count=1
- client: go test -tags='walletdkrpc swapruntime' ./darepod ./swapwallet ./sdk/walletdk ./cmd/darepocli/darepoclicommands/devrpc -count=1
- client: make lint-changed-local
- client: git diff --check
- parent: focused exit-plan/concurrent-fanout itests
- parent: make systest
- parent: make fmt-changed GOPATH=$(go env GOPATH)
- parent: make lint-changed-local GOPATH=$(go env GOPATH)
- parent: git diff --check

levmi:
**Triage linkage check** — this PR currently says `Fixes #672`, but #672 is the open **P0 epic** for the externally-driven PSBT broadcast mode (`GetUnrollPlan` / `GetUnrollSweepPsbt` / `SubmitUnrollSweepSignature` + `external_broadcast` mode + regtest itest). This PR adds `GetExitPlan` + `SweepWallet` only, so it shouldn't auto-close the epic on merge.

Suggest changing `Fixes #672` → **`Part of #672`**.

litbot-9000:

Roasbeef:
@claude review this

PR #698, opened 2026-06-05
round: Round-scoped VTXO-tree + connector signing keys

ellemouton:
Part of the operator-key-rotation epic (lightninglabs/darepo#525). Pairs
with the server PR lightninglabs/darepo#533.

## What

Stops the client from assuming the operator's global `GetInfo` key when
participating in a round. The server now derives a fresh, per-round
ephemeral VTXO-tree MuSig2 signing key (independent of its persistent
identity key) and hands it — plus the operator key its connector tree was
built with — to the client in `ClientBatchInfo`. This lets a client that
is still on a previous operator-key epoch agree with the server on the
tree-signing key and reconstruct the same connector outpoints, so an
operator key rotation no longer breaks in-flight round participation.

## Changes

- **`roundpb`**: add `tree_cosign_key` + `connector_operator_key` to
  `ClientBatchInfo` (proto + regenerated stubs).
- **`round`**: carry both keys from `CommitmentTxBuilt` through
  `CommitmentTxReceivedState` and use them for VTXO-tree `ValidatePath`
  and connector-ancestry reconstruction instead of
  `env.OperatorTerms.PubKey`. Fall back to the global operator key when a
  field is absent (older server), so the change is wire-compatible.

The leaf policy template still commits to the global identity key — only
the tree cosigner and connector reconstruction move to the round-delivered
keys.

## Testing

`round` unit tests pass, including a new `FromProto` test covering key
parsing, the absent-field fallback, and malformed-key rejection. Verified
end-to-end against the server change via systests (boarding, refresh,
leave) and the `TestLeaveAfterOperatorKeyRotation` itest.

### DIFF
--- a/round/events.go
+++ b/round/events.go
@@ -183,6 +183,19 @@ type CommitmentTxBuilt struct {
 	// instead of the global operator key. Nil when talking to an older
 	// server; callers fall back to the global operator key.
 	ConnectorOperatorKey *btcec.PublicKey
+
+	// SweepKey is the operator sweep key for this round's VTXO-tree sweep
+	// leaf, delivered per round so the client agrees with the server on the
+	// sweep branch even across an operator key rotation. It replaces the
+	// global GetInfo sweep key entirely; a well-formed round always carries
+	// it.
+	SweepKey *btcec.PublicKey
+
+	// SweepDelay is this round's batch-wide absolute-timelock in blocks for
+	// the VTXO-tree sweep leaf, delivered per round alongside SweepKey. It
+	// replaces the global GetInfo sweep delay and drives batch-expiry
+	// computation for VTXOs created in this round.
+	SweepDelay uint32
 }
 
 func (e *CommitmentTxBuilt) clientEventSealed() {}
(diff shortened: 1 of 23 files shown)


ellemouton:
@claude review this

round/events.go
```diff
@@ -168,6 +168,21 @@ type CommitmentTxBuilt struct {
...
 	TreeOpts []roundpb.TreeFromProtoOption
+
+	// TreeCosignKey is the operator's MuSig2 cosigner key for this round's
+	// VTXO output tree, derived fresh per round by the server. The client
+	// aggregates and validates the tree against this key instead of the
+	// operator's global GetInfo key, so an operator key rotation does not
+	// change the tree key the client must agree on. Nil when talking to a
+	// server that predates this field; callers fall back to the global
+	// operator key.
+	TreeCosignKey *btcec.PublicKey
```
Roasbeef, line 179: Do we need a 3rd key here? So the vtxo sign key? or the tree sign key is the same for all users in a round?

Roasbeef (approved):
LGTM 🥦

round/events.go
ellemouton: addded below

PR #700, opened 2026-06-05
multi: condense and dedupe the test suite, drop ~3k LOC

Roasbeef:
In this PR, we take a pass over the test suite to cut out redundant and
duplicated test code without giving up any coverage. The motivation was
simple: the suite had grown a lot of one-test-function-per-case sprawl,
copies of the same encode/decode roundtrip sitting right next to a `rapid`
property that already subsumes it, and a handful of "demonstration" tests
that only assert tautologies. The net result is **-2,876 test LOC across 40
packages** with zero per-function coverage loss.

The work came in two waves, and every change was gated the same way: capture
a per-function `go tool cover -func` profile before and after, and only keep
the change if no function's statement coverage drops and the package still
passes `go test`. See each commit message for the per-package detail w.r.t
what moved and why.

## Wave 1: drop redundant and stale tests

The first wave deletes tests that add no unique signal. Most fall into a few
buckets: an example test whose every input and assertion is already covered
by a sibling `rapid` property (the oorpb/roundpb proto roundtrips, the
lib/recovery codec), a "type safety" or "shape stable" test that only asserts
something the compiler already guarantees (chainsource, round's
refresh_psbt_size), or a security-suite "FINDING" test that loops a verify
over the same signature 100x (indexer).

A couple of these were more subtle. In lib/recovery,
`TestMarkConfirmedRequiresParents` advertises a parents-confirmed guard but
never broadcasts first, so it can only ever hit the unbroadcast branch the
sibling already covers, and the guard it claims to test is unreachable
through the public API anyway (`MarkBroadcasted` refuses first). In round,
the triage wanted us to drop `TestValidateDelayParametersExitDelayTooLarge`
but keep its sweep-branch twin, except both assert the stronger "exceeds
maximum reasonable" string that the table row only checks loosely, so we
first strengthen the table row, then drop the standalone.

## Wave 2: compact the survivors

The second wave shrinks what's left by folding one-func-per-case clusters
into data-only table-driven tests (one shared runner, struct rows of inputs
and expected values, no per-row closures) and pulling duplicated fixture
setup into shared helpers. The big ones: oor's four resume-from-snapshot
ceremonies collapse into one table (-273), lib/tree's ten `partitionLeaves`
cases become one table (-97), and wallet's eight `SendVTXOs` admission tests
merge (-92).

Six packages got skipped in this wave because they couldn't shrink without
either losing a distinct case or growing LOC. That's the honest outcome: a
clean skip beats a forced merge that bloats the file.

## Coverage: before vs after

The point of all this is that none of it costs coverage. We measured each
touched package on `origin/main` (the production code is identical, only the
tests differ) and again on this branch. Every package holds or improves, with
gateway picking up +2.8% as a side effect of folding the BrowserHeaders cases
into a table that also exercises the passthrough row. Zero drops:

| package | before | after | Δ |
|---|---:|---:|---:|
| `arkrpc` | 8.8% | 8.8% | 0 |
| `chainbackends` | 27.8% | 27.8% | 0 |
| `chainsource` | 89.7% | 89.7% | 0 |
| `cmd/darepocli/darepoclicommands` | 31.4% | 31.4% | 0 |
| `daemonrpc` | 1.0% | 1.0% | 0 |
| `darepod` | 30.0% | 30.0% | 0 |
| `db` | 70.8% | 70.8% | 0 |
| `db/actordelivery` | 68.8% | 68.8% | 0 |
| `fraud` | 79.1% | 79.1% | 0 |
| `gateway` | 75.0% | 77.8% | +2.8 |
| `indexer` | 38.9% | 38.9% | 0 |
| `internal/actortest` | 89.2% | 89.2% | 0 |
| `ledger` | 74.1% | 74.1% | 0 |
| `lib/arkscript` | 65.8% | 65.8% | 0 |
| `lib/bip322` | 80.3% | 80.3% | 0 |
| `lib/recovery` | 82.0% | 82.0% | 0 |
| `lib/tree` | 83.8% | 83.8% | 0 |
| `lib/tx` | 79.4% | 79.4% | 0 |
| `lib/tx/checkpoint` | 73.6% | 73.6% | 0 |
| `lib/tx/oor` | 74.8% | 74.8% | 0 |
| `lib/types` | 58.4% | 58.4% | 0 |
| `lwwallet` | 56.5% | 56.5% | 0 |
| `mailbox/conn` | 83.2% | 83.2% | 0 |
| `mailbox/rpc` | 82.2% | 82.2% | 0 |
| `oor` | 67.7% | 67.7% | 0 |
| `round` | 76.1% | 76.1% | 0 |
| `rpc/oorpb` | 23.6% | 23.6% | 0 |
| `rpc/roundpb` | 14.9% | 14.9% | 0 |
| `scripts/verify-schema-registry` | 68.8% | 68.8% | 0 |
| `sdk/swaps` | 65.4% | 65.6% | +0.2 |
| `sdk/walletdk` | 42.0% | 42.0% | 0 |
| `serverconn` | 78.5% | 78.5% | 0 |
| `serverconn/mailboxpull` | 96.8% | 96.8% | 0 |
| `timeout` | 81.1% | 81.1% | 0 |
| `txconfirm` | 81.0% | 81.0% | 0 |
| `unroll` | 67.6% | 67.6% | 0 |
| `unrollplan` | 86.6% | 86.6% | 0 |
| `vhtlcrecovery/coordinator` | 47.3% | 47.3% | 0 |
| `vtxo` | 67.4% | 67.4% | 0 |
| `wallet` | 57.6% | 57.6% | 0 |

## A note on the rebase

This branch was rebased onto current `origin/main`, which has since merged the
Read/Stage/Commit migration (#688) and the observed-unilateral-exit work
(#602). Both restructured the exact unroll and vtxo tests we'd condensed: the
`TestResume*` ceremonies grew an `adaptTx` Exec wrapper, and the force-unroll
/ status transitions flipped `UnilateralExitState` to non-terminal. Rather
than re-condense on top of a moving target, we let upstream's newer tests win
there, so the unroll condensation and vtxo's force-unroll/status portion are
dropped (that's the difference between the original -3,075 and the current
-2,876). The vtxo `statusToState` table that survived was updated to match the
#602 non-terminal semantics.

## What's next

A few directions to take this further, roughly in order of value:

- **Mutation testing to validate the kept tests.** The coverage-diff gate
  proves we didn't drop *statement* coverage, but the stronger question is
  whether the survivors still kill the same mutants. Running `gremlins
  unleash` on the touched money-path packages (round, vtxo, lib/arkscript,
  txconfirm) would confirm the condensation didn't quietly weaken an
  assertion, and surface any survivor zone worth a real test.
- **The strengthen pass.** The triage flagged ~30 assertion-free or weak
  spots (the #270 `IsChange` gap in lib/types, the #269 per-input fee path in
  wallet, lib/tree's `VerifySigned` sitting at 0%, txconfirm's discarded
  notify-error returns). These *add* LOC so they're out of scope for a
  LOC-reduction PR, but they're where the actual bug-catching power lives.
- **Selective rapid reshapes.** A few spots are property tests in disguise
  (mailboxpull's `RetryDelay` 100-iteration loop, some of the proto
  roundtrips). Converting those to `rapid` would harden invariant coverage.
  We held off because an auto-generated reshape pass both grew LOC and
  produced failing properties, so this wants to be done by hand, one at a
  time.
- **Underscore test-name cleanup.** baselib/actor, serverconn, mailbox/conn,
  and indexer still carry `TestFoo_Bar` names. Worth a mechanical sweep to
  match the no-underscore convention.

## Heads up for review

Two commits bundle two packages each: `eb91f20` carries serverconn's
TLV-table change under the lwwallet message, and `9277434` carries db's
ledger-table change under the mailboxpull message. The content is correct and
verified, the labels just under-describe it. Happy to split them into clean
per-package commits if preferred.

All packages pass `go test`. Three failures only show up under full-suite
parallel load and pass in isolation (sdk/ark bufconn transport, sdk/swaps
overdue-invoice, db/actordelivery cross-key independence); they're
pre-existing flakes this PR doesn't touch.


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

Issue #681, opened 2026-06-03
Client should prove forfeit connectors are rooted in the current commitment tx


yyforyongyu:
## Summary

The client signs forfeit transactions using flat operator-supplied connector leaf
data, but it never proves that the connector leaf is part of a connector tree
rooted in the current round's commitment transaction.

This is independent from the VTXO-tree binding issue. Even if the new VTXO tree
is correctly bound to the commitment tx, an unbound connector can make the
old-VTXO forfeit usable without the replacement round ever committing.

## Impact

The round's atomicity is supposed to be: old VTXO forfeited if and only if the new
VTXO is created. The connector enforces that because the forfeit transaction
spends the old VTXO plus a connector leaf that descends from this round's
commitment tx.

If the Lightning Labs operator implementation (`darepo`) has a connector
construction / serialization bug, or if the running compatible operator service is
compromised or misconfigured, the client can receive a connector leaf that exists
independently of this round's commitment tx. The issue is not that we expect the
operator to act maliciously; it is that the client currently fails unsafe if the
operator side ever emits inconsistent connector data. The client signs the
forfeit over that connector. Once the independent connector is spendable, the old
VTXO can be claimed without broadcasting the commitment tx that would create the
client's new VTXO.

In our expected honest-operator deployment, the main concern is fail-safe behavior
under operator implementation bugs, serialization bugs, misconfiguration, or
compromise. Under a fully untrusted-operator model, the same missing proof becomes
a direct theft path.

## Evidence

- `rpc/roundpb/round.proto` defines `ConnectorLeafInfo` as `leaf_outpoint` and
  `leaf_output`; `ClientBatchInfo.connector_leaf_map` carries no connector tree,
  root output index, leaf index, or proof.
- `round/from_proto.go` converts the proto directly into `ForfeitMappings` with
  `ConnectorOutpoint`, `ConnectorPkScript`, and `ConnectorAmount`, without an
  ancestry check.
- `round/transitions.go` carries these mappings into signing states after VTXO
  tree validation, but no connector binding check runs there.
- `vtxo/transitions.go` builds and signs the forfeit transaction from the supplied
  connector fields. The supplied connector output is used as BIP-341 prevout
  material when signing.
- `lib/tx/forfeit.go` builds the forfeit with input 0 as the old VTXO and input 1
  as the connector. The VTXO collaborative signature uses `SigHashDefault`, so the
  client signs the exact connector outpoint supplied by the operator.
[...]

PR #712, opened 2026-06-09
round: Bind VTXO trees to the commitment tx before signing

### LINKED ISSUE #680: Client round validation should bind VTXO trees to the commitment tx
## Summary

The round path validates an operator-supplied VTXO tree for internal consistency,
but it does not verify that the tree's `BatchOutpoint` and `BatchOutput` match
the actual commitment transaction output the client is about to sign into.

This means a self-consistent tree can validate and be co-signed even if it is not
rooted in the commitment tx output that receives the client's funds.

## Impact

If the Lightning Labs operator implementation (`darepo`) has a construction /
serialization bug, or if the running compatible operator service is compromised or
misconfigured, the client can receive a VTXO tree rooted at the wrong
outpoint/output. The issue is not that we expect the operator to act maliciously;
it is that the client currently fails unsafe if the operator side ever emits
inconsistent round data. The client can then sign its boarding input or forfeit
old VTXOs into the round while believing it has a valid unilateral-exit path.

Once the commitment tx confirms, the persisted tree does not spend the real batch
output. The client's new VTXOs are not recoverable through that tree. In the
conservative case the real batch output is swept by the operator after batch
expiry; in the substituted-script case the commitment output could be redirected
to a non-canonical script that is not derived from the agreed tree root unless the
root script is also recomputed and checked.

In our expected honest-operator deployment, the main concern is fail-safe behavior
under operator implementation bugs, serialization bugs, misconfiguration, or
compromise. Under a fully untrusted-operator model, the same missing check becomes
a direct theft path.

## Evidence

- `round/transitions.go` stores `CommitmentTx`, `TxID`, and operator-supplied
  `VTXOTreePaths`, then validates paths and anchors without checking the tree's
  `BatchOutpoint` or `BatchOutput` against `s.CommitmentTx.UnsignedTx`.
- `lib/tree/tree.go` verifies `t.Root.Input == t.BatchOutpoint`, but both fields
  come from the same operator-supplied tree.
- `rpc/roundpb/convert.go` decodes `BatchOutpoint` and `BatchOutput` from the
  proto; `round/from_proto.go` validates the PSBT shape but does not bind those
  fields to the commitment tx.
- MuSig2 signing and verification use the tree's claimed `BatchOutput` as the
  prevout, so a valid signature proves consistency with the claimed prevout only.
- `oor/incoming_vtxo.go` has the analogous ancestry check for OOR
  (`TreePath.BatchOutpoint.Hash == CommitmentTxID`); the round path lacks it.
- Honest `darepo` tree construction already checks the expected output: production
  code builds VTXO batch outputs with `tree.BuildBatchOutput`, checks commitment
  tx output value and pkScript, and sets `BatchOutpoint` from the commitment txid
  and matched output index.

## Suggested Fix

Add validation before nonce generation/signing in `CommitmentTxReceivedState`.

For each `outputIdx, vtxoTree := range s.VTXOTreePaths`:

- Recompute the commitment txid from `s.CommitmentTx.UnsignedTx.TxHash()` and
  either use that value directly or assert it matches `s.TxID`.
- Verify `vtxoTree.BatchOutpoint.Hash == commitmentTxID`.
- Verify `uint32(outputIdx) == vtxoTree.BatchOutpoint.Index`.
- Verify the index is in range for `s.CommitmentTx.UnsignedTx.TxOut`.
- Verify the commitment output at that index exactly matches
  `vtxoTree.BatchOutput.Value` and `vtxoTree.BatchOutput.PkScript`.
- Also verify the output script is the expected tree-root script by recomputing
  the root P2TR script from the tree root cosigner set and sweep tapscript root,
  then comparing it with both `BatchOutput.PkScript` and `TxOut[idx].PkScript`.

Also update confirmation registration paths that currently assume `TxOut[0]` so
they use the validated batch-output index.

## Test Coverage

Add tests that reject:

- Mismatched `BatchOutpoint.Hash`.
- Mismatched map key vs `BatchOutpoint.Index`.
- Out-of-range `BatchOutpoint.Index`.
- `BatchOutput` value mismatch.
- `BatchOutput` script mismatch.
- Commitment output script that matches `BatchOutput` but not the recomputed
  tree-root script.

Add positive fixtures for honest single-output and multi-output rounds.

## Related Issues

Targeted searches did not find an existing issue for this client-side round tree
binding gap. Nearby items such as
[`darepo-client#51`](https://github.com/lightninglabs/darepo-client/issues/51)
and
[`darepo-client#61`](https://github.com/lightninglabs/darepo-client/issues/61)
are related but do not track this client validation before signing.

This is the VTXO-tree counterpart to the companion issue, "Client should prove
forfeit connectors are rooted in the current commitment tx." Both issues are
needed to restore the round atomicity invariant: fixing only VTXO-tree binding
still leaves the old-VTXO forfeit usable through an unbound connector, while
fixing only connector binding still leaves the new VTXO tree unbound from the
commitment tx output. After both issues are filed, they should be linked by issue
number.

ellemouton:
Fixes #680.

## Problem

The round path validated each operator-supplied VTXO tree for internal
self-consistency (leaf script present, client signing key on the path, anchors
well-formed) but **never checked that the tree's `BatchOutpoint` and
`BatchOutput` matched the commitment-transaction output the client was about to
sign into**. Every field the client trusted came from the same operator-supplied
blob as the rest of the tree, so a self-consistent tree rooted at the *wrong*
outpoint — or describing a batch output that did not match the real commitment
output — passed validation and was co-signed.

Once the commitment tx confirms, such a tree does not spend the real batch
output, and the client's new VTXOs (whose outpoints derive from the tree root)
are unrecoverable through it. Under an honest operator this is a fail-unsafe gap
on a construction/serialization bug or misconfiguration; under an untrusted
operator it is a fund-redirect path.

The OOR receive path already closes the analogous gap (`validateIncomingAncestry`),
and the connector counterpart landed in #681 (`validateConnectorAncestry`). This
PR brings the round VTXO-tree path to parity. Together #680 and #681 restore the
round atomicity invariant.

## Changes

**`round: Bind VTXO trees to the commitment tx before signing`**
- New `validateVTXOTreeBinding`, called in `CommitmentTxReceivedState` before
  the per-VTXO `ValidatePath` loop (so trees are bound before their contents are
  trusted) and before nonce generation. For each `(outputIdx, tree)` it asserts:
  - the tree's `BatchOutpoint` names this commitment tx,
  - the map key agrees with `BatchOutpoint.Index`,
  - the index is in range,
  - the committed output byte-matches the tree's `BatchOutput`, and
  - the committed script equals the taproot script **recomputed** from the tree
    root's cosigner set and sweep tapscript root (rejecting a substituted but
    self-consistent script, since `BatchOutput.PkScript` is itself
    operator-supplied).
- A mis-rooted tree fails the round into `ClientFailedState{Recoverable:false}`.
- Test harness now builds commitment-bound fixtures (`bindTreeToCommitment`),
  mirroring how the honest operator constructs a round.

**`round: Watch the validated batch output for confirmation`**
- The confirmation-registration paths assumed `TxOut[0]`. New
  `confirmationWatchScript` watches the validated batch-output index instead
  (falling back to output 0 for rounds with no VTXO trees), wired into both FSM
  outbox registrations and the actor resume/checkpoint registration. Not the
  core safety fix — `TxOut[0]` still works for txid-based detection — but
  correct for multi-output rounds and script-filtering backends (Neutrino).

## Testing

New `round/vtxo_tree_binding_test.go`:
- Six negative cases (outpoint-hash mismatch, map-key/index disagreement,
  out-of-range index, batch-output value mismatch, batch-output script mismatch,
  substituted-but-self-consistent script). Red-before/green-after confirmed by
  neutering the helper.
- Positive single-output case.
- Multi-output case (batch at index 1) asserting both that binding passes and
  that the confirmation watch script tracks the batch output, not output 0.
- An FSM-level test confirming an unbound tree fails the round with a binding
  reason rather than being co-signed.

`go test ./round/` green; `make fmt-changed` + `make lint-changed-local` clean;
`make commitmsg-lint` OK.

### DIFF
--- a/round/actor.go
+++ b/round/actor.go
@@ -24,6 +24,7 @@ import (
 	"github.com/lightninglabs/darepo-client/ledger"
 	"github.com/lightninglabs/darepo-client/lib/actormsg"
 	"github.com/lightninglabs/darepo-client/lib/arkscript"
+	"github.com/lightninglabs/darepo-client/lib/tree"
 	"github.com/lightninglabs/darepo-client/lib/types"
 	"github.com/lightninglabs/darepo-client/serverconn"
 	"github.com/lightninglabs/darepo-client/timeout"
@@ -911,9 +912,11 @@ func forfeitsMatch(forfeits []types.ForfeitRequest,
 
 // registerCommitmentConfirmation registers for confirmation monitoring of a
 // commitment transaction with the chainsource actor. The commitmentTx is used
-// to extract the pkScript for LND's confirmation tracking.
+// to extract the pkScript for LND's confirmation tracking, and vtxoTrees (the
+// round's validated VTXO trees, possibly nil) selects which output to watch.
 func (a *RoundClientActor) registerCommitmentConfirmation(ctx context.Context,
-	txid chainhash.Hash, commitmentTx fn.Option[*psbt.Packet]) {
+	txid chainhash.Hash, commitmentTx fn.Option[*psbt.Packet],
+	vtxoTrees map[int]*tree.Tree) {
 
 	callerID := fmt.Sprintf("commitment-tx-%s", txid.String())
 
@@ -929,12 +932,16 @@ func (a *RoundClientActor) registerCommitmentConfirmation(ctx context.Context,
 		},
 	)
 
-	// Extract pkScript from the commitment transaction's first output.
-	// LND requires a pkScript for confirmation tracking.
+	// Extract the pkScript LND needs for confirmation tracking. Watch the
+	// validated batch output (the output that receives this client's
+	// funds) rather than assuming output 0; confirmationWatchScript falls
+	// back to output 0 when the round carries no VTXO trees.
 	var pkScript []byte
 	commitmentTx.WhenSome(func(packet *psbt.Packet) {
-		if packet.UnsignedTx != nil && len(packet.UnsignedTx.TxOut) > 0 {
-			pkScript = packet.UnsignedTx.TxOut[0].PkScript
+		if packet.UnsignedTx != nil {
+			pkScript = confirmationWatchScript(
+				packet.UnsignedTx, vtxoTrees,
+			)
 		}
 	})
 
@@ -1164,6 +1171,7 @@ func (a *RoundClientActor) Start(ctx context.Context) error {
 			a.commitmentTxIndex[roundFSM.TxID] = keyStr
 			a.registerCommitmentConfirmation(
 				ctx, roundFSM.TxID, round.CommitmentTx,
+				round.VTXOTreePaths.UnwrapOr(nil),
 			)
 
 			a.log.InfoS(ctx, "Resumed round awaiting confirmation",
@@ -2331,6 +2339,7 @@ func (a *RoundClientActor) processOutbox(ctx context.Context,
 			a.commitmentTxIndex[txid] = keyStr
 			a.registerCommitmentConfirmation(
 				ctx, txid, roundFSM.CommitmentTx,
+				inputSigState.VTXOTreePaths,
 			)
 
 			a.log.InfoS(ctx, "Round checkpoint processed",
--- a/round/harness_test.go
+++ b/round/harness_test.go
@@ -781,55 +781,144 @@ func (h *boardingTestHarness) newTestVTXOTreeForIntents(
 	return vtxtTree
 }
 
-// newCommitmentTxBuiltEvent creates a CommitmentTxBuilt event simulating what
-// the server sends after building the commitment transaction with all boarding
-// inputs and the batch VTXT tree output. The transaction is returned as a PSBT
-// with WitnessUtxo populated for all inputs, which is required for correct
-// Taproot sighash computation.
-func (h *boardingTestHarness) newCommitmentTxBuiltEvent(roundID RoundID,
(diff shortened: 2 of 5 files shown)


ellemouton:
@claude review this

ellemouton:
Thanks for the reviews. Addressed both comments (folded into the relevant commits via autosquash):

1. **`confirmationWatchScript` nil guard** — added `commitmentTx == nil` for symmetry with `validateVTXOTreeBinding`'s defensive posture. As @claude noted, the current callers already deref `UnsignedTx` upstream so nothing reachable today triggers it, but it makes the helper safe on any future unvalidated path.

2. **`int64` index comparison** — switched `uint32(outputIdx) != BatchOutpoint.Index` to `int64(outputIdx) != int64(BatchOutpoint.Index)`. As @claude observed this isn't reachable in production (`VTXOTreePaths` keys come from an int32 proto field with negatives already rejected in `from_proto.go`), but the int64 form is lossless defense-in-depth and lets the separate `< 0` guard go.

Neither was a live bug; both are cheap robustness improvements. `go test ./round/`, `make fmt-changed`, `make lint-changed-local`, and `make commitmsg-lint` are all clean.

PR #715, opened 2026-06-09
chainsource: reconnect block epoch subscriptions

### LINKED ISSUE #714: darepod can miss confirmed boarding UTXOs after block epoch stream exits
## Problem

A darepod instance can stop promoting funded boarding addresses into confirmed boarding intents if its LND block epoch stream exits. The funds remain safe and visible to LND, but darepod never records the confirmed boarding UTXO, so callers see no boardable funds.

We hit this on signet with the swapd auto-boarder:

- swapd called `NewAddress` and funded the returned boarding address.
- LND later showed the output as confirmed and unspent.
- darepod still reported `boarding_confirmed_sat=0` and `Board` returned `no_boarding_utxos`.
- The auto-boarder reservation stayed in `sent` because it waits for darepod to report confirmed boarding balance before calling `Board`.

## Observed production evidence

The funded output was present in LND:

```text
outpoint: bea491fd6c79cc7203d7616a268114305bef7e34e108e12a98fd3849d925b88a:1
address: tb1p9z6ewff64gzfeng9q6qxv6jzushr8hy2dk557ky5v05t9lgt5has3q64xk
amount_sat: 5000000
pk_script: 512028b597253aaa049ccd050680666a42e42e33dc8a6da94f589463e8b2fd0ba5fb
confirmations: 8
```

The darepod DB had the boarding address row for that script, but no matching `boarding_intents` row and no `wallet_utxo_log` deposit row for the funded outpoint.

The logs show the wallet scanner processing tips until the LND notifier shut down, then no later boarding-wallet tip scans:

```text
2026-06-09 14:38:15 ARKW: Processing new chain tip height=308107 last_processed=308106
2026-06-09 14:38:15 ARKW: ListUnspent returned UTXOs height=308107 utxo_count=42
2026-06-09 14:41:34 LNDC: Block epoch error err="rpc error: code = Unknown desc = chain notifier shutting down"
2026-06-09 14:41:34 LNDC: Block epoch forwarder goroutine exiting
2026-06-09 16:11:10 ARKW: Created new boarding address address=tb1p9z6ewff64gzfeng9q6qxv6jzushr8hy2dk557ky5v05t9lgt5has3q64xk exit_delay=512
```

After confirmation, `GetBalance` returned:

```text
boarding_confirmed_sat: 0
boarding_unconfirmed_sat: 0
vtxo_balance_sat: 0
```

`Board` returned:

```text
status: no_boarding_utxos
vtxo_count: 0
```

## Code path

`NewAddress` only creates and persists the boarding address. It does not create a boarding intent. The intent is created later when the wallet scanner sees a confirmed UTXO paying to a known boarding script.

In `wallet/wallet.go`, `handleProcessTipTick` depends on the latest block epoch, calls `ListUnspent(min_confs=1)`, and then calls `processUtxo`. `processUtxo` is where the `BoardingStatusConfirmed` intent is inserted.

`GetBalance` also explains the observed disappearing balance:

- unconfirmed boarding balance is computed by directly scanning wallet UTXOs with `Confirmations == 0`;
- confirmed boarding balance is computed only from persisted `BoardingStatusConfirmed` intents.

So an output can appear as `boarding_unconfirmed_sat` before confirmation, then disappear from both buckets if the block scanner never creates the confirmed intent.

`Board` has the same dependency: it fetches `BoardingStatusConfirmed` intents from the store. If the scanner missed the UTXO, `Board` cannot discover it by itself and returns `no_boarding_utxos`.

The chain notification side currently exits on stream failure:

- `chainbackends/lndclient_adapters.go` returns from the forwarder goroutine when `RegisterBlockEpochNtfn` sends an error.
- That closes the epoch channel.
- `chainsource/block_epoch_actor.go` returns when the block epoch channel closes.
- The wallet actor subscribed once at startup and does not resubscribe.

I checked latest `origin/main` against the deployed signet image commit (`main-3666c20`). The block epoch subscription files were unchanged between deployed and latest main, so this failure mode does not appear to be fixed yet.

## Expected behavior

A transient LND block epoch stream failure should not permanently strand boarding detection. A confirmed UTXO paying to a known boarding address should eventually become a `BoardingStatusConfirmed` intent without requiring a manual darepod restart.

## Suggested fix

We should make the boarding detection path resilient to notification stream loss. Two complementary approaches would cover this:

1. Make block epoch subscriptions resubscribe with backoff when the backend epoch channel closes or the LND notifier reports an error.
2. Add a periodic or startup catch-up scan that uses current best height plus `ListUnspent(min_confs=1)` to detect confirmed UTXOs for known boarding scripts even if the exact block epoch was missed.

The catch-up scan is important because it heals already-confirmed UTXOs that were missed while the stream was down.

bhandras:
## Summary

Fixes #714.

This PR keeps the block epoch actor alive when the backend epoch stream closes after the initial subscription succeeds. Instead of treating the closed stream as terminal, the actor now cancels the closed registration, waits with bounded backoff, and registers a fresh stream.

The production symptom was that darepod could keep serving RPCs after LND's chain notifier stream exited, but the wallet stopped processing new chain tips. A boarding address funded after that point could confirm on-chain without being promoted into a confirmed boarding intent, so `GetBalance` and `Board` still reported no boarding UTXOs until the daemon restarted.

## Test Coverage

Added a regression test that closes the first backend block epoch registration, verifies it is canceled, and then confirms the same subscriber receives a block from the replacement registration.

Local checks:

```bash
make unit pkg=./chainsource case=TestBlockEpochActorReconnectsAfterBackendStreamCloses timeout=30s
make unit pkg=./chainsource timeout=2m
make lint-changed-local
```

### DIFF
--- a/chainsource/block_epoch_actor.go
+++ b/chainsource/block_epoch_actor.go
@@ -5,11 +5,25 @@ import (
 	"fmt"
 	"log/slog"
 	"sync"
+	"time"
 
 	"github.com/btcsuite/btclog/v2"
 	"github.com/lightninglabs/darepo-client/baselib/actor"
 	"github.com/lightninglabs/darepo-client/build"
 	fn "github.com/lightningnetwork/lnd/fn/v2"
+)
+
+const (
+	// defaultBlockEpochReconnectBackoff is the initial delay before a
+	// long-lived block subscription tries to re-register after its backend
+	// stream closes. LND can close all chain notifier streams during
+	// restart or internal notifier churn; callers such as the boarding
+	// wallet need the subscription to heal without a daemon restart.
+	defaultBlockEpochReconnectBackoff = time.Second
+
+	// defaultBlockEpochMaxReconnectBackoff caps reconnect backoff so a
+	// backend outage does not spin, but recovery still happens promptly.
+	defaultBlockEpochMaxReconnectBackoff = 30 * time.Second
 )
 
 // BlockEpochConfig holds configuration for BlockEpochActor.
@@ -21,6 +35,15 @@ type BlockEpochConfig struct {
 	// falls back to extracting a logger from context via LoggerFromContext,
 	// or uses btclog.Disabled if no logger is found.
 	Log fn.Option[btclog.Logger]
+
+	// ReconnectBackoff is the initial delay before reconnecting a block
+	// epoch subscription whose backend stream closed after the initial
+	// registration succeeded. Zero uses defaultBlockEpochReconnectBackoff.
+	ReconnectBackoff time.Duration
+
+	// MaxReconnectBackoff caps the exponential reconnect delay. Zero uses
+	// defaultBlockEpochMaxReconnectBackoff.
+	MaxReconnectBackoff time.Duration
 }
 
 // WithLogger returns a new config with the given logger set.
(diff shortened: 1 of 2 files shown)


bhandras:
@claude review

PR #716, opened 2026-06-09
multi: Make round intents restart-safe via a generic pending-intents outbox

### LINKED ISSUE #660: SendOnChain: make the send intent restart-safe (durable round actor)
Follow-up from review of #646 (cc @Roasbeef).

## Problem

`SendOnChain` (the wallet-shaped onchain payment from #634) registers a leave+change intent with the round actor and then triggers round registration. If the daemon dies after the intent is accepted but before the next round actually starts, the send intent is lost — the user asked to send onchain, got a \"submitted\" response, but nothing reaches a round on restart.

This is the same class of gap that durable Board replay closed: when a user says \"board these UTXOs\", we persist that intent and replay it on startup until a round adopts it.

## Options discussed

- Persist the SendOnChain intent (mirroring the Board replay machinery) and replay on startup.
- Or make the round actor itself fully durable — an open question raised in the review. If the round actor were durable, `TriggerRoundRegistration` returning nil would be a reliable guarantee that we always join a round, and per-intent replay machinery wouldn't be needed.

## Scope

Out of scope for #634 (which fixes the exact-amount overpay). Tracking here so the restart-safety gap isn't lost.

Roasbeef:
In this PR, we close the restart-safety gap tracked in #660: a daemon that died after `SendOnChain` registered its intent with the round actor, but before the next round actually started, would silently drop the user's onchain send even though the RPC had already returned "submitted". Rather than bolting a second bespoke replay path next to the Board one, we generalize the Board replay machinery from #416 (which was already a hand-rolled transactional outbox specialized to one intent type) into a single generic pending-intents outbox, fold Board into it, delete the ad-hoc version, and then register `SendOnChain` as the second kind.

The shape is persist-before-publish: the wallet writes the intent to the outbox before the round actor ever sees it, so every crash window either leaves no row (the caller saw an error, clean retry) or a replayable row. Each intent is anchored to the set of outpoints the round consumes when it adopts it (confirmed boarding outpoints for `board`, the reserved forfeit set for `send_onchain`), and `RoundPersistenceStore.CommitState` clears those anchors inside the same point-of-no-return transaction that checkpoints the round. Adoption and outbox-clear commit together, hence replay-after-adoption is structurally impossible without making any actor durable.

The idempotency key is just a hash of the input: `sha256(kind || sorted_anchors || payload)`. No client-supplied UUIDs, no new RPC plumbing. Re-persisting the same logical intent is a natural upsert, and the anchors double as the structural dedup since a round can only spend a given forfeit set once.

Fixes #660.

## Replay is reconcile, not redelivery

On startup (after the round-client actor is on the receptionist), the daemon issues one `ReplayPendingIntentsRequest` Ask and the wallet walks a per-kind replayer registry. Replay deliberately does not redeliver frozen bytes: the Board replayer re-walks which anchors are still `Confirmed` and re-runs `handleBoard` via self-Tell (preserving the FIFO-before-gRPC ordering from the original fix), while the SendOnChain replayer re-reserves the _exact_ persisted forfeit outpoints (never re-running coin selection, which would silently change the intent's identity), rebuilds the package from the TLV payload with a freshly derived change key, and registers with `TriggerRegistration=true` since there is no RPC caller around to fire the second step. A reservation failure on replay means some anchor was consumed by another spend path, so the intent is provably stale and gets dropped rather than failing every future boot.

A synchronous round rejection on the live RPC path deletes the row alongside the reservation release: the caller was told the send failed, so it must not silently resurrect on the next start.

## Why not the baselib `OutboxPublisher`

The actor framework already ships an `OutboxPublisher`, and we considered building on it. It doesn't fit this problem yet, for one load-bearing reason: its completion contract marks a row done as soon as the `Tell` durably lands in the _target's_ mailbox, with exactly-once provided by the target `DurableMailbox` deduping on the outbox ID. The round actor is a plain actor with an in-memory mailbox, so "delivered" would fire while the intent is still RAM-only, which is exactly the crash window this PR exists to close. What we need is completion at round _adoption_, a cross-subsystem fact the publisher has no way to observe. The replay semantics differ too (reconcile-then-rebuild vs byte redelivery).

## The road not taken (yet): a durable round system

The other route from the #660 discussion would resolve this end to end: make the round actor (and the wallet handoff into it) durable. At that point `TriggerRoundRegistration` returning nil becomes a real guarantee for _every_ intent type at once, the framework's mailbox-level at-least-once plus `DeduplicationTTL` replaces all per-intent replay machinery, and the replayer registry in this PR gets deleted wholesale. We didn't take it here because it's a much bigger blast radius on a hot actor, but the persistence contract was designed to survive that swap: the anchors, the hashed intent ID, and the clear-on-`CommitState` join point all carry over unchanged, only the publisher leg gets replaced. Tracking the durable-actor migration in the darepo#527 line of work.

## Test plan

- [x] Board replay suite ported property-for-property onto the generic store (persist-with-anchors, persist-failure-skips-Tell, all-stale sweep, live self-Tell with decoded target, `NoPersist` opt-out, empty-store no-op).
- [x] New send-replay coverage: payload codec roundtrip, persist-before-publish ordering, rejection-deletes-row, stale-anchor drop, exact-anchor re-registration.
- [x] `CommitState` integration test against a real DB: committing a round whose forfeit set covers a send intent's anchors deletes the row atomically while unrelated intents survive.
- [x] `go test -tags "walletrpc swapruntime" ./...`, `make lint-changed-local` (0 issues), `make commitmsg-lint` across the series.


Roasbeef:
@claude review this

Roasbeef:
Addressed the review feedback and re-pushed (history collapsed to 4 clean logical commits; tree unchanged).

**P1 (Codex + Claude) — caller cancellation deleted a still-landing send:** `handleSendOnChain` now distinguishes a caller-await cancellation (`ctx.Err() != nil`) from a genuine synchronous round rejection. On caller-cancel it sets `committed` so neither the outbox-row delete nor the forfeit-reservation release fires — the detached `Ask` may still land the intent, and startup replay (or the `CommitState` checkpoint) reconciles it. The row is deleted only on a real round rejection. New test `TestHandleSendOnChainCallerCancelKeepsIntent` cancels the ctx during the persist and asserts no delete + no reservation release.

**Medium (Gemini + Claude) — replay Ask had no deadline:** the replay-path round `Ask` is now wrapped in `context.WithTimeout(context.WithoutCancel(ctx), 30s)`, so a slow round actor at startup can't head-of-line-block the single-worker wallet mailbox. The error path already keeps the row and releases the reservation, so a timeout just retries next start.

**Minor (Gemini) — `cmp.Compare`:** already in (`wallet/pending_intent.go`).

**Minor (Claude) — stale Board rows lingering when some are live:** the Board replayer now deletes fully-stale intents by id during reconcile, not only in the all-stale bulk-clear branch. Covered by `TestReplayDeletesStaleBoardIntentAlongsideLiveOne`.

All wallet + db suites green, lint clean.

Roasbeef:
@claude review this

PR #717, opened 2026-06-09
multi: enforce operator amount limits on receive and boarding

Roasbeef:
In this PR, we wire the operator's amount limits into the client ahead of the mainnet launch: a per-VTXO maximum and an advisory cap on the total balance a single user holds in the system. The per-VTXO ceiling already rides in the existing `max_boarding_amount` terms field (the server has always populated it from its `MaxVTXOAmount`), so the wire change here is a single new `max_user_balance` field on `GetInfo` and the daemon-level `ServerInfo`, threaded through `OperatorTerms` and the SDK.

The user-facing enforcement lands in two places. On the Recv path, the wallet layer now pre-flights the invoice amount against both limits before a swap session (and invoice) ever exists, so the user gets a clean error naming the limit and their current balance instead of a stuck swap. The balance term counts live VTXOs plus every boarding bucket, since those funds are already committed to entering the system.

On the boarding path, `handleBoard` clamps the confirmed boarding balance to the remaining headroom under the cap and splits the boarded amount so every VTXO target fits under the per-VTXO maximum. The interesting case is a deposit larger than the headroom (e.g. a 2 BTC boarding output under a 1 BTC cap): rather than rejecting it, the wallet boards the headroom and ships the remainder as a change leave output paying a fresh boarding script. The round machinery already supports this shape (outputs <= inputs, leaves riding alongside boarding intents), so the clipped remainder re-confirms as a brand-new boarding intent once the commitment tx hits the chain, ready to board again when headroom frees up. When no usable headroom remains, the board fails with a typed `ErrBoardingCapReached` and a friendly maximum-balance message.

Both checks are zero-means-disabled and fail open when terms are unavailable: the operator re-validates every limit server-side, so the client-side pass exists for the error message, not the security boundary. Wallets without the new seams (harnesses, operators advertising no caps) board exactly as before.

The server-side halves of this live in the companion darepo and swapdk-server PRs; the darepo systest exercises the full partial-board / change-leave / cap-rejection story end to end. See each commit message for the incremental details.


Roasbeef:
Pushed a follow-up that makes the rejection routing typed end to end: oorpb gains `OOR_REJECT_OUTPUT_POLICY`, `ClassifySubmitError` maps it onto a typed `ErrOutputPolicyViolation`, and the darepod ingress adapter now routes every typed rejection through the classifier so `GetOORSession` / `ListOORSessions` surface the classified cause as `failure_reason`.

The new path also exposed (and fixes) a stranding gap: a session that failed terminally used to leave its reserved inputs in SpendingState until the startup sweep. The FSM now emits a `ReleaseInputsRequest` when the failure lands before the point of no return, so the inputs return straight to LIVE.

litbot-9000:

Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #726, opened 2026-06-10
db: configurable commit durability and fullfsync, pg test-DB reuse

Roasbeef:
In this PR, we expose the SQLite storage durability levers that the OOR
benchmarking campaign showed were the first two ceilings on payment
throughput, and we make the postgres test harness reuse its database across
daemon restarts so restart-style systests work on both backends.

The headline knob is commit durability: `synchronous` is now configurable,
defaulting to NORMAL on the client (was FULL). Under WAL, NORMAL omits the
per-commit WAL fsync (the WAL is synced only before a checkpoint), so a power
loss can roll back the last few commits but can never corrupt the database; an
OS or app crash loses nothing. That contract composes cleanly with the durable
mailbox model, where every cross-actor message is redelivered until acked, so a
dropped tail commit replays rather than strands. On the high-contention stress
shape this single default was worth 1.3-2.3x throughput with p50 falling from
~6s to ~700ms: the per-commit fsync was the latency tail.

The second knob is macOS specific: sqlite on darwin issues F_FULLFSYNC (a full
disk-barrier flush), a hidden second fsync tax that profiling showed cost ~35%
throughput on the bench host. Under NORMAL it governs the WAL checkpoint sync
rather than a per-commit fsync, but checkpoints recur continuously under
sustained write load, so the `fullfsync` pragma is now configurable. The
default is unchanged.

Both knobs are exposed on the daemon under the `db.sqlite.*` namespace
(`db.sqlite.synchronous`, `db.sqlite.nofullfsync`), matching the dotted config
convention used elsewhere; the `db.postgres.*` namespace is reserved.
Postgres-side durability tuning (`synchronous_commit`) is deliberately deferred
to a separate change, and the operator-side default is a separate decision that
stays FULL over there (darepo#553).

The test-infra commit keys the postgres test database off the dbPath so a
restarted daemon lands on the same database rather than a fresh one, which the
restart-recovery systests depend on.

This PR builds on #693 and is part of the OOR optimization train.

### DIFF
--- a/cmd/darepod/main.go
+++ b/cmd/darepod/main.go
@@ -214,6 +214,23 @@ func newRootCmd() *cobra.Command {
 			"otherwise",
 	)
 
+	// SQLite database durability knobs (db.sqlite.* namespace). The
+	// db.postgres.* namespace is reserved for a future Postgres-tuning
+	// change; the daemon always opens SQLite.
+	f.String(
+		"db.sqlite.synchronous", cfg.DB.Sqlite.Synchronous, "the "+
+			"SQLite synchronous (commit durability) level: one "+
+			"of full, normal, off; empty defaults to normal, "+
+			"which under WAL mode omits the per-commit fsync "+
+			"of full for higher write throughput",
+	)
+	f.Bool(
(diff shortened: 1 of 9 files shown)


Roasbeef:
@claude review this

PR #727, opened 2026-06-10
baselib/actor+db: leaseless single-worker consume, transactional outbox

Roasbeef:
In this PR, we extend the durable actor runtime with two framework-layer
optimizations that came straight out of profiling the OOR payment path: a
leaseless consume path for single-worker actors, and a transactional fold for
outbox delivery.

The leaseless path attacks transaction #1 on every consume. The classic path
opens a write transaction just to lease the next mailbox message
(`LeaseNextMessage` is an UPDATE under BEGIN IMMEDIATE), then a second one to
commit the turn's effects. For an actor with `numWorkers == 1` there is no
competing consumer, so the lease buys nothing: we add a read-only
`PeekNextMailboxMessage` plus by-id ack/nack variants and gate them strictly on
the single-worker case via a mailbox flag. Routing is keyed on the empty lease
token, so the multi-worker path is byte-for-byte unchanged. At its point in the
optimization chain this was worth ~1.6x: the peek is a pure read, so idle polls
go from one write transaction to zero and stop grabbing the single sqlite writer
under BEGIN IMMEDIATE, which is exactly what was serializing all the concurrent
session pollers. The head-to-head lives in lightninglabs/darepo#538's Experiment
6.

Because the peek takes no lease, it also doesn't pre-bump attempts; that bump
moves to nack-by-id, so a message that fails by returning an error still climbs
toward dead-lettering just like the leased path. The one case it doesn't cover
is a message that crashes the worker *process* on every attempt: it skips the
nack each time and so re-peeks forever instead of dead-lettering. That's
deliberate, not a gap to "fix" by bumping attempts at claim, which would put
every poll (including the empty ones that dominate under contention) back under
BEGIN IMMEDIATE and hand the 1.6x right back. The single-worker actors on this
path carry their own higher-level retry/backoff and treat the dead-letter table
as a manual sink, so retrying a process-poisoning message beats silently
dropping it. A wedged-DB nack now also backs off on the poll ticker instead of
tight-spinning re-peeks of the same row.

The outbox fold introduces `TxAwareDeliveryStore` and folds outbox delivery
(enqueue into the destination mailbox + mark-complete) into one write
transaction. Before the fold those were two commits ordered for safety, which
left a crash window where an enqueued message could be redelivered because its
completion never landed; the fold closes that window structurally. Folding the
enqueue inside the publisher's tx means the row isn't visible to the target
until commit, so the in-process wake from `Send` races ahead of visibility and
finds nothing. We restore the snappy same-process handoff with a post-commit
wake: once a folded enqueue commits, we rouse the local mailbox receive loops to
re-poll. It's a coarse broadcast (every local mailbox re-polls, a non-target
does one empty poll) with the poll ticker as the correctness fallback, so a
missed wake only ever costs a poll interval, never a message. The same `TxAware`
mechanism is what the serverconn ingress fold builds on.

Rounding things out: drip-box and concurrent-mailbox throughput benchmarks for
the delivery store, and a systest driver fix so the incoming outbox driver
handles `ScheduleRetryRequest`.

This is part of the OOR optimization train; see each commit message for a
detailed description w.r.t the incremental changes.

### DIFF
--- a/baselib/actor/CLAUDE.md
+++ b/baselib/actor/CLAUDE.md
@@ -26,7 +26,7 @@ crash-safe at-least-once delivery with exactly-once deduplication.
 - `Receptionist` — Service locator mapping `ServiceKey` → `ActorRef` for decoupled actor wiring.
 - `Message` — Sealed interface for all actor messages (must embed `BaseMessage`).
 - `MessageCodec` — TLV-based codec for message serialization/deserialization.
-- `DeliveryStore` / `TxAwareDeliveryStore` — Interfaces for durable mailbox persistence (enqueue, claim, ack, dead-letter).
+- `DeliveryStore` / `TxAwareDeliveryStore` — Interfaces for durable mailbox persistence (enqueue, claim, ack, dead-letter). The leaseless single-worker fast path adds `PeekNextMessage` (read-only claim, no lease, no attempts bump; yields an empty lease token), `AckMessageByID` (unfenced delete), and `NackMessageByID` (unfenced release that increments attempts). A `DurableActor` enables it (via `DurableMailboxConfig.SingleWorkerLeaseless`) strictly when `NumWorkers == 1` AND the behavior is the Read/Commit (Right/`TxBehavior`) path, eliminating the per-message lease write transaction. The multi-worker pool and the classic path are byte-for-byte unchanged: they keep `LeaseNextMessage` and the lease-fenced ack. Ack/nack route to the by-ID ops automatically whenever the delivery's lease token is empty; `Delivery.ShouldDeadLetter` counts the in-flight attempt as `Attempts + 1` on the leaseless path so the dead-letter boundary matches the leased path (where attempts is pre-incremented at lease).
 - `DurableActor` — Actor variant with crash-safe mailbox backed by SQL persistence. Provides `Wait(ctx)` to block until the actor stops and `StopAndWait(ctx)` to request a graceful shutdown and then wait.
 - `DurableActorConfig[M, R]` — Configuration struct for `DurableActor`: behavior, store, codec, clock, DLO, WaitGroup, `TellRetryPolicy`, lease/heartbeat/poll durations, max attempts, cleanup timeout, deduplication TTL, and `NumWorkers`.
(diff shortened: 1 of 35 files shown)


Roasbeef:
Finished the adversarial follow-up pass and pushed an autosquashed update (`6be5a1df`). Main fixes:

- Leaseless peek now normalizes every peeked row to an empty lease token/zero lease deadline at the store adapter boundary, even if the DB row still has stale expired lease metadata from an older leased claim.
- `NackMailboxMessageByID` now clears stale `lease_token`/`lease_until` while incrementing attempts, keeping the persisted row aligned with the leaseless state machine.
- Added `Delivery.EffectiveAttempts()` and routed Tell retry/dead-letter policy decisions through it on the normal, tx-aware, and fallback result paths. This counts the in-flight leaseless attempt before a nack can bump the row to `max_attempts`.
- Added commit-level logging around tx-aware outbox delivery, so begin/commit/rollback-level `ExecTx` failures are visible even when the inner enqueue + complete operations returned nil.
- Updated the p-model docs for the two key invariants: the leaseless consume ownership model and the outbox fold model.

Validation run after autosquash:

- `make unit pkg=./baselib/actor`
- `make unit pkg=./db/actordelivery`
- `make unit pkg=./internal/actortest`
- `make lint-changed-local`
- `make commitmsg-lint range="1047332e521d4798a8f168e9609ad3b1e43b9653..HEAD"`
- `git diff --check 1047332e521d4798a8f168e9609ad3b1e43b9653..HEAD`

Second adversarial pass focused on stale lease metadata, in-flight attempt accounting, tx boundary observability, and the distributed ownership assumption behind `SingleWorkerLeaseless`. I don't see remaining blockers under the documented invariant that leaseless mailboxes have exactly one live runtime owner, or an external singleton/ownership fence if more than one daemon/process could otherwise drain the same mailbox.

Roasbeef:
Follow-up correction: I had updated the package invariant docs, but missed the dedicated `p-models/` tree on the first pass. That is now fixed and pushed in the autosquashed branch (`8e93cfba`).

P-model additions:

- Added explicit leaseless mailbox operations to `p-models/durableactor/src/mailbox_fifo.p`: `PeekNext`, `AckByID`, and `NackByID`.
- Added a green P scenario, `TestDurableMailboxSpec_LeaselessPeekMasksStaleLease`, covering the review bug shape: a row is leased once, the lease expires without maintenance clearing metadata, then the leaseless path peeks it as an empty-token delivery and nacks it by ID.
- Added a bridge trace, `mailbox_leaseless_peek_stale_lease.json`, replayed against the real `db/actordelivery` store.
- Extended the Go bridge to replay `peek`, `ack_by_id`, and `nack_by_id`, and to assert attempts/token values on leased and peeked deliveries.
- Updated the p-model READMEs to document the leaseless empty-token/by-ID retry contract.

Validation run after autosquash:

- `./p-models/scripts/check.sh`
- `go test ./p-models/durableactor/bridge`
- `make unit pkg=./baselib/actor`
- `make unit pkg=./db/actordelivery`
- `make unit pkg=./internal/actortest`
- `make lint-changed-local`
- `make commitmsg-lint range="1047332e521d4798a8f168e9609ad3b1e43b9653..HEAD"`
- `git diff --check 1047332e521d4798a8f168e9609ad3b1e43b9653..HEAD`

The full p-model script now compiles the P project, checks all green cases with 0 bugs, confirms the intentional negative counterexamples still find bugs, and runs the Go bridge conformance replay successfully.

Roasbeef:
@claude review this

PR #728, opened 2026-06-10
multi: cut VTXO listing and coin selection CPU

Roasbeef:
In this PR, we take the CPU cost out of the VTXO listing and coin
selection paths. Profiling the stress workload at ~10/s showed
`ListVTXOsByStatus` alone eating 32% of daemon CPU, almost all of it
decode work and ancestry loads on rows the caller never needed in full.

Coin selection now runs on a lightweight selection-candidate projection
instead of fully materialized VTXOs: a new sqlc query returns just the
fields selection needs, and the manager consumes that shape directly. The
immutable descriptor decode is memoized by outpoint, so the expensive
parse happens once per VTXO lifetime rather than once per listing. RPC
listings skip the ancestry load outright, and `ListVTXOs` callers can opt
out of checkpoint PSBT population via a new `exclude_checkpoint_psbts`
flag (the checkpoint PSBT package loads were the dominant share of the
listing cost). Together these took `ListVTXOsByStatus` from 32% to 4.8% of
CPU and killed the sqlite external-sorter temp-file spill that accounted
for 61% of raw syscalls on the profile.

Two changes here are correctness hardening that the faster listing path
made visible: live VTXOs whose session actor is gone are self-healed
during coin selection (respawned through the registry rather than failing
the selection), and spend reservations are delivered tell-style off the
manager turn so reservation delivery can't deadlock against the manager's
own mailbox.

This PR builds on #727 and is part of the OOR optimization train.

### DIFF
--- a/daemonrpc/daemon.pb.go
+++ b/daemonrpc/daemon.pb.go
@@ -1635,9 +1635,15 @@ type ListVTXOsRequest struct {
 	// If VTXO_STATUS_UNSPECIFIED (default), all statuses are returned.
 	StatusFilter VTXOStatus `protobuf:"varint,1,opt,name=status_filter,json=statusFilter,proto3,enum=daemonrpc.VTXOStatus" json:"status_filter,omitempty"`
 	// min_amount_sat excludes VTXOs below this value.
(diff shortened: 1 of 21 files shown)


Roasbeef:
@claude review this

Roasbeef:
@claude review this

PR #730, opened 2026-06-10
serverconn: fold ingress dispatch and checkpoint into one transaction

Roasbeef:
In this PR, we fold the serverconn ingress path's durable dispatches and
cursor checkpoint into a single write transaction. Before this, a pulled
batch of k envelopes paid k+1 commits: each dispatcher Tell was its own
`EnqueueMessage` commit, then the cursor checkpoint paid one more. The
commit ledger (#729) measured the connection actors' checkpoints at ~21
writes per payment, the single biggest bucket nobody's estimate had
caught.

The fold reuses the ambient-transaction pattern from the outbox fold
(#727): the dispatch Tells join the transaction through the context, so
the cursor can never run ahead of the enqueues. That is an upgrade, not
just a saving. The pre-fold design was only safe by commit ordering; the
fold makes cursor-covers-enqueues atomic. The ack-path checkpoint becomes
lazy and rides the next dispatch commit, with an idle flush for quiet
connections; a crash costs at most one redundant, idempotent AckUpTo.

One contract worth calling out for review: the fold is two-phase. In-memory
KIND_RESPONSE envelopes deliver before the transaction opens, and only
durable dispatches plus the checkpoint fold inside it. Responses are what
unary callers block on with RPC deadlines, so gating them behind the
writer lock turns write contention into payment-wide timeout collapse (we
measured exactly that with a first cut that folded everything).

The second commit adds a P model spec for the no-loss contract: a
persisted cursor must never cover an envelope whose enqueue did not
commit. The spec passes against the folded design under nondeterministic
batch sizes, rollbacks, and crashes, and ships with two counterexamples
(eager cursor after rollback, checkpoint-before-enqueue) that the checker
correctly flags, so the monitor demonstrably has teeth.

On the bench: counted writes per payment fell 77.9 to 59.8 (the fold's own
commit runs on the uninstrumented TxAware path, so the true cut is
~18-20%), p95 fell 12-20%, with 4/4 clean runs at 100% success. The
operator-side clientconn deliberately does not get this fold; its
dispatchers run synchronous handler work, and folding them held the
operator writer lock across whole handlers (darepo#552 tracks the purity
split needed first). It gets the lazy ack only, in the darepo-side PR.

This PR builds on #729 and closes out the OOR optimization train.

### DIFF
--- a/mailbox/conn/response_registry.go
+++ b/mailbox/conn/response_registry.go
@@ -147,6 +147,24 @@ func (r *ResponseRegistry) RemoveWaiter(id CorrelationID) {
 	}
 }
 
+// HasWaiter reports whether an active in-memory waiter is currently registered
+// for correlation ID id. It lets the ingress loop classify a KIND_RESPONSE at
+// split time: a response with a live waiter can be delivered on the fast
+// pre-transaction path, while a response without one must fold into the durable
+// dispatch transaction so its enqueue commits atomically with the cursor. Stale
+// waiters are pruned before the check so an expired entry never masquerades as
+// a live one.
+func (r *ResponseRegistry) HasWaiter(id CorrelationID) bool {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.pruneStaleLocked(time.Now())
+
+	_, ok := r.waiters[id]
+
+	return ok
+}
+
 // RemovePending drops any buffered early response for the correlation ID.
 func (r *ResponseRegistry) RemovePending(id CorrelationID) {
 	r.mu.Lock()
--- a/mailbox/conn/response_registry_test.go
+++ b/mailbox/conn/response_registry_test.go
@@ -124,3 +124,48 @@ func TestResponseRegistry_DeliverNilReturnsFalse(t *testing.T) {
 	require.Equal(t, DeliveryDropped, registry.DeliverResponse("any",
 		nil))
 }
+
+// TestResponseRegistryHasWaiterTracksRegistration verifies HasWaiter reflects
+// the live-waiter state the ingress split path relies on: false before
+// registration, true while a waiter is registered, and false again once the
+// waiter is removed or a buffered (waiterless) response is delivered.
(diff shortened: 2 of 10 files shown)


Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
@claude Dug into the P3 (waiter-delivered response re-dispatched durably on a whole-batch rollback). It's structurally plausible but not reachable in the current wiring — the two conditions it needs are disjoint:

- **Responses with an in-memory waiter** come only from `UnaryFacade`: `RegisterWaiter` has exactly two call sites, both on the low-latency `SendRPC`/`AwaitRPC` path.
- **Responses that durably enqueue** require a registered `cfg.Dispatchers[{service,method}]`; otherwise the waiterless branch is `WarnS("Failed to deliver response envelope")` + `break` — no enqueue (`ingress.go:373-393`).

The OOR indexer query responses named as the example are actually immune: they flow through `SendUnaryRequest` → the EventRouter durable route and never register a waiter, so they sit permanently in the "durable dispatcher, no waiter" set. The intersection (live waiter **and** durable dispatcher) is empty today, so the dup can't fire:

- low-latency RPC response → has a waiter, no dispatcher → warn + advance on the retry, never enqueued;
- durable query response → has a dispatcher, never had a waiter → delivered exactly once, durably.

It is a genuine *latent* edge if a future response type is ever wired into both mechanisms (which the `SendUnaryRequest` design does contemplate — "if no in-memory waiter exists, falls back to durable route"). Even then it's bounded: it's just the response-flavored instance of the already-accepted whole-batch-rollback re-dispatch, and the durable route is the crash-replay path, so its consumer is replay-idempotent and absorbs the dup.

So I'm not adding the cross-iteration "delivered-this-batch" tracking to guard a case that can't currently occur — that's fragile state for no live benefit. Behavior stays as-is. Good shape to flag though; the reasoning holds, just not for the example given.

PR #731, opened 2026-06-10
swaprpc: ack accepted out-swap HTLC events

bhandras:
## Summary

This adds an explicit receiver ACK for out-swap HTLC events.

Before this change, the receive side only ACKed the mailbox cursor after it accepted an HTLC event. That proved the message was consumed, but it did not give the swap server a durable signal that the receiver had validated and persisted the event before the server funded the Ark vHTLC.

The new `AcknowledgeOutSwapHtlc` RPC lets the receiver ACK the event back to the swap server after the mailbox ACK succeeds. The receive session keeps the existing retry boundary: if the mailbox ACK or server ACK fails, the pending cursor remains and resume retries the missing acknowledgement. Same-Ark receives skip the server ACK because the swap server does not fund those vHTLCs.

This PR is the client-side half of the fix for lightninglabs/swapdk-server#29. The matching swapdk-server PR updates the server state machine to wait for this ACK before funding.

## Stack

This PR is based on lightninglabs/darepo-client#717 (`max-vtxo-amt`). The server currently needs the newer darepo/client API surface from that stack, so keeping this PR stacked avoids reintroducing the stale submodule baseline that swapdk-server was pinned to.

## Tests

- `go test ./sdk/swaps -count=1`
- `go test ./rpc/restclient -count=1`

### DIFF
--- a/rpc/restclient/clients.go
+++ b/rpc/restclient/clients.go
@@ -188,6 +188,20 @@ func (c *SwapServiceClient) AuthorizeInSwapRefund(ctx context.Context,
 	return out, err
 }
 
+// AcknowledgeOutSwapHtlc tells the swap server an out-swap receiver durably
+// accepted the HTLC event.
+func (c *SwapServiceClient) AcknowledgeOutSwapHtlc(ctx context.Context,
+	in *swaprpc.AcknowledgeOutSwapHtlcRequest, _ ...grpc.CallOption) (
+	*swaprpc.AcknowledgeOutSwapHtlcResponse, error) {
+
+	out := new(swaprpc.AcknowledgeOutSwapHtlcResponse)
+	err := c.client.Post(
+		ctx, "/v1/swap/acknowledge-out-swap-htlc", in, out,
+	)
+
+	return out, err
+}
+
 // NewDaemonServiceClient creates a DaemonService REST client.
 func NewDaemonServiceClient(addr string,
 	opts ...Option) daemonrpc.DaemonServiceClient {
--- a/sdk/swaps/client.go
+++ b/sdk/swaps/client.go
@@ -371,6 +371,11 @@ type SwapServerConn interface {
 		paymentHash lntypes.Hash, amountSat btcutil.Amount,
 		expirySeconds uint32) (*OutSwapQuote, error)
 
+	// AcknowledgeOutSwapHTLC tells the server this receiver validated and
+	// durably accepted the out-swap HTLC event.
+	AcknowledgeOutSwapHTLC(ctx context.Context, paymentHash lntypes.Hash,
+		vhtlcPubkey *btcec.PublicKey) error
+
 	// CreateInSwap initiates an Ark->LN swap on the server.
 	CreateInSwap(ctx context.Context, invoice string, maxFeeSat uint64,
 		clientVhtlcPubkey *btcec.PublicKey) (*InSwapConfig, error)
--- a/sdk/swaps/grpc_conn.go
+++ b/sdk/swaps/grpc_conn.go
@@ -79,6 +79,30 @@ func (g *GRPCSwapServerConn) RequestChannelID(ctx context.Context,
 	}, nil
 }
 
+// AcknowledgeOutSwapHTLC tells the swap server this receiver validated and
+// durably accepted the out-swap HTLC event.
+func (g *GRPCSwapServerConn) AcknowledgeOutSwapHTLC(ctx context.Context,
+	paymentHash lntypes.Hash, vhtlcPubkey *btcec.PublicKey) error {
+
+	if vhtlcPubkey == nil {
+		return fmt.Errorf("vHTLC pubkey must be provided")
+	}
+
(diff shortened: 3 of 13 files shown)


bhandras:
@claude pls review. Context: this is the client-side half of lightninglabs/swapdk-server#29. It adds AcknowledgeOutSwapHtlc and has receive sessions call it after mailbox ACK succeeds, so the swap server can wait for receiver acceptance before funding. Same-Ark receives skip the server ACK because swapdk-server does not fund those vHTLCs.

bhandras:
Review context on the ACK retry contract:

The unconditional client retry is intentional. The receive session keeps `pendingHTLCAckCursor` until both the mailbox ACK and server ACK have succeeded, so a crash anywhere before cursor-clear replays the ACK sequence.

The matching server PR now pins the load-bearing idempotency contract explicitly:

- duplicate ACKs return OK for already-ACKed, funding-initiated, funded, timed-out, refund-initiated, completed, failed, refunded, and needs-intervention states;
- ACKs cannot overwrite a concurrent no-ACK failure because state-only transitions now use a guarded state update;
- ACKs arriving after the settle buffer fail the swap instead of moving it to `HTLCEventAcked`.

For the empty `settlementType` case: that is intentional backward-compatible behavior for any older persisted receive rows that predate the explicit settlement type. New Lightning and same-Ark paths persist `settlementType` in the same mutation that records the pending cursor, and same-Ark sessions skip the server ACK.

The Gemini note is against the carried multi-recipient SendOOR commits that this branch inherits from the current swapdk-server submodule pointer; it is outside the ACK change in `b164449`. I’m leaving it out of this ACK PR unless we decide to respin those carried commits here too.

bhandras:
@claude pls re-review with the context above. The matching server PR now explicitly makes duplicate ACKs idempotent for already-advanced and terminal states, guards state-only transitions so ACK cannot overwrite a no-ACK failure, and fails ACKs that arrive after the settle buffer instead of advancing to funded. Please focus on whether the client retry/durability behavior is now acceptable given that server contract.

bhandras:
@claude pls re-review. Context: this PR was rebuilt as a stacked PR on #717 (`max-vtxo-amt`) so the diff is now one ACK commit (`91e573e5`) relative to that base. The swapdk-server side depends on this RPC and now points its submodule at this head.

bhandras:
Replying to the late-ACK wire-contract point: the server PR now pins this explicitly with `TestAcknowledgeOutSwapHtlcLateAckFailsSwapWithOK`. If the ACK arrives after the settle buffer, the server fails the swap durably but the ACK RPC returns OK, which lets the client clear its pending cursor instead of retrying a terminal late ACK forever. So the client-side retry boundary remains unchanged.

bhandras:
@claude FYI latest force-push is formatting-only after local `make fmt-check`; head is now `627ae894`. The ACK flow and tests are unchanged from your last review. Local validation: `make fmt-check`, `make tidy-module-check`, `go test ./sdk/swaps -count=1`, and `go test ./rpc/restclient -count=1` pass.

bhandras:
@claude pls re-review current head `83ee54e2`. Context from the local review: client-side server ACK errors were previously all retried forever. Folded fix: `acknowledgeOutSwapHTLC` now classifies authoritative server rejections (`InvalidArgument`, `NotFound`, `PermissionDenied`) as terminal receive failures, while keeping `FailedPrecondition` retryable for the server not-yet-published state. Added `TestReceiveSessionClassifiesOutSwapServerAckStatus`. Local validation: `make fmt-check`, `make tidy-module-check`, `go test ./sdk/swaps -count=1`, and `go test ./rpc/restclient -count=1` pass.

bhandras:
@claude pls re-review latest head `ca6ce820`. Context: this force-push is lint-only after the previous LGTM. It fixes the test fake receiver style in `grpc_conn_test.go` and wraps the long `VHTLCConfig` literal in `out_swap_test.go`; no ACK behavior changed. Local validation: `make fmt-check`, `make tidy-module-check`, `go test ./sdk/swaps -count=1`, `go test ./rpc/restclient -count=1`, and `make commitmsg-lint` pass.

bhandras:
@claude pls re-review latest head `133d1eb4`. Context: this force-push is lint-only after your previous LGTM. It avoids gofmt expanding the long `VHTLCConfig` literal in `out_swap_test.go` by assigning the two long fields after construction; ACK behavior is unchanged. Local validation: `make fmt-check`, `make tidy-module-check`, `go test ./sdk/swaps -count=1`, `go test ./rpc/restclient -count=1`, and `make commitmsg-lint` pass.

bhandras:
@claude pls re-review latest head `7ff74240`.

Context: this PR has been rebuilt onto latest `main` instead of being stacked on `max-vtxo-amt`. The branch now contains only the ACK commit, and the code change is the same out-swap receiver ACK behavior already reviewed.

Local checks after rebase:
- `make fmt-check`
- `make tidy-module-check`
- `go test ./sdk/swaps -count=1`
- `go test ./rpc/restclient -count=1`
- `make commitmsg-lint`

PR #732, opened 2026-06-10
db: Generalize wallet key triples into a shared internal_keys table

### LINKED ISSUE #706: db: Introduce a shared internal_keys table for KeyLocator+pubkey triples (FK refactor)
## Summary

Several tables in the client DB inline the same key-locator triple: a
33-byte compressed `*_pubkey` BLOB plus a `*_key_family` / `*_key_index`
pair (an lnd `keychain.KeyLocator`). The triple lets us reconstruct the
signing descriptor for a wallet key, so it rides along with anything we
have to re-sign or re-derive after the fact. Today each table re-spells
those columns and each store re-implements the same encode/decode in Go.

Unlike the server (which already has an `operator_keys` table that `vtxos`
FK-references), the client has **no canonical key table** — this is a
greenfield refactor with no existing FK pattern to follow.

## Proposal

Introduce a shared `internal_keys` table keyed by `(pubkey, key_family,
key_index)` with UNIQUE/CHECK guards (one canonical row per key; reject
inconsistent re-registration), and have the tables below reference it via
a nullable `*_key_id` FK instead of inlining three columns. Mirror the
server's `operator_keys` design so the two repos converge on one shape.

## Inlined triples to migrate

Ranked by row volume / payoff.

- [ ] **`vtxos`** — `client_pubkey` / `client_key_family` /
  `client_key_index` (+ `operator_pubkey`), migration 000003. Highest
  volume table; biggest win. `client_*` is the local-ownership key,
  `operator_pubkey` the co-signer.
- [ ] **`round_vtxo_requests`** — `signing_pubkey` /
  `signing_key_family` / `signing_key_index`, plus `owner_key_family` /
  `owner_key_index` (sentinel `-1` for foreign-owned VTXOs), plus
  `client_pubkey` / `operator_pubkey`, migration 000003. Two distinct
  locators (owner vs. signing) → two FKs; owner FK nullable.
- [ ] **`boarding_addresses`** — `client_pubkey` / `client_key_family` /
  `client_key_index` (+ `operator_pubkey`), migration 000002. Low volume
  but a clean complete triple.
- [ ] **`owned_receive_scripts`** — `client_pubkey` / `client_key_family`
  / `client_key_index` (+ `operator_pubkey`), migration 000004.
  **Type inconsistency:** uses `BIGINT` for family/index while the other
  tables use `INTEGER`. The FK refactor is a good moment to settle on one
  (`BIGINT`, since `KeyLocator` fields are `uint32`).
- [ ] **`vhtlc_recovery_jobs`** — `signer_key_family` / `signer_key_index`
  + `signer` pubkey, migration 000015. Only the *signer* key is a wallet
  locator triple → one FK. `sender_pubkey` / `receiver_pubkey` /
  `server_pubkey` are policy participants (no locator) → leave inline.

## Pubkey-only — likely leave as-is

These store a pubkey with no locator, so there is no triple to normalize
unless they later need a signing descriptor.

- `round_boarding_intents` — `client_key` / `operator_key` (000003).
- SDK swap sessions `receive_swaps` / `pay_swaps`
  (`sdk/swaps/migrations/000001`) — policy/destination pubkeys only.

## Notes

- Settle the `INTEGER` vs `BIGINT` divergence as part of this; standardize
  on `BIGINT` to match `keychain.KeyLocator`'s `uint32` fields.
- Go stores to update: `db/vtxo_store.go`, `db/round_store.go`,
  `db/boarding_wallet.go`, `db/oor_artifact_store.go`,
  `db/vhtlc_recovery_store.go`.
- Sibling tracking issue for the server repo (which already has the
  `operator_keys` / `vtxos.operator_key_id` pattern to generalize):
  lightninglabs/darepo#542.

ellemouton:
Closes #706.

Client twin of the server's internal_keys FK refactor (darepo #542/#549). Introduces a shared `internal_keys` table and converts the inlined key-locator triples (33-byte `*_pubkey` BLOB + `*_key_family` + `*_key_index`) into a nullable `*_key_id` FK.

## internal_keys
Keyed by `(pubkey, key_family, key_index)` with `UNIQUE` + `CHECK(length(pubkey)=33)`, BIGINT family/index. Unlike the server, the client has no key *role* concept, so there is no role column or roles table — the canonical identity is the bare triple. Registration is idempotent + race-free via `ON CONFLICT (...) DO UPDATE ... RETURNING id`. Created at the top of migration `000002`, before its earliest referencer (`boarding_addresses`), so the Postgres FK target exists when each constraint is declared. New Go helpers `RegisterInternalKeyTx` / `InternalKeyDescByIDTx` mirror the server's `db/internal_key.go`.

## Tables converted
- `vtxos` → `client_key_id` (`operator_pubkey` left inline — pubkey only, no locator).
- `round_vtxo_requests` → `signing_key_id` + nullable `owner_key_id` (NULL replaces the old `-1/-1` foreign-owned locator sentinel); `client_pubkey`/`operator_pubkey` left inline.
- `boarding_addresses` → `client_key_id`.
- `owned_receive_scripts` → `client_key_id`; settled the INTEGER/BIGINT divergence the issue noted onto BIGINT.

All folded into the existing CREATE TABLE migrations (pre-GA, deployed DBs are nuked — no forward migrations).

## Deviation from the issue: `vhtlc_recovery_jobs` NOT converted
The issue lists converting `signer_key_family`/`signer_key_index` **plus a `signer` pubkey** into `signer_key_id`. But there is no signer pubkey — neither a column in migration `000015` nor a `SignerPubKey` field on `vhtlcrecovery.RecoveryJob` (only `SignerKeyFamily`/`SignerKeyIndex`, a bare locator populated from the RPC request). `internal_keys` mandates a NOT-NULL 33-byte pubkey as its canonical anchor, so there is no key material to register. I left `vhtlc_recovery_jobs` unchanged rather than invent a pubkey or relax the registry invariant. The `sender`/`receiver`/`server` pubkeys there are policy participants (no locator) and stay inline, as the issue intended. Happy to revisit if the signer pubkey can be sourced.

## Verification
- `make sqlc` regenerates cleanly; `go build` clean.
- db unit tests pass on **both** backends: sqlite and postgres (postgres exercises the FK declarations + ordering).
- Consumer packages (`round`, `vtxo`, `oor`, `wallet`) pass on sqlite.
- `lint-changed-local` 0 issues; `commitmsg-lint` clean.


ellemouton:
@claude review this

ellemouton:
Thanks @claude — addressed (force-pushed `b803dc1c` → `9f7fdb73`):

**1. Signing-key locator narrowing — confirmed intentional + safe.** `SigningKey` is the client's per-round MuSig2 cosigning key (family `VTXOSigningKeyFamily = 45`); a locally-composed `VTXORequest` always carries its pubkey (it has to, to sign the VTXO tree). The narrowing only affects a hypothetical nil-pubkey request, and there dropping the lone locator is harmless: a MuSig2 signing descriptor is unusable without the pubkey, so the old behavior (persisting a locator with a nil pubkey) carried nothing reconstructable downstream. The registry's pubkey-NOT-NULL anchor makes that tightening explicit.

**2. Owner-key gate vs. registered pubkey — intentional, documented.** The gate `req.OwnerKey.PubKey != nil` is the "is this request locally owned?" signal (a foreign-owned request leaves `owner_key_id` NULL, replacing the old `-1/-1` sentinel). The registered descriptor deliberately pairs the *policy* owner key (`params.OwnerKey`, formerly the `client_pubkey` column) with the *request*'s owner locator — that's exactly the descriptor the old read path reconstructed, so it's preserved. As you note, it's strictly safer than `main` (which would have nil-panicked). The field-sourcing rationale is in the comment at `round_store.go:1225-1230`.

**3. Concurrent multi-key registration deadlock — already covered by the retry layer.** `parsePostgresError` maps `pgerrcode.DeadlockDetected` (40P01) → `ErrDeadlockError` (`db/sqlerrors.go:136`), and `ExecTx` retries the whole transaction on `IsSerializationOrDeadlockError` (`db/interfaces.go:305`), which includes deadlock — with the standard 10× exponential backoff. SQLite serializes writes so it's a Postgres-only concern, and it's retryable there.

**4. Orphaned `internal_keys` rows — acknowledged, pre-GA.** Keys are immutable shared identity and the table only grows; no GC for now, consistent with the server side. Worth a follow-up if it ever matters.

**5. Aspirational nullable-FK comments — fixed.** Reworded both (`000002_boarding_tables.up.sql` and `000004_oor_artifact_store.up.sql`) to say the column is declared nullable only for uniformity with the genuinely-optional FKs (`vtxos`, `round_vtxo_requests`); in practice every boarding address / owned receive script has a client key, so the write path always registers it first and the read path treats a NULL as an error.

Folded the two comment rewords into their owning FK commits. The server submodule bump on darepo#558 was re-pointed at the new tip `9f7fdb73`.

PR #28, opened 2025-11-27
Closures

sputn1ck:
##  Summary

  This PR introduces a flexible closure-based system for VTXO (Virtual Transaction Output) scripts, replacing the hardcoded script
  construction with a modular architecture that supports custom spend conditions.

  - Add closure package (lib/closure/) - Core closure types (CSVSigClosure, CSVMultisigClosure, MultisigClosure, CLTVMultisigClosure,
  ConditionMultisigClosure, ConditionCSVMultisigClosure) with Script/Decode/Witness methods
  - Replace existing vtxo scripts with closures - Migrate lib/scripts/, lib/tree/, and lib/tx/ to use the new closure system
  - Add DefaultVTXOTapScript for compatibility - Bridge function for code requiring waddrmgr.Tapscript type

 ## Motivation

  The previous VTXO implementation had hardcoded 2-of-2 multisig + CSV timeout structure. This PR enables:
  - Custom VTXO scripts with arbitrary closure combinations
  - Additional spend conditions (CLTV timelocks, conditional scripts)
  - Flexible validation rules that can be extended per-closure type

 ## Key Changes

  New lib/closure/ Package

  | File           | Purpose                                                            |
  |----------------|--------------------------------------------------------------------|
  | closure.go     | 6 closure types with Script/Decode/Witness methods                 |
  | vtxo_script.go | TapscriptsVtxoScript container with Validate/TapTree/Encode/Decode |
  | locktime.go    | BIP68 relative and absolute locktime encoding                      |
  | script.go      | Script evaluation helpers                                          |
  | *_test.go      | Comprehensive test coverage (~2100 lines)                          |

 ## Closure Types

  - CSVSigClosure - Single-sig exit with CSV timelock
  - CSVMultisigClosure - Multi-sig exit with CSV timelock
  - MultisigClosure - Collaborative path (CHECKSIG or CHECKSIGADD variants)
  - CLTVMultisigClosure - Collaborative with absolute timelock
  - ConditionMultisigClosure - Collaborative with custom condition script
  - ConditionCSVMultisigClosure - Exit with condition + CSV timelock

 ## Validation Rules

  - Server/signer key must be present in all forfeit (collaborative) closures
  - Exit closures must meet minimum locktime requirement
  - Block-type locktimes can be restricted via parameter

PR #737, opened 2026-06-12
swaps: accept MPP out-swap events

bhandras:
## Summary

This PR adds the darepo-client side of MPP out-swap receive.

Before this change, the client treated an out-swap HTLC event as one Lightning payment part with one final-hop onion blob. That is enough for a single-part payment, but it does not describe a Lightning MPP payment. In an MPP receive, LND delivers several HTLC shards. Each shard has its own amount and onion payload, but all shards belong to one out-swap receive and one Ark vHTLC.

The client now understands that shape. It can accept a mailbox event that carries several HTLC parts, validate each part, and only then durably accept and acknowledge the out-swap event.

## New model

For one receive invoice:

```text
single-part payment -> legacy onion_blob path
multi-part payment  -> parts[] with one onion blob and amount per HTLC shard
accepted event      -> every part validates and the part amounts sum to the invoice amount
```

The legacy path stays intact. If `parts` is empty, the client treats `onion_blob` as the single payment part, and that one part must forward the full invoice amount on its own.

For MPP events, each part carries the millisatoshi amount for that shard and the raw final-hop onion blob for that shard. The receive session decodes every onion with the receive auth key, checks that every shard belongs to the expected invoice, and checks that the aggregate forwarded amount matches the out-swap amount before accepting the event.

## What changes

1. Adds `OutSwapHtlcPart` to the client model and adds a repeated `parts` field to `OutSwapHtlcEvent` in the wire proto.
2. Keeps `onion_blob` as the compatibility field for legacy single-part events. When `parts` is populated, the proto still carries the whole out-swap amount in `amount_sat`.
3. Decodes mailbox out-swap events into either the legacy single-onion shape or the new per-part shape.
4. Advertises optional MPP support on generated receive invoices while preserving the existing route-fee hint behavior.
5. Validates every HTLC shard against the expected payment hash, payment address, MPP total, and per-part forwarded amount.
6. Accepts the receive event only when the decoded shard amounts sum to the invoice amount.
7. Adds coverage for valid MPP parts, legacy single-part events, underpaid legacy events, malformed MPP sets, and invoice feature bits.

## Why this is safe

The client still fails closed at the receive-event boundary. It does not acknowledge the mailbox event until the out-swap HTLC event has been validated and durably accepted.

For an MPP event, every shard must decode with the receive auth key and bind to the expected payment hash. Every shard must also carry the expected payment address and MPP total. The plaintext part amount from the server must match the amount committed in that shard's onion, so the server cannot lie about the split.

The aggregate check is still required. A set of individually valid shards is accepted only if the forwarded amounts add up to the invoice amount. A missing part, underpayment, overpayment, bad payment address, bad MPP total, zero-amount shard, or mismatched part amount rejects the event.

Single-part events remain compatible because an event without `parts` still uses the old `onion_blob` field. That legacy shard must forward the full invoice amount.

The funding and aggregation side lives in swapdk-server PR #123. That server PR collects the held Lightning shards, funds one Ark vHTLC after the invoice total is covered, and publishes the per-part onions that this client PR validates.

## Validation

- `go test ./sdk/swaps -run 'TestReceiveSessionValidateOnionPayload|TestInvoiceGeneratorIncludesRouteFeeHints|TestMailboxOutSwapEventReceiverPullsAndAcks' -count=1`
- `go test ./sdk/swaps ./swaprpc -count=1`
- `make commitmsg-lint commit=d6c56801`
- `make commitmsg-lint commit=57124441`
- `git diff --check`

## Dependency / coordination

This is the darepo-client side of MPP out-swap receive and pairs with lightninglabs/swapdk-server#123.

Expected order:

1. merge this client PR
2. update the swapdk-server #123 client submodule pointer to the merged client commit
3. merge the server-side aggregation and funding changes in swapdk-server #123

### DIFF
--- a/sdk/swaps/client.go
+++ b/sdk/swaps/client.go
@@ -14,6 +14,7 @@ import (
 	"github.com/lightningnetwork/lnd/invoices"
 	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
 )
 
 // SwapDirection identifies which Lightning/Ark direction one persisted swap
@@ -264,6 +265,23 @@ type OutSwapHtlcEvent struct {
 
 	// VHTLCConfig contains the script parameters for the funded vHTLC.
 	VHTLCConfig VHTLCConfig
+
+	// Parts lists the individual HTLC shards of a multi-part payment set.
+	// When empty the event is a legacy single-part payment carried by
+	// OnionBlob.
+	Parts []OutSwapHtlcPart
+}
+
+// OutSwapHtlcPart carries one HTLC shard of a multi-part out-swap payment
+// set. Each shard has its own final-hop onion that the client validates
+// before acknowledging the event.
+type OutSwapHtlcPart struct {
+	// AmountMsat is the millisatoshi amount forwarded by this shard.
+	AmountMsat lnwire.MilliSatoshi
+
+	// OnionBlob is the raw final-hop onion blob forwarded by the server
+	// for this shard.
+	OnionBlob []byte
 }
 
 // OutSwapHtlcNotification carries one mailbox-delivered out-swap HTLC event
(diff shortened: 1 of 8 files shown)


bhandras:
@claude pls review. Context: this is the darepo-client/proto half of swapdk-server issue #28 (MPP receive). It adds optional MPP invoice support and validates per-part out-swap HTLC event onions while preserving the legacy single-onion event path. The server PR is lightninglabs/swapdk-server#123 and points its client submodule at this branch.

PR #743, opened 2026-06-15
multi: Add Ark protocol version negotiation and enforcement

ellemouton:
## Summary

PR A of the protocol-versioning work tracked by
https://github.com/lightninglabs/darepo/issues/526. It owns the protobuf
contract and all client-side behavior for the **Ark session protocol** version.
Production behavior remains Ark v1; this PR does not implement Ark v2 semantics.

## Scope

Ark **session protocol** negotiation and admission **only**. It does **not**
implement VTXO construction-version or script-version semantics (how a specific
VTXO is built and spent) — that is a separate, out-of-scope future concern.
There is no deprecation lifecycle, disable deadline, upgrade-URL signalling,
mailbox transport v2, or old-server fallback.

## What the client does

- advertises its supported Ark protocol versions during the direct `GetInfo`
  bootstrap;
- refuses startup unless the operator selects a compatible, non-disabled,
  non-zero version (an empty or no-overlap response is fatal — there is no
  fallback to v1);
- binds the selected Ark version and the mailbox transport version immutably to
  the mailbox runtime;
- stamps that bound pair onto every outbound envelope (durable replay, unary
  calls, events, heartbeats, and locally built responses);
- validates every inbound envelope against the bound pair before dispatch or
  acknowledgement;
- treats version failures as permanent: stops retries and ingress, fails
  pending callers, dead-letters the failed durable send, and reports the daemon
  disconnected;
- pins later `GetInfo` refreshes to the bound version and never renegotiates.

## Approach / structure

All version/compatibility logic is centralized in `serverconn`, the mailbox
connector, rather than spread across the daemon and the individual send/recv
paths:

- **Negotiation lives in `serverconn`.** `ArkVersionNegotiator` owns bootstrap
  version selection over the operator's direct `GetInfo`, and
  `ValidateRefreshSelection` owns the refresh-only check. The decision logic is
  pure and unit-tested; the daemon only supplies the gRPC client and parses
  domain terms from the response.
- **One response/status check.** Every dispatch/recv path funnels its edge
  outcome through a single `edgeResponseError` helper (transport error →
  nil-response guard → non-OK status → typed `StatusError`) instead of
  repeating the same checks at each call site.
- **Stamping via an edge interceptor.** Outbound version stamping is applied in
  one interceptor-style decorator wrapped around the mailbox edge (mirroring the
  existing auth decorator), so no individual send path can forget to stamp.

## Wire contract

`ArkVersionPolicy` carries a `version` and an `ACTIVE`/`DISABLED` `state`.
`GetInfoRequest` carries the client's `supported_ark_versions`.
`GetInfoResponse` carries `selected_ark_version` and the per-version
`ark_version_policies` — the operator's enabled set is simply its `ACTIVE`
policies, so there is no separate supported-versions response field. Every
`Envelope` carries `ark_protocol_version`. Permanent status codes:
`MAILBOX_VERSION_UNSUPPORTED`, `ARK_VERSION_UNSUPPORTED`, `ARK_VERSION_MISMATCH`,
`UPGRADE_REQUIRED`. Generated stubs come from `make rpc`.

### DIFF
--- a/arkrpc/ark.proto
+++ b/arkrpc/ark.proto
@@ -15,6 +15,37 @@ service ArkService {
 }
 
 message GetInfoRequest {
+    // supported_ark_versions lists the Ark protocol versions this client can
+    // use, ordered by client preference (most preferred first). The operator
+    // selects the first operator-preferred version present in this list. An
+    // empty list means the client advertises no Ark version support, so it
+    // overlaps with nothing and the operator selects zero.
+    repeated uint32 supported_ark_versions = 1;
+}
+
+// ArkVersionPolicy advertises the operator's lifecycle policy for a single Ark
+// protocol version: whether it is ACTIVE (selectable) or DISABLED (advertised
+// so clients learn it was retired, but never selected or accepted).
+message ArkVersionPolicy {
+    // State is the lifecycle state of an Ark protocol version.
+    enum State {
+        // STATE_UNSPECIFIED is the zero value and must not be used as a real
+        // policy state.
+        STATE_UNSPECIFIED = 0;
+
+        // STATE_ACTIVE means the version may be selected and admitted.
+        STATE_ACTIVE = 1;
+
+        // STATE_DISABLED means the version must never be selected or accepted;
+        // it is advertised only so clients learn the version was retired.
+        STATE_DISABLED = 2;
+    }
+
+    // version is the Ark protocol version this policy describes.
+    uint32 version = 1;
+
+    // state is the lifecycle state of the version.
+    State state = 2;
 }
 
 message GetInfoResponse {
@@ -93,6 +124,17 @@ message GetInfoResponse {
     // receive and boarding flows before funds enter the system. Zero
     // means no cap.
     int64 max_user_balance = 20;
+
+    // selected_ark_version is the Ark protocol version the operator selected
+    // for this client from its supported_ark_versions request. Zero means no
+    // common version exists; the client must not start its mailbox runtime.
+    uint32 selected_ark_version = 21;
+
+    // ark_version_policies carries the lifecycle policy (ACTIVE or DISABLED)
+    // for each of the operator's known Ark protocol versions. The ACTIVE
+    // entries are the operator's enabled, selectable versions; DISABLED
+    // entries are advertised only so clients learn a version was retired.
+    repeated ArkVersionPolicy ark_version_policies = 22;
 }
 
 // EstimateFeeRequest asks the server to estimate the fee for a
--- a/darepod/operator_negotiation.go
+++ b/darepod/operator_negotiation.go
@@ -1,0 +1,148 @@
+package darepod
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/lightninglabs/darepo-client/arkrpc"
+	"github.com/lightninglabs/darepo-client/lib/types"
+	mailboxconn "github.com/lightninglabs/darepo-client/mailbox/conn"
+	"github.com/lightninglabs/darepo-client/serverconn"
+)
+
+// arkVersionNegotiation is the outcome of the bootstrap GetInfo negotiation.
+// connectAndBootstrapMailbox is the sole owner of this negotiation: it caches
+// every value below before constructing the mailbox runtime, so the later
+// round-actor initialization and refresh-only GetInfo calls never renegotiate.
+type arkVersionNegotiation struct {
+	// operatorPubKey is the operator's main public key.
+	operatorPubKey *btcec.PublicKey
+
+	// selectedArkVersion is the Ark protocol version to bind to the
+	// runtime. It is always non-zero on a successful negotiation.
+	selectedArkVersion uint32
+
+	// terms is the initial operator terms parsed from the same response, so
+	// the round actor does not need a second negotiating call.
+	terms *types.OperatorTerms
+}
+
+// operatorTermsFromResponse parses operator terms from a GetInfo response. It
+// is shared by the bootstrap negotiation and the refresh-only path so both
+// produce identical terms from the same fields.
+func operatorTermsFromResponse(resp *arkrpc.GetInfoResponse) (
+	*types.OperatorTerms, error) {
+
+	if resp == nil {
+		return nil, fmt.Errorf("nil GetInfo response")
+	}
+
+	if len(resp.Pubkey) == 0 {
+		return nil, fmt.Errorf("operator pubkey is missing")
+	}
+
+	pubKey, err := btcec.ParsePubKey(resp.Pubkey)
+	if err != nil {
+		return nil, fmt.Errorf("parse operator pubkey: %w", err)
+	}
+
+	minVTXOAmount := resp.MinVtxoAmountSat
+	if minVTXOAmount < resp.DustLimit {
+		minVTXOAmount = resp.DustLimit
+	}
+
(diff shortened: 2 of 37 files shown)


ellemouton:
@claude review this

arkrpc/ark.proto
```diff
@@ -15,6 +15,37 @@ service ArkService {
 }
 
 message GetInfoRequest {
+    // supported_ark_versions lists the Ark protocol versions this client can
+    // use, ordered by client preference (most preferred first). The operator
+    // selects the first operator-preferred version present in this list. An
+    // empty list means the client advertises no Ark version support, so it
+    // overlaps with nothing and the operator selects zero.
+    repeated uint32 supported_ark_versions = 1;
```
Roasbeef, line 23: Instead of a list of versions, what if we went with something more granular here? Eg more decomposed by features, similar to how things are on LN. 

So we'd declare an enum type here of diff protocol shapes/features, then make this a slice of that. 

As an example, maybe a future version has a distinct way of handling dust recv/send. If they client shows that they understand that, then rather than reject dust, the server would utilize the new protocol. 

We may be able to copy some proto fragments from lnd if we like this direction.

arkrpc/ark.proto
```diff
@@ -81,6 +112,19 @@ message GetInfoResponse {
...
+    // common version exists; the client must not start its mailbox runtime.
+    uint32 selected_ark_version = 19;
+
+    // supported_ark_versions lists the Ark protocol versions the operator
+    // currently has enabled, in operator-preference order.
+    repeated uint32 supported_ark_versions = 20;
+
+    // ark_version_policies carries the lifecycle policy (ACTIVE or DISABLED)
+    // for each of the operator's known Ark protocol versions.
+    repeated ArkVersionPolicy ark_version_policies = 21;
```
Roasbeef: Could combine this field into a single struct as IICU it's 1:1 with `supported_ark_versions`?

arkrpc/ark.proto
Roasbeef: Alternatively, as this is a client+sever protocl, a single version may be sufficient, as in the short term we only expect a single client that won't sparsely implement new featurse.

darepod/operator_negotiation.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/lightninglabs/darepo-client/lib/types"
+	mailboxconn "github.com/lightninglabs/darepo-client/mailbox/conn"
+	mailboxpb "github.com/lightninglabs/darepo-client/mailbox/pb"
+)
+
+// arkVersionNegotiation is the outcome of the bootstrap GetInfo negotiation.
+// connectAndBootstrapMailbox is the sole owner of this negotiation: it caches
+// every value below before constructing the mailbox runtime, so the later
+// round-actor initialization and refresh-only GetInfo calls never renegotiate.
+type arkVersionNegotiation struct {
```
Roasbeef, line 20: We could put this in `serverconn` perhaps?

serverconn/runtime.go
```diff
@@ -75,6 +97,21 @@ func NewRuntime(cfg ConnectorConfig) (*Runtime, error) {
...
 
+// permanentAwareTellRetryPolicy is the durable actor's retry policy for
+// serverconn egress. A permanent version error is non-retryable, so the
+// failing durable message is dead-lettered immediately rather than retried
+// forever; every other (transient) failure keeps the default
+// exponential-backoff schedule.
+func permanentAwareTellRetryPolicy(err error,
+	attempts int) (bool, time.Duration) {
+
+	if mailboxconn.IsPermanentVersionError(err) {
```
Roasbeef, line 108: Nice, important to ensure that we just shed load of incompatible clients rather than having them continue to hammer the server

ellemouton:
@claude review this

ellemouton:
Thanks for the thorough re-review! Addressed the findings and force-pushed — fixes folded into the relevant commits so the stack stays 3 commits.

**🔴 Permanent inbound mismatch never drove the terminal transition on the transactional path** — fixed. The `txOK` branch of `ingressLoop` now calls `checkPermanentStatus(foldErr)` before backing off, mirroring the legacy `dispatchBatch` path, so a permanent inbound mismatch drives `markIncompatible` (caches the error, cancels ingress, `FailAll`, `OnIncompatible`) instead of re-pulling the same batch forever. Confirmed the production `ExecTx` (`db/actordelivery/store_impl.go`) returns the closure error unwrapped, so `errors.As` still resolves the `*StatusError`. Added `TestRunFoldedDispatchRejectsMismatchedEnvelope` that drives an inbound mismatch through `runFoldedDispatch` and asserts the transition fires.

**🟠 Waiter-backed responses skipped `validateInboundEnvelope` on the tx path** — fixed. `runFoldedDispatch` now validates every split-time response envelope before in-memory delivery; a mismatch surfaces as the same permanent error and drives the transition (covered by the new test's waiter-response case).

**🟡 Reserve `Status` tag 6** — done. Added `reserved 6; reserved "upgrade_url";` to the `Status` message and regenerated.

Verified locally: build, full-tree `make lint-native` (0 issues), and serverconn `-race` all green. Thanks also for confirming Roasbeef's three design points are resolved.

litbot-9000:

Roasbeef (approved):
## Approve

Reviewed via a multi-agent pass (code quality, offensive security, differential/regression, deep function analysis, spec compliance, API safety). **Zero Critical, zero High.** Core invariants independently confirmed by multiple agents: the version pair binds immutably, zero/disabled/downgrade selections all fail closed, every inbound path is validated before ack (both the durable `dispatchBatch` partition and the waiter-response partition in `runFoldedDispatch`), permanent errors dead-letter instead of retry-storming, and the line-by-line diff of the rewritten files (`ingress.go`, `actor.go`, `server.go`, `unary_facade.go`) showed **no dropped validation/ack-cursor/retry/cleanup behavior**. Spec compliance came back full_match on all 7 behavioral claims + 5 wire-contract clauses. Nice work.

Approving — none of the below block merge. Three follow-up items worth addressing:

### M-1 — Response-path stamping bypasses the central stamping decorator
The "no send path can forget to stamp" invariant has one structural exception: `handleInboundRPC` sends through the **raw, undecorated** edge at `darepod/server.go:3235` and relies solely on the explicit `s.runtime.StampEnvelope(responseEnv)` at `darepod/server.go:3233` (dispatchers are constructed before `NewServerConnectionActor` wraps `cfg.Edge`). Correct today and idempotent, but a future locally-built `edge.Send` on this path would silently ship `ark_protocol_version=0`. Suggest routing this path through the decorator, or adding a prominent comment at the raw-edge send documenting the manual-stamp invariant.

### M-3 — Operator-controlled `ark_version_policies` list is unbounded
`selectedArkPolicy` (`serverconn/ark_version.go:206`) and `activeArkVersions` (`serverconn/ark_version.go:227`) linearly scan the operator-supplied `resp.ArkVersionPolicies` repeated field with no length cap. A malicious operator could return a multi-million-entry policy list in the bootstrap `GetInfo` response to force a connect-time CPU/alloc spike. Self-limited (gRPC max-recv, connecting client only, no amplification, no fund risk), but a sanity cap on policy count before iterating would be cheap hardening.

### M-4 — `ValidateRefreshSelection` does not self-defend against a zero `boundVersion`
At `serverconn/ark_version.go:151`, if `boundVersion` were ever 0 a `resp.SelectedArkVersion == 0` refresh would match (`0 == 0`) and pass. Unreachable in the wired flow today (`NewRuntime` rejects a zero ark version at `serverconn/runtime.go:59-61`, so a runtime can't exist with a zero bound version), but the validator leans on a caller invariant rather than re-asserting it. A one-line non-zero guard at the top of `ValidateRefreshSelection` would make it self-defending.

PR #745, opened 2026-06-16
darepod: support custom vHTLC refresh signatures

bhandras:
## What this PR does
This PR adds the darepod and SDK support needed for cooperative custom vHTLC refresh.

A vHTLC refresh is the path where swapdk-server keeps a live swap vTXO alive by moving it into a fresh Ark round before the old batch expires. That keeps the swap on the normal Lightning path instead of immediately escalating to recovery.

The important detail is that a vHTLC is not a normal single-party wallet vTXO. Its future forfeit transaction can require signatures from more than one participant. The daemon that queues the refresh must therefore be able to collect a participant signature after the round builder has assigned the connector and built the exact forfeit transaction.

This branch teaches darepod to do that.

## What changed
The PR adds four pieces of plumbing.

1. Round and vTXO code can carry multiple participant signatures for one forfeit transaction.
2. Custom vTXO refresh requests can include explicit forfeit signing context, so darepod knows how a later participant-signature request should be answered.
3. The daemon RPC exposes pending connector-bound forfeit-signature requests and accepts submitted participant signatures for those requests.
4. The bundled swap runtime signs the correct side of the request:
   - if darepod has a local participant signer, it signs inside the daemon;
   - if another participant must sign, darepod publishes a pending request and waits for the external signature to be submitted.

There is also a small cleanup path for failed custom refresh admission. If darepod queued temporary custom refresh actors but the later trigger step fails, the RPC asks the wallet to drop those temporary actors so stale signer contexts do not linger.

## The RPC contract
The protocol is intentionally route-based, not swap-direction-based. A custom vHTLC may be used by swaps today, but the daemon RPC does not need to know whether the caller thinks of the vHTLC as a swap in or a swap out.

`RefreshCustomVTXOs` takes the old custom vTXO, the replacement output, and a `ForfeitSigningContext`:

- `payment_hash` is an opaque 32-byte correlation id. Swap users set it to the swap hash, but darepod only copies it into later pending requests so the external coordinator can find its own state.
- `signing_route` tells darepod how to answer the future connector-bound participant-signature request.

There are two routes:

- `LOCAL_SIGNER`: when the round later supplies the exact forfeit transaction, darepod asks its configured local signer to sign the vTXO input immediately.
- `PENDING_REQUEST`: darepod stores the exact transcript, returns it from `ListPendingForfeitParticipantSignatureRequests`, blocks the temporary VTXO actor, and resumes only after `SubmitForfeitParticipantSignatures` supplies the participant signature for that `request_id`.

That split is necessary because `RefreshCustomVTXOs` runs before a round has assigned a connector. At that time the final forfeit transaction does not exist, so the participant signature cannot be collected yet. The later pending request includes the connector outpoint, connector amount, connector pkScript, selected forfeit spend path, unsigned forfeit transaction, and server forfeit output script. The external participant signs that exact VTXO input transcript.

## Function-level flow
`RefreshCustomVTXOs` validates the caller-supplied custom input and queues a temporary refresh actor. That actor stores the old vTXO, the replacement output, and the requested signing route, but it cannot ask anyone to sign yet because the connector has not been assigned.

When the round actor later builds the exact connector-bound forfeit transaction, it calls into the temporary actor with the full transcript. The actor then chooses the route:

- for `LOCAL_SIGNER`, it calls the local forfeit signer and returns the participant signature directly to the round path;
- for `PENDING_REQUEST`, it registers the transcript in the in-process broker and waits on `SubmitForfeitParticipantSignatures`.

`ListPendingForfeitParticipantSignatureRequests` is the polling API for external coordinators. It returns only unanswered requests and uses `next_sequence` as a cursor. Answered later requests no longer advance that cursor past an earlier pending request, so an external signer cannot accidentally skip a still-unanswered transcript.

`SubmitForfeitParticipantSignatures` matches the `request_id`, wakes the blocked actor, and removes the answered request from the broker. Repeated submits for an already-answered request are treated as idempotent success only when the submitted signatures match the stored answer. A different answer for the same request id is rejected as `AlreadyExists`.

## Why this is safe
The participant signs an exact forfeit transaction transcript. That keeps the signature scoped to the connector-bound transaction that the operator will use for the refreshed round.

The pending-request path is durable at the coordinator layer: the request contains the opaque payment hash and a stable request id, so swapdk-server can recover its own state and resubmit the participant signature if it restarts in the middle.

Daemon-local signing contexts are removed after a successful external submit or synchronous local signer response, so successful refresh signing does not leave stale contexts behind.

Ordinary wallet-managed VTXOs keep using the existing behavior. The new signing path is only used for custom refresh inputs that carry a forfeit signing context.

## Stack
This is the darepo-client side of the active vHTLC refresh stack. The matching darepo PR carries the round-wire fields, and the swapdk-server PR uses these APIs when a funded vHTLC is approaching batch expiry.

## Validation
- `make fmt-changed-check base=origin/main`
- `go test ./darepod ./db -count=1`
- Earlier on this branch: `go test ./darepod ./wallet ./daemonrpc ./sdk/swaps -count=1`

The matching darepo and swapdk-server PRs add full integration coverage for the multi-participant refresh path. Those tests prove that the extra participant signature is requested, submitted, carried into the round input artifacts, accepted by the operator, and confirmed in the replacement round.


bhandras:
@claude pls review this refreshed custom vHTLC refresh support. Context: the latest force-push folds the lint fixes and the PR body now calls out the multi-participant refresh integration coverage in the companion darepo and swapdk-server PRs. Please focus on durable signing/ack ordering, participant forfeit signature plumbing, and whether the daemon RPC surface is sufficient for swapdk-server active refresh.

bhandras:
@claude pls re-review with context: since the last request I folded the local review finding that successful forfeit signing left daemon-local signing contexts behind. The broker now clears the context after external signatures are accepted and after synchronous in-swap signing succeeds, with tests for both paths. I also updated the PR body/validation to reflect the folded tests. Please re-check durable signing/ack ordering and context lifecycle.

bhandras:
@claude small extra context: I force-pushed once more after the prior re-review request to fold the static-check fix. The new head adds a tiny generated-schema comment refresh and gofmt/import formatting; local `make fmt-changed-check base=origin/main`, `make sqlc-check`, commitmsg-lint, and smoke compile are now green. Please review the current head.

bhandras:
@claude pls re-review with context: I folded the custom refresh signing API away from swap-direction terminology and into route terminology. The daemon proto now documents the full flow: RefreshCustomVTXOs queues the custom vTXO before a connector exists, LOCAL_SIGNER signs inside darepod when the connector-bound ForfeitRequest arrives, and PENDING_REQUEST publishes the exact transcript through ListPendingForfeitParticipantSignatureRequests until SubmitForfeitParticipantSignatures answers it. Client head is 5d5a26d2. Local validation: make fmt-changed-check base=origin/main; make commitmsg-lint range=origin/main..HEAD; go test ./darepod ./wallet ./daemonrpc ./sdk/swaps -count=1.

bhandras:
@claude pls re-review. Context since the last review: I expanded the daemon proto/RPC comments and PR body to describe the route-based signing flow in detail; fixed the broker cursor bug so answered later requests do not advance next_sequence past an earlier pending request; added TestForfeitSignatureBrokerListDoesNotAdvancePastEarlierPendingRequest; and documented that custom policy refresh outputs may fail standard policy decode while still persisting policy_template and pk_script as authoritative script data. Current head is 7cb5bb65.

bhandras:
@claude pls re-review with context.

Folded the last local/CI cleanup on top of the expanded proto/RPC documentation:

- wrapped the long `ForfeitSigningRoute_PENDING_REQUEST` case in `darepod/forfeit_signature_broker.go` so lint should no longer fail on the 81-character line;
- kept the broker cursor fix and test from the prior push intact;
- PR body already reflects the clarified idempotency contract: repeated submit with matching signatures is idempotent, different signatures return `AlreadyExists`.

Latest pushed client head: `6b0ebfdab8df5477947ae982095de5774121f81d`.

Local validation after the fold:

- `make fmt-changed-check base=origin/main`
- `go test ./darepod ./db -count=1`

bhandras:
@claude pls re-review with context. I folded the remaining review concerns on the client side:

- bounded pending external forfeit-signature waits using the existing forfeit collection timeout, so a missing receiver-side response cannot block the VTXO actor indefinitely;
- cleaned up temporary broker signing contexts on transcript-build failure, request-id conflict, and wait cancellation;
- made duplicate signature submission idempotence order-insensitive for the same participant signature set;
- added a defensive nil-response guard around SignVTXOForfeit in the SDK handler.

Focused local checks passed:

- go test ./darepod -run 'TestForfeitSignatureBroker|TestRefreshCustomVTXOs|TestSignVTXOForfeit' -count=1
- go test ./sdk/swaps -run 'Test.*Forfeit|Test.*Refresh' -count=1
- make fmt-changed-check base=origin/main

bhandras:
Folded the remaining client-side review items and force-pushed the branch.

What changed:
- `SignVTXOForfeit` now requires the requested VTXO outpoint to exist in the daemon's local VTXO store, and it rejects transcripts whose amount / pkScript / policy template do not match that local state.
- `forfeitSignatureBroker.submit` now verifies submitted participant signatures against the exact pending connector-bound transcript before storing them or waking the blocked VTXO actor.
- broker tests now use real forfeit signatures over the actual VTXO-input sighash, plus coverage for invalid-signature rejection and idempotent resubmission of the same valid participant set.

Focused validation rerun:
- `go test ./darepod -run 'TestForfeitSignatureBroker|TestSignVTXOForfeit' -count=1`
- `go test ./sdk/swaps -run 'Test.*Forfeit' -count=1`
- `make lint-changed-local`

@claude please re-review when you get a chance. The two substantive client blockers from the last review round should now be addressed.

litbot-9000:

PR #747, opened 2026-06-16
coinselect: unify coin selection, fix onchain send preview outflow

Roasbeef:
In this PR, we fix a misleading onchain send preview and, in the process,
consolidate coin selection behind a single shared abstraction. The bug
that kicked this off: a bounded `--amt 9000` send to an onchain address
would preview an "Expected total outflow" of 101196 sats. The 9000 sat
send hadn't suddenly grown a 90k+ fee, the preview was just reporting the
gross sum of the VTXO bundle it selected to cover the send, not the sats
that actually leave the wallet. Under the seal-time change-VTXO semantics
(#634, #270) a bounded send lands exactly the requested amount at the
destination and the residual comes back as a change VTXO, so the real net
outflow is amount + fee. The preview now reports that.

While digging in we found the root cause was deeper than a label. The swap
wallet router had its own greedy `selectVTXOsForAmount` walk that existed
only to build the preview (the real send re-selects in the daemon and
ignores it entirely). That's a parallel coin-selection implementation the
`swapwallet` deps contract explicitly says this layer should never grow.
So rather than patch the number in place, we pull selection into one
place and route everyone through it.

## coinselect

We add a new `coinselect` leaf package with a single, coin-type-agnostic
`LargestFirst`. It depends only on `btcutil`, so both the VTXO manager's
reservation path and the swap wallet's preview select through the same
code instead of each maintaining its own accumulate-until-covered loop.
It's generic over the candidate type via an `AmountFunc[T]` accessor, does
largest-first selection with optional dust-change avoidance, supports a
sweep-all mode, and reports failures as typed errors (shortfall,
dust-change, empty set) that carry the covered total so callers can map
them onto their own diagnostics.

## vtxo + swapwallet

In `vtxo`, the manager drops its private `selectLargestFirst*` helpers and
calls `coinselect.LargestFirst`, keeping its actor-coupled reservation and
rollback. This is behavior-preserving: the algorithm was already
largest-first with dust avoidance.

In `swapwallet`, we delete the duplicate greedy selector and have the
preview select through `coinselect` too, so it now mirrors the daemon's
largest-first selection rather than a divergent first-fit walk. Selection
stays local over the daemon's live-VTXO view so a preview still works when
the operator is offline.

We also replace the hardcoded "cooperative-leave quote support is not
available yet" warning with a real fee estimate. We prefer the operator's
dynamic `EstimateFee` quote (which folds in the on-chain share, liquidity,
and margin charged at seal time) and mark that preview COMPLETE. When the
operator is unreachable, we fall back to a local batch-size-1 floor
derived from the daemon's cached operator terms: the larger of the minimum
operator fee and the on-chain cost of this leave's footprint at the
operator's target feerate. That floor can't see the operator's
liquidity/congestion components, so it's a lower bound the caller sees as
LOCAL_ONLY with an explanatory warning.

One caveat worth flagging for review: the local floor uses approximate
taproot vbyte constants (`leaveInputVBytes = 58`, etc.) and is only shown
when the operator can't be reached. The binding fee is still set
server-side at seal time.

changes.

### DIFF
--- a/coinselect/selector.go
+++ b/coinselect/selector.go
@@ -1,0 +1,183 @@
+// Package coinselect provides a single, coin-type-agnostic coin-selection
+// algorithm shared across the client. It deliberately holds no wallet,
+// actor, or RPC dependencies so every layer that needs to pick a covering
+// subset of coins — the VTXO manager's reservation path and the swap
+// wallet's send preview alike — selects through the same code rather than
+// growing parallel implementations.
+package coinselect
+
+import (
+	"errors"
+	"sort"
+
+	"github.com/btcsuite/btcd/btcutil"
+)
+
+// Selection errors. The selector stays policy-free: it reports why a pass
+// failed and the covered total, and leaves layer-specific diagnostics (for
+// example distinguishing locked liquidity from a true shortfall) to
+// callers.
+var (
+	// ErrSelectionShortfall means the candidate set cannot cover the
+	// requested target even when every candidate is selected. The
+	// returned Result.Total carries the full candidate sum so callers
+	// can render a precise message.
+	ErrSelectionShortfall = errors.New("insufficient candidates to cover " +
+		"target")
+
+	// ErrChangeBelowMin means a covering selection exists but its
+	// non-zero change falls below the requested minimum, and no
+	// exact-fit (zero-change) selection was found. The returned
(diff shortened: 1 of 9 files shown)


Roasbeef:
@claude review this

PR #748, opened 2026-06-16
tlvutil: dedup durable-actor TLV serialization behind shared helpers

Roasbeef:
In this PR, we cut down the pile of near-identical TLV serialization code
that every durable-actor message codec was carrying. Each `Encode`/`Decode`
hand-wrote the same `tlv.NewStream(...)` / `if err != nil` / `Encode(w)` (or
`DecodeWithParsedTypes`) dance, and a handful of domain types (outpoints
especially) were re-encoded from scratch in each package. We pull all of that
behind one small shared package, `baselib/tlvutil`, and migrate the codecs
onto it.

The package lives under `baselib` on purpose: it's the one module both this
repo and the server (darepo) already import, so the same helpers can be reused
on both sides. The server-side migration is a follow-up PR that bumps the
submodule to pick this up (see below).

## What's in tlvutil

Two things, and deliberately not more.

First, the stream helpers: `EncodeRecords`, `DecodeRecords`, and the
byte-slice variants. These are thin wrappers over `tlv.NewStream` that collapse
the repeated scaffolding into a single call. They emit byte-for-byte the same
output as the inlined form they replace, so adopting them is purely mechanical.
There's a property test that checks this equivalence directly against the old
`NewStream`/`Encode` shape.

Second, `OutPointRecord`. This is the one domain type worth centralizing: a
`wire.OutPoint` has no native `tlv` primitive, so every codec re-rolled the
"32-byte hash plus 4-byte little-endian index" encoding by hand. We give it a
single shared constructor that takes a pointer to the caller's existing
`wire.OutPoint` field, so nothing has to change the field's type to adopt it.

Notably, we *don't* add helpers for pubkeys, hashes, or the integer widths.
`tlv.MakePrimitiveRecord` already handles `**btcec.PublicKey`, `*[32]byte`,
`*[33]byte`, the `uint` types, and `[]byte` natively, so those just call the
native primitive directly (a `chainhash.Hash` is a `(*[32]byte)` cast away).
There's a test that locks that reuse in so we don't regrow wrappers later.

## Wire compatibility

This is the important part: durable mailbox state and on-disk blobs are
persisted, so the wire format has to stay byte-identical. It does. The records
passed to the helpers are unchanged, only the surrounding scaffolding moves, so
the encoding can't shift. On top of the property tests, we add a golden test
that pins the exact wire bytes of a fully-populated `VTXOSentMsg` (which goes
through `OutPointRecord`), so any accidental format change breaks loudly.

The `ledger` package had zero codec-level tests before this, so we also add
round-trip coverage for all six of its durable messages while we're here.

## Numbers

The migrated codecs drop ~570 lines net (739 removed, 169 added back as call
sites), against +154 for the one-time shared library, for ~−416 net in
production code in this repo. The new tests add ~486 lines of coverage that
didn't exist before.

See each commit message for the per-package detail.

### DIFF
--- a/baselib/actor/ask_response.go
+++ b/baselib/actor/ask_response.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/lightninglabs/darepo-client/baselib/tlvutil"
 	"github.com/lightningnetwork/lnd/tlv"
 )
 
@@ -63,8 +64,8 @@ func (m AskResponse) Encode(w io.Writer) error {
 	resultBlob := m.ResultBlob
 	errorText := []byte(m.ErrorText)
 
-	records := []tlv.Record{
-		tlv.MakePrimitiveRecord(
+	return tlvutil.EncodeRecords(
+		w, tlv.MakePrimitiveRecord(
 			askResponseCorrelationIDType, &correlationID,
 		),
 		tlv.MakePrimitiveRecord(
@@ -73,14 +74,7 @@ func (m AskResponse) Encode(w io.Writer) error {
 		tlv.MakePrimitiveRecord(
 			askResponseErrorTextType, &errorText,
 		),
-	}
-
-	stream, err := tlv.NewStream(records...)
-	if err != nil {
-		return err
-	}
-
-	return stream.Encode(w)
+	)
 }
 
 // Decode deserializes the message from the provided reader.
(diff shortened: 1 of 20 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #753, opened 2026-06-17
db: Fix postgres error mapping under the pgx v5 driver

ellemouton:
## Problem

darepo-client's postgres error mapping is broken whenever the pgx v5
stdlib driver backs the `database/sql` connection: unique-violation and
other classified postgres errors leak through `MapSQLError` as raw
`*pgconn.PgError` instead of the database-agnostic sentinels
(`ErrSQLUniqueConstraintViolation`, etc.). It is the client-side mirror
of lightninglabs/darepo's server fix.

## Root cause

The client links **two** pgx stdlib drivers: `pgx/v5/stdlib` (imported
by `db/postgres.go`) and `pgx/v4/stdlib` (transitive, via lnd's
`kvdb/sqlbase`). Each registers the bare `"pgx"` `database/sql` driver
name **only if no other major version claimed it first**, so which one
backs `sql.Open("pgx", …)` is an init-order coin flip. A pgx v4
connection surfaces errors as `*github.com/jackc/pgconn.PgError`; pgx v5
surfaces `*github.com/jackc/pgx/v5/pgconn.PgError` — two distinct,
non-interchangeable types. `MapSQLError` only matched the v4 type.

## Fix

1. `MapSQLError` now matches **both** pgconn error types and classifies
   by the shared string SQLSTATE code.
2. Open with the version-qualified `"pgx/v5"` driver name for a
   deterministic backend matched to the imported stdlib.

## Verification

- Build / vet / `fmt-changed` / `lint-changed-local` all clean.
- `test_postgres`: migration tests (`TestMigrationSteps`,
  `TestMigrationDowngrade`) and the unique-conflict tests
  (`TestActiveBoardingSweepInputUnique`, two `TestOORArtifactStore…Conflict`)
  pass — the latter exercise the mapping path directly.

## Update: root-cause hardening (3rd commit)

`build: Bump lnd kvdb to v1.5.1` removes the duplicate driver at its source.
The pinned `kvdb v1.4.16` imported `pgx/v4/stdlib` (a competing `"pgx"`
`database/sql` driver); kvdb v1.5.0+ is on pgx v5 (a released tag — lnd master
is on v5 too). After the bump, `pgx/v4/stdlib` is no longer in the build graph
(pgx v4 lingers only via `pgxpool`, which registers no `database/sql` driver),
so only `pgx/v5/stdlib` claims `"pgx"` — the init-order race is gone.

Note: this does **not** replace the `MapSQLError` fix — with the bump the
backend is deterministically v5, so the v5 mapping is now strictly required.
The pin + dual-type mapping stay as defense in depth. Raises the `go` directive
to 1.25.10 (required by kvdb v1.5.1).


ellemouton:
Closing — superseded by the planned lnd dependency upgrade, not abandoned.

Bumping `lnd` + `lndclient` to master removes `github.com/jackc/pgx/v4` from
`go.mod` entirely (verified locally), giving a v5-only graph. At that point the
fix here collapses into the upgrade: `MapSQLError` simplifies to v5-only and the
explicit `pgx/v5` pin + `kvdb` bump become redundant.

Tracked in lightninglabs/darepo#581. The branch/commits are preserved and this
can be reopened as a stopgap if the postgres-unit flake bites before the upgrade
lands.

PR #755, opened 2026-06-17
[codex] lwwallet: pass seed birthday to btcwallet

bhandras:
## Summary

Fresh lwwallet creation was decoding aezeed mnemonics down to raw seed entropy and dropping the aezeed birthday. The daemon then started btcwallet with a nonzero recovery window but without a birthday, which can leave btcwallet without a useful lower bound for recovery scans and make a new wallet behave like an unbounded historical restore.

This PR carries the seed birthday through the fresh InitWallet path and passes it into btcwallet-backed wallets.

## Changes

- Add mnemonic decoding helpers that return both raw seed entropy and aezeed birthday.
- Thread the birthday from `InitWallet` into self-managed wallet startup.
- Pass the birthday into LND `btcwallet.Config` for both lwwallet and the neutrino-backed btcwallet backend.
- Add tests proving the aezeed birthday is preserved through mnemonic decoding and InitWallet seed persistence.

## Impact

A newly created lwwallet should no longer need to scan from genesis just because recovery window is enabled. Restored mnemonics also get a bounded scan from the aezeed birthday instead of an unset birthday.

Existing unlock/auto-unlock from the encrypted raw seed still passes a zero birthday, relying on btcwallet's persisted wallet DB metadata when present.

## Verification

- `make fmt-changed`
- `go test ./darepod ./lwwallet ./btcwbackend ./walletcore`
- `go test ./...`
- `go test -tags=walletdkrpc,swapruntime ./darepod ./cmd/darepod ./swapwallet ./sdk/walletdk ./rpc/walletdkrpc`
- `make lint-changed-local`


bhandras:
@claude review

PR #761, opened 2026-06-18
round: route boarding failures by round id (darepo#571)

ellemouton:
## Summary

Client-side half of [darepo#571](https://github.com/lightninglabs/darepo/issues/571) (proposal #3): *"surface round-build failures back to the client so a deposit does not sit silently pending."*

### Root cause

When the operator fails to build a round, it sends every client a `ClientRoundFailedResp` carrying the round id. On the client this becomes a `BoardingFailed` event, and the transition table maps `BoardingFailed → ClientFailedState` from every state — so **if the event reaches the FSM**, the round fails cleanly and stops projecting pending VTXOs.

But `BoardingFailed.FromProto` **discarded the round id**. With no id, `extractRoundID` has no `*BoardingFailed` case, so routing fell to `routeServerMessageToPending`, which only matches *temp-keyed* rounds via `findPendingRound`. After `RoundJoined` re-keys the FSM to its server-assigned round id, that lookup misses, and delivery fell back to a heuristic that routes the failure **only when exactly one round is tracked**. Terminal FSMs (`ClientFailedState`/`ConfirmedState`) are never evicted from the rounds map, so as soon as any prior round lingered, `len(a.rounds) != 1`, the heuristic bailed, and the failure was **silently dropped** (the durable `Tell` had already acked, so the ingress cursor advanced and the envelope was GC'd).

The live round then stayed stuck in a waiting state (e.g. `RoundJoinedState`), whose `liveRoundDetails` projects a synthetic `VTXO_STATUS_PENDING_ROUND` — so the boarding deposit appeared **"pending" forever**.

### Fix

- Carry the round id on `BoardingFailed` (`fn.Option[RoundID]`), populated from the 16-byte `ClientRoundFailedResp.round_id`. It stays `None` for failures that predate round assignment (e.g. `ClientErrorResp`, which has no round id).
- In `routeServerMessageToPending`, route a `BoardingFailed` **deterministically by its round id** (looking up the re-keyed FSM directly), regardless of how many rounds are tracked. The sole-round heuristic is retained only as a fallback for failures that carry no id.

With this, a server round failure always reaches the correct FSM → `ClientFailedState` → the phantom pending VTXO disappears, and `ListRounds`/`GetRound` surface `ROUND_STATE_FAILED` with the reason.

## Testing

- New `TestBoardingFailedRoutesByRoundIDWithLingeringRound`: stages a lingering terminal round **plus** a live re-keyed round (so `len(a.rounds) == 2`), delivers the failure, and asserts it routes to the live round (→ `ClientFailedState`) while the lingering round is untouched. Verified that reverting only the routing change makes this test fail with the exact dropped-event symptom (`no pending round for event *round.BoardingFailed`).
- `FromProto` tests for round-id present (→ `Some`), empty round-id (→ `None`), and `ClientErrorResp` (→ `None`).
- `go build`, `go test ./round/...`, `make fmt-changed`, `make lint-changed-local` (0 issues), `make commitmsg-lint` all pass.

## Scope / follow-up

This fixes the **"pending forever"** mechanism (the headline symptom). One further enhancement from the issue — **auto-retrying** a failed boarding on a *running* daemon (today a confirmed boarding intent is only re-boarded on daemon-restart replay, not after a live round failure) — is intentionally left as a separate change; it is a larger wallet-orchestration feature and is independent of this routing correctness fix.

### DIFF
--- a/round/events.go
+++ b/round/events.go
@@ -14,6 +14,7 @@ import (
 	"github.com/lightninglabs/darepo-client/lib/tree"
 	"github.com/lightninglabs/darepo-client/lib/types"
 	"github.com/lightninglabs/darepo-client/rpc/roundpb"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
 )
 
 // ClientEvent is a sealed interface for all events that can be processed by
@@ -346,6 +347,13 @@ func (e *BoardingConfirmed) clientEventSealed() {}
 // BoardingFailed is emitted when an error occurs during the boarding
 // process.
 type BoardingFailed struct {
+	// RoundID is the server-assigned round id when the failure carries
+	// one (ClientRoundFailedResp). It is None for failures that arrive
+	// before a round was assigned (e.g. ClientErrorResp). When set, the
+	// actor routes the failure deterministically to the matching FSM
+	// instead of relying on the sole-round heuristic.
+	RoundID fn.Option[RoundID]
+
 	// Reason is a human-readable description of the failure.
 	Reason string
 
--- a/round/from_proto.go
+++ b/round/from_proto.go
@@ -12,6 +12,7 @@ import (
 	"github.com/lightninglabs/darepo-client/lib/tree"
 	"github.com/lightninglabs/darepo-client/lib/types"
 	"github.com/lightninglabs/darepo-client/rpc/roundpb"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/lightningnetwork/lnd/keychain"
 	"google.golang.org/protobuf/proto"
 )
@@ -428,6 +429,18 @@ func (e *BoardingFailed) FromProto(p proto.Message) error {
 	case *roundpb.ClientRoundFailedResp:
 		e.Reason = pb.Reason
 		e.Recoverable = true
+
+		// Carry the server-assigned round id when present so the actor
+		// can route this failure deterministically to the matching FSM.
+		// A failure that arrives before the round was assigned
+		// legitimately has no id (len != roundIDLen), in which case
+		// RoundID stays None and routing falls back to the sole-round
+		// heuristic.
+		if len(pb.RoundId) == roundIDLen {
+			var rid RoundID
+			copy(rid[:], pb.RoundId)
+			e.RoundID = fn.Some(rid)
+		}
 
 		return nil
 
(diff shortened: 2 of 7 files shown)


ellemouton:
@claude review this

round/events.go
```diff
@@ -346,6 +347,13 @@ func (e *BoardingConfirmed) clientEventSealed() {}
 // BoardingFailed is emitted when an error occurs during the boarding
 // process.
 type BoardingFailed struct {
+	// RoundID is the server-assigned round id when the failure carries
+	// one (ClientRoundFailedResp). It is None for failures that arrive
+	// before a round was assigned (e.g. ClientErrorResp). When set, the
+	// actor routes the failure deterministically to the matching FSM
+	// instead of relying on the sole-round heuristic.
+	RoundID fn.Option[RoundID]
```
Roasbeef, line 355: 👍

round/from_proto.go
```diff
@@ -429,6 +430,18 @@ func (e *BoardingFailed) FromProto(p proto.Message) error {
 		e.Reason = pb.Reason
 		e.Recoverable = true
 
+		// Carry the server-assigned round id when present so the actor
+		// can route this failure deterministically to the matching FSM.
+		// A failure that arrives before the round was assigned
+		// legitimately has no id (len != 16), in which case RoundID
+		// stays None and routing falls back to the sole-round
+		// heuristic.
+		if len(pb.RoundId) == 16 {
```
Roasbeef: lift up to constant? uuid size iiuc

Roasbeef (approved):
LGTM 🥝

ellemouton:
Addressed both review comments (force-pushed):

- **@Roasbeef** (`from_proto.go` — lift `16` to a constant): added `roundIDLen = len(RoundID{})` next to the `RoundID` type in `interfaces.go` (derived from the type so it tracks automatically) and use it for the round_id length check.
- **gemini** (sole-round fallback should require no RoundID): the fallback now also gates on `bf.RoundID.IsNone()`, so a failure carrying a RoundID that matches no tracked round is treated as a genuine miss instead of being misrouted to the sole tracked round. Added `TestBoardingFailedMismatchedRoundIDDoesNotMisroute` to lock it in. (The pre-`RoundJoined` case is unaffected — `findPendingRound` handles temp-keyed rounds before the fallback.)

`go test ./round/...`, `make fmt-changed`, `make lint-changed-local` (0 issues), `make commitmsg-lint` all green.

PR #762, opened 2026-06-18
chainsource: escalate unrecoverable block epoch loss to the daemon

Roasbeef:
## Motivation

A production signet incident left the `darepod` ark client wedged after its
backing LND bounced. The block epoch reconnect work (#698, "chainsource:
reconnect block epochs") correctly treats a closed stream as transient and
re-registers with bounded backoff — but it **retries forever**. When the
backend connection is genuinely stuck, dozens of block subscribers all
re-register against the dead notifier every 30s indefinitely (the log filled
with bursts of "Registering for block epoch notifications"), no subscription
ever heals, and the daemon serves RPCs as a zombie while its readiness probe
stays green.

The downstream effect: a Lightning → Ark in-swap never completed. The swap
server kept intercepting the HTLC but failed with `get server pubkey:
identity pubkey is empty`, because the wedged client could not surface the
operator identity. HTLCs were canceled back; from the payer's side the swap
looked stuck.

## What this does

Bounds the reconnect instead of retrying forever:

- `BlockEpochActor` tracks how long a subscription has been **continuously
  down** — stamped on first loss, cleared only once a replacement stream
  actually *delivers a block*. A backend that hands back streams which
  immediately close therefore still counts as down (no false "healthy").
- Once the down streak outlives `FatalReconnectTimeout` (default 5m), the
  actor stops retrying and escalates via a new `OnFatal` hook.
- Because every block subscriber (`vtxo`, `unroll`, `wallet`, `txconfirm`)
  is served by this one actor, the escalation lives in a single shared place.
- The hook is threaded from the daemon through `ChainSourceConfig`. The
  server wires it to a `context.CancelCause` on the run context, so `run()`
  returns the cause → non-zero exit → the orchestrator restarts `darepod`
  with a fresh backend connection. Normal signal-driven shutdown still
  returns nil.

No `os.Exit` in subsystem code; the failure bubbles up the normal error path.
A clock is injected so the timeout is exercised without real waits.

## Testing

- `go test ./chainsource/...` green, `go vet ./chainsource/ ./darepod/` clean.
- New test: a backend that can never sustain a stream escalates through
  `OnFatal` (clock-driven). Existing reconnect/iterator tests still pass.

### DIFF
--- a/chainsource/block_epoch_actor.go
+++ b/chainsource/block_epoch_actor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"math/rand/v2"
 	"sync"
 	"time"
 
@@ -24,7 +25,28 @@ const (
 	// defaultBlockEpochMaxReconnectBackoff caps reconnect backoff so a
 	// backend outage does not spin, but recovery still happens promptly.
 	defaultBlockEpochMaxReconnectBackoff = 30 * time.Second
+
+	// defaultBlockEpochFatalTimeout bounds how long a subscription may stay
+	// continuously down before the actor stops retrying and escalates. A
+	// notifier restart heals in seconds; a streak this long means the
+	// backend connection is stuck (the production symptom was dozens of
+	// subscribers re-registering against a dead notifier every 30s forever,
+	// which no amount of in-process retrying can recover). Escalating lets
+	// the daemon exit and restart with a fresh backend connection.
+	defaultBlockEpochFatalTimeout = 5 * time.Minute
+
+	// defaultBlockEpochFatalJitter is the maximum random padding added to
+	// the fatal timeout on the default-timeout path. A backend outage tends
+	// to drop every daemon's streams at the same instant; without jitter a
+	// whole fleet would hit the 5m budget together and restart in lockstep
+	// (thundering herd). Spreading the effective budget over a ~1m window
+	// decorrelates those restarts.
+	defaultBlockEpochFatalJitter = time.Minute
 )
+
+// errBlockEpochUnrecoverable is the placeholder cause used when a subscription
+// stays down past the fatal timeout without a more specific reconnect error.
+var errBlockEpochUnrecoverable = fmt.Errorf("block epoch backend unreachable")
 
 // BlockEpochConfig holds configuration for BlockEpochActor.
 type BlockEpochConfig struct {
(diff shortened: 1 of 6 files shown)


Roasbeef:
@claude review this

Roasbeef:
Thanks — addressed the two correctness issues, folded into the existing commits (force-pushed `54f70758`).

**#1 (🔴 silent subscription death without `OnFatal`) — fixed.** The fatal `return` is now gated on `a.cfg.OnFatal.IsSome()`: only callers that opted into bounded escalation take the exit path; hookless callers (systest, SDK/embedded) keep the original unbounded-retry contract instead of having the subscription silently killed after the budget. New test `TestBlockEpochActorRetriesWithoutOnFatal` asserts a hookless actor against a permanently-dead backend keeps reconnecting (attempt count climbs well past where an escalating actor would have stopped) rather than exiting.

**#2 (🟠 spurious escalation on shutdown) — fixed.** Added an `a.ctx.Err() != nil` early-return at the top of the `registration == nil` block, before `reconnectExhausted`, so a stop that coincides with a long outage exits cleanly and never latches `fatalFlag` (which would otherwise flip liveness/readiness to failing). I didn't add a dedicated unit test for this one: the trigger is an inherent goroutine-timing boundary (cancellation observed precisely between the loop top and the exhaustion check), so any test would be racy/flaky; the guard itself is a cheap, obviously-correct ordering check.

**#3 (🟡 open-but-silent stream) — documented, not implemented.** Agreed it's narrower than the closing-stream symptom this PR targets, and `downSince` is only ever stamped on a close, so a healthy backend (which delivers blocks well within the budget) is unaffected. Added a comment on the `downSince`/budget block noting that an open stream that goes silent without closing is out of scope. Happy to add a bounded idle-wait on the open stream in a follow-up if you'd prefer full coverage.

Nits confirmed (no change needed): the `fatalCancel` happens-before is sound, and the jitter-on-default-path design is intentional. `go test -race ./chainsource/... ./darepod/...` green; `make lint-changed-local` and `commitmsg-lint` clean.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

Issue #582, opened 2026-05-28
darepocli: activity inspect prints payment_hash twice


ViktorT-11:
## Summary

`darepocli activity inspect <id>` prints the `payment_hash` field twice
for the same entry.

## Reproduction

```
$ darepocli --no-tls activity inspect bdaeb8a1100a27410a86ec49b05e7edb7758e465a3bb6855ae840b665523048c
Activity
- last_update: 2026-05-28 13:22:20.000
- created: 2026-05-28 13:21:44.000
- kind: RECV
- status: COMPLETE
- amount: 5000 sat
- fee: 0 sat
- phase: confirmed
- id: bdaeb8a1100a27410a86ec49b05e7edb7758e465a3bb6855ae840b665523048c
- invoice: lntbs50u1p4psf2gpp5hkht3ggspgn5zz5xa3ymqhn7mdm43er95waks4dwss9kv4frqjxqdqqcqzpgxqrrssrzjqvjsxpqvh5h6mk09fng6phzfgqz3e25xfnfkwywf2mua7gltce07dttk0hd5m2vpdyqqqqqqqqqqqqqq8ssp5jqtux9p2xpdw2v9w9n6jzj7a2nmm922ye5mzkauya5q42228s4qq9qrsgqvnkgc9stexvchv45x7ezj2q4ahkr9reehxtf0hgc7fgn7sqm5evh9tlvnevv8q2jtclt8rzx9qwjkpsu5j2hfmxqp0slcanlvvmx9wgqfmp69a
- payment_hash: bdaeb8a1100a27410a86ec49b05e7edb7758e465a3bb6855ae840b665523048c
- payment_hash: bdaeb8a1100a27410a86ec49b05e7edb7758e465a3bb6855ae840b665523048c
- progress_vtxo: 0b02c92e32692b03e4f8a336c3e21406cf68fa4057f699e6f95b5020d2fc800a:0
```

Note the two consecutive `payment_hash:` lines with identical values.

## Expected

`payment_hash` should be printed once.

## Environment

- Client: darepocli (current `main`)
- Network: signet

PR #764, opened 2026-06-18
systest: cover post-admission round failure handling

Roasbeef:
In this PR, we add two systests for how the client handles a round that
fails *after* it's been admitted. This stacks on #761 (the round-id failure
routing) and exercises it together with the forfeit-release fix in #763.

The fake operator in the systest harness gains an admit-then-fail mode: when
it receives a `JoinRound`, it pushes a `RoundJoined` that echoes the round's
accepted inputs (so the client re-keys its pending round to the server id,
the way `handleRoundJoined` matches by outpoint), and then pushes a
`ClientRoundFailedResp` carrying that same id. The failure routes back to the
re-keyed round (the #761 path) and drives it into `ClientFailedState`. This
is the smallest faithful reproduction of the production incident: admit,
then fail, with no quote or commitment dance.

## Forfeit release

`TestForfeitReleasedOnPostAdmissionRoundFailure` reserves a VTXO via a
cooperative leave (so it lands in `PENDING_FORFEIT`), then has the operator
admit and fail the round. We disable the registration timeout, so the only
way the VTXO can get back to `LIVE` is the post-admission forfeit release from
#763. On a daemon without that fix it stays stranded; we confirmed the test
catches it by neutering the release and watching the assertion time out
(~73s) where the fixed daemon recovers in ~14s.

## Boarding failure routing

`TestBoardingRoundFailureSurfacedNotStuck` funds a real boarding deposit,
lets the eager-join wallet register it, and has the operator admit then fail
the round. Before #761 the failure carried no round id and got dropped for
the re-keyed round, so the boarding sat projecting a synthetic
`PENDING_ROUND` VTXO forever. Here we assert the round surfaces as
`ROUND_STATE_FAILED` and the phantom pending VTXO clears.

See each commit message for the incremental details. Note this is based on
the #761 branch, so the shared release-fix commit drops out once #761 and
#763 land in main.

### DIFF
--- a/round/actor.go
+++ b/round/actor.go
@@ -1792,12 +1792,38 @@ func (a *RoundClientActor) routeServerMessageToPending(ctx context.Context,
 
 	roundFSM := a.findPendingRound()
 
-	// Round failures can arrive after the round is keyed by a
-	// server-assigned RoundID. When there is exactly one tracked
-	// round, route the failure there.
+	// Round failures can arrive after the round has been re-keyed by a
+	// server-assigned RoundID, so findPendingRound (which only matches
+	// temp-keyed rounds) misses them. Route them deterministically by the
+	// RoundID the failure carries, falling back to the sole-round heuristic
+	// for failures that predate round assignment (e.g. ClientErrorResp).
 	if roundFSM == nil {
-		_, isBoardingFailed := msg.(*BoardingFailed)
-		if isBoardingFailed && len(a.rounds) == 1 {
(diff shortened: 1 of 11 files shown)


Roasbeef:
Closing in favor of a server-side e2e systest in darepo. The client-side fake-operator approach proves the routing + release in isolation, but we actually want to assert the client-visible outcome (balance intact) against a *real* server failing a round, so the test belongs in darepo's ./systest harness on top of #582. The release fix itself stays in #763.

PR #772, opened 2026-06-23
Add sat credit send support

sputn1ck:
## Summary
- Add msat-native credit RPC fields across raw swap, swapclient, and walletdk APIs.
- Add durable credit sessions in `sdk/swaps` for receive, Ark top-up, redeem, and pay flows.
- Route walletdk `Recv`, `PrepareSend`, and `Send` through credits for below-dust receives, credit-only pays, and mixed pays.
- Document the credit account model, top-up and redemption flows, walletdk behavior, and server operation states in `docs/credit_system.md`.

## Server PR
- Used by lightninglabs/swapdk-server#134.

## Validation
- `go test ./sdk/swaps ./sdk/walletdk`
- `go test -tags="swapruntime walletdkrpc" ./swapclientserver ./swapwallet`
- `make lint-local`
- `make lint-changed-local`
- From `swapdk-server`: `make itest-swapruntime icase="TestWalletDK(SubDustPayNoCreditsTopsUpThenPays|SubDustPayExistingCreditsSkipsTopUp|MixedPayUsesCreditsAndVTXO)" ITEST_PARALLEL=1`


rpc/swapclientrpc/swap_client.proto
```diff
@@ -73,6 +112,11 @@ message QuotePayRequest {
     // Zero means no explicit caller cap. Quotes still return when the
     // estimated fee exceeds this cap so callers can render the fee.
     uint64 max_fee_sat = 2;
+
+    // max_credit_msat is the maximum credit amount the caller is willing to
+    // use for this quote. Zero means only required-credit shortfalls are
+    // reported.
+    uint64 max_credit_msat = 3;
```
Roasbeef: So basically we allow a combo of the credit and actual sats to pay an invoice?

rpc/swapclientrpc/swap_client.proto
```diff
@@ -123,6 +173,85 @@ message StartPayResponse {
     SwapSummary swap = 2;
 }
 
+message CreateCreditRequest {
+    string idempotency_key = 1;
```
Roasbeef, line 178: `credit_nonce`?

Roasbeef: same below

rpc/swapclientrpc/swap_client.proto
```diff
@@ -123,6 +173,85 @@ message StartPayResponse {
...
+    CreditFundingSource source = 2;
+    uint64 amount_msat = 3;
```
Roasbeef: Can just make it mSats right? As you can map between them. 

Stepping back, why expose msats at all? They exist on LN, but mostly so we can do proper accounting for fees. But in this case the server always works in whole sats amount for refreshes and also liquidity fees.

rpc/swapclientrpc/swap_client.proto
```diff
@@ -19,6 +19,15 @@ service SwapClientService {
     // continue the claim flow after the invoice is paid.
     rpc StartReceive (StartReceiveRequest) returns (StartReceiveResponse);
 
+    // CreateCredit starts a durable credit funding session.
+    rpc CreateCredit (CreateCreditRequest) returns (CreateCreditResponse);
```
Roasbeef, line 23: If a client tries to recv an amt below the min receivable amt, will they automatically hit this `CreateCredit` endpoint? 

Wondering if we need to expose low level knobs like this at all. Can the client just detect that credit is needed then fall back?

rpc/walletdkrpc/wallet.proto
```diff
@@ -339,6 +299,10 @@ message PrepareSendResponse {
     // warning carries a concise human-facing caveat for local-only
     // previews.
     string warning = 14;
+
+    // credit_preview is populated when the invoice send will or can use
+    // msat-native server credits.
+    CreditPreview credit_preview = 15;
```
Roasbeef, line 347: 👍

sdk/swaps/client.go
```diff
@@ -42,8 +42,126 @@ const (
...
+	// a reserved credit balance.
+	SettlementTypeMixed SettlementType = "mixed"
+)
+
+// CreditQuote describes how a pay quote uses wallet credits.
+type CreditQuote struct {
+	MustUseCredit       bool
+	CreditAppliedMsat   uint64
+	CreditShortfallMsat uint64
+	CreditTopupSat      uint64
```
Roasbeef: Is this mutually exclusive with shortfall? Perhaps we actually want two structs here (one for send and one for recv?).

sdk/swaps/client.go
```diff
@@ -42,8 +42,126 @@ const (
...
+type CreditFundingSource string
+
+const (
+	// CreditFundingLightningReceive means the server creates and owns the
+	// Lightning invoice, then credits the wallet account after settlement.
+	CreditFundingLightningReceive CreditFundingSource = "lightning_receive"
+
+	// CreditFundingArkTopUp means the server returns a pubkey-backed Ark
+	// destination and credits the account after the OOR top-up is visible.
+	CreditFundingArkTopUp CreditFundingSource = "ark_topup"
```
Roasbeef, line 74: So this is for OOR credit sends? Perhaps the var name can refelct that.

sdk/swaps/client.go
```diff
@@ -42,8 +42,126 @@ const (
...
+	LedgerEntries []CreditLedgerEntry
+}
+
+// CreditRedemption records the result of a successful credit redemption.
+type CreditRedemption struct {
+	Operation     CreditOperation
+	DebitedMsat   uint64
+	RedeemedSat   uint64
+	RemainderMsat uint64
+	SessionID     string
```
Roasbeef: oor session ID?

sdk/swaps/grpc_conn.go
```diff
@@ -497,9 +666,205 @@ func inSwapQuoteFromProto(resp *swaprpc.QuoteInSwapResponse) (*InSwapQuote,
...
+
+	case CreditFundingArkTopUp:
+		name = "CREDIT_FUNDING_SOURCE_ARK_TOPUP"
+
+	default:
+		return 0, fmt.Errorf("unknown credit funding source %q",
+			source)
+	}
+
+	value, ok := swaprpc.CreditFundingSource_value[name]
```
Roasbeef: Could just assign the variable above in the switch instead of looking up here. Less brittle that way.

sdk/swaps/in_swap.go
```diff
@@ -543,9 +558,22 @@ func (s *paySession) createSwap(ctx context.Context) error {
 		return fmt.Errorf("get client pubkey: %w", err)
 	}
 
-	cfg, err := s.client.server.CreateInSwap(
-		ctx, s.invoice, s.maxFeeSat, clientKey,
-	)
+	var cfg *InSwapConfig
+	if server, ok := s.client.server.(interface {
```
Roasbeef, line 562: Can just accept the interace instead of type assertion ?

sdk/swaps/in_swap_quote.go
```diff
@@ -41,25 +40,61 @@ func extractInvoiceAmountSat(msat *lnwire.MilliSatoshi) (uint64, error) {
 	if amountMSat == 0 {
 		return 0, fmt.Errorf("invoice amount must be positive")
 	}
-	if amountMSat%1000 != 0 {
-		return 0, fmt.Errorf("invoice amount must be whole satoshis")
+
+	return amountMSat, nil
+}
+
+func ceilMsatToSat(msat uint64) uint64 {
```
Roasbeef, line 47: Can use `lnwire.MiliSatoshi` for this

sdk/swaps/in_swap_quote.go
```diff
@@ -105,30 +142,52 @@ func validateInSwapPreview(invoice string, quote *InSwapQuote,
...
 	}
 
-	amountSat, err := extractInvoiceAmountSat(decoded.MilliSat)
+	amountMsat, err := extractInvoiceAmountMsat(decoded.MilliSat)
 	if err != nil {
 		return err
 	}
 
-	if amountSat != quote.InvoiceAmountSat {
+	expectedInvoiceSat := amountMsat / 1000
```
Roasbeef: Same here, should use the proper type throughout.

sdk/swaps/in_swap_quote.go
```diff
@@ -160,10 +219,20 @@ func validateInSwapQuote(invoice string, maxFeeSat uint64, cfg *InSwapConfig,
 		return fmt.Errorf("in-swap payment hash does not match invoice")
 	}
 
-	amountSat, err := extractInvoiceAmountSat(decoded.MilliSat)
+	amountMsat, err := extractInvoiceAmountMsat(decoded.MilliSat)
 	if err != nil {
 		return err
 	}
+	amountSat := amountMsat / 1000
+	if amountMsat%1000 != 0 {
```
Roasbeef: msat type

swapwallet/router.go
```diff
@@ -162,6 +164,26 @@ func (r *router) prepareInvoice(ctx context.Context, invoice string,
...
+		if creditQuote.GetCreditShortfallMsat() >
+			^uint64(0)-intent.maxCreditMsat {
+
+			intent.maxCreditMsat = ^uint64(0)
+		} else {
+			intent.maxCreditMsat += creditQuote.GetCreditShortfallMsat()
+		}
+		if quote.GetCreditQuote().GetMustUseCredit() {
+			if amountSat <= ^uint64(0)/1000 {
+				requiredCreditMsat := uint64(amountSat) * 1000
```
Roasbeef: msat type, above too

swapwallet/router.go
```diff
@@ -197,12 +224,128 @@ func (r *router) sendInvoiceIntent(ctx context.Context,
...
+	}
+
 	return &walletdkrpc.SendResponse{
 		Entry:           entry,
-		ActualAmountSat: startResp.GetSwap().GetAmountSat(),
+		ActualAmountSat: actualAmountSat,
 	}, nil
 }
 
+func (r *router) fundInvoiceCreditShortfall(ctx context.Context,
```
Roasbeef: `godoc` comments thru the entire diff would be nice.

swapwallet/router.go
```diff
@@ -197,12 +224,128 @@ func (r *router) sendInvoiceIntent(ctx context.Context,
...
+	}
+
+	if creditResp.GetState() != swapclientrpc.
+		CreditOperationState_CREDIT_OPERATION_STATE_CREDITED {
+
+		if len(creditResp.GetDestinationPubkey()) == 0 {
+			return fmt.Errorf("credit top-up destination missing")
+		}
+
+		_, err = r.deps.RPCServer.SendOOR(
```
Roasbeef: So I don't really understand this top off concept. Why is it needed? 

We want to pay an invoice, but why do we require new credits at all? Is this to cover a case where after paying you create _change_ that's below dust? I wonder if we should reject such sends

Roasbeef: I think we should also make a doc in `docs/` that explains teh credit system and its mechanics. The `/technical-writing` skill would be great for this case.

docs/credit_system.md
```diff
@@ -0,0 +1,307 @@
+# Credit System
```
Roasbeef, line 1: Ofc here it is! Ignore other comment

rpc/swapclientrpc/swap_client.proto
sputn1ck: create credit is not for the receive path, it is for the send path. in the receive case the server will simply create an invoice and on payment give you the credits.

for sending sub dust we first need the client to create credits by sending a vtxo to the swapserver

sputn1ck:
@claude please review

Roasbeef:
@claude review this

Roasbeef:
## Status update: durability rework + review fixes + a boarding fix

Pushing an update on where this branch stands. Head is now `7fa1daee` (20 commits).

### Credit ops now run on a durable actor subsystem

The credit send/receive/redeem flows no longer live as ad-hoc goroutines. They run on a small durable-actor subsystem so an operation survives a daemon crash and resumes exactly where it left off:

- A plain **supervisor** (`credit.Registry`, in-memory mailbox) owns admission and routing, and spawns one **durable per-operation child** per credit op (`op-credit-<opID>` mailbox, backed by `credit_operations` in sqlite via sqlc).
- Each op is a checkpointed FSM. Pay walks `Quoting → TopupCreating → TopupFunding → TopupAwaitingCredit → Paying → PayAwaitingSettlement → Completed`; receive and redeem have their own state lines. Every external effect is preceded by a persisted checkpoint (the durable framework's `Stage`), so resume after a crash replays the work-not-yet-done instead of re-firing effects.
- Idempotency keys are stable where it matters (`pay:<paymentHash>`) so a resumed pay re-attaches to the same server-side operation rather than starting a new one.

### Multi-agent review pass, folded into the feature commits

Ran a code review over the subsystem and addressed the findings in place (history folded so each fix lives in the commit that introduced the code, not as a trailing patch series):

- **C-1** — credit-only pay was marked complete the instant `StartPay` returned, before the Lightning leg settled. Added `PayAwaitingSettlement`, which reconciles against the server ledger (correlating the pay op by invoice payment hash via `ListCredits`) before reporting success.
- **H-1** — deterministic bad input (unknown state, missing destination/invoice) now terminal-fails the op instead of wedging in redelivery.
- **H-2** — redeem persists its allocated receive script *before* calling `RedeemCredit`, so a crash can't strand the redemption watching the wrong pkScript.
- **H-3** — auto-redeem is opt-out, and the sweep now subtracts credits a pending credit-backed send has earmarked (interlock wired from the wallet's prepared-send store into the redeemer).
- **H-4** — added crash-recovery tests that drive the real `Receive`/commit path (including injected `Stage`/`Commit` failures).
- **M-1/M-2/M-3/M-4/M-5/M-7/M-8** — persist-before-effect via `Stage`; bounded awaiting-settlement polling with a configurable cap; corrected `op_key` semantics doc; receive admission timeout; settlement reconcile; a saturating/ceil arithmetic wrapper around the sat figures; and the poll cap surfaced as an option.

One open dependency worth calling out: C-1 assumes the swap server surfaces the credit pay op in `ListCredits` with the payment hash and `debited`/`released`/`failed` states. That contract is being confirmed on the server side; it's noted in the design doc.

### Boarding fix found via the cross-repo e2e (`7fa1daee`)

Standing up the walletdk boarding e2e surfaced a pre-existing boarding bug (independent of credits, but it blocked the e2e). A boarding round was failing with arkd `ROUND_STATE_FAILED` / "vtxo[0] pkScript echo mismatch".

Root cause: `wallet.handleBoard` ships the full confirmed boarding set on every trigger, and the round actor mints a fresh VTXO owner key per registration. A boarding intent stays `Confirmed` from the trigger until the round's commitment-tx checkpoint, so when a second deposit confirms before the first round adopts, the eager per-block trigger re-ships the still-confirmed first outpoint, which the round actor re-registers under a *new* owner key. The server quotes one registration and the client validates the other, so the echo check fails. It only bites with two funding sources confirming across ticks, which is why no existing test caught it (the rest fund via OOR and never board).

Fix: make boarding registration idempotent per outpoint, with the wallet as the single authority. The wallet tracks in memory the outpoints it has handed to the round actor, excludes still-confirmed already-shipped ones from later triggers, skips the dispatch entirely when nothing new remains, and names the inputs each trigger covers in a new `TriggerBoardMsg.Outpoints` field. The round actor filters its own confirmed-boarding fetch to that set so the proven inputs stay coherent with the amounts. The shipped set is in-memory on purpose, so it clears on restart and the board replayer still re-boards an outpoint stranded by a failed round. Covered by new wallet and round unit tests; the swap-server boarding e2e is the end-to-end gate and is being re-run now.

PR #785, opened 2026-06-23
multi: harden TLV decoders against unbounded-allocation DoS, add fuzzing

Roasbeef:
### Motivation

Every TLV decoder in the client parses bytes that either cross the
client/server trust boundary or are replayed from a durable actor
mailbox on restart. A single malformed payload that reaches one of these
decoders can take the process down, and because durable messages are
replayed from disk on every restart, one poisoned mailbox entry becomes
a permanent crash loop.

### The bug

`lnd/tlv` sizes its value buffers with `make([]byte, declaredLength)`
before reading a single value byte. This happens in two spots:

- `DVarBytes`, used for every known `[]byte` record, and
- the unknown-record discard path in `stream.go`, which fires under
  `DecodeWithParsedTypes`.

Neither path on the non-P2P decode API caps the declared length. So
roughly ten attacker-controlled bytes that declare a record length near
2^63 panic the decoder with `makeslice: cap out of range`, and an
in-range-but-huge length drives the process to OOM. The unknown-record
path is the nasty part: even a message whose own fields are all
fixed-width scalars is vulnerable, because an injected unknown odd type
carries its own attacker-chosen length. We confirmed this is reachable
from essentially every decoder in the client.

### The fix

Each package gets a bounded-decode guard:

- decoders that use `DecodeWithParsedTypes` read the payload into a
  size-capped buffer and pre-validate the `(type, length)` framing,
  rejecting any record whose declared length exceeds the bytes
  physically present, before the real decoder ever allocates;
- hand-rolled `make([]T, count)` / `make([]byte, size)` sites bound the
  count or length against the bytes remaining in the reader;
- a couple of short-read-prone `reader.Read` calls become `io.ReadFull`.

The caps are per-package and generous (OOR, for instance, is sized for
many aggregated 64 KiB PSBT blobs, which `DecodeP2P`'s fixed per-record
cap would wrongly reject), so well-formed payloads of any realistic size
still decode; only genuinely corrupt framing is rejected, with a clean
wrapped error.

### Fuzzing

Each affected package gets a native Go fuzz harness over its decode
entry points. Every target seeds a valid encoding, the empty input, and
the canonical malicious-length payload, so the bound runs as a
deterministic regression on every `go test`. The crashers that surfaced
during development are kept as regression corpus.

### Packages hardened

`ledger`, `mailbox/conn`, `serverconn`, `db` (tree codec), `oor`,
`baselib/actor` + `internal/actortest`, `lib/types`, `lib/recovery`
(+ a `lib/bip322` signature fuzz harness), `lib/tx/checkpoint`,
`lib/tx/oor`, `unroll`, and `unrollplan`. `indexer` is encode-only and
needs no change.

### Testing

- `go build` and `go vet` clean across both modules.
- Existing unit tests pass.
- Every fuzz target runs clean (no panic, no OOM) after the fixes.

Commits are atomic per package.


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #786, opened 2026-06-24
rpc: fuzz the OOR and round proto parsers for panic safety

Roasbeef:
### Motivation

The darepo server's OOR and round RPC handlers parse client-controlled
proto bytes through helpers that live in this module (`rpc/oorpb`,
`rpc/roundpb`). A panic in one of those parsers is a server-crash vector
reachable by any client, so they deserve dedicated fuzz coverage.

This PR adds it. It is independent of the TLV-codec hardening in #785.

### What this adds

Fuzz harnesses over the RPC proto parsers, driven both field-level
(fuzzer-controlled PSBT / descriptor / outpoint / signature / tree-node
bytes) and via the realistic `proto.Unmarshal` wire path:

- **`rpc/oorpb`**: `ParseSubmitPackageRequest`,
  `ParseFinalizePackageRequest`, and the response parsers.
- **`rpc/roundpb`**: `OutpointFromProto`, `TxOutFromProto`,
  `PSBTFromBytes`, `SchnorrSigFromBytes`, `TxIDFromHex`,
  `MsgTxFromBytes`, and `TreeFromProto`.

All targets assert the no-panic contract; seeds include valid values,
empty input, off-length hashes/sigs, and (for the tree) oversize child
indices and large node counts.

### Findings

No bugs. These parsers already length/shape-check before fixed-size
copies, guard nil sub-messages, and feed btcd's PSBT/wire/schnorr
parsers only through a bounded reader. `TreeFromProto` is iterative (no
decode recursion) and enforces its `WithMaxTreeNodes` cap before any
per-node allocation, which bounds both node count and total allocation;
the pre-order `childIdx > i` invariant keeps the produced graph acyclic.
Fuzzing confirmed all of this holds on hostile input.

One out-of-scope note for a possible follow-up: `lib/tree`'s
`Node.Depth()` / `ForEach` are recursive, so a valid ~50k-node linear
chain accepted by the parser would recurse deeply in downstream
traversal. That is a `lib/tree` concern (not these parsers) and is
bounded in practice by the node cap and Go's growable stacks.

### Testing

- `go build ./rpc/...` and `go vet ./rpc/oorpb/ ./rpc/roundpb/` clean.
- All fuzz targets run clean (no panic, no OOM).

Commits are atomic per package.

### DIFF
--- a/rpc/oorpb/payloads_fuzz_test.go
+++ b/rpc/oorpb/payloads_fuzz_test.go
@@ -1,0 +1,247 @@
+package oorpb
+
+import (
+	"testing"
+
+	"google.golang.org/protobuf/proto"
+)
+
+// The parsers in this file (ParseSubmitPackageRequest and
+// ParseFinalizePackageRequest) run on the SERVER's OOR RPC handlers
+// against fully client-controlled proto bytes. A panic, OOM, slice
+// out-of-bounds, or integer overflow anywhere in these decode paths is a
+// remote server-crash vector: a single hostile client could take down
+// the operator. These fuzz targets assert ONLY the no-panic invariant
+// (the parsers are free to return errors); they are deliberately not
+// round-trip checks. Seeds cover a representative valid value, empty
+// input, and off-length edge cases (31/33-byte session hashes) that
+// historically trip fixed-length copies.
+
+// validSessionID is a representative 32-byte session hash used to seed the
+// fuzzers with a value that passes the length gate so the fuzzer explores
+// past it.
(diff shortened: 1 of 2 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #787, opened 2026-06-24
multi: Add per-object construction and flow versioning

ellemouton:
## Construction / flow versioning

Follow-on to the protocol-versioning work (#743, now merged). Where #743
negotiates a per-**session** Ark protocol version, this PR adds a
per-**object**, permanent **construction/flow version** to the three durable
objects a future protocol change would alter independently of session
negotiation:

| Object | Version | Persisted in |
|--------|---------|--------------|
| VTXO | `construction_version` | `vtxos` |
| Round | `flow_version` | `rounds` |
| OOR session | `flow_version` | `oor_session_registry` |

Each version is **stamped at creation, persisted, carried on the wire, and read
back with a fail-closed guard**. V1 is the only version that exists today, so
this is purely the *mechanism* — **nothing branches on the version yet**. The
point is that a future, genuinely different construction/flow can be introduced
additively without ambiguity about how an existing object was built.

### Pattern (per axis)
- A `version.go` with `XVersionV1`, `NormalizeXVersion` (0 → V1), and
  `ValidateXVersion` (fail closed on anything unknown), with the constant
  living next to its type.
- DB: stamp on insert (normalized, immutable/write-once), reject on load.
- Wire:
  - **VTXO** rides `arkrpc` indexer responses.
  - **Round** flow version is validated on receipt
    (`CommitmentTxBuilt.FromProto`) and threaded through the FSM onto the
    persisted `round.Round`, so the stored value mirrors what the operator
    stamped on the wire.
  - **OOR** rides `oorpb.SubmitPackageRequest` (client declares, operator
    validates at admission).

### Notes
- The base protocol-versioning PR **#743 has merged**; this PR now targets
  `main` directly.
- Schema columns are **folded into the existing `CREATE TABLE`** rather than
  added via forward migrations — darepo is pre-GA and DBs are recreated from
  scratch, so there is no on-disk table at an older version to `ALTER`.
- OOR's flow version is deliberately distinct from the existing
  `oor_session_registry.snapshot_version`, which versions only the resume
  blob's encoding.

### Testing
Unit Normalize/Validate tests per axis, persist→load round-trip assertions, and
synthetic-V2 rejection tests proving each guard fires in the *real* load/restore
path (not just the pure validator). All packages build, `lint-changed-local`
clean, `commitmsg-lint` clean.


ellemouton:

ellemouton:
Addressed the review feedback (folded into the relevant axis commits, no new commits):

**Docs honesty + `TODO(v2)` markers** (the over-claim every review flagged):
- `Round.FlowVersion` doc now states the wire value is validated *on receipt* but the persisted value is stamped V1 today, not threaded from `CommitmentTxBuilt`. Added a `TODO(v2)` at the drop point in `round/transitions.go` and on the field.
- `Descriptor.ConstructionVersion` doc now notes the indexer-wire read is deliberately deferred (`IncomingVTXOEvent` carries no version field), with a `TODO(v2)` to add it + thread it when a V2 exists.

**Polish:**
- VTXO load guard now wraps the error with the outpoint (`vtxo %s: %w`), matching the round path.
- Dropped the redundant `NormalizeFlowVersion(FlowVersionV1)` identity calls (OOR submit + registry bridge) in favour of the constant.
- Restored the displaced `TestRoundStoreCommitAndFetch` doc comment.

**Deliberately out of scope:** fully threading the wire value into the persisted object (and adding a version field to `IncomingVTXOEvent`) is the deferred coexistence work — doing it now would contradict the "mechanism only, V1" scope. The `TODO(v2)`s mark every site so the V2 implementer can't miss them. The DB load guards still fail closed on any unknown version regardless.

ellemouton:
Addressed the round-3 finding:

**Round-created VTXOs now stamp V1, not 0.** New commit `db: Stamp construction version on round-created VTXOs` — `domainVTXOToInsertParams` (the `SaveVTXOs` path that writes the initial row for round VTXOs) now stamps the normalized construction version instead of leaving the column unset (which persisted a write-once `0`). "Stamped at creation" is now true for round VTXOs, and the pre-field-row sentinel is no longer ambiguous.

The wire-guard observations (VTXO field-20 read deferred; round `FlowVersion` validated-on-receipt-but-not-threaded) remain intentional and `TODO(v2)`-marked, per the mechanism-only scope.

vtxo/version.go
```diff
@@ -0,0 +1,40 @@
...
+
+import "fmt"
+
+// ConstructionVersionV1 is the only VTXO construction version this build
+// understands. The construction version is a per-VTXO, permanent property,
+// stamped when a VTXO is first created and read back whenever it is loaded, so
+// that a future, genuinely different VTXO construction can be introduced
+// additively without ambiguity about how an existing VTXO must be spent or
+// exited. Until that future version exists, every VTXO is V1.
+const ConstructionVersionV1 uint32 = 1
```
Roasbeef: Should use an enum type here.

vtxo/interfaces.go
```diff
@@ -394,6 +394,25 @@ type Descriptor struct {
...
+	//
+	// TODO(v2): the operator stamps construction_version on the indexer
+	// VTXO wire message (arkrpc.VTXO), but the client materialization path
+	// does not yet read it onto this field (the IncomingVTXOEvent the
+	// client actually materializes from carries no version field), so a
+	// freshly materialized VTXO is stamped V1 at persistence. The DB load
+	// guard still protects against a locally-corrupted row. When a second
+	// construction exists, add the field to IncomingVTXOEvent and thread it
+	// here so the receive path mirrors the operator's value.
+	ConstructionVersion uint32
```
Roasbeef: Same re enum type.

vtxo/version.go
Roasbeef: Any reason to not zero index?

round/transitions.go
```diff
@@ -1636,6 +1636,13 @@ func (s *RoundJoinedState) processEvent(ctx context.Context, event ClientEvent,
 			slog.Int("vtxo_tree_count", len(evt.VTXOTreePaths)),
 		)
 
+		// TODO(v2): evt.FlowVersion was validated on receipt in
```
Roasbeef: Why not resolve now/

round/version.go
```diff
@@ -0,0 +1,38 @@
...
+
+import "fmt"
+
+// FlowVersionV1 is the only round flow version this build understands. The flow
+// version is a per-round, permanent property describing the choreography rules
+// under which a round was created. The operator stamps it and the client
+// records the same value, so a future, genuinely different round flow can be
+// introduced additively without ambiguity about how an existing round was
+// conducted. Until that future version exists, every round is V1.
+const FlowVersionV1 uint32 = 1
```
Roasbeef: Same here re if should zero index

rpc/roundpb/version.go
```diff
@@ -0,0 +1,38 @@
...
+// introduced additively without ambiguity about how an existing round was
+// conducted. Until that future version exists, every round is V1.
+const FlowVersionV1 uint32 = 1
+
+// NormalizeFlowVersion maps a zero flow version (a value that predates the
+// field) to the V1 default. Any non-zero value is returned unchanged so an
+// unknown version still surfaces through ValidateFlowVersion rather than being
+// silently rewritten to V1.
+func NormalizeFlowVersion(v uint32) uint32 {
+	if v == 0 {
```
Roasbeef: Wouldn't need this is it zero indexeed

db/vtxo_store.go
```diff
@@ -751,6 +769,16 @@ func (s *VTXOPersistenceStore) rowToDescriptor(ctx context.Context,
 		clientKey.PubKey = derived.clientPubkey
 	}
 
+	// Fail closed on a VTXO built under a construction version this build
+	// does not understand. A zero value predates the column and is treated
+	// as V1; any other unknown value is rejected rather than loaded.
+	constructionVersion := uint32(row.ConstructionVersion)
+	if err := vtxo.ValidateConstructionVersion(
```
Roasbeef: I think we also need to validate the version here: https://github.com/lightninglabs/darepo-client/blob/bc831a689a939a9c37d6855d84fabde10d90a135/db/round_store.go#L1524

vtxo/version.go
Roasbeef: Don't see where we adopt/validate this when we handle `IncomingVTXOEvent` (server ntfn of recv).

arkrpc/indexer.proto
```diff
@@ -442,6 +442,12 @@ message VTXO {
     // operator_pubkey is the compressed operator public key committed to this
     // VTXO's collaborative spend path.
     bytes operator_pubkey = 19;
+
+    // construction_version is the per-VTXO construction version: the rules
+    // under which this VTXO was built and must be spent or exited. The
+    // operator stamps it at creation; the receiving client records the same
+    // value so both sides agree. Today the only understood value is 1.
+    uint32 construction_version = 20;
```
Roasbeef, line 453: The comment above says "the receiving client records the same value so both sides agree", but no client path actually reads this field yet. Seed recovery (`recoveryDescriptorFromIndexer` in `darepod/wallet_recovery.go:339`) takes the `arkrpc.VTXO` and ignores `construction_version` entirely, rebuilding the descriptor under hard-coded V1 rules and persisting `ConstructionVersion=0`, which the store normalizes to V1, write-once. The live receive path is worse off: it materializes from `VTXOEvent`/`IncomingVTXOEvent`, and neither message carries the field at all, so there is nothing to validate even if the handler wanted to.

Net effect: a recipient on this build that receives or seed-recovers a V2-constructed VTXO stamps it as a valid V1 row, every later DB-load guard passes, and the wallet reports funds it can't spend or unilaterally exit, with no error anywhere. Because the mis-stamp is write-once, this is strictly worse than a missing guard: it permanently defeats the guards this PR installs.

The recovery path is fixable in this PR with no proto change (the field is already on this message): validate `GetConstructionVersion()` fail-closed in `recoveryDescriptorFromIndexer` and thread the normalized value onto the descriptor. The event protos growing the field can be a fast-follow, but until then the comment here (and the matching one in the server's indexer store) overstates what the receiving side does.

db/round_store.go
```diff
@@ -844,9 +853,21 @@ func (s *RoundPersistenceStore) dbRoundToDomainRound(ctx context.Context,
 		return nil, fmt.Errorf("parse round ID: %w", err)
 	}
 
+	// Fail closed on a round conducted under a flow version this build does
+	// not understand. A zero value predates the column and is treated as
+	// V1.
+	if err := round.ValidateFlowVersion(
+		uint32(dbRound.FlowVersion),
+	); err != nil {
+		return nil, fmt.Errorf("round %s: %w", roundID, err)
```
Roasbeef: This validation covers `dbRoundToDomainRound`, but `reconstructInputSigSentState` (line ~1099) rebuilds the FSM state from the same `dbRound` and never sets `FlowVersion` on the `round.InputSigSentState` it constructs. A client restarting mid-round resumes with `state.FlowVersion=0`, every later transition propagates the zero, and `CommitState` silently normalizes it back to V1 at persistence. Harmless today (the row is write-once and stays correct), but in a V2 world any behavior branch on the state's version takes the V1 path after any restart: a silent downgrade rather than an error.

It also shows the sharp edge in threading the version by hand through 7 state structs and a dozen transition constructions: a forgotten copy can never fail loudly, because zero means V1. Setting it here is a one-liner, but longer term it may be worth carrying the version once on the shared `Round`/environment instead of per-state copies.

round/version.go
Roasbeef: Same note as on the server PR: this file (and `oor/version.go`, `vtxo/version.go`) now exists in near-identical form in both repos, six copies plus six matching test files, all guarding values that cross the wire in proto packages that live in this module and that the server already imports. The two existing protocol versions each live once here for exactly that reason (`arkrpc.ArkProtocolVersionV1`, `mailboxpb.MailboxProtocolVersionV1`), so the two sides can't drift.

Hosting each constant + Normalize/Validate next to its wire field (`roundpb`, `oorpb`, `arkrpc`) needs zero new import edges on either side, and removes the failure mode where a future V2 bump lands in one module but not its mirror, which for the round/OOR axes means fleet-wide rejection of every batch announcement or submit. The load-bearing subtlety ("Normalize must NOT clamp unknown values") would then live in one place instead of six.

Roasbeef (commented):
Client half of the review pass I just left on lightninglabs/darepo#601 (the server PR pins its submodule at this branch's head, and I verified field numbers and validation symmetry across the pair). Same overall read: the stamp/persist/carry/guard pattern is applied consistently, the guards fire in the real load/restore paths, and the tests back that up.

The comments below cover the client-specific seams: the VTXO axis is the one place the fail-closed story is actually fail-open (the receiving side never reads the version the operator advertises, and a wrong write-once V1 stamp defeats every later guard), the crash-restore path rebuilds FSM state without the version the same function just validated, and the version constants now live in six near-identical copies across the two repos. The first one is the pre-merge item; the rest are fix-now or fast-follow.

db/round_store.go
```diff
@@ -296,6 +297,14 @@ func (s *RoundPersistenceStore) CommitState(ctx context.Context, r *round.Round,
 			CreationTime:          nowUnix,
 			LastUpdateTime:        nowUnix,
 			StartHeight:           int32(r.StartHeight),
+
+			// Stamp the round flow version, normalizing a zero
+			// (a round recorded before this field existed) to V1.
+			FlowVersion: int32(
```
Roasbeef: nit, applies to all the insert-side stamps in this PR (here, the VTXO descriptor path in `db/vtxo_store.go`, and the record-store insert): the write side runs `Normalize` but never `Validate`, and Normalize deliberately passes an unknown non-zero value through. So a bug that puts a bogus version on an in-memory object persists silently and only surfaces as a load failure at some later restart, far from the write that caused it. Validating before the stamp keeps the failure attributable; the rejection tests can keep injecting bad rows via raw sqlc.

oor/session_registry_bridge.go
```diff
@@ -114,6 +114,11 @@ func outgoingRegistryRecord(sessionID SessionID,
 		LastError:       lastError,
 		SnapshotData:    raw,
 		SnapshotVersion: int32(snapshot.Version),
+
+		// Stamp the OOR flow version this session is conducted under.
+		// It is write-once on the DB row, so re-stamping V1 on every
+		// staged write is a no-op after the first insert.
+		FlowVersion: FlowVersionV1,
```
Roasbeef: nit: this stamps the build constant rather than a version carried on the session, which is the "re-defaulted to the constant" shape the PR description warns about. Write-once semantics protect re-stamps of an existing row, but the first insert of a session is not protected: when a second flow exists, a V2 session staged through this bridge gets stamped V1 at creation unless someone remembers to thread the session's own version through here (and the incoming twin below). Fine today since V1 is the only flow, but worth a TODO so this site shows up in the V2 grep.

Roasbeef (approved):
LGTM 🐡

db/round_store.go
```diff
@@ -282,6 +284,16 @@ func (s *RoundPersistenceStore) CommitState(ctx context.Context, r *round.Round,
 
 		nowUnix := s.clock.Now().Unix()
 
+		// Validate before stamping so a bogus in-memory flow version
+		// fails at the write that introduced it, keeping the blast
+		// radius at this one object rather than converting into an
+		// opaque load failure on some later restart.
+		if err := roundpb.ValidateFlowVersion(
```
Roasbeef: Same here as the other PR, can do w/o enforcing this at the db level. Now we coupling rpc semantics with simple storage.

PR #801, opened 2026-06-25
walletdk: expose the wallet SDK to mobile via gomobile

### LINKED ISSUE #713: sdk/walletdk: expose the wallet SDK to mobile via gomobile
## Summary

Expose the `sdk/walletdk` SDK to mobile (iOS + Android) via `gomobile bind`,
so React Native / Swift / Kotlin hosts can drive an embedded `darepod` wallet
in-process without standing up a separate daemon binary or opening a socket.

We already do the hard part inside `sdk/walletdk`: `Start` boots the daemon
in-process and dials it over a private `bufconn` gRPC transport
(`sdk/walletdk/embedded.go:112`). So unlike the `falafel` path, we do **not**
need a `protoc` plugin generating a parallel binding layer. We have a
hand-written, wallet-shaped facade already; the work is to add a thin
mobile-safe wrapper package on top of it that respects `gomobile`'s type
restrictions, plus a high-level message-pump API that meshes with the mailbox
ingress loop.

## Background: how falafel / lnd did it

`lightninglabs/falafel` is a `protoc` plugin that flattens gRPC services into a
`gomobile`-safe API and wires it to the real server over an in-memory
`bufconn` listener (no open port). The generated convention is:

- **bytes-in / bytes-out** — every method takes a serialized protobuf `[]byte`
  and hands results back as `[]byte`; the host (de)serializes with
  swift-protobuf / protobuf-javalite.
- **callback-out for everything async** — you cannot return a Go channel, so
  results are pushed through host-implemented listener interfaces:
  - `Callback{ OnResponse([]byte); OnError(error) }` — unary, fired once.
  - `RecvStream{ OnResponse([]byte); OnError(error) }` — server-streaming, EOF
    delivered as a terminal `OnError`.
  - `SendStream{ Send([]byte) error; Stop() error }` — caller-driven side of a
    bidi stream.
- **lifecycle** — `Start(args string, ready Callback)`, singleton-guarded with
  a CAS, daemon on a goroutine, `ready.OnResponse(nil)` once the RPC server is
  serving; `RecreateListeners()` on each restart because `bufconn` listeners
  close on shutdown.

We can adopt the *shape* (bytes-in / callback-out) without adopting the
*generator*, since our facade already owns the bufconn wiring.

## `gomobile bind` type restrictions (the constraint we design around)

From `golang.org/x/mobile/cmd/gobind`, exported symbols may only use:

- signed integer and floating-point types;
- `string` and `bool`;
- `[]byte` (passed by reference, mutable — copy before handing to a goroutine);
- function types whose params/results are all supported, returning 0, 1, or
  `(T, error)`;
- interfaces whose exported methods are all supported function types;
- structs whose exported methods/fields are all supported.

**No** `uint*`, **no** maps, **no** slices other than `[]byte`, **no**
channels, **no** `context.Context`, **no** arbitrary struct params. Panics do
**not** cross the boundary — an unrecovered panic exits the process, so every
goroutine body must `recover()` and route failures to `OnError`. The
`(value, error)` convention maps to Swift `throws` / Java checked exceptions.

## Gap analysis: what in `sdk/walletdk` is not mobile-safe today

The current public surface (`client.go`) violates the restrictions in several
ways. Inventory of what a wrapper must hide or reshape:

| Issue | Where | Fix in the wrapper |
|-------|-------|--------------------|
| `context.Context` param on every RPC method + `Start`/`Connect` | all of `GetInfo`, `CreateWallet`, … `Status` | drop ctx from the mobile signature; hold an internal `context.Context` keyed off `Start`/`Stop` lifecycle |
| `Subscribe` returns `(<-chan Entry, <-chan error, error)` | `client.go:499` | replace with a `RecvStream`-style callback registration |
| `Wait() <-chan error` | `client.go:94` | replace with a `ready Callback` / terminal-error callback |
| Escape hatches return external interfaces (`GRPCConn`, `WalletRPC`, `ArkRPC`, `SwapRPC`, `Btcwallet*`) | `client.go:103-148` | exclude entirely from the mobile package |
| `Start(ctx, cfg, opts ...Option)` — variadic func type | `embedded.go:112` | mobile `Start(cfgJSON string, ready Callback)`; fold options into config |
| `Config.DaemonConfig *darepod.Config`, `Config.LogWriter io.Writer`, `ConnectConfig.DialOptions`, `ConnectConfig.HTTPClient` | `embedded_config.go`, `types.go` | mobile config is a flat, scalar-only struct (or a JSON string) |
| `[]string` / `[]EntryKind` / `[]Entry` / `[]WalletVTXO` / `[]OnchainTx` fields on DTOs | `types.go` | bytes-in/bytes-out: marshal DTOs to protobuf/JSON `[]byte` |
| `time.Time` fields (`Entry.CreatedAt`, etc.) | `types.go` | int64 unix seconds in the wire form |
| `uint64` amount fields (`AmountSat`, `MaxFeeSat`, …) | `types.go` | int64, or string for >2^53 safety on JS hosts |

### 1. New package: `sdk/walletdk/mobile` (built with `//go:build mobile` + `walletdkrpc`)

A flat facade that owns a `*walletdk.Client` and translates at the boundary.
Two viable conventions for the request/response payloads:

- **(A) bytes-in / bytes-out** — mirror lnd: each method takes the serialized
  `walletdkrpc` request proto and returns the serialized response proto. Hosts
  generate the proto classes from our existing `.proto` files. Maximally
  faithful, zero hand-maintained DTO drift, but pushes (de)serialization onto
  every host.
- **(B) scalar/JSON facade** — keep the rich DTOs but JSON-encode them to
  `string`/`[]byte` at the boundary, and pass simple calls (e.g. `Balance`) as
  flat scalars. Friendlier for a quick React Native bridge; more
  hand-maintained surface.

**Recommendation:** lead with (A) for the RPC verbs (it's what
swift-protobuf/protobuf-javalite hosts expect and what falafel/lnd settled on),
and offer a small number of (B)-style scalar conveniences for the hottest paths
(`Balance`, `Status`) where a host shouldn't need a proto runtime. Decide this
explicitly before building — see open questions.

### 2. Lifecycle

```go
// All //go:build mobile.

// Start boots the embedded daemon from a serialized config and fires
// ready.OnResponse(nil) once gRPC is serving (or OnError on failure).
// Singleton-guarded via CAS; a second Start calls OnError.
func Start(cfg []byte, ready Callback)

// Stop tears down the daemon and releases the bufconn transport. Idempotent.
func Stop() error
```

This wraps the existing `walletdk.Start` / `Client.Stop`. The internal
`context.Context` is created here and cancelled by `Stop`, so the mobile API
never has to express a ctx. Honor the existing invariant that daemon lifetime
is owned by walletdk's `runCtx`, not a per-call ctx
(`sdk/walletdk/CLAUDE.md` invariants). Singleton state must reset on daemon
exit so a host can restart after backgrounding.

### 3. Async results: callback interfaces

```go
type Callback interface {
    OnResponse(b []byte)
    OnError(err error)
}

type RecvStream interface {
    OnResponse(b []byte)
    OnError(err error) // EOF is terminal
}
```

`Subscribe` becomes `Subscribe(req []byte, stream RecvStream)`; the wrapper
runs the existing channel-based `Client.Subscribe` in a goroutine
(`client.go:499`), forwarding each `Entry` to `OnResponse` and the terminal
error to `OnError`. Every goroutine body gets a `recover()` that routes to
`OnError` — panics crashing the process is a real gomobile failure mode.

### 4. High-level `ProcessNextMsg` — meshing with the mailbox loop

The mailbox ingress loop (`serverconn/ingress.go:28`, started by
`StartIngress` at `serverconn/actor.go:1069`) currently runs as a background
goroutine: pull a batch (`pullBatch`, `ingress.go:214`) → dispatch
(`dispatchBatch`, `ingress.go:252`) → ack + checkpoint, with exponential
backoff for self-healing across TCP RSTs (`serverconn/mailboxpull/pull.go:50`).
Cursor state is persisted, so a dropped connection just retries from the same
position — there is no explicit "reconnect" event.

On mobile, a host that is frequently suspended/resumed by the OS benefits from
**driving** that loop rather than relying on a always-on background goroutine
that the OS may freeze. Proposal: expose a synchronous, host-driven pump:

```go
// ProcessNextMsg pulls and processes a single mailbox envelope (or returns
// after a long-poll timeout with nothing to do), advancing the ack
// watermark and checkpointing. Returns an opaque status the host can use to
// decide whether to loop again, back off, or sleep.
func ProcessNextMsg() (status []byte, err error)
```

This requires a small refactor in `serverconn`: factor the body of
`ingressLoop` into a `processOnce(ctx, state) (state, outcome, error)` step
that the existing goroutine loop calls, and that the mobile pump can also call
directly. The background loop stays the default for the daemon/server; mobile
opts into host-driven stepping. Key correctness constraints to preserve:

- the four-cursor ack watermark invariant `AckCommittedTo <= DispatchCommittedTo`
  (`mailbox/conn/ack_state.go`) — never ack past non-durable local work;
- checkpoint after dispatch, before advancing ack;
- backoff/self-heal still applies (the host should not have to reimplement it);
  `processOnce` returns enough in `outcome` for the host to know whether to
  retry immediately or sleep.

Open question: do we expose `ProcessNextMsg` from `serverconn` and let the
mobile wrapper call it, or keep mobile purely on the walletdk facade and run
the standard background ingress goroutine? The host-driven pump is the more
mobile-native shape (cooperative with OS suspension, async/await friendly), but
it's a larger change. See open questions.

### 5. Richer-type wrappers for Swift / Java

With bytes-in/bytes-out, rich types live in protobuf and we lean on
swift-protobuf / protobuf-javalite generated classes — no hand-written ObjC/
Swift/Java wrappers needed for the data model. We should still ship thin,
ergonomic native wrappers (a Swift `WalletClient` actor, a Kotlin
`WalletClient` class) that:

- adapt the `Callback`/`RecvStream` interfaces to Swift `async`/`AsyncStream`
  and Kotlin coroutines/`Flow`;
- map the `(value, error)` boundary to Swift `throws` / Kotlin exceptions;
- own the singleton `Start`/`Stop` lifecycle and tie it to app
  foreground/background.

These are small and live in the host repos, but we should provide reference
implementations + a usage doc here.

## Work breakdown

- [ ] Decide payload convention: bytes-in/bytes-out (A) vs JSON facade (B) vs hybrid.
- [ ] Add `sdk/walletdk/mobile` package behind `//go:build mobile,walletdkrpc`.
- [ ] Flat, serializable mobile `Config` (no `DaemonConfig`/`io.Writer`/dial opts).
- [ ] `Start(cfg, ready Callback)` / `Stop()` singleton lifecycle with CAS + reset-on-exit.
- [ ] `Callback` / `RecvStream` interfaces; convert `Subscribe` and `Wait` to callbacks.
- [ ] Wrap each RPC verb (drop ctx, bytes/scalars across the boundary, `recover()` in goroutines).
- [ ] (If pursuing the pump) refactor `serverconn` `ingressLoop` into a reusable `processOnce` step and expose `ProcessNextMsg`.
- [ ] `gomobile bind` build targets + a `gen_bindings.sh` (ios xcframework, android aar); CI job to keep it compiling.
- [ ] Reference Swift + Kotlin wrappers and an integration doc under `docs/`.
- [ ] Tests: a `mobile`-tagged unit suite exercising the bytes boundary and the callback/`recover` paths against a bufconn-embedded daemon.

## Open questions

1. **Payload convention** — full protobuf bytes-in/bytes-out, JSON, or hybrid
   (proto for verbs + scalars for `Balance`/`Status`)? This drives the whole
   wrapper.
2. **Mailbox pump** — host-driven `ProcessNextMsg` (cooperative with OS
   suspension, needs the `serverconn` `processOnce` refactor) vs. standard
   always-on background ingress goroutine inside the embedded daemon? The
   former is more mobile-native; the latter is zero new mailbox surface.
3. **Build/target ownership** — does the `gomobile bind` output (xcframework /
   aar) get built + published from this repo's CI, or consumed from a host
   mobile repo?
4. **Secret handling** — `Mnemonic`/`WalletPassword`/`SeedPassphrase` are
   cloned + zeroable at the SDK boundary today. Confirm the bytes-in path
   preserves that (copy-then-zero) so secrets aren't pinned by a host proxy
   object across the boundary.

## References

- `sdk/walletdk/embedded.go:112` — existing in-process `bufconn` wiring.
- `sdk/walletdk/client.go:499` — `Subscribe` channel shape to convert.
- `serverconn/ingress.go:28` / `serverconn/actor.go:1069` — mailbox ingress loop.
- `mailbox/conn/ack_state.go` — ack watermark invariants the pump must respect.
- `lightninglabs/falafel`, `lightningnetwork/lnd/mobile` — prior-art binding shape.
- `golang.org/x/mobile/cmd/gobind` — type restrictions.

Roasbeef:
This series exposes `sdk/walletdk` to iOS and Android through `gomobile
bind`, so a React Native, Swift, or Kotlin host can drive an embedded
`darepod` wallet in-process, with no separate daemon binary and no open
socket.

We already do the hard part in `sdk/walletdk`: `Start` boots the daemon and
dials it over a private bufconn gRPC transport. So rather than a
falafel-style protoc plugin, we add a thin hand-written facade,
`sdk/walletdk/mobile`, that translates at the boundary and respects
gomobile's type restrictions.

The facade is callback-free. gomobile carries only a narrow set of types, so
the rich `walletdk.Client` surface (context, channels, maps, unsigned ints,
time.Time, slices of structs) cannot cross directly. Instead:

- RPC verbs take and return JSON bytes; a few hot paths return scalars.
- `Start` is synchronous and singleton-guarded. It returns once gRPC is
  serving, which `walletdk.Start` already guarantees, unlike `lnd.Main`,
  which blocks forever and forces lnd-mobile to use a callback.
- The one streaming verb hands back a pull-based `Subscription` instead of a
  channel or a host callback.
- Every entry point recovers panics into errors, since a panic does not
  cross the gomobile boundary and would otherwise kill the host process.

The package builds only under `mobile && walletdkrpc && swapruntime`, so
default builds are untouched. `make mobile-android` / `mobile-ios` drive
`gen_bindings.sh`.

### What is here

- `sdk/walletdk/mobile`: the facade plus its boundary tests.
- `make mobile-*` targets and `gen_bindings.sh`.
- `docs/walletdk_mobile.md`.
- A Mobile Bindings CI workflow: a fast host-tagged compile on PRs and the
  merge queue, with the heavy `.aar` / `.xcframework` builds gated to
  `mobile-v*` tags so routine merges do not trigger them.

### What is not here

The sample apps and the idiomatic Kotlin (coroutines / `Flow`) and Swift
(`async` / `AsyncThrowingStream`) wrappers live in
[lightninglabs/damobile](https://github.com/lightninglabs/damobile). Both
were validated end to end against signet: the embedded wallet boots
in-process, connects to the operator mailbox, syncs from Esplora, and runs
on the Android emulator and the iOS simulator.

Closes #713.

### DIFF
--- a/.github/workflows/mobile-bindings.yml
+++ b/.github/workflows/mobile-bindings.yml
@@ -1,0 +1,144 @@
+name: Mobile Bindings
+
+# The mobile binding package compiles on every relevant PR (a fast, host-tagged
+# build), but the expensive gomobile .aar / .xcframework builds (which cross
+# compile the whole embedded daemon for every ABI) run only when a
+# `mobile-v*` tag is pushed, so routine merges do not trigger them.
+on:
+  push:
+    tags:
+      - "mobile-v*"
+  pull_request:
+    paths:
+      - "sdk/walletdk/**"
+      - "swapwallet/**"
+      - "Makefile"
+      - ".github/workflows/mobile-bindings.yml"
+  merge_group:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: mobile-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  # If you change this please also update GO_VERSION in Makefile.
+  GO_VERSION: 1.25.5
+  MOBILE_TAGS: "mobile walletdkrpc swapruntime"
+
+jobs:
+  ########################################################################
+  # Fast gate: the facade must compile and its boundary tests must pass.
+  # Runs on PRs and the merge queue, not on tag builds.
+  ########################################################################
+  unit:
+    name: Compile + test (host)
+    if: github.event_name != 'push'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v5
+
+      - name: Setup Go ${{ env.GO_VERSION }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Test mobile facade
+        run: go test -tags="${MOBILE_TAGS}" ./sdk/walletdk/mobile/...
+
+  ########################################################################
+  # Android .aar — tag builds only.
+  ########################################################################
+  android:
+    name: Build Android .aar
+    if: startsWith(github.ref, 'refs/tags/mobile-v')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v5
+
+      - name: Setup Go ${{ env.GO_VERSION }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Setup JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "17"
+
+      - name: Setup Android SDK
+        uses: android-actions/setup-android@v3
+
+      - name: Setup Android NDK
+        uses: nttld/setup-ndk@v1
+        id: ndk
+        with:
+          ndk-version: r27c
+
+      - name: Install gomobile
+        run: |
+          go install golang.org/x/mobile/cmd/gomobile@latest
+          go install golang.org/x/mobile/cmd/gobind@latest
+          echo "$(go env GOPATH)/bin" >> "$GITHUB_PATH"
+
+      - name: Build .aar
+        env:
+          ANDROID_NDK_HOME: ${{ steps.ndk.outputs.ndk-path }}
+        run: |
+          gomobile init
+          make mobile-android
+
+      - name: Upload .aar
+        uses: actions/upload-artifact@v4
+        with:
+          name: Walletdk-aar
+          path: sdk/walletdk/mobile/build/android/Walletdk.aar
+          if-no-files-found: error
+
+  ########################################################################
+  # iOS .xcframework — tag builds only (macOS runner for Xcode).
+  ########################################################################
+  ios:
+    name: Build iOS .xcframework
+    if: startsWith(github.ref, 'refs/tags/mobile-v')
+    runs-on: macos-latest
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v5
+
+      - name: Setup Go ${{ env.GO_VERSION }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Install gomobile
+        run: |
+          go install golang.org/x/mobile/cmd/gomobile@latest
+          go install golang.org/x/mobile/cmd/gobind@latest
+          echo "$(go env GOPATH)/bin" >> "$GITHUB_PATH"
+
+      - name: Build .xcframework
+        run: |
+          gomobile init
+          make mobile-ios
+
+      - name: Package .xcframework
+        run: |
+          tar -czf Walletdk.xcframework.tar.gz \
+            -C sdk/walletdk/mobile/build/ios Walletdk.xcframework
+
+      - name: Upload .xcframework
+        uses: actions/upload-artifact@v4
+        with:
+          name: Walletdk-xcframework
+          path: Walletdk.xcframework.tar.gz
+          if-no-files-found: error
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 .PHONY: ast-lint ast-grep-fix
 .PHONY: unit unit-cover unit-race unit-swapruntime check-go-version build install clean release
 .PHONY: build build-swapruntime build-swapclient build-walletdkrpc rpc install install-swapruntime install-walletdkrpc help clean-networks
+.PHONY: mobile mobile-android mobile-ios
 .PHONY: systest systest-verbose
 .PHONY: commitmsg-lint commitmsg-fmt commitmsg-reword
 
@@ -435,6 +436,18 @@ build-swapclient: build-swapruntime #? Alias for build-swapruntime
 build-walletdkrpc: #? Build debug binaries with walletdkrpc + swapruntime enabled
 	@$(call print, "Building debug binaries with walletdkrpc and swapruntime.")
 	$(MAKE) build tags="walletdkrpc swapruntime"
+
+mobile: #? Build gomobile bindings for sdk/walletdk (target=android|ios|all)
+	@$(call print, "Building gomobile walletdk bindings.")
+	./sdk/walletdk/mobile/gen_bindings.sh $(or $(target),android)
+
+mobile-android: #? Build the Android .aar for sdk/walletdk
+	@$(call print, "Building Android .aar for walletdk.")
+	./sdk/walletdk/mobile/gen_bindings.sh android
+
+mobile-ios: #? Build the iOS .xcframework for sdk/walletdk
+	@$(call print, "Building iOS .xcframework for walletdk.")
+	./sdk/walletdk/mobile/gen_bindings.sh ios
 
 install: #? Build and install binaries to GOPATH/bin
 	@$(call print, "Installing binaries.")
--- a/docs/walletdk_mobile.md
+++ b/docs/walletdk_mobile.md
@@ -1,0 +1,187 @@
+# walletdk on mobile (gomobile)
+
+`sdk/walletdk/mobile` is a [gomobile](https://pkg.go.dev/golang.org/x/mobile/cmd/gomobile)-safe
+facade over the wallet SDK. It lets an Android (Kotlin/Java) or iOS
+(Swift/ObjC) host drive an embedded `darepod` wallet **in-process** — no
+separate daemon binary, no open socket — by reusing the private `bufconn` gRPC
+transport that [`walletdk.Start`](../sdk/walletdk/embedded.go) already sets up.
+
+It borrows the *bytes-out* idea from `lightninglabs/falafel` and `lnd/mobile`
+**without** the protoc generator or their callback interfaces, because the
+`walletdk` facade already owns the wiring and `walletdk.Start` returns once gRPC
+is serving (`lnd.Main` blocks forever, so lnd needs a callback — we do not).
+
+## Why a separate package
+
+`gomobile bind` only carries a narrow set of types across the FFI boundary:
+signed ints, floats, `string`, `bool`, `[]byte`, and interfaces/structs whose
+members are all of those. It cannot carry `context.Context`, channels, maps,
+`uint*`, `time.Time`, slices other than `[]byte`, or tagged-union structs — all
+of which appear throughout `walletdk`'s public API. So `mobile` is a flat
+translation layer; it never exposes the rich `walletdk` types directly.
+
+The package is gated behind three build tags — `mobile`, `walletdkrpc`, and
+`swapruntime` — so it only compiles into a `gomobile bind` output. An ordinary
+`go build ./...` sees only the unconstrained `doc.go`.
+
+## The payload contract (hybrid)
+
+| Shape | Convention | Verbs |
+|-------|-----------|-------|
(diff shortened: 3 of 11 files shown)


Roasbeef:
Follow-up: #802 tracks the mobile mailbox wake-and-drain lifecycle (push-driven wake instead of the always-on ingress loop), which was open question #2 from #713. This PR ships the bindings with the existing always-on loop; #802 scopes the `processOnce` factoring and the push integration on top.

Roasbeef:
@claude review this

Roasbeef:
Addressed the review feedback and force-pushed (branch rebased onto latest
`main`, fixups autosquashed into their target commits, so the history stays at
five clean commits).

- **Lifecycle race in `Start`/`Stop`** (gemini + codex, P1): replaced the
  single started/stopped CAS with a four-state machine (stopped / starting /
  started / stopping) plus a stored `startCancel` and a generation counter. A
  `Stop` that races an in-progress `Start` now cancels the boot; the `Start`
  then tears down any client it produced instead of publishing it under a reset
  guard, and `Stop` only returns after `client.Stop()` completes. (in
  `walletdk: add gomobile-safe mobile binding package`)
- **`gen_bindings.sh` empty-NDK glob** (gemini): the NDK lookup now uses `find`
  instead of a `*` glob, so an empty `ndk/` directory no longer aborts the
  script under `set -euo pipefail`. (same commit)
- **Negative signed config** (codex, P2): `parseConfig` now rejects a negative
  `wallet_poll_interval_seconds` (which would otherwise panic the tip poller's
  ticker), recovery window, fee, and buffer size, with a test covering them.
  (config in the package commit, test in `walletdk: test the mobile binding
  boundary`)
- **Doc JSON field names** (codex, P2): corrected `docs/walletdk_mobile.md` to
  state that the verb DTOs have no `json` tags, so the wire keys are the Go
  field names (PascalCase: `Version`, `ConfirmedSat`, …), with the snake_case
  `Start` config called out as the one tagged exception. (in `docs: add
  walletdk mobile bindings guide`)

The host wrappers in lightninglabs/damobile already map the PascalCase response
keys, so that side is consistent.

Roasbeef:
@claude review this

Roasbeef:
Thanks for the thorough re-review. Addressed the open item and the polish
notes, and fixed the `fmt-changed` failure (had run only `gofmt` locally, not
`llformat`). Force-pushed; still five clean commits.

- **`Subscription.Next` cancel-vs-EOF** (the behavioral one): `Next` now stores
  the subscription's `ctx` and, when `updates` closes with a non-nil `errs`,
  returns `io.EOF` if `s.ctx.Err() != nil` (a self-initiated `Close`/`Stop`)
  rather than the wrapped `context.Canceled`. So a `Stop` that fires while a
  `Flow`/`AsyncStream` is collecting (the app-suspend path) now completes the
  host loop cleanly, matching the doc. A genuine upstream error still
  propagates. (in the package commit)
- **`IsRunning()` boot window**: now backed by the lifecycle status, so it
  reports `true` across the whole `starting` window (not only once gRPC is
  serving), with the doc noting a `true` during boot does not yet imply RPCs
  succeed (use `WalletReady`).
- **Recovery-window upper bound**: `validate()` now also rejects
  `wallet_recovery_window > math.MaxUint32`, so the `uint32` narrowing is total;
  covered by a new test.
- **Wire-name test**: `TestEntryRoundTripsAsJSON` now asserts the literal
  PascalCase keys (`"ID"`, `"Kind"`, `"AmountSat"`) so the documented public
  contract is pinned, not just a Go-to-Go round trip.
- **`Next` marshal consistency**: switched from `json.Marshal` to the shared
  `marshal` helper.

Verified locally: `make fmt-changed-check` clean, mobile test gate green
(including the new cases), all five commits pass `commitmsg-lint`.

The unit failures in the previous run were a flaky `db/actordelivery`
worker-pool test (`TestDurableActorPerKeyFIFOUnderWorkerPool`) unrelated to
this branch, which touches no non-mobile code.

Roasbeef:
@claude review this

PR #803, opened 2026-06-25
walletdk: browser (js/wasm) wallet bindings

Roasbeef:
In this PR, we add the WebAssembly wallet bindings for `walletdk`, so the
full embedded daemon (wallet + swaps + OOR) runs inside a single browser VM
with no separate gateway process. This supersedes #431, which had drifted out
of sync with `main` over the past few months: the OOR actor set was rewritten
(the old global signing-effect actor is gone), the `walletdk.Client` send path
moved to a prepare/dispatch split, and `sdk/walletdk` itself was reorganized
around the `walletdkrpc` rename. Rather than untangle that by hand, we rebuilt
the wasm portions from scratch against current `main`.

This work stacks on top of #713 (the `sdk/walletdk/mobile` gomobile facade), so
the base branch here is `walletdk-mobile-gomobile`, not `main`. The browser
bridge drives that same JSON facade, which means the wasm and gomobile bindings
now share one source of truth, and the bridge can't drift from the `Client` API
the way the old hand-written `syscall/js` switch did. That drift is exactly what
broke #431 (`client.Send` no longer exists).

## What's in here

We pulled in the OPFS-backed SQLite storage layer from the original branch (the
`db`, `internal/sqlbase`, and per-subsystem `*_wasm.go` stubs), then did two
things differently from #431.

First, there's no inline OOR actor. #431 hand-rolled an `inlineOORActorRef` to
run the legacy global OOR actor synchronously in the browser, but `main` deleted
that actor entirely in favor of the durable registry + per-session actors. It
turns out those new durable actors compile and run under `js/wasm` as-is on top
of the OPFS SQLite mailbox, so the whole inline-actor concept just goes away.

Second, the browser bridge in `cmd/walletdk-wasm` is a thin `syscall/js` adapter
over the `sdk/walletdk/mobile` JSON facade. Every verb takes a JS request object
and resolves a `Promise` with the decoded JSON response. It's ~280 lines instead
of the old ~700, and nothing in it reaches into `walletdk.Client` directly.

On the build side, we drop the `!js` walls that fenced the embedded daemon
(`embedded.go`, `errmap.go`, the `swapruntime` / `walletdkrpc` tag pairs) off
from `js/wasm`, disable the public HTTP gateway on the embedded path (a browser
has no TCP listener to bind), and split `sdk/ark` into native + wasm halves.
`make wasm-wallet` then builds the stripped blob and drops it next to
`wasm_exec.js` and the go-wasmsqlite OPFS runtime assets. None of the
app-specific bits from #431 came along: no React demo, no Playwright harness, no
Pages workflow.

See each commit message for the incremental detail.

## Validation

```
make wasm-wallet
make build
env GOOS=js GOARCH=wasm go test -c ./db
env GOOS=js GOARCH=wasm go test -c ./sdk/swaps
env GOOS=js GOARCH=wasm go test -c -tags swapruntime ./sdk/walletdk
make unit pkg=./db
make unit pkg=./sdk/swaps
make unit pkg=./sdk/walletdk
make lint-changed-local
```

cc @jamaljsr


Roasbeef:
@claude review this

Roasbeef:
## Verified end-to-end in a real browser

The build passing didn't prove the OPFS/worker runtime path actually works, so I drove the `make wasm-wallet` blob in a cross-origin-isolated Chromium with a throwaway harness (a vanilla `index.html` running the Go wasm on the main thread, plus a COOP/COEP static server). No app code, just the blob + the go-wasmsqlite library assets the make target emits.

| Step | Result |
|------|--------|
| `crossOriginIsolated` | `true` |
| wasm decompress + boot (`walletdk-ready`) | ✅ the 132MB blob loads and runs |
| OPFS SQLite open | ✅ `vfsType: opfs, persistent: true` |
| OPFS SQLite write → read roundtrip | ✅ `hello-opfs` |
| OPFS persistence across 2 page reloads | ✅ row count = 2 (durable) |
| `isRunning` (pre-start) | ✅ `false` |
| **embedded daemon `start`** | ✅ **boots in-browser** (opens its DBs over OPFS, runs migrations, brings up the OOR durable actors + swap runtime) |
| `getInfo` | ✅ real state: `0.0.2-alpha / regtest / lwwallet` |
| `createWallet` | reaches password validation → seed gen → wallet DB open (the `internal/sqlbase` fork) → fails only at `get tip height: http://127.0.0.1:1/...: fetch() failed` |
| `balance` | `wallet is not ready` (create didn't finish) |

The two non-green steps aren't binding bugs: the harness points the light wallet at a dead Esplora (`127.0.0.1:1`). `createWallet` gets all the way through the OPFS-backed wallet DB to the chain-sync tip poller before the fetch fails, which exercises the `internal/sqlbase` + `walletdb_wasm` + localStorage-seed path end to end (and proves wasm `net/http` `fetch()` works). A fully-green create would need a reachable Esplora, which COEP blocks for external hosts, so it's out of scope for a smoke test.

So the whole thing genuinely works: the blob boots, OPFS SQLite persists across reloads, the full embedded daemon comes up in one browser VM, and the bridge round-trips real RPCs.

One fix from this: `make wasm-wallet` failed on re-run because it `cp`'d read-only module-cache assets over existing read-only files. Folded a clean-first + `chmod -R u+w` into the target.

Roasbeef:
@claude review this

Roasbeef:
Addressed the last open item: `internal/sqlbase/readWriteBucket.Get` now guards `value == nil` before the length check, so a `Get` on a key that names a sub-bucket (NULL value column) returns nil per the walletdb contract instead of panicking on the nil pointer. Consistent with the `value == nil` guards already used elsewhere in the file.

Validated with the `GOOS=js` build of `internal/sqlbase` + `cmd/walletdk-wasm`, the `db` wasm test-compile, native build + `db` tests, and a full browser e2e re-run (daemon still boots over OPFS, wallet DB opens through sqlbase, no regression). Folded into the `db:` commit.

PR #807, opened 2026-06-26
oor: add OOR_REJECT_USER_BALANCE typed reject code + classifier

Roasbeef:
In this PR, we add a dedicated typed reject code for the operator's per-user
balance cap on the OOR path, plus the client-side classifier that routes it.

The arkd server rejects an OOR submit that would push a recipient mailbox
over its `MaxUserBalance`, but until now it borrowed `OOR_REJECT_OUTPUT_POLICY`
on the wire to avoid a proto change. That left a custodial sender (e.g.
swapdk) unable to tell a balance rejection from a per-VTXO policy rejection
without string-matching the reason, and the two want opposite recovery.

A per-VTXO policy rejection is permanent for a given output shape, so the
caller must restructure the outputs. A balance rejection is transient: it
clears once the recipient spends or refreshes its balance down, so a
custodian should retain the value and retry later rather than restructuring.

We add `OOR_REJECT_USER_BALANCE` to the `OORRejectCode` enum and teach
`ClassifySubmitError` to map it onto a new typed `ErrUserBalanceExceeded`,
documented as transient, so a wallet routes recovery on the code alone. A
classifier test pins the mapping and asserts the balance and output-policy
errors don't alias each other.

The arkd server side that emits this code (and the end-to-end itest) lives in
the darepo PR that bumps this submodule.

### DIFF
--- a/oor/errors.go
+++ b/oor/errors.go
@@ -64,6 +64,35 @@ func (e *ErrOutputPolicyViolation) Is(target error) bool {
 	return ok
 }
 
+// ErrUserBalanceExceeded is the client-facing typed error returned when
+// the operator rejects an OOR submit because a recipient mailbox's
+// aggregate VTXO balance would exceed the operator's MaxUserBalance cap.
+//
+// Unlike ErrOutputPolicyViolation, this is NOT terminal for the same
+// output shape: the rejection clears once the recipient spends or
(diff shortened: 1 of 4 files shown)


Roasbeef:
@claude review this

PR #815, opened 2026-06-29
rpc: secure daemon RPC auth

sputn1ck:
## Summary

- add a shared `rpcauth` package backed by `lndclient.MacaroonService` and lnd macaroons
- default `darepod` local gRPC/REST gateway paths to TLS plus macaroon auth
- forward gateway macaroons and pass operator macaroons on outbound gRPC/REST clients
- expose TLS cert, macaroon path, and insecure-mode connection settings through `darepocli` and `walletdk`

## Commit Structure

1. `rpcauth: add TLS and macaroon helpers`
2. `darepod: secure local RPC by default`
3. `sdk: pass daemon RPC auth credentials`

## Validation

- `make fmt-changed`
- `make lint-changed-local`
- `go test ./rpcauth ./darepod ./cmd/darepocli/darepoclicommands ./rpc/restclient ./sdk/walletdk ./sdk/ark`
- `make unit pkg=darepod timeout=5m`


sputn1ck:
@claude review

sputn1ck:
@claude rereview please

sputn1ck:
@claude do another review pass, focussing on the new sqlite usage

rpcauth/auth.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	if err := os.MkdirAll(filepath.Dir(dbPath), 0o700); err != nil {
+		return nil, fmt.Errorf("create macaroon db dir: %w", err)
+	}
+	if err := os.MkdirAll(filepath.Dir(macaroonPath), 0o700); err != nil {
+		return nil, fmt.Errorf("create macaroon dir: %w", err)
+	}
+
+	rootKeyStore, db, err := lndclient.NewBoltMacaroonStore(
```
Roasbeef: Why use bolt here instead of `sqlc`? One less db that way. 

IIRC there also should be a interface in lnd we can just implement.

darepod/rpc_auth.go
```diff
@@ -0,0 +1,191 @@
...
+
+const darepodMacaroonEntity = "darepod"
+
+var (
+	darepodReadOp = bakery.Op{
+		Entity: darepodMacaroonEntity,
+		Action: "read",
+	}
+	darepodWriteOp = bakery.Op{
+		Entity: darepodMacaroonEntity,
```
Roasbeef: I think we can make this more granular, we have quite a few logical "entities" here.

Roasbeef: Done in 536a1f87. Split the single `darepod` entity into ten logical entities (`info`, `vtxo`, `address`, `onchain`, `oor`, `round`, `swap`, `recovery`, `fees`, `activity`), modeled on lnd's per-entity map — each method now names the specific entity + action it needs, grouped by domain. Also baked a `readonly.macaroon` alongside the admin one (read op per entity; authorizes every query, no mutating call).

Issue #776, opened 2026-06-23
[EPIC] Canonical append-only activity log: stable ids, pending→settled reconciliation, resumable SubscribeWallet


darioAnongba:
### Thesis

SwapDK's low-level primitives are durable and resumable — the double-entry `ledger`, and the swapd↔client mailbox (monotonic `event_seq`, dedup, ack cursors, one immutable row per event). The user-facing "activity" feed is not: it is **recomputed on every read**, merging 4–6 live sources with no canonical record, no pending→settled reconciliation, and a **lossy, non-resumable** `SubscribeWallet`. This epic promotes the mailbox's append-only durability discipline up into the wallet activity surface.

### The structural gap

- Derive-on-read merge: `swapwallet/history.go` `listActivity` joins `ListSwaps` + `ListTransactions` + `GetBalance` + `ListVTXOs` + `runtime.pendingSnapshot` per call, then sorts by `updated_at` and offset-paginates.
- Lossy subscribe: `swapwallet/runtime.go` `emit` drops updates to slow consumers by design (non-blocking `select`/`default`).
- Non-resumable subscribe: `swapwallet/service.go` `SubscribeWallet` has no cursor; `include_existing` re-snapshots on each connect.
- `swapwallet/doc.go` "V1 LIMITATIONS" records the consequence: EXIT/DEPOSIT pending and confirmed rows do not share an id, and "the right home for that link is the daemon-side persistence … not a process-local map".

Three consequences follow: (1) no stable cross-restart id per operation; (2) no pending→settled reconciliation, so rows stick at PENDING; (3) a lossy, non-resumable subscription that silently drops updates, possibly terminal ones.

### Scope

This epic covers the **activity-identity and reconciliation** surface: stable cross-restart ids, pending→settled reconciliation, a stable pagination cursor, and a gap-tolerant resumable subscribe — all read from one canonical store.

Balance accounting is **out of scope** and tracked in #816. The "balance reads zero mid-operation" bugs (#648, #760, #765) are daemon-side missing-balance-bucket gaps in `GetBalance`, on a path that never touches `listActivity`; a canonical activity log does not fix them.

### Sub-work / child issues

- [x] **C1 — canonical store with stable ids + append-only event log** → #774 (subsumes #610). Two tables: a current-state `activity_entries` projection for `List`, and an append-only `activity_events` log with a monotonic `event_seq` for `Subscribe`. Design: #798.
[...]

Issue #774, opened 2026-06-23
swapwallet: canonical append-only activity record with stable cross-restart ids


darioAnongba:
> Child of the event-log tracking issue #776.

### Problem

`WalletEntry` is rebuilt from scratch on every `List` by merging 4–6 live sources, with no canonical persisted record.
A consequence: EXIT/DEPOSIT pending and confirmed rows get **different ids** and can't be correlated across restart,
and an on-chain leave entry's id is the *consumed VTXO outpoint* — which is destroyed seconds later when the round
seals, so the handle disappears and can't represent a multi-input sweep.

### Evidence (verified against HEAD `53c81d63`)

- Derive-on-read merge: `swapwallet/history.go:30-38` (the consumer-side merger), `listActivity` (`:101-214`) joins
  `ListSwaps` + `ListTransactions` + `GetBalance` + `ListVTXOs` + `runtime.pendingSnapshot` on every call.
- Documented v1 gap: `swapwallet/doc.go:37-58` — only SEND-invoice / RECV (Lightning `payment_hash`) carry a stable id;
  "EXIT and DEPOSIT operations DO surface a row at submit time and a separate confirmed ledger row later, but the two
  rows do NOT share an id in v1 because there is no daemon-side notification hook." It explicitly defers a "v2
  canonical-id projection … when the daemon exposes the missing hooks."
- The exact symptom is filed as **#610** (open): on-chain-send entry id = consumed VTXO outpoint, vanishes after the
  round seals, can't represent multi-input sweeps. This issue **subsumes #610**.

### Proposed work

- Persist a sequence-numbered `activity_entries` table keyed by a **stable cross-restart canonical id** per entry,
  written as state advances. `List`/`Subscribe` read from it instead of re-joining live sources; derivation becomes a
  one-time projection, not a per-read merge.
- Requires the daemon-side notification hooks `doc.go` calls out (exit outpoints → sweep txid; deposit boarding address
  → boarding txid).

### Acceptance criteria

- Every entry kind (incl. EXIT/DEPOSIT and multi-input on-chain sends) has one stable id from initiation to
  confirmation, surviving restart.
- `List`/`Subscribe` read from the canonical store; no per-read multi-source merge.
- Regression coverage with golden activity fixtures.

### Links

Subsumes **#610**. Underpins the activity-pagination (`has_more`/cursor) work and the resumable-subscribe child.
Prior point-fixes that papered over id collisions on-read: #566, #612 (both closed).

[...]

Issue #610, opened 2026-05-28
wallet: 'send --onchain' activity entry id is the consumed VTXO outpoint; vanishes after the round seals and can't represent multi-input sweeps


jtobin:
## Environment

- `darepo-client`: `cd035b84` (also reproducible on current main `32b33fb5`)
- Network: signet
- Wallet backend: `lnd`
- Ark server: `arkd-signet.testnet.lightningcluster.com:443`

## Summary

`darepocli send --onchain` returns an activity entry whose `id` is the **outpoint of the VTXO being consumed** by the leave round. That outpoint is destroyed within seconds when the round seals. The id also can't represent the case of `--sweep-all`, where multiple VTXOs are consumed but the entry has a single id slot. Callers trying to look up the entry by id after the round seals get NotFound.

## Expected Behavior

Use a stable identifier for the activity entry: either the **round commitment txid** (allocated and known once the leave round produces a commitment tx; observable on-chain forever) or a **synthetic session id** allocated at request time (analogous to how OOR sessions and swap sessions are identified). The id must persist past the moment the underlying VTXO ceases to exist, and must be able to represent N-input leaves cleanly.

## Actual Behavior

Sending 50,000 sat on-chain from signet node a:

```sh
darepocli send tb1psayylun9kuxnh4pfwrlkenn036smu75napvkz2usxdw46cp7f4xsh0h4w4 \
   --onchain --amt 50000
```

Response (excerpt):

```jsonc
{
  "entry": {
    "id": "4b1c4b32a51e37173cd2046caa559592f304c6caa355c57add75d37c69d24d7f:1",
    "kind": "ENTRY_KIND_EXIT",
    "status": "ENTRY_STATUS_PENDING",
    ...
  },
  "actual_amount_sat": "964745"
}
```

The id is `<vtxo_outpoint>` of the input the wallet picked. v1 has whole-VTXO semantics (#15 / #16 in the source repo's filings family), so the entire 964,745-sat VTXO was consumed. After the round commitment tx confirmed:

- The outpoint `4b1c4b32...:1` ceased to exist.
- `darepocli activity inspect 4b1c4b32a51e37173cd2046caa559592f304c6caa355c57add75d37c69d24d7f:1` returns **NotFound**.
- The only durable identifier for the leave is the **round's commitment txid** (`84320a3068a88fa608b0a0393ead201bc7b928e0a96145fc3576e50486674255` in this run), which is what eventually landed on chain.

[...]

PR #817, opened 2026-06-29
db+swapwallet: canonical activity log store + dual-write projector (#774)

darioAnongba:
Foundation for #774 (C1), the first child of the event-log epic #776. Design: #798.

**Additive, no read-behavior change.** This lands the persisted canonical activity store and dual-writes into it, while `List`/`SubscribeWallet` keep deriving from the legacy merge in `history.go`. The read-path cutover, the proto cursor/`has_more` fields, and the daemon-side stable-id hooks for EXIT/DEPOSIT (the #610 fix) are deferred to follow-up PRs.

### What's here

- **Schema (migration 000022).** Two tables modelled on the design doc:
  - `activity_entries` — the current-state projection, one row per operation keyed by a stable `canonical_id`, read by `List`.
  - `activity_events` — the append-only transition log with a monotonic `event_seq`, the basis for a resumable `SubscribeWallet`.
  - `activity_kinds` / `activity_statuses` enum tables seed the proto enum integers, and `kind`/`status` are foreign keys to them — so a row can only hold a defined wire `EntryKind`/`EntryStatus` and the projection needs no internal-to-wire mapping. Money columns are `BIGINT` (a bare `INTEGER` is 32-bit on the Postgres backend); empty BLOB handles are stored as NULL.
- **Store (`db/activity_store.go`).** `ActivityPersistenceStore.ProjectEntry` upserts the current-state row and appends the transition event in one transaction (FK-safe order). Keyset `ListEntries` (immutable `created_at` cursor) and `PullEvents` (`event_seq > cursor`) read helpers back the eventual cutover.
- **Projector (`swapwallet/projector.go`).** A `Runtime`-owned projector maps each emitted `WalletEntry` to the store and writes it project-then-emit at the three existing emit sites (swap monitor, cooperative-leave submit, deadline overlay), plus a one-time idempotent startup backfill from the existing collectors. Projection is best-effort: a store error is logged and never blocks or fails the emit, and a nil store (tests, no-DB builds) disables it.
- **Wiring.** The daemon builds the store next to the VTXO store and injects it into the walletdkrpc subserver in-process via a programmatic `SwapWalletConfig` field.

### Notes

- The Codex review's two points on #798 are handled by construction here: `kind`/`status` mirror the wire enums via the FK (no `ONCHAIN_SEND`/`CONFIRMED` divergence) — an on-chain send is an `EXIT`-kind row, as today.
- Until the daemon stable-id hooks land, a cooperative-leave EXIT is still keyed on the consumed outpoint and a DEPOSIT on `txid:vout`, exactly as the legacy merge behaves; the dual-write test asserts the store mirrors the merge for the stable-id rows.

### Tests

- `db/activity_store_test.go`: atomic upsert+append, in-place update preserving `created_at` and prior handles, enum FK rejection, NULL-blob handling, keyset paging, event-pull cursor.
- `swapwallet/projector_test.go`: `WalletEntry` → projection mapping with a lossless `entry_json` round-trip, project-then-emit ordering, nil-store and store-error safety.
- `swapwallet/activity_dualwrite_test.go`: backfill mirrors the legacy merge for stable-id rows and is idempotent.

`make sqlc`, `make build`, `make lint-changed-local`, and the `db` + `swapwallet` unit suites pass.


darepod/config.go
```diff
@@ -250,6 +251,15 @@ type Config struct {
 	// walletdkrpc and swapruntime tags.
 	SwapWallet *SwapWalletConfig `mapstructure:"swapwallet"`
 
+	// ActivityStore is the canonical activity-log projector (issue #774)
```
darioAnongba: do not reference issues in the code or comments like this

darepod/config.go
```diff
@@ -250,6 +251,15 @@ type Config struct {
...
+	// the walletdkrpc subserver writes to as wallet state advances. It is
+	// injected programmatically by the server, never deserialized, and it
+	// is a top-level field (not under SwapWallet) because the subserver is
+	// registered by build tag regardless of whether the operator supplied
+	// a [swapwallet] config section. A nil value disables projection, so
+	// builds and tests without a database keep working.
```
darioAnongba:
```suggestion
	// a [swapwallet] config section. A nil value disables projection.
```

darepod/server.go
```diff
@@ -1186,6 +1187,15 @@ func (s *Server) run(ctx context.Context, shutdownFn func()) error {
 	)
 	s.vtxoStore = dbStore.NewVTXOStore(s.clk)
 
+	// Build the canonical activity-log store (issue #774) and hand it to
```
darioAnongba: same remark. This comment is not needed right? too verbose

db/sqlc/migrations/000023_canonical_activity_log.up.sql
```diff
@@ -0,0 +1,117 @@
+-- The canonical activity log is the persisted source of truth for the wallet
+-- activity feed, replacing the derive-on-read merge in swapwallet/history.go.
```
darioAnongba:
```suggestion
-- activity feed.
```

db/sqlc/migrations/000023_canonical_activity_log.up.sql
```diff
@@ -0,0 +1,117 @@
...
+-- It is two tables: activity_entries holds the current state of each operation
+-- (one row, updated in place, keyed by a stable canonical id) and backs List;
+-- activity_events is an append-only log of lifecycle transitions (one immutable
+-- row per transition, ordered by a monotonic event_seq) and backs a resumable
+-- SubscribeWallet. See docs/canonical_activity_log_design.md (issue #774).
```
darioAnongba:
```suggestion
-- SubscribeWallet. See docs/canonical_activity_log_design.md.
```

PR #819, opened 2026-06-29
vtxo+oor: multi-parent lineage gate + OOR-received registration (C7)

ellemouton:
## C7 — OOR-received VTXO multi-parent lineage governance

Part of the reorg-safety epic (lightninglabs/darepo#454). Stacked on C6 (#818).
Unlocks acceptance **F4/F5** (OOR-received VTXO reorg → limbo → usable / ancestor
conflict → unavailable).

### C7a — gate the full multi-parent lineage (`vtxo`)
The C5 admission gate read only a candidate's *direct* commitment txid, leaving
cross-commitment OOR VTXOs (which descend from more than one batch) ungoverned
by their ancestor batches. `lineageCommitmentTxids` now collects the direct
commitment txid plus every distinct `Descriptor.Ancestry[].CommitmentTxID`
(deduped, zero-skipped), and `LineageBlocked` takes the worst state across all
of them — so a multi-input OOR VTXO is excluded from selection if ANY
contributing batch is reorged out or conflict-invalidated. Completes the
multi-parent follow-up C5 left open.

### C7b — register received lineage at OOR receive (`oor`)
The receiver did not participate in the ancestor rounds, so its manager has no
record of those batches and the gate would stay permissive. At OOR
materialization the session actor now registers every commitment batch in the
received VTXOs' lineage with the `BatchCanonicalityManager` (best-effort,
post-commit, gated behind an optional `SessionActorConfig.BatchCanonicality`
ref — `None` = dormant). The batch-output pkScript comes from each ancestry
fragment's tree root (`BatchOutput`), so confirmation/reorg detection works on
script-filtering light-client receivers (Esplora/Neutrino).

`ConsumedInputs` are intentionally empty (the receiver doesn't hold the ancestor
commitment txs' inputs at this seam): a reorg-out of an ancestor batch is still
detected via its confirmation watch and marks the VTXO limbo (F4); per-input
double-spend watches and per-fragment CSV delta are documented follow-ups.

### Activation (darepod) — deferred
Like C5/C6, the gate ships dormant. Constructing/starting the canonicality
manager and threading the store/ref to the vtxo + round + oor actors is the
shared integration step (ties #795 + #796 + C6/C7/C8) and needs the daemon
harness to validate.

### Tests
Unit tests: `lineageCommitmentTxids` (dedup/zero-skip/direct-only) and a
multi-parent selection case; OOR lineage registration (distinct/shared
ancestors with dependents, dormant no-op, incomplete-fragment skip). F4/F5 land
[...]

PR #820, opened 2026-06-29
unroll: gate admission on source-lineage canonicality (C8)

ellemouton:
## C8 — gate unroll admission on source-lineage canonicality

Part of the reorg-safety epic (lightninglabs/darepo#454). Stacked on C7 (#819).

### What this does
An unroll broadcasts the VTXO's exit tree, which spends a commitment batch
output. If a batch in the VTXO's source lineage is **permanently invalidated**
(a consumed input was double-spent past finality), the exit tree can never
confirm. `EnsureUnroll` now consults the target VTXO's full source lineage (the
direct commitment txid plus every ancestor commitment txid) before spawning a
fresh child and refuses with `ErrSourceLineageUnavailable` when the lineage is
`Invalidated`.

### Why it blocks only the terminal verdict
It deliberately blocks **only** `Invalidated`, not the transient `LimboReorg` /
`LimboConflict` states:
- A reorged-out batch (no input conflict) is expected to re-confirm on its own;
  a not-yet-final conflict may still resolve in the batch's favor.
- The critical-expiry safety net reaches this gate via a **fire-and-forget
  Tell**, so a refusal can't be observed or retried — blocking a transient,
  self-healing condition could permanently drop a needed exit.
- An already-admitted unroll tolerates a transiently-absent parent by
  reconciling its own anchors (#410), so a fresh safety exit is admitted for
  the same condition rather than refused.

For `Invalidated` there is genuinely nothing to retry (the funds are gone via
the finalized conflict), so refusing — and a dropped fire-and-forget refusal —
is safe.

The gate is also **fail-permissive**: a descriptor-load or canonicality-lookup
error logs and admits rather than blocking an exit. Gated behind an optional
`RegistryConfig.BatchCanonicality` store (nil = dormant, matching C5–C7); only
fresh admissions are gated.

### Follow-ups (when darepod wires the gate — deferred, gate dormant today)
- `recoverOrphanedUnrollJobs` treats `EnsureUnroll` errors as fatal at startup;
  it should classify `ErrSourceLineageUnavailable` as non-fatal/deferred.
- `unrollLineageCommitmentTxids` duplicates the vtxo selection-gate helper;
  consider exporting one shared helper.

### Tests
Unit tests cover blocked-on-invalidated-ancestor, permitted transient-reorg and
canonical cases, unregistered + load-failure (permissive) cases, the dormant
no-op, and the `errors.Is`-matchable wrapped sentinel.

### DIFF
--- a/chainsource/finality.go
+++ b/chainsource/finality.go
@@ -21,16 +21,6 @@ var finalityBlockSubscriptionBackoffs = []time.Duration{
 	2 * time.Second,
 }
 
-// finalityBlockSubscriptionAttemptTimeout bounds each individual
-// RegisterBlocks attempt. Without it a single hung RegisterBlocks call
-// (e.g. a wedged lndclient gRPC stream) would block the conf/spend
-// monitoring goroutine indefinitely — stalling Confirmed/Reorged/Done
-// delivery on that watch — since the retry schedule only bounds the gaps
-// between attempts, not the attempts themselves. 10s mirrors the per-call
-// registration timeout used in conf_actor.go's handleRegisterConf so the
-// whole file behaves consistently under a slow backend.
-const finalityBlockSubscriptionAttemptTimeout = 10 * time.Second
-
 // registerBlocksForFinality registers a block-epoch subscription used
 // to synthesize a Done signal at FinalityDepth past an observed
 // confirmation or spend. The call is retried with a short bounded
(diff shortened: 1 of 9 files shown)


ellemouton:
Superseded by #896 as part of condensing the reorg-safety client stack (epic lightninglabs/darepo#454) from 12 PRs into 3. The commits are carried over unchanged; see #896. Branch retained as a backup.

PR #821, opened 2026-06-29
credit: Run the operation FSM on protofsm, fold redeem into receive

Roasbeef:
In this PR, we move the credit per-operation state machine off the hand-rolled state-string `switch` it shipped with in #772 and onto `baselib/protofsm`, the same engine `round` uses, so the credit subsystem is built like the rest of the codebase. This is the follow-up flagged at the end of #772: the FSM rolled its own driver instead of using protofsm and the outbox system.

The durable per-operation actor stays exactly where it was. It's still the execution substrate that owns the lease-fenced `Stage`/`Commit` writes and the exactly-once mailbox; we only swap out the FSM core it drives. Each protofsm state's `ProcessEvent` runs its idempotent effect, records the result on the operation row, and returns the next state plus a set of outbox directives. The durable turn then executes those directives through its `Exec` handle: a `stageRecord` becomes an `ax.Stage` checkpoint, and the per-turn `ax.Commit` folds the final ack. So the FSM dictates *what* to persist and signal, while the actor keeps owning *how*. That's the bit worth calling out, since it's how we get the protofsm structure without giving up the durable-actor guarantees: persist-before-effect falls out of flushing a `stageRecord` before the next state runs its effect.

This builds on #772 (now merged); see each commit message for the incremental detail.

## What moved

The hand-rolled `state.go` switch, the `runFSM`/`advance` loop, and the per-step methods in `op_steps.go` are gone. In their place we have the usual protofsm shape, mirroring `round`: `states.go` for the sealed state types, `events.go` for the single drive event and the outbox directives (`stageRecord`, `parkOp`, `triggerRedeem`), `transitions.go` for the `ProcessEvent` logic, and `transition_table.go` for the `CreditTransitions` table that documents every transition and serves as a validation target (like `round`'s `BoardingClientTransitions`).

The states, effects, and crash-safety guarantees are unchanged; only the structure moved. The one new wrinkle is that the redeem reservation splits into `redeem_reserving` (allocate the destination, checkpoint it) and `redeem_submitting` (reserve against it), so the destination checkpoint lands strictly before the reservation effect. That's the same H-2 invariant from #772, now expressed as a state boundary instead of a mid-step `Stage`.

## Redeem folds into the receive FSM

The second half of this PR ties auto-redeem into the state machine, which kills the periodic background sweep. A settled receive is the only thing that grows the available balance, so a receive that clears the earmark-adjusted watermark emits a `triggerRedeem` that the actor signals to the registry once the turn commits. The registry applies the no-pending-pay/redeem interlock and admits the redeem. The receive FSM signals intent; the registry owns the decision.

The `autoRedeemer` shrinks from a ticker loop to a single boot-time reconcile, for the one case the receive trigger can't cover: a balance already sitting over the watermark at start, which no new receive will re-evaluate. The earmark interlock from #772 is unchanged and shared by both the receive trigger and the boot reconcile. One gap left for a later pass: a mid-session balance bump from a *released* pay reservation is only re-checked at the next receive or restart, not right away.

## Testing

`go test ./credit/...` passes (including `-race`), and the existing crash-recovery and idempotency tests carry over unchanged since the durable harness is the same. New tests cover the receive-driven redeem trigger (over and under the watermark, and disabled) and the watermark + earmark math. `make lint-changed-local` is clean, and the full build with `walletdkrpc swapruntime` is green.


Roasbeef:
@claude review this

Roasbeef:
Ran a multi-agent review of this branch (code-quality, offensive-security, differential/ToB, deep-function-analysis, spec-compliance, API-safety). Verdict was **approved with conditions** — no Critical findings, no client-reachable fund-loss, and the central migration risk came back clean: an in-flight `credit_operations` row written by the merged #772 code resumes correctly under the new FSM (all persisted state strings preserved byte-for-byte). Persist-before-effect, the post-commit `triggerRedeem` crash window, the earmark/watermark math (no uint64 underflow, strict-greater), and the fail-safe config defaults all verified.

Addressed the actionable findings in the latest commits:

- **H-1 — interlock test coverage.** The no-pending-pay/redeem interlock moved out of the deleted `redeemDecision` into `considerRedeem` with no test. Added `TestConsiderRedeemInterlock`: a pending pay or redeem blocks, a pending receive does not, a clear balance admits exactly one redeem.
- **M-1 — redeemed amount.** `awaitingOORState` completed on any live vTXO at the destination without checking its amount, recording the requested rather than the materialized value. Now it reconciles the landed amount against the reservation, records what actually landed (so the wallet entry is accurate), warns on divergence, and guards against a spurious zero-value match. It never terminal-fails after the vTXO lands — the value is already home.
- **M-3 — dead config.** Removed the now-unread auto-redeem sweep `Interval` knob (credit config, daemon config + `autoredeeminterval` flag, sample conf) and refreshed the stale "sweep" comments.
- **M-5 — dropped redeem signal.** The post-commit Tell and the registry admit are best-effort; raised their skip paths from debug to warn and captured the previously-discarded `admit` result.
- **Docs.** Rewrote `credit_durable_actor_design.md` around the shipped protofsm + redeem-on-receive design (folding the old follow-up section in, demoting the version comparison to a decision record), and corrected the transition-table comment that claimed a validating test.

Deliberately deferred: the redeem-vs-credit-send TOCTOU (the documented earmark gap) is the separately-tracked single-authority follow-up, not a regression here; the transition-table validation test (M-2) is left for that pass.

Roasbeef:
Folded the review feedback (this comment thread + the agent review) into the two base commits and force-pushed; the branch is now 2 clean commits again.

Addressed:

- **Corrupt persisted state wedge (item 1 / Codex P2).** Fixed. `decodeCreditState` now returns `(state, ok)`; on an unrecognized `state` string `dispatch` drives the row through `applyState(failedState)` + `LastError` so it terminal-commits and is reaped, instead of treating the in-memory `failedState` as already committed and restoring the row on every boot. Covered by `TestDispatchFailsCorruptState`.
- **Ignored `admit` result (item 3 / Gemini).** Already addressed in this push: `considerRedeem` unpacks the `fn.Result` and warns on a failed admission.
- **Dropped trigger while a pay is in flight (item 2).** Made non-silent: the interlock-deferral path now logs (with the blocking op id), and the code comments the gap and its causal closure. Kept as a tracked follow-up (the single-authority change that routes credit-backed send reservations through the registry), as the body notes.
- **Stale `available` nit.** Added a comment on why reserving against a slightly stale snapshot is safe (the server revalidates the reservation, so an unsupported amount fails cleanly rather than over-materializing).

The transition-table validation test (table-can't-drift) is the one item I left for the single-authority follow-up rather than this PR.

Separately, rewrote the design doc to be self-contained: it now describes the system as it is, with no PR numbers, version history, or alternatives-considered record.

Roasbeef:
@claude review this

PR #832, opened 2026-06-30
txconfirm: CPFP fee-bump a funded-anchor parent + BumpNow override

Roasbeef:
In this PR, we teach the `txconfirm` broadcaster to fee-bump a *funded*-anchor
parent: a v2 transaction that pays its own miner fee and confirms standalone,
with a non-zero P2A anchor reserved as a CPFP handle. Until now txconfirm only
knew the zero-value ephemeral anchor on a v3/TRUC parent, which it must always
ride out as a package because that parent pays zero fee. The server side
(darepo) wants to attach a funded anchor to every round/commitment transaction
so an operator can unstick a round whose original fee rate the mempool has
since outpaced, without forcing the commitment tx itself to be a 0-fee TRUC
package.

## arktx predicates

We add two predicates alongside the existing zero-value `IsAnchorOutput`.
`IsP2AAnchorScript` matches the P2A anchor by script regardless of value, and
`IsFundedAnchorOutput` is the exact complement of `IsAnchorOutput` over P2A
outputs (non-zero value). A funded anchor is one whose value sits above the P2A
dust threshold (240 sats per BIP-433), so its parent pays its own fee rather
than relying on a CPFP descendant to fund the package.

## Funded-anchor CPFP path

`Submit` now branches on the anchor form. A funded parent broadcasts directly
on its first pass: no CPFP child, no fee-input reservation, since we may never
need to bump. It only builds a v2 CPFP child when a fee bump is actually
requested (the interval-paced bump or a forced one). The TRUC version gate
stays in force for the ephemeral path but no longer rejects a v2 funded parent.

The fee math subtracts the parent's own fee, threaded through as `ParentFee`,
so the combined parent+child fee lands on the target rate instead of
overshooting it by the parent's fee. `ParentFee` defaults to zero, which is
exactly right for the zero-fee ephemeral case and a safe overpay when a caller
doesn't supply it.

## BumpNow override

We add a `BumpNowReq` message so an operator can force an immediate CPFP at a
chosen fee rate (clamped to the configured max) rather than waiting for the
interval-paced bump. `broadcastTrackedTx` now carries the one-shot target rate
and the fee-bump intent and returns the `BroadcastResult` so the bump handler
can surface the child txid. The existing fraud-responder path (zero-value
ephemeral, v3/TRUC) is unchanged: all of its tests stay green.

changes.

### DIFF
--- a/lib/tx/arktx/canonical.go
+++ b/lib/tx/arktx/canonical.go
@@ -30,6 +30,36 @@ func IsAnchorOutput(out *wire.TxOut) bool {
 	return bytes.Equal(out.PkScript, arkscript.AnchorPkScript)
 }
 
+// IsP2AAnchorScript returns true if pkScript is the keyless P2A anchor
+// script, regardless of the spending output's value. Unlike IsAnchorOutput,
+// this does not require the zero-value ephemeral-dust form: it also matches a
+// "funded" anchor whose value has been lifted above the P2A dust threshold
+// (see arkscript.WithAnchorValue) so that its parent transaction can pay a
+// non-zero miner fee and confirm on its own. Use this when locating a CPFP
+// anchor to spend on a parent that is already independently valid, where the
+// anchor is a spare fee-bump handle rather than the parent's only fee source.
+func IsP2AAnchorScript(pkScript []byte) bool {
+	return bytes.Equal(pkScript, arkscript.AnchorPkScript)
+}
+
+// IsFundedAnchorOutput returns true if out is a P2A anchor output carrying a
+// non-zero value (the "funded" anchor form). A funded anchor lets its parent
+// pay its own fee and confirm standalone, with the anchor reserved as an
+// optional CPFP handle for fee-bumping only when the parent is stuck. This is
+// the complement of IsAnchorOutput, which matches only the zero-value
+// ephemeral form used by TRUC parents that pay zero fee themselves.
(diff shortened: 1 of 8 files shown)


Roasbeef:
@claude review this

Roasbeef:
All three findings addressed in 7eded93c (the first two overlapped with our own review pass; the anchor-burn bug was a new catch):

- **Anchor value burned as fee**: confirmed real. `BuildCPFPChild` now credits `anchorOutput.Value` into the change, so the child's effective fee equals exactly the `childFee` asked for (the misnamed `totalFee` param is renamed `childFee` while at it). `TestBuildCPFPChildCreditsAnchorValue` asserts `inputs - outputs == childFee` for a funded anchor, guarding the fix.
- **Over-restrictive fee-input selection**: selection now targets `childFee + DustLimit - anchorValue` (floored at 1 sat) via `childFeeInputTarget`, computed before selection; the precise-vsize reselect uses the same child-based target. The ephemeral path reduces to the old `totalFee + DustLimit` arithmetic since `parentFee` and `anchorValue` are both zero there.
- **Forced bump reporting success on the direct fallback**: the funded path already refuses the fallback outright (581aee93); the ephemeral path now reports `Bumped: false` with a reason whenever the submission produced no CPFP child, so a fallback rebroadcast can no longer read as a successful bump.

Roasbeef:
@claude review this

PR #836, opened 2026-06-30
darepod: Restrict lwwallet CPFP fee inputs to default account

### LINKED ISSUE #831: lwwallet: CPFP child fails to finalize taproot fee input (PSBT is not finalizable)
## Summary

On the `lwwallet` (esplora) backend, CPFP fee-bumping is silently broken: the
CPFP child always fails to finalize with

```
finalize PSBT: btcwallet FinalizePsbt: error finalizing PSBT: PSBT is not finalizable
```

at `stage=sign_cpfp_child`. The broadcaster then logs `CPFP unavailable;
broadcasting parent directly` and falls back to broadcasting the bare,
zero-fee ephemeral-anchor parent, which won't relay on its own. So every
unroll on this backend loses its fee bump.

Surfaced in the field via darepo-client#828 (the boarding-sweep 404 in that
issue is a separate, infra-side bug; this is the second failure in the same
log):

```
2026-06-30 17:57:20.514 [INF] TXCF: trackedtx(9f6774df...) State transition from_state=New to_state=Broadcasting
2026-06-30 17:57:20.591 [WRN] TXCF: CPFP unavailable; broadcasting parent directly err="finalize PSBT: btcwallet FinalizePsbt: error finalizing PSBT: PSBT is not finalizable" stage=sign_cpfp_child label=unroll-node-9f6774df...
```

The wallet had `utxo_count=2` at the time, and the finalize failed in ~77ms,
so this isn't a "no funds" case: nothing was signed at all.

## Root cause: fee-input selection and finalization disagree on key scope

The CPFP child has two inputs: the parent's ephemeral P2A anchor (pre-finalized
empty witness) and one wallet-owned fee input. The fee input is selected and
finalized through two different paths that use different account/scope rules.

**Selection** — `lwUnrollWallet.ListUnspent` (`darepod/server.go`) first lists
the default account, and when that's empty falls back to listing across *all*
accounts so it can surface wallet-managed P2TR outputs:

```go
lnUtxos, err := w.Wallet.BtcWallet.ListUnspentWitness(minConfs, maxConfs, lnwallet.DefaultAccountName)
...
// btcwallet's default-account filter can miss wallet-managed P2TR
// outputs in lwwallet mode ... Fall back to the broader query so
// harness-funded fee inputs remain available for CPFP package relay.
if len(lnUtxos) == 0 {
    lnUtxos, err = w.Wallet.BtcWallet.ListUnspentWitness(minConfs, maxConfs, "")
}
```

**Finalization** — `lwUnrollWallet.FinalizePsbt` -> `Wallet.FinalizePsbtDirect`
(`lwwallet/wallet.go`) always finalizes against the default account:

```go
func (w *Wallet) FinalizePsbtDirect(packet *psbt.Packet) error {
    return w.BtcWallet.FinalizePsbt(packet, lnwallet.DefaultAccountName)
}
```

And in lnd's btcwallet, `DefaultAccountName` resolves to a *nil* key scope,
which covers only NP2WKH and P2WKH (`lnwallet/btcwallet/psbt.go`):

```go
// If the default/imported account name was specified, we'll provide a
// nil key scope to FundPsbt, allowing it to sign inputs from both key
// scopes (NP2WKH, P2WKH).
case lnwallet.DefaultAccountName:
    accountNum = defaultAccount
```

Taproot (BIP86 / P2TR) is not in that set. But lwwallet's own outputs are
taproot: `lwUnrollWallet.NewWalletPkScript` returns "a fresh wallet-managed
**taproot** pkScript suitable for change and sweep outputs", and
`signCPFPChild` even special-cases the fee input as taproot
(`cpfpFeeInputSighash` returns `SigHashDefault` for P2TR).

So selection hands back a P2TR fee input via the all-accounts fallback, then
finalization pins itself to the default account, can't derive the key for that
P2TR input, signs nothing, and `MaybeFinalizeAll` reports "PSBT is not
finalizable". The two sides are simply inconsistent about scope.

## Proposed fix

Make finalization scope-aware so it can sign the same P2TR fee inputs that
selection returns. A few options, roughly in order of preference:

1. Finalize against the fee input's actual key scope. `signCPFPChild` already
   knows the fee input's `PkScript`; thread the right account/scope (BIP86 for
   P2TR) into `FinalizePsbtDirect` instead of hardcoding `DefaultAccountName`.
2. Have `FinalizePsbtDirect` finalize across all wallet scopes (including
   taproot) so it matches the breadth of the `ListUnspent` fallback.
3. As a narrower stopgap, restrict CPFP fee selection to scopes the default
   finalize path can actually sign. This keeps the two sides consistent but
   gives up P2TR fee inputs, which on lwwallet is most/all of the wallet, so
   it's not really viable on its own.

(1) or (2) is the real fix.

## Test

Add a regression test for `signCPFPChild` with a **P2TR** fee input on the
lwwallet path, asserting the child finalizes and the witness is copied back.
The existing finalize tests don't exercise a taproot fee input, which is why
this slipped through.

## Scope / environment

- Backend: `lwwallet` (esplora). The LND backend path uses LND's own
  `WalletKit().FinalizePsbt` and is likely unaffected, but worth confirming.
- Impact: CPFP fee-bumping unavailable for unrolls on lwwallet; unrolls fall
  back to direct parent broadcast of a zero-fee ephemeral-anchor tx that does
  not relay standalone.

Roasbeef:
In this PR, we fix a silent CPFP failure on the `lwwallet` (esplora)
backend where every unroll quietly lost its fee bump. The CPFP child
would fail to finalize with `PSBT is not finalizable` at
`stage=sign_cpfp_child`, after which the broadcaster fell back to
broadcasting the bare, zero-fee ephemeral-anchor parent on its own. That
parent doesn't relay standalone, so the bump was effectively dropped on
the floor.

Fixes #831.

## Root cause

The interesting part here is that this is _not_ a finalization-scope
bug, which is where the issue initially pointed. The wallet's own
derived taproot outputs finalize fine through the default account:
btcwallet's nil-scope `FinalizePsbt` signs P2TR via `ComputeInputScript`
just like p2wkh, independent of the account name. I confirmed this by
porting btcwallet's own `TestFinalizePsbt` to a P2TR fee input against
our pinned version, and it signs cleanly.

The real problem is fee-input _selection_. `lwUnrollWallet.ListUnspent`
fell back to an all-accounts query (`""`) whenever the default account
came up empty:

```go
if len(lnUtxos) == 0 {
        lnUtxos, err = w.Wallet.BtcWallet.ListUnspentWitness(
                minConfs, maxConfs, "",
        )
        ...
}
```

That fallback surfaced imported, watch-only taproot outputs as CPFP fee
inputs. These are boarding outputs imported via `ImportTaprootScript`
(`TaprootFullKeyOnly`): they live in the _imported_ account, and carry
no BIP32 derivation and no spendable key. So btcwallet's `FinalizePsbt`
can't sign them in any scope, skips the input, and `MaybeFinalizeAll`
reports the PSBT as unfinalizable. This also explains the field
symptom, where `ListUnspent` on the default account came back empty
(`utxo_count=2` were both boarding outputs) but the all-accounts fallback
happily handed one back.

I reproduced the exact field log against the pinned btcwallet by
importing a `TaprootFullKeyOnly` output, funding it, and finalizing:

```
default-account utxos=0  all-account utxos=1
FinalizePsbt err = error finalizing PSBT: PSBT is not finalizable
```

## The fix

We drop the all-accounts fallback and let the default account stand on
its own. The default-account filter already excludes the unspendable
imported outputs while still surfacing every wallet-derived,
key-spendable fee input (lwwallet derives its change and sweep addresses
via `NewAddress(TaprootPubkey, DefaultAccountName)`, i.e. BIP86 account
0, which is exactly "default"). The sibling `btcwUnrollWallet.ListUnspent`
path already does precisely this, with a comment saying as much, so this
brings the two backends back in line and matches the function's own
doc comment.

## Test

The regression test in `darepod` builds a real `lwwallet` on a mock
Esplora pinned at the regtest genesis tip, funds an imported boarding
output as the wallet's _only_ UTXO (the field scenario, where the
default account is empty), and asserts fee selection never offers it. It
then funds a wallet-derived taproot output and asserts that one _is_
offered. The test only catches the bug when the imported output is the
sole UTXO: a co-present derived output keeps the default account
non-empty, which hides the fallback entirely. Verified it fails on the
old code and passes on the fix.

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -4883,24 +4883,20 @@ func (w *lwUnrollWallet) ListUnspent(ctx context.Context, minConfs,
 		return nil, fmt.Errorf("wait for wallet sync: %w", err)
 	}
 
+	// Restrict CPFP fee selection to the default account. Imported,
+	// watch-only taproot outputs (e.g. boarding outputs imported via
+	// ImportTaprootScript) live in the imported account: they carry no
+	// BIP32 derivation and no spendable key, so the FinalizePsbt path
+	// cannot sign them and finalization fails with "PSBT is not
+	// finalizable". The default-account filter already surfaces every
+	// wallet-derived (key-spendable) fee input, so a broader all-accounts
+	// query would only readmit the unspendable imported outputs we must
+	// avoid.
 	lnUtxos, err := w.Wallet.BtcWallet.ListUnspentWitness(
 		minConfs, maxConfs, lnwallet.DefaultAccountName,
 	)
 	if err != nil {
 		return nil, err
-	}
-
-	// btcwallet's default-account filter can miss wallet-managed P2TR
-	// outputs in lwwallet mode even though the same UTXOs are visible when
-	// enumerating across all accounts. Fall back to the broader query so
-	// harness-funded fee inputs remain available for CPFP package relay.
-	if len(lnUtxos) == 0 {
-		lnUtxos, err = w.Wallet.BtcWallet.ListUnspentWitness(
-			minConfs, maxConfs, "",
-		)
-		if err != nil {
-			return nil, err
-		}
 	}
 
 	result := make([]*wallet.Utxo, 0, len(lnUtxos))
--- a/darepod/unroll_fee_input_test.go
+++ b/darepod/unroll_fee_input_test.go
@@ -1,0 +1,235 @@
+package darepod
+
+import (
+	"bytes"
+	"encoding/json"
+	"math"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btclog/v2"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	btcwalletbase "github.com/btcsuite/btcwallet/wallet"
+	"github.com/btcsuite/btcwallet/walletdb"
+	"github.com/btcsuite/btcwallet/wtxmgr"
+	"github.com/lightninglabs/darepo-client/lwwallet"
+	"github.com/lightninglabs/darepo-client/wallet"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/lnwallet"
+	"github.com/stretchr/testify/require"
+)
+
+// TestLwUnrollWalletListUnspentExcludesImported is a regression test for
+// darepo-client#831. On the lwwallet (esplora) backend, CPFP fee selection
+// must only surface wallet-derived, key-spendable outputs. Imported,
+// watch-only taproot outputs (boarding outputs imported via
+// ImportTaprootScript) have no BIP32 derivation and no spendable key, so the
+// FinalizePsbt path cannot sign them and finalization fails with "PSBT is not
+// finalizable", forcing a zero-fee parent broadcast that does not relay.
+//
+// The default-account ListUnspentWitness filter already excludes these
+// imported outputs. The bug was an all-accounts fallback that re-admitted
+// them; this test pins the corrected behaviour: the wallet's own derived
+// output is offered as a fee input, while an imported boarding output is not.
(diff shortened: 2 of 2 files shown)


Roasbeef:
@claude review this

PR #846, opened 2026-07-01
darepocli: emit a compact JSON summary on successful send

Roasbeef:
In this PR, we clean up what `da send` prints on success. Right now a
completed send is a wall of text: first the full dispatch entry dumped as
JSON, then the entire markdown inspection trace (the activity block, the
swap block, the whole VTXO set, every ledger row, and the correlation
notes). The invoice string alone shows up four times across roughly ninety
lines, and the one thing you actually want back, the payment preimage, is
buried somewhere in the middle.

We collapse all of that down to a single compact JSON object. On a terminal
completion we print the distilled summary and stop:

```json
{
  "status": "COMPLETE",
  "kind": "SEND",
  "amount_sat": -1000,
  "fee_sat": 0,
  "settlement": "IN_ARK",
  "destination": "lntbs10u1p4ytrgqpp56wep0fmm4l8d9",
  "payment_hash": "d3b217a77bafced29fece043f417dafe91fd8c060f238e2681628e9f9033f14d",
  "preimage": "a6837e6d587a7a297fa7a35fbf22a67fdcbed60b9c1ddc12cac22723fc22c1c9",
  "vtxo_outpoint": "5d2f5b892ca87f7857e9bf80a9a013afe5bacb05cf51326670183f2c74f3ff65:0",
  "id": "d3b217a77bafced29fece043f417dafe91fd8c060f238e2681628e9f9033f14d"
}
```

That's the amount that left the wallet, the fee, the settlement rail, and,
for a Lightning send, the preimage that proves the invoice was paid. The
`id` is the handle you feed to `da activity inspect <id>` when you do want
the full ledger, VTXO, and swap breakdown, so nothing is lost, it just
moves off the happy path.

## What changed

The dispatch-time proto dump is dropped from the waiting path. The `phase:`
transitions already narrate progress on stderr, so echoing the raw entry
before we even start waiting was pure noise. `--no-wait` now echoes the same
compact receipt instead of the raw proto, so the two modes match.

The failed path is deliberately left alone: a send that fails is exactly
when you want the expanded trace, since the ledger and VTXO rows are what
you'll be staring at to figure out what went sideways.

stdout stays machine-readable JSON and the `phase:` lines stay on stderr, so
a shell pipeline can keep consuming the body while a human watching the
terminal still sees progress.

changes.

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_send.go
+++ b/cmd/darepocli/darepoclicommands/cmd_send.go
@@ -207,12 +207,18 @@ func walletSend(cmd *cobra.Command, args []string) error {
 					err)
 			}
 
-			if err := printWalletProto(resp); err != nil {
-				return err
-			}
-
 			if noWait, _ := cmd.Flags().GetBool("no-wait"); noWait {
-				return nil
+
+				// Without waiting there is no terminal state to
+				// summarize yet, so echo the dispatched entry
+				// as a compact pending receipt rather than the
+				// full proto dump.
+				return printSendResult(
+					cmd.OutOrStdout(),
+					sendResultFromEntry(
+						resp.GetEntry(),
+					),
+				)
 			}
 
 			return waitForSendTerminal(cmd, resp.GetEntry())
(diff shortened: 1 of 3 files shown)


Roasbeef:
@claude review this

PR #849, opened 2026-07-02
lndbackend+txconfirm: fix CPFP fee-input selection on the LND backend

### LINKED ISSUE #831: lwwallet: CPFP child fails to finalize taproot fee input (PSBT is not finalizable)

Roasbeef:
In this PR, we fix the LND-backend half of #831: CPFP fee-bumping for
unilateral exits silently failing with `PSBT is not finalizable`. Together
with #836 (the lwwallet half), this closes out the issue across all three
wallet adapters.

## The bug

The boarding backend imports boarding and exit taproot scripts into the same
LND wallet that funds CPFP children, and its unfiltered `ListUnspent`
surfaces those watched outputs alongside the wallet's own coins. LND's
`FinalizePsbt` cannot sign an imported tapscript output, so whenever one won
the smallest-first fee-input pick, the child failed to finalize and the fee
bump degraded to a direct broadcast of a zero-fee anchor parent, which cannot
relay on its own. The `lnd / exits-leave-sweep` itest shard shows this
chronically: a green baseline run on master carries 195 finalize warnings,
with retries recovering only when a signable coin happens to arrive. Under a
real relay-fee floor that means a stuck exit, not a slower one.

## The fix

Two commits. First, the backend gains `ListUnspentDefaultAccount`, which pins
the walletKit query to LND's default account via `WithUnspentAccount`, and
`lndUnrollWallet.ListUnspent` routes through it. The unfiltered enumeration
keeps its all-accounts semantics for the wallet actor, sweep, and recovery
callers that legitimately need to see watched outputs. This mirrors the
account scoping the btcwallet adapter already has and the lwwallet adapter
gains in #836.

Second, defense in depth in the broadcaster: a failure at the
`sign_cpfp_child` stage marks the selected outpoint as suspect, and selection
prefers coins with a clean signing history from then on. Suspects only lose
ties, they are not banned; when nothing else qualifies the broadcaster still
settles for one after the sync-lag poll window, so a transient signing
failure cannot permanently starve fee bumping, and a successful sign clears
the mark. Even if a future backend leaks an unsignable coin into fee
selection again, one bad coin can no longer shadow the whole wallet.

## Testing

`TestListUnspentAccountScoping` pins the account filter on the new backend
method (and the absence of one on the general enumeration).
`TestSelectFeeInputDeprioritizesSuspects`,
`TestFallbackDirectBroadcastMarksSuspects`, and
`TestSubmitClearsSuspectAfterSuccessfulSign` cover the selection tiers, the
stage-scoped marking, and the end-to-end recover-and-clear path. Full suites
for `txconfirm`, `lndbackend`, and `darepod` pass, lint is clean.

Closes #831 (together with #836).

### DIFF
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -4783,12 +4783,21 @@ type lndUnrollWallet struct {
 	boardingBackend *lndbackend.BoardingBackend
 }
 
-// ListUnspent delegates to the boarding backend's wallet UTXO
-// enumeration.
+// ListUnspent returns UTXOs from LND's default wallet account only.
+//
+// The boarding backend's unfiltered enumeration also surfaces imported
+// watch-only script outputs (boarding and exit scripts tracked via
+// ImportTaprootScript). Those are not signable by LND's FinalizePsbt, so
+// offering them as CPFP fee inputs makes the child PSBT unsignable and the
+// fee bump fails with "PSBT is not finalizable". Restricting fee selection
+// to the default account keeps selection aligned with what the finalize
+// path can actually sign, mirroring the lwwallet and btcwallet adapters.
 func (w *lndUnrollWallet) ListUnspent(ctx context.Context, minConfs,
 	maxConfs int32) ([]*wallet.Utxo, error) {
 
-	return w.boardingBackend.ListUnspent(ctx, minConfs, maxConfs)
+	return w.boardingBackend.ListUnspentDefaultAccount(
+		ctx, minConfs, maxConfs,
+	)
 }
 
 // NewWalletPkScript returns a fresh wallet-managed taproot pkScript suitable
--- a/darepod/unroll_fee_input_test.go
+++ b/darepod/unroll_fee_input_test.go
@@ -140,7 +140,8 @@ func newFundedLwWallet(t *testing.T) *lwwallet.Wallet {
 	t.Cleanup(esplora.Close)
 
 	w, err := lwwallet.New(lwwallet.Config{
-		Seed:           seed,
+		Seed:           seed[:],
+		WalletPassword: []byte("test-password"),
 		EsploraURL:     esplora.URL,
 		ChainParams:    &chaincfg.RegressionNetParams,
 		PollInterval:   time.Hour,
--- a/lndbackend/boarding_backend.go
+++ b/lndbackend/boarding_backend.go
@@ -19,6 +19,7 @@ import (
 	"github.com/lightninglabs/lndclient"
 	fn "github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/lightningnetwork/lnd/lnwallet"
 )
 
 // BoardingBackend implements the wallet.BoardingBackend interface by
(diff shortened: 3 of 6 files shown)


Roasbeef:
@claude review this

PR #868, opened 2026-07-06
multi: fix inverted swap-direction wording in wallet facade docs

jamaljsr:
## Summary

The wallet-facing facade describes `Recv` (Lightning receive) as opening a
"swap-in" and the invoice `Send` / `PrepareSend` (pay) as an "out-swap." That is
backwards relative to the swap subsystem's own canonical convention, where an
**in-swap is the Ark-to-Lightning pay direction** and an **out-swap is the
Lightning-to-Ark receive direction** (see `swaprpc` / `sdk/swaps`, e.g.
`CreateInSwap initiates an Ark-to-Lightning swap` and the `out-swap HTLC
notification for one receive invoice`). The wrong terminology flowed downstream
into the generated SDK API reference and confused a protocol reviewer.

The fix is split by layer:

- `wallet.proto` is documented as a "simplified, swap-vocabulary-free wallet
  API," and these two comments were the only place leaking swap-direction jargon
  into it. They are reworded to drop the jargon entirely and describe the
  operations as receive / send.
- The internal Go comments and CLI help legitimately use the swap subsystem's
  vocabulary, so those are left using it: just the inverted direction is
  corrected (a receive is an `out-swap`).

## Changes

- `rpc/walletdkrpc/wallet.proto`: reword the `Recv` RPC and
  `PrepareSendRequest.invoice` comments to describe receive / send without
  "swap-in" / "out-swap." Updated the generated `wallet.pb.go` and
  `wallet_grpc.pb.go` doc comments to match.
- `swapwallet/service.go`, `swapwallet/recv.go`, `swapwallet/doc.go`: correct
  the `Recv` doc comments from "swap-in" to "out-swap" (its actual direction).
- `cmd/darepocli/darepoclicommands/cmd_recv.go`: same direction fix in the
  `recv --offchain` CLI help text.
- No behavior change; comments only. The swap subsystem's own in-swap /
  out-swap vocabulary (`swaprpc`, `swapclientserver`, `sdk/swaps`) is correct
  and left untouched.

## Testing

- `git grep -nE 'swap-in' rpc/walletdkrpc swapwallet cmd` returns no matches,
  and the remaining `out-swap` usages describe the receive direction correctly.
- Comment-only change; no build or runtime behavior affected.
- The generated `.pb.go` comments were synced by hand to match the proto;
  running `make rpc` should confirm no drift.

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_recv.go
+++ b/cmd/darepocli/darepoclicommands/cmd_recv.go
@@ -17,7 +17,7 @@ func newRecvCmd() *cobra.Command {
 		Short: "Receive a payment (offchain invoice / onchain addr)",
 		Long: "Asks the daemon to materialize an inbound payment " +
 			"surface. With --offchain (default) the daemon opens " +
-			"a swap-in and returns a BOLT-11 invoice signed " +
+			"an out-swap and returns a BOLT-11 invoice signed " +
 			"with a daemon-managed key. With --onchain the " +
 			"daemon returns a fresh boarding address; fund it " +
 			"and the daemon rolls the boarding output into the " +
--- a/rpc/walletdkrpc/wallet.pb.go
+++ b/rpc/walletdkrpc/wallet.pb.go
@@ -1029,9 +1029,9 @@ type isPrepareSendRequest_Destination interface {
 }
 
 type PrepareSendRequest_Invoice struct {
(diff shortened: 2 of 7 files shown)


Roasbeef (approved):
LGTM 🐦‍⬛

PR #869, opened 2026-07-07
harness: add opt-in --requireinterceptor for the primary lnd

calvinrzachman:
Add `Options.LNDRequireInterceptor`, threaded through `startLNDInstance` into the primary lnd container args as `--requireinterceptor`. Only the primary swap node sets it; additional nodes (payer, extra) never do, so their forwarding is unaffected.

The swapdk server's out-swap disconnect itests use this to run the swap lnd under lnd's retain-and-replay-on-reconnect interceptor policy, which out-swap fund safety depends on when a held HTLC's subscriber drops.

### DIFF
--- a/harness/harness.go
+++ b/harness/harness.go
@@ -302,6 +302,12 @@ type Options struct {
 	// LNDImage is the docker image:tag to use for lnd.
 	LNDImage string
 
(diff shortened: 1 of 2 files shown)


Roasbeef (approved):
LGTM 🐴

PR #872, opened 2026-07-07
oor: avoid stale self-transfer reaps

bhandras:
## Summary

This PR fixes an intermittent OOR self-transfer race observed by `swapdk-server` in `TestInSwapLateFundAfterExpiryClientUnilateralRefund`.

When a client sends an OOR transfer to one of its own receive scripts, the outgoing OOR session and the incoming recipient session share the same deterministic session id. The registry already has logic to admit the incoming replacement only after the outgoing session reaches a terminal state, but terminal notifications only carried the session id. That let a stale terminal notification from the outgoing child race with the replacement incoming child and reap whichever child was currently active under that id.

The failure mode is:

1. The outgoing OOR session commits `completed`.
2. A terminal notification is sent asynchronously to the registry.
3. The incoming self-transfer hint is admitted and an incoming child is installed under the same session id.
4. The stale outgoing terminal notification reaches the registry before the incoming child commits its first snapshot.
5. The registry sees a terminal durable row for that session id and drops the active child, which is now the incoming replacement.
6. The incoming `ResolveIncomingTransferRequest` stays in the session mailbox, so the recipient VTXO exists on the operator/indexer side but never becomes a live local wallet VTXO.

## Changes

- Track the direction of each active OOR session child alongside the active child map.
- Ignore terminal notifications whose durable row direction does not match the currently active child direction.
- Keep historical behavior for hand-constructed/test-only active entries that do not have direction metadata.
- Add a regression test that models a terminal outgoing row, an active replacement incoming child, and a stale terminal notification racing in afterward.

## Validation

- `go test ./oor -run 'TestOORRegistry(SelfTransferParkAndRedrive|StaleTerminalDoesNotReapReplacementIncoming|SessionTerminal|RestoreNonTerminal)' -count=1`
- `go test ./oor ./darepod -count=1`
- `make lint`

The first `make lint` attempt hit a transient Docker/compiler resource failure (`compile: signal: killed`) while typechecking an unrelated package. A rerun with warm caches completed successfully with zero issues.

Fixes lightninglabs/swapdk-server#195

### DIFF
--- a/oor/registry.go
+++ b/oor/registry.go
@@ -214,10 +214,11 @@ func NewOORRegistryActor(cfg OORRegistryConfig) (*OORRegistryActor, error) {
 	cfg.Limits = normalizeReceiveLimits(cfg.Limits)
 
 	behavior := &oorRegistryBehavior{
-		cfg:      cfg,
-		log:      cfg.Log.UnwrapOr(btclog.Disabled),
-		active:   make(map[SessionID]*OORSessionActor),
-		incoming: make(map[SessionID]struct{}),
+		cfg:        cfg,
+		log:        cfg.Log.UnwrapOr(btclog.Disabled),
+		active:     make(map[SessionID]*OORSessionActor),
+		activeDirs: make(map[SessionID]clientdb.OORSessionDirection),
+		incoming:   make(map[SessionID]struct{}),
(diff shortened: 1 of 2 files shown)


bhandras:
@claude review

PR #882, opened 2026-07-08
swapwallet: keep pay activity live

### LINKED ISSUE #881: Activity Log: Successful Lightning payment shows up as `FAILED`
I made a Lightning payment (`r_hash b960600a4e03672eefd18d12604a1e5b932c57dc2670f62ffb6ac17a4c6b31f2` from darepocli. It failed silently, without any mention in the activity.

I then restarted `darepod`, and the payment started showing up as `FAILED` (`2026-07-08 08:35:21.000  SEND     FAILED    -1001   1    failed     b960600a4e03672eefd18d12604a1e5b932c57dc2670f62ffb6ac17a4c6b31f2`) in the activity log.

However, I can see from my Lightning node that the Lightning payment actually succeeded (`"settled":  true`).

Furthermore, the balance was not adjusted.

I've attached the logs, but please note that the change to `debuglevel=debug` only occurred after the restart.

[darepod.log](https://github.com/user-attachments/files/29808112/darepod.log)

bhandras:
## Summary

Fixes #881.

This PR fixes two client-side edges that can make a direct Lightning pay confusing or misleading in wallet activity:

- Direct Lightning sends now project the admitted `StartPay` row into the canonical activity store immediately.
- Pay sessions that are already funded keep polling when a local block-height lookup fails instead of durably marking the pay `FAILED`.

## Root Cause

The direct send path returned the `StartPay` summary to the caller, but it did not write that row into the canonical activity store. The path implicitly relied on the swap monitor. The monitor sends existing rows when the daemon subscribes and publishes again when a background worker exits, so a newly admitted direct pay could be absent from `darepocli activity` until restart backfill derived it from the swap store.

Separately, after a pay-side vHTLC was funded and the client was waiting for the server claim/preimage, a `BlockHeight` read error was returned as an ordinary action error. The generic FSM failure handler can persist ordinary action errors as `PayStateFailed`. That is unsafe after funding, because the server may still settle Lightning and claim the vHTLC, and a terminal `FAILED` row will no longer resume to observe the eventual preimage.

## Changes

- Project and emit the direct-pay `WalletEntry` immediately after `StartPay` succeeds.
- Treat block-height lookup failures during claim waiting as a transient poll condition.
- Treat block-height lookup failures during refund-maturity waiting as a transient poll condition.
- Add a router regression proving direct pays land in activity immediately.
- Add a pay-session regression proving a post-funding height-read failure does not persist `FAILED`, and that a resumed session can still complete once the claim preimage is indexed.

## Validation

```bash
go test ./sdk/swaps
go test -tags 'walletdkrpc swapruntime' ./swapwallet
git diff --check
```

I could not fetch the log attachment from #881 because GitHub returns `404 Not Found` for the linked file, so the fix is based on the reported activity row plus the code path that can produce the described symptoms.

### DIFF
--- a/sdk/swaps/in_swap.go
+++ b/sdk/swaps/in_swap.go
@@ -1044,7 +1044,24 @@ func (s *paySession) waitForClaimPreimage(ctx context.Context) error {
 
 		height, err := s.client.daemon.BlockHeight(ctx)
 		if err != nil {
-			return fmt.Errorf("get block height: %w", err)
+			s.client.log.DebugS(
+				ctx,
+				"Unable to query block height while waiting "+
+					"for in-swap claim",
+				slog.String("err", err.Error()),
+				btclog.Hex("hash", s.cfg.PaymentHash[:]),
+			)
+			if err := s.waitForNextPoll(ctx); err != nil {
+				if errors.Is(err, errSwapExpired) {
+					return s.initiateRefund(
+						ctx, "claim deadline elapsed",
+					)
+				}
+
+				return err
+			}
+
+			continue
 		}
 		if height >= s.cfg.VHTLCConfig.RefundLocktime {
 			return s.initiateRefund(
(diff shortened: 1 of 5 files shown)


darioAnongba:
See comment: https://github.com/lightninglabs/darepo-client/issues/881#issuecomment-4919452112.

PR #892, opened 2026-07-09
multi: enrich exit status with progress, fees, and an exit summary

Roasbeef:
In this PR, we make `da exit status` actually useful. Today a unilateral
exit that's grinding its way on-chain reports almost nothing:

```json
{
  "found": true,
  "status": "EXIT_JOB_STATUS_MATERIALIZING",
  "sweep_txid": "",
  "last_error": ""
}
```

A user watching that has no idea where they are: how much of the recovery
tree has confirmed, how many transactions are still left to broadcast,
what they've spent on fees, how far off the CSV timeout is, or roughly how
many blocks until the coin is finally back in their wallet. The status
even lags reality, since the registry answers from a coarse cached record
rather than the live actor, so the phase can read `MATERIALIZING` well
after the target has confirmed.

We fix both. The same call now returns the full picture:

```json
{
  "found": true,
  "status": "EXIT_JOB_STATUS_MATERIALIZING",
  "phase_detail": "materializing recovery tree: layer 3 of 5, 2/5 txs confirmed (1 in flight, 0 ready, 2 blocked)",
  "progress": {
    "confirmed_txs": 2, "in_flight_txs": 1, "ready_txs": 0, "blocked_txs": 2,
    "total_txs": 5, "current_layer": 2, "total_layers": 5,
    "target_confirmed": false, "all_proof_confirmed": false
  },
  "csv": null,
  "fees": {
    "cpfp_fee_sat": "1550", "sweep_fee_sat": "400", "total_cost_sat": "1950",
    "spent_so_far_sat": "620", "vtxo_amount_sat": "177591",
    "net_recovered_sat": "177191", "fee_rate_sat_vbyte": "2",
    "sweep_fee_actual": false
  },
  "best_case_blocks_remaining": 148,
  "current_height": 312354
}
```

Everything here is derived, not newly persisted, so there are no schema
changes anywhere in the PR.

## Where the data comes from

The per-target unroll actor already holds the immutable proof, the pure
`unrollplan.Planner`, and its durable state, so it can run one `Plan` at
the current best height and project the resulting snapshot into a compact
`ExitProgress`: the confirmed / in-flight / ready / blocked transaction
counts, the current frontier layer of the recovery tree, the CSV maturity
countdown (once the target confirms), and an optimistic block estimate to
a confirmed sweep. The fee breakdown reuses the same cost model as
`GetExitPlan` (`unroll.PlanExitFunding`), so `da exit plan` and
`da exit status` agree on the numbers.

Fees are an estimate by design. The CPFP total comes from the cost model
(per-child fees aren't persisted, so there's nothing exact to report), and
`spent_so_far_sat` prorates that total over the proof transactions already
broadcast, then folds in the sweep fee once the sweep exists. The one
number we can pin down exactly is the sweep fee itself, derived from the
built sweep tx (target value minus the swept output) and flagged with
`sweep_fee_actual` when it's real rather than modeled. A completed or
failed exit whose actor has already been evicted still answers from the DB
store with its phase line and a fresh fee estimate, so the endpoint stays
useful after the exit is done.

## Detail is opt-in so the pollers stay cheap

A `GetStateRequest` against a durable child actor becomes a persisted
mailbox row, and the history and reconciler loops hit `GetUnrollStatus`
on every pending exit, so we can't have every status read waking the
child. All of the rich detail is gated behind a `detailed` request flag.
The interactive CLI sets it (so `da exit status` is detailed by default,
with `--detailed=false` as the cheap phase-only escape hatch), the
automated pollers don't. Even on the detailed path the registry Asks the
child under a short local timeout and falls back to the coarse cached
record, so a wedged child can never stall the single-goroutine registry
on a user-triggered probe.

## A wallet-wide view: `da exit summary`

Looking at one exit doesn't tell you how much is still tied up across all
of them. So we add `da exit summary`, a portfolio of every in-progress
exit plus aggregate totals: the amount still being recovered, the
estimated fees, and the estimated net recoverable. It only walks the
non-terminal jobs (a completed exit has nothing left to recover) and stays
cheap by reading the coarse job rows and the descriptor, never probing the
live actors.

```json
{
  "exits": [ { "outpoint": "...", "status": "EXIT_JOB_STATUS_MATERIALIZING",
               "vtxo_amount_sat": "177591", "est_total_fee_sat": "1950",
               "est_net_recovered_sat": "177191" } ],
  "total_exits": 1,
  "total_vtxo_amount_sat": "177591",
  "total_est_fee_sat": "1950",
  "total_est_net_recovered_sat": "177191"
}
```

## Review notes

The change runs bottom-up across the stack (unroll actor/registry, then
the protos, then the regenerated stubs, then the daemon RPC, then the
walletdk projection, then the CLI, then the SDK facades), split into
seven atomic commits. See each commit message for the detailed
description w.r.t the incremental changes. The generated-stubs commit
also carries the hand-written REST transport mirror so the
`WalletServiceClient` interface stays satisfied, which keeps the default
`go build` bisectable at every commit.

The last commit threads the same view through the `walletdk` SDK, its
gomobile JSON facade, and the browser WASM entrypoint, so embedded hosts
get the detailed status and the `exitSummary` verb too, not just the
`darepocli` user.

The progress math (frontier layer, best-case block countdown,
spent-so-far proration, actual sweep fee) is covered by table tests in
`unroll` and `darepod`.

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_exit.go
+++ b/cmd/darepocli/darepoclicommands/cmd_exit.go
@@ -53,6 +53,7 @@ func newExitCmd() *cobra.Command {
 			"preview, non-zero on validation failure")
 
 	cmd.AddCommand(newExitStatusCmd())
+	cmd.AddCommand(newExitSummaryCmd())
 	cmd.AddCommand(newExitPlanCmd())
 
 	return cmd
@@ -120,10 +121,17 @@ func newExitStatusCmd() *cobra.Command {
 		Use:   "status",
 		Short: "Query the status of an exit (unroll) job",
 		Long: "Returns the current status of a unilateral exit job " +
-			"for the specified VTXO outpoint, including the " +
-			"job phase, sweep txid, and any errors.\n\n" +
+			"for the specified VTXO outpoint. By default the " +
+			"response includes recovery-tree progress (layer and " +
+			"transaction counts), the CSV maturity countdown, a " +
+			"best-case block estimate to full exit, and the " +
+			"on-chain fee breakdown, alongside the job phase, " +
+			"sweep txid, and any errors. The endpoint stays " +
+			"queryable after the exit completes.\n\n" +
 			"Example:\n" +
-			"  darepocli exit status --outpoint TXID:VOUT",
+			"  darepocli exit status --outpoint TXID:VOUT\n" +
+			"  darepocli exit status --outpoint TXID:VOUT " +
+			"--detailed=false",
 		Args: cobra.NoArgs,
 		RunE: walletExitStatus,
 	}
@@ -131,6 +139,10 @@ func newExitStatusCmd() *cobra.Command {
 	cmd.Flags().String("outpoint", "",
 		"VTXO outpoint to query (txid:vout)")
 	_ = cmd.MarkFlagRequired("outpoint")
+	cmd.Flags().Bool("detailed", true,
+		"include tree/CSV progress, best-case block countdown, and "+
+			"the fee breakdown; pass --detailed=false for a "+
+			"coarse, cheaper phase-only status")
 
 	return cmd
 }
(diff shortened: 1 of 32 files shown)


Roasbeef:
@claude review this

Roasbeef:
Thanks all — addressed the review feedback in three follow-up commits (`872037f`, `1f996ba`, `2f7e64b`), appended rather than folded so the reviewed commits and inline threads stay anchored.

**Correctness**
- **`actualSweepFeeSat` used `desc.Amount`, flagged `sweep_fee_actual`** (claude): now derives the input value from `b.proof.TargetOutput().Value`, the exact output the sweep spends, so the "actual" fee can't diverge from the built tx.
- **Best-case ignores an exit policy's absolute locktime** (codex P2): a vHTLC refund-without-receiver gates the sweep on an `nLockTime` the standard policy lacks, so once CSV matured the estimate reported `1` while the sweep was still many blocks out. We now cache the policy's `RequiredLockTime()` when `startSweep` resolves it and take `max(csv, locktime)` for the post-target wait.
- **`spentSoFarSat` integer-division truncation** (gemini/claude): switched to multiply-before-divide so per-child truncation can't zero the estimate.

**Robustness / polish**
- **`unrollPhaseDetail` could render "layer N+1 of N"** (claude): clamped the 1-based layer display to `TotalLayers`.
- **`ExitSummary` hard-failed on a fee-estimate error** (gemini): now degrades to a zero fee rate (amounts/phases still come from the store) instead of sinking the whole summary — symmetric with `enrichExitFees`.
- **`exit summary` missing from schema/MCP** (codex P2): registered `exit.summary` in `schema_registry.go` and as an MCP tool; also gave the MCP `exit.status` tool `detailed=true` to match the CLI default.

**Acknowledged, no change**
- The 3s bounded child-Ask in `detailedChildState` (claude, low) is a deliberate, documented tradeoff with an always-available coarse fallback.
- `unrollProgressToProto`'s `int → uint32` cast is safe by the planner invariant (`ConfirmedTxs = totalTxs − unconfirmed ≥ 0`).

All builds green (default + `walletdkrpc swapruntime`), new table cases cover the locktime `max` and the input-value fee.

Roasbeef:
Rebased onto current `main` and fixed the CI break. The failing "Mobile Bindings" job was not actually a bindings problem: that job compiles `darepod`, and CI builds the PR *merged with `main`*, where #888 (`03d4909`) had changed `unroll.PlanExitFunding` to take a `mat *LineageMaterial` parameter. My two new callers still used the old 3-arg form, so the merge didn't compile.

Adopted the new signature: the single-exit `enrichExitFees` resolves lineage via `resolveExitLineage` so the detailed `exit status` fee stays exact and in agreement with `exit plan`, while the aggregate `exit summary` passes `nil` (the descriptor-only ChainDepth approximation) to keep the portfolio cheap.

The rebase also ran the full `darepod` suite, which caught a gap the earlier partial runs missed: `TestDarepodRPCPermissionsCoverDaemonServices` requires every registered method to have a macaroon permission, so `ExitSummary` now gets the same `entityOnChain` read grant as `ExitStatus`/`GetExitPlan`.

Force-pushed (linear history preserved), so the inline threads above now point at pre-rebase blobs. All 10 commits build against current `main` (default + `walletdkrpc swapruntime` + `js/wasm`); unroll/darepod/swapwallet/cmd/sdk tests pass.

Roasbeef:
@claude review this

Issue #899, opened 2026-07-09
swapwallet: durable leave record for restart-survivable cooperative-leave EXIT (#776)


darioAnongba:
> Child of #776. Originally scoped as "C5: startup reconciliation at the chain tip" — that startup-at-tip mechanism was **investigated and decided against** (keep the ordering-independent 60s reconciler poll; block-epoch triggering fires before the wallet processes the block and doesn't reduce loops — see comments). The remaining epic work is the AC2 residual below.

## Problem (AC2 residual)

The on-chain cooperative-leave EXIT is keyed by a stable `send_job_id` (a deterministic hash of the consumed outpoints; `swapwallet/router.go`), and its PENDING row is projected at submit. But that pending row is **process-local** (the runtime's in-memory pending map): there is no durable leave-job → forfeit link. So after a mid-flight restart:

- the in-memory pending row is gone, and
- the confirmed sweep resurfaces from the ledger keyed by its **txid** (a different id, via `ledgerActivityID`),

leaving the `send_job_id` PENDING→COMPLETE transition unrecoverable — cooperative-leave completion is best-effort, not restart-survivable. (`swapwallet/doc.go`; `decorateCooperativeLeaveEntry` matches the retained VTXO outpoint against the forfeited set, which only works while the in-memory row survives.)

## Proposed work

Persist a durable leave record linking `send_job_id` → retained consumed-outpoint(s) (and, once known, the forfeit/sweep txid) so a restarted daemon can rebuild the PENDING→COMPLETE transition under the stable `send_job_id`, instead of losing it or re-surfacing the sweep under a txid id.

Sketch (refine during implementation):
- A small durable record written at leave submit: `{send_job_id, retained_outpoints, created_at}`.
- On startup backfill / reconcile, rebuild the EXIT row from that record + the forfeited-VTXO set, keyed by `send_job_id`, so the flip survives restart.
- `ledgerActivityID` keys the confirmed sweep by `send_job_id` (via the durable link) rather than txid, so the confirmed row merges with the pending one.

## Acceptance
- A cooperative-leave EXIT submitted, then interrupted by a daemon restart before the round seals, still flips PENDING→COMPLETE under the stable `send_job_id` after restart (no orphaned txid-keyed row).

## Decided against (see comments)
[...]

PR #901, opened 2026-07-09
swapwallet: reconcile activity at the chain tip on startup (C5a)

darioAnongba:
Implements **C5a** of #899 (startup reconciliation at the chain tip).

## Problem

The startup backfill (`backfillActivity` in `resumeAll`) seeds the activity
store at wallet-ready. In `lwwallet`/`lnd` modes that hook fires **before** the
chain backend has synced forward to the tip, so the backfill re-derives pre-tip
state. A DEPOSIT that confirmed or an EXIT that was swept while the daemon was
down reads PENDING until the periodic reconciler's first tick — up to a full
`reconcileInterval` (60s) later, plus sync lag. A UI reconnecting right after a
restart sees a settled op look stuck, then silently jump to terminal.

## Change

A one-shot startup goroutine (`swapwallet/tip_reconciler.go`, launched from
`start()`) that polls `GetInfo` until the best-block height is **stable across a
poll interval and the wallet is READY**, then runs one reconciler pass
(`reconcileActivity` — DEPOSIT/EXIT, since SEND/RECV self-heal via the live
monitor `SubscribeSwaps` stream). A timeout backstop reconciles once anyway if
the backend never settles; the periodic reconciler owns steady state afterward.
`tipPollInterval`/`tipReconcileTimeout` are `Runtime` fields so tests drive the
loop without real-time waits.

The "caught up" test is a heuristic: `GetInfo.block_height` is the backend's
best-known height and `wallet_state` flips READY at wallet-ready, so there is no
precise "wallet synced to tip" signal on the current RPC surface. This is a
strict improvement (worst case it runs one extra reconcile early; the periodic
reconciler covers any residual wallet-processing lag). A precise tip signal
(e.g. a `GetInfo` target-height field or a block-epoch subscription) is noted as
a follow-up in #899.

## Tests

`swapwallet/tip_reconciler_test.go` (all pass under `-race`): fires on
stable-height + READY, waits while SYNCING, timeout backstop fires when the
height never settles, and is a no-op without a store / RPCServer.

## Scope

This is **C5a**. **C5b** — the durable leave record (the AC2 residual:
restart-survivable cooperative-leave EXIT completion) — is the heavier,
independent follow-up tracked in #899.

### DIFF
--- a/swapwallet/mocks_test.go
+++ b/swapwallet/mocks_test.go
@@ -34,6 +34,8 @@ type fakeRPCServer struct {
 	listTxLastReq    *daemonrpc.ListTransactionsRequest
 	getInfoResp      *daemonrpc.GetInfoResponse
 	getInfoErr       error
+	getInfoHook      func(call int) (*daemonrpc.GetInfoResponse, error)
+	getInfoCalls     int
 	getBalanceResp   *daemonrpc.GetBalanceResponse
 	getBalanceErr    error
 	newAddressResp   *daemonrpc.NewAddressResponse
@@ -151,6 +153,11 @@ func (f *fakeRPCServer) ListTransactions(_ context.Context,
 
 func (f *fakeRPCServer) GetInfo(_ context.Context,
 	_ *daemonrpc.GetInfoRequest) (*daemonrpc.GetInfoResponse, error) {
+
+	f.getInfoCalls++
+	if f.getInfoHook != nil {
+		return f.getInfoHook(f.getInfoCalls)
+	}
 
 	if f.getInfoResp == nil && f.getInfoErr == nil {
 		return &daemonrpc.GetInfoResponse{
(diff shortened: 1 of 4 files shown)


darioAnongba:
Closing in favor of an event-driven design. As discussed: this adds a third polling loop, against the goal of reducing them, and relies on a shaky "at tip" heuristic. The DEPOSIT/EXIT transitions being reconciled are all on-chain confirmations, the daemon already has an internal block-epoch notification (chainbackends, `RegisterBlockEpochNtfn`) not yet exposed to the swapwallet contract, and `doc.go` already names a "per-block/confirmation reconcile trigger" as the intended design.

The better design — plumb the block-epoch feed through the swapwallet `Deps` contract and drive DEPOSIT/EXIT reconciliation off block events (coalesced), retiring the 60s C2 poll and getting startup catch-up for free — is being scoped into #899 (C5).

PR #902, opened 2026-07-09
swapwallet: make cooperative-leave EXIT restart-survivable (C5, #776)

### LINKED ISSUE #899: swapwallet: durable leave record for restart-survivable cooperative-leave EXIT (#776)

darioAnongba:
Implements the durable-leave-record work in #899 (the last remaining piece of the activity-log epic #776): make cooperative-leave EXIT completion restart-survivable.

## Problem (AC2 residual)

The on-chain cooperative-leave EXIT is projected as a durable PENDING row at submit, keyed by its stable `send_job_id` with the retained consumed outpoint in `vtxo_outpoint`. But its PENDING→COMPLETE flip is driven by matching that outpoint against the forfeited-VTXO set, and the only source the completion pass reads is the runtime's **in-memory pending map** — populated only by the (RPC-driven) submit path. A restart empties that map, so a leave interrupted mid-flight, or one that sealed while the daemon was down, was stranded PENDING in the store forever.

## Change

`rehydrateWalletLocalPending` (called at the top of `resumeAll`, before the swap-backend guard and before `backfillActivity`) pages the store's current-state rows and re-tracks the PENDING EXIT ones into the in-memory pending map. The existing derive → `decorateCooperativeLeaveEntry` → reproject → clear pipeline then flips the row COMPLETE under its stable id on the first backfill/reconcile pass after a restart — no behavior change to that pipeline.

**No new table / SQL.** The `send_job_id → retained-outpoint` link already round-trips through the `activity_entries` row (`canonical_id` + `vtxo_outpoint`); rehydration reuses the existing `ActivityStore.ListEntries` keyset scan. Also corrected the now-stale `doc.go` note (completion is restart-survivable; counterparty/note/amount survive in the store).

## Design notes

- This is what remained of C5 after the startup-reconciliation-at-tip mechanism was dropped (the ordering-independent 60s reconciler is kept; block-epoch triggering was rejected as the wrong layer — a block epoch fires before the wallet processes the block into the derive sources; see #899).
- A cooperative leave books a hidden `round`-type ledger row with **no txid**, so there is no duplicate txid-keyed EXIT row — the bug was exactly one stranded PENDING row, so no dedup is needed.

## Tests

- `TestReconcileCompletesLeaveAfterRestart` — models a restart (PENDING row in the store, empty in-memory map), runs `resumeAll` (rehydrate — also asserting it runs *before* the swap-backend guard) + one reconcile, and asserts the row completes under `send_job_id`, the outpoint is preserved, and the pending record is cleared.
- `TestRehydrateWalletLocalPending` — only PENDING EXIT rows (both `send_job_id`- and outpoint-keyed) are re-tracked; a COMPLETE EXIT and a PENDING SEND are not.

All pass under `-race`; full `swapwallet` suite green; `lint-changed-local` clean.

Closes #899.

### DIFF
--- a/darepod/config.go
+++ b/darepod/config.go
@@ -664,6 +664,13 @@ type ActivityStore interface {
 	ListEntries(ctx context.Context, cursorCreated int64, cursorID string,
 		limit int32) ([]sqlc.ActivityEntry, error)
 
+	// ListEntriesByKindStatus returns up to limit rows of the given kind
+	// and status, paged by canonical_id ascending after cursorID. It backs
+	// the startup rehydration of the wallet-local pending map, scanning
+	// only the matching rows rather than decoding the whole feed.
+	ListEntriesByKindStatus(ctx context.Context, kind, status int64,
+		cursorID string, limit int32) ([]sqlc.ActivityEntry, error)
+
 	// PullEvents returns up to limit append-only transition rows with
 	// event_seq strictly greater than cursor, oldest-first. It is the
 	// resumable-subscribe replay primitive: a reconnecting client passes
--- a/db/activity_store.go
+++ b/db/activity_store.go
@@ -32,6 +32,12 @@ type ActivityStore interface {
 		error,
 	)
 
+	ListEntriesByKindStatus(ctx context.Context,
+		arg sqlc.ListEntriesByKindStatusParams) (
+		[]sqlc.ActivityEntry,
+		error,
+	)
+
 	PullActivityEvents(ctx context.Context,
 		arg sqlc.PullActivityEventsParams) ([]sqlc.ActivityEvent, error)
 }
@@ -248,6 +254,34 @@ func (s *ActivityPersistenceStore) ListEntries(ctx context.Context,
 				CursorCreated: cursorCreated,
 				CursorID:      cursorID,
 				LimitCount:    limit,
+			},
+		)
+
+		return err
+	})
+
+	return rows, err
+}
+
+// ListEntriesByKindStatus returns up to limit entries of the given kind and
+// status, paged by canonical_id ascending after cursorID. Filtering in SQL
+// keeps a scan for a specific kind/status (e.g. the startup rehydration of
+// PENDING EXIT rows) proportional to the matching rows, and the unique
+// canonical_id cursor is strictly monotonic.
+func (s *ActivityPersistenceStore) ListEntriesByKindStatus(ctx context.Context,
+	kind, status int64, cursorID string, limit int32) ([]sqlc.ActivityEntry,
+	error) {
+
+	var rows []sqlc.ActivityEntry
+
+	err := s.db.ExecTx(ctx, ReadTxOption(), func(q ActivityStore) error {
+		var err error
+		rows, err = q.ListEntriesByKindStatus(
+			ctx, sqlc.ListEntriesByKindStatusParams{
+				Kind:       kind,
+				Status:     status,
+				CursorID:   cursorID,
+				LimitCount: limit,
 			},
 		)
 
--- a/db/sqlc/activity_log.sql.go
+++ b/db/sqlc/activity_log.sql.go
@@ -160,6 +160,77 @@ func (q *Queries) ListActivityEntries(ctx context.Context, arg ListActivityEntri
 	return items, nil
 }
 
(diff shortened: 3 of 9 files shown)


darioAnongba:
Addressed the review (and the gemini bot's startup-scan note):

**Fixed**
- **M-1 (full-table scan/decode)** — replaced the generic `ListEntries` scan with a filtered `ListEntriesByKindStatus` query (`kind = EXIT AND status = PENDING`, paged by the unique `canonical_id`). Rehydration now reads only the matching rows instead of decoding the whole feed. This also **structurally removes L-1** (the `canonical_id` cursor is strictly monotonic, so the `created_at == 0` reset can't happen) and **L-4** (no Go-side kind/status filter left to wrap). Generated sqlc committed separately.
- **M-2 (pager/error branches untested)** — `rehydratePageSize` is now an injectable field; added `TestRehydratePagesMultipleBatches` (5 rows, page size 2 → 3 pages; asserts each restored exactly once and the loop terminates) and `TestRehydrateReturnsOnScanError` (scan error → clean abort, empty map).
- **L-3** — dropped the `#899` token from the test comment.

**Not changing (with reasoning)**
- **L-2 (`!looksLikeOutpoint` to skip unilateral EXITs)** — the filter is intentionally kind+status, not id-shape: a *fallback* cooperative leave (when the daemon returns no `send_job_id`) is *also* outpoint-keyed, so `!looksLikeOutpoint` would silently drop those. Re-tracking a unilateral EXIT is harmless (`dedupeByID` + `GetUnrollStatus` routing); the ghost-PENDING edge it cites is the pre-existing no-FAILED-path gap (I-4), not introduced here.
- **L-5 (retry on startup scan error)** — adding a retry loop for a rare local-DB startup read error runs against the "fewer loops" direction for marginal benefit; left as a one-shot (a scan error is logged and the next restart re-runs it).

No fund/consensus/signing exposure and no false-positive completion, per the review's verified informational findings.

PR #905, opened 2026-07-09
wallet: fix onchain sweep-all send failing before it leaves the daemon

### LINKED ISSUE #889: Unable to cooperatively exit with `send --onchain --sweep-all`
I am running `darepocli send --onchain --sweep-all` to the address `tb1qn6w5rynua82wfmx6x9jj4hmzmkn8umpdc5v6km` on signet, using btcwallet as the backend.

The payment shows as pending in the activity log, but the balance is returned to me.

The same happens when I run the command again, but no new entry is made to the activity log.

First attempt:

[2026-07-08_darepod_02.log](https://github.com/user-attachments/files/29820853/2026-07-08_darepod_02.log)

After a restart, I attempt again:

[darepod.log](https://github.com/user-attachments/files/29820846/darepod.log)

Roasbeef:
In this PR, we fix a bug where a `da send <addr> --onchain --sweep-all` never made it out of the daemon. The send would be accepted, the forfeits reserved, the intent registered, and then the round would die on the very next turn:

```
ROND: temp(ff45310c) State transition from_state=PendingRoundAssembly to_state="ClientFailed: no VTXO output amount"
ROND: Round failed err="total VTXO output is zero" reason="no VTXO output amount" recoverable=true
DRPD: SendOnChain completed status=submitted sweep_all=true actual_amount_sat=1317437
```

So the CLI cheerfully reported `status=submitted` while the round had already failed underneath it. Every sweep-all onchain send hit this.

## Root cause

The break is in the intent builder (`buildSendOnChainIntentPackage`). On the sweep-all path we shipped the single leave output with no value at all, on the theory that the server stamps the real amount onto it at seal time since it's the `IsChange` slot. But the round FSM's `IntentRequested` transition sums the leave output values to compute the intent's total output, and hard-fails a zero total (`"total VTXO output is zero"`) before the intent ever reaches the operator.

The bounded-mode change VTXO right below already understood this: it ships a real, non-zero placeholder above the operator's dust floor precisely so it clears admission. The sweep-all leave just broke that invariant by leaving `Value` at zero.

## The fix

In `wallet`, we stamp the pre-fee `Σ(inputs)` onto the sweep-all leave. `IsChange` is still set, so the server rewrites the slot with the seal-time residual (`Σin - fee`) regardless of the placeholder we send. The value only has to be non-zero for the client FSM and above dust for the operator's `ValidateLeaveRequest` floor, both of which `Σ(inputs)` satisfies by construction (each forfeited VTXO already cleared dust). The change lands in the builder shared by the live send and the restart-replay path, so both produce the identical shape.

I checked the server side (darepo `rounds/seal_time_fee_builder.go`) to be sure: it excludes the `IsChange` leave from the fixed-output total and stamps `residual = Σin - fee` onto it, so our placeholder never distorts the fee math. The bug is entirely client-side and the server needs no change.

## Testing

The wallet unit tests cover both the live and replay builder paths, and both fail on a bare-zero leave (verified by reverting the one-line fix). The `round` FSM test drives a forfeit-funded, leave-only, zero-VTXO intent (the exact sweep-all shape) through `IntentRequested` and asserts it clears into `IntentSent` rather than failing the round.

The gap that let this ship was that nothing drove a real sweep-all through the round FSM, so we also add `TestSendOnChainSweepAllEndToEnd`: it seeds a live VTXO, submits a sweep-all `SendOnChain` against the full daemon stack, and asserts a `JoinRound` request egresses to the fake operator mailbox, which only happens once the FSM clears `IntentRequested`. Against a live daemon the round now walks `PendingRoundAssembly -> IntentSent` and publishes the registration:

```
ROND: temp(f7a8054e) Intent balance check passed total_input="0.00100000 BTC" total_output="0.00100000 BTC" estimated_operator_fee="0 BTC"
ROND: temp(f7a8054e) Sending JoinRoundRequest to server forfeit_requests=1 leave_requests=1
ROND: temp(f7a8054e) State transition from_state=PendingRoundAssembly to_state=IntentSent
```

Note this is stacked on `send-invoice-eager-projection`.

Fixes https://github.com/lightninglabs/darepo-client/issues/889

### DIFF
--- a/round/transitions_test.go
+++ b/round/transitions_test.go
@@ -826,6 +826,62 @@ func TestPendingRoundAssemblyState(t *testing.T) {
 		require.Contains(t, failedState.Reason, "no VTXO output amount")
 	})
 
+	t.Run("sweep_all_leave_only_intent_passes", func(t *testing.T) {
+		t.Parallel()
+
+		h := newTestHarness(t)
+
+		// A sweep-all onchain send has no VTXO requests: the input
+		// comes entirely from a forfeited VTXO and the sole output is
+		// one IsChange leave. buildSendOnChainIntentPackage stamps the
+		// pre-fee Σ(inputs) onto that leave so this intent clears the
+		// IntentRequested output validation instead of tripping the
+		// "no VTXO output amount" guard on a bare-zero leave — the
+		// exact failure that killed every sweep-all send before the
+		// fix.
+		forfeitOutpoint := wire.OutPoint{
+			Hash: chainhash.Hash{
+				0xaa,
+			},
+			Index: 0,
+		}
+		const sweepAmount = btcutil.Amount(50000)
+		h.vtxoStore.On(
+			"GetVTXO", mock.Anything, forfeitOutpoint,
+		).Return(&ClientVTXO{
+			Outpoint: forfeitOutpoint,
+			Amount:   sweepAmount,
+		}, nil)
+
+		leavePkScript := append(
+			[]byte{0x51, 0x20}, bytes.Repeat([]byte{0xcd}, 32)...,
+		)
+		h.withState(&PendingRoundAssembly{
+			Forfeits: []types.ForfeitRequest{{
+				VTXOOutpoint: &forfeitOutpoint,
+				Amount:       sweepAmount,
+			}},
+			Leaves: []*types.LeaveRequest{{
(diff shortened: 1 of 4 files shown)


Roasbeef:
@claude review this

PR #910, opened 2026-07-10
multi: Bound credit awaiting polls to fail fast

### LINKED ISSUE #880: CLI hangs after confirming external offchain payment "credit shortfall"
I generated a 500 sat Lightning invoice from an external node and am trying to pay it from `darepocli send --offchain`

I'm shown the warning: `Warning: credit shortfall requires 546 sat top-up`

I can nonetheless confirm the payment in the dialogue, but the payment then indefinitely hangs.

It is never made.

ellemouton:
## Problem

Fixes #880. `darepocli send --offchain` on a **credit-shortfall** invoice (the CLI warns e.g. `Warning: credit shortfall requires 546 sat top-up`) lets the user confirm, then hangs indefinitely — with no error and the payment never made.

## Root cause

A credit-backed send routes through the durable credit subsystem (`swapwallet/router.go` → `sendCreditInvoiceIntent` → `credit.Registry`). The per-operation FSM performs the Ark top-up then the pay, parking in awaiting states (`topupAwaitingCredit`, `payAwaitingSettlement`) that poll the server until it reports a terminal state.

The FSM already has a fail-fast valve — `MaxAwaitingPolls` → `awaitExhausted()` → `b.fail(...)`, covered by `credit.TestAwaitingPollCapFailsOp`. **But the production daemon wiring (`darepod/credit_registry.go`) never set `MaxAwaitingPolls`, so it defaulted to `0` = unlimited**, disabling the cap in production. When the top-up funding or the credit pay never converges server-side, the op polls forever, the wallet entry stays `PENDING` forever, and the CLI hangs.

## Fix

Wire a bounded default into the credit registry:

- `credit.DefaultMaxAwaitingPolls = 120` (~4 min per awaiting state at the 2s poll interval, deliberately under the CLI's 5-min `defaultSendWaitTimeout` so the clear failure reason surfaces first). Documented rationale in-comment.
- New configurable `CreditConfig.MaxAwaitingPolls` daemon knob (`credit.maxawaitingpolls`); `darepod` coerces `0` → the default so production can never run unbounded. The library keeps its `0 = unlimited` contract for tests/embedders.

A stuck credit-shortfall send now **terminal-fails** with a clear reason (`"top-up credit not finalized within poll cap"` / `"credit pay not settled within poll cap"`); the projector flips the same payment-hash-keyed entry to `FAILED`, and the CLI surfaces `SEND_FAILED` instead of hanging.

## Testing

- `go build` (credit + darepod, `swapruntime,walletdkrpc`), `go vet`, `make lint-changed-local` (0 issues), `make fmt-changed` — clean.
- `go test ./credit/` and the new `darepod` wiring test pass.
- Note: `arktest` cannot reproduce this path — its topology has no swap-server/credit/Lightning-invoice wiring — so verification rests on the static analysis plus the FSM-level (`TestAwaitingPollCapFailsOp`) and wiring-level unit tests.

## Known adjacent limitation

For a *mixed* credit+Lightning pay (`creditOnly=false`), the projector deliberately skips the row (the swap monitor owns it). If a mixed pay's top-up stalls, the op now fails but the entry isn't flipped by the projector, so the CLI bounds at its 5-min wait rather than showing the credit reason. The reported case is credit-only (shortfall covers the full amount → `creditOnly=true`) and is fully resolved.

### DIFF
--- a/credit/config.go
+++ b/credit/config.go
@@ -377,4 +377,24 @@ const (
 	// DefaultAdmitTimeout bounds the synchronous server work an admission
 	// performs on the supervisor goroutine.
 	DefaultAdmitTimeout = 30 * time.Second
+
+	// DefaultMaxAwaitingPolls bounds how many reconciliation polls a single
+	// awaiting state (top-up funding or credit-pay settlement) may take
+	// before the operation terminal-fails. It is the production backstop
+	// against a credit-backed send parking forever when the server never
+	// reports a terminal state for the top-up or the pay — the hang behind
+	// darepo-client#880, where a confirmed credit-shortfall send never
+	// completes and never fails.
+	//
+	// At DefaultPollInterval (2s) this is ~4 minutes of awaiting per state:
+	// generous enough to absorb a slow OOR top-up credit or real-Lightning
+	// settlement, yet deliberately under the CLI's 5-minute default send
+	// wait (cmd/darepocli defaultSendWaitTimeout) so a stuck operation's
+	// clear terminal failure reason usually surfaces before the CLI falls
+	// back to a generic wait timeout. Zero (the OpActorConfig /
+	// RegistryConfig field default) still means "unlimited" for tests and
+	// embedders that rely on server-reported terminal states; production
+	// wiring coerces zero to this cap so the fail-fast bound is never
+	// silently disabled.
+	DefaultMaxAwaitingPolls uint32 = 120
 )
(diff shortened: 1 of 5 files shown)


ellemouton:
Addressed the gemini-code-assist suggestion: the `creditMaxAwaitingPolls` helper is now a `CreditConfig.MaxAwaitingPollsOrDefault()` method next to the config type (call site and test updated). Force-pushed.

@claude review this

ellemouton:
Thanks @claude. Addressed:
- **sample-darepod.conf**: added the `swap.credit.maxawaitingpolls` entry next to the other credit knobs (this also fixes the Static Checks `sample-conf-check` failure).
- **doc duplication**: trimmed the call-site comment to a one-liner that points at `MaxAwaitingPollsOrDefault`.

The mixed-pay projector limitation stays scoped as a follow-up as discussed.

PR #911, opened 2026-07-10
unroll: Bound proof-node confirmation-watch height floor

### LINKED ISSUE #884: Excessive neutrino log activity after unilateral exit
I exited a vtxo unilaterally using the btcwallet backend.

Shortly after, the logs show only:

`2026-07-08 12:06:30.358 [DBG] BTCW: Received cfilter from 208.68.4.71:38333 (outbound)`

In at least one instance 19 times in one millisecond. The logs grew from 800kB to 80MB in just a few minutes.

I've attached only a truncated version of the logs, but it's easy to imagine the rest!

[2026-07-08_darepod.log](https://github.com/user-attachments/files/29814578/2026-07-08_darepod.log)

ellemouton:
Fixes #884.

## Problem

After a unilateral exit on the btcwallet/neutrino backend, the unroll
actor registered a confirmation watch for each proof-graph node with a
height hint of `1` (genesis). lnd's neutrino chain notifier treats that
hint as "the tx may have confirmed anywhere from genesis to tip" and
dispatches a historical rescan that walks every block from the current
tip down toward block 1, issuing a `GetCFilter` per block. Because the
v3/TRUC exit tx is rejected by signet peers it never confirms, so the
"never give up on a no-mempool tx" invariant keeps the confirmation
watch (and its descending rescan) alive — flooding the logs with
`Received cfilter` lines (800 kB → 80 MB in minutes).

The genesis floor was chosen deliberately (commit `0bd33f4c`) because
proof roots / OOR-checkpoint ancestors can confirm *before* the target
VTXO's `CreatedHeight`, so `CreatedHeight` is not a safe floor. The
intent is correct; using genesis as the floor is what turns
"conservative" into "scan the whole chain."

## Fix

Replace the bare `const proofNodeHeightHint uint32 = 1` with a helper
`proofNodeHeightHint(currentHeight)` that floors at
`max(1, currentHeight - proofNodeHeightHintLookback)`, applied at both
registration sites (`ensureNodeConfirmed`, `watchDeferredCheckpoint`).
This bounds the historical rescan to a fixed lookback window regardless
of chain length while preserving the existing safety property (we still
do not use `CreatedHeight` as the floor).

`proofNodeHeightHintLookback` is a named, documented constant sized as a
safe multiple of a generous batch-lifetime window (no single hard max
batch lifetime exists — it is operator-configured server-side).

## Follow-up

The tighter, long-term-correct bound is the commitment-tx height (from
`Descriptor.Ancestry.CommitmentTxID`), which a proof ancestor cannot
predate. This PR ships the bounded-lookback stopgap; the commitment-height
floor can follow.

## Testing

- `TestProofNodeHeightHintBoundedBelowTip` / `TestProofNodeHeightHintClampsToGenesis`
- `make unit pkg=unroll`, `fmt-changed`, `lint-changed-local` clean
- E2E path exercised by the neutrino unilateral-exit itest
  (`itest/unroll_test.go::TestUnilateralExitManualStartSingleParentTree`),
  where the `Fetching filters for heights=[a,b]` span should now be
  bounded to the lookback window rather than spanning tip→1.

### DIFF
--- a/unroll/AGENTS.md
+++ b/unroll/AGENTS.md
@@ -149,9 +149,15 @@ For field-level detail, use `go doc github.com/lightninglabs/darepo-client/unrol
 - `watchDeferredCheckpoint(ctx, txid, node)` — registers confirmation
   watch for fraud-triggered checkpoints while the actor waits for
   operator confirmation of the proof node.
-- `proofNodeHeightHint = 1` — height hint for proof-node
-  spend/confirmation watches (proof roots and intermediate ancestors
-  can confirm before the target VTXO's creation height).
+- `proofNodeHeightHint(currentHeight)` — confirmation-watch height floor
+  for proof-graph nodes: `max(1, currentHeight - proofNodeHeightHintLookback)`
+  (10000 blocks). Proof roots and intermediate ancestors can confirm before
+  the target VTXO's creation height, so the floor is anchored to the current
+  tip minus a lookback (never `CreatedHeight`). This bounds the neutrino
+  historical rescan instead of scanning to genesis (darepo-client#884). The
+  `behavior.proofNodeConfHeightHint` wrapper warns when the VTXO's age exceeds
+  the lookback (the floor may then miss an already-confirmed ancestor).
+  Stopgap for the tighter commitment-tx-height bound.
 
 ## Relationships
 
--- a/unroll/CLAUDE.md
+++ b/unroll/CLAUDE.md
@@ -149,9 +149,15 @@ For field-level detail, use `go doc github.com/lightninglabs/darepo-client/unrol
 - `watchDeferredCheckpoint(ctx, txid, node)` — registers confirmation
   watch for fraud-triggered checkpoints while the actor waits for
   operator confirmation of the proof node.
-- `proofNodeHeightHint = 1` — height hint for proof-node
-  spend/confirmation watches (proof roots and intermediate ancestors
-  can confirm before the target VTXO's creation height).
+- `proofNodeHeightHint(currentHeight)` — confirmation-watch height floor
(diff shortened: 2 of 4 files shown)


ellemouton:
@claude review this

ellemouton:
Thanks for the thorough review. Addressed all three:

1. **`age > lookback` silent-stall observability** — added a `behavior.proofNodeConfHeightHint(ctx, txid)` wrapper around the pure `proofNodeHeightHint` helper that emits a `WarnS` breadcrumb at watch-registration when the target VTXO's age (`currentHeight - CreatedHeight`) reaches `proofNodeHeightHintLookback` — i.e. exactly when the floor could rise to/above an ancestor's real confirmation height and the exit could stall silently. Both registration sites (`ensureNodeConfirmed`, `watchDeferredCheckpoint`) route through it. The tighter commitment-tx-height floor remains the documented follow-up (kept out of this stopgap PR).
2. **Stale docs** — updated the `proofNodeHeightHint` bullet in both `unroll/CLAUDE.md` and `unroll/AGENTS.md` to describe the bounded-lookback function (and the new warn wrapper), no longer the `= 1` constant.
3. **Redundant assertion** — dropped the `GreaterOrEqual` line in `TestProofNodeHeightHintBoundedBelowTip` (entailed by the preceding `Equal`).

Folded into the single commit (fixup + autosquash), `make unit pkg=unroll` + lint green. @claude review this

ellemouton:
Combined into #927 (retargeted to `main`), which now carries the bounded-lookback logic as the fallback beneath the commitment-height floor — clearer as a single PR than a stopgap stacked under the real fix. The branch `ellemouton/884-bound-proof-conf-height` is retained; nothing is lost.

PR #913, opened 2026-07-10
ci: automate and harden doc-gardening (model pin + per-PR advisory)

Roasbeef:
In this PR, we make the doc-gardening automation more autonomous, and
more trustworthy while we're at it. Two changes, one per commit.

First, we harden the nightly sweep. It was the only Claude workflow with
no `--model`, so it silently rode the action's default (Sonnet) and
would drift again the moment the action bumps that default. The `@claude`
and code-review workflows are already pinned to Opus, so the writer
running unattended was, backwards, the least capable and least pinned of
the three. We pin it to `opus-4-8`, and fold in what the recent manual
sweep learned the hard way: the prompt now mirrors every `CLAUDE.md` to
its `AGENTS.md` as an explicit gate (the root pair is the one most often
missed, and `doc-check` fails on a byte diff), scrubs leaked tool
artifacts (stray `</content>`/`</invoke>` tags) before committing, and
forbids `go build`/`go test` so a compiled binary never gets swept into
`git add -A`.

Second, we add a per-PR advisory that catches drift in lockstep with the
change that caused it, instead of waiting up to 24h for the nightly.
When a PR touches Go code, we run the skill scoped to just the packages
whose `.go` files changed (only ones that already have a `CLAUDE.md`, so
we never create new docs here, that stays the nightly's job) and post the
proposed fix back as a single sticky suggested-diff comment via
`--edit-last`. It never commits, pushes, or opens a PR.

The advisory is advisory by design: the comment is the whole signal, and
the check exits 0 so it can't block merge. There's a marked one-line
toggle to flip it to a non-required soft-fail later if the comment alone
proves too easy to ignore. We deliberately skipped a hard "changed pkg
must update its doc" gate, since most code changes (bug fixes, refactors,
test-only) don't warrant a doc edit and the escape valve (touch the file
to bypass) just trains people to make meaningless edits.

One cost knob worth flagging: the advisory burns an Opus run on every
Go-touching PR. If that gets noisy, easy dials are a cheaper model for
the scoped diff, or a `paths-ignore` for `**/*_test.go`.

The sibling client-side PR is lightninglabs/darepo-client (same two
commits, adjusted for `main` and no submodule).

### DIFF
--- a/.github/workflows/doc-gardening-nightly.yml
+++ b/.github/workflows/doc-gardening-nightly.yml
@@ -165,9 +165,28 @@ jobs:
                makes the sweep actually look at everything that changed
                since the last merged nightly (not just "new packages").
 
-            2. **Validate.** Phase 7 runs `make doc-check`. If it fails,
-               iterate to fix the errors — never commit broken docs. If you
-               cannot make it pass after reasonable effort, print
+            2. **Mirror and scrub, then validate.** Before running
+               `make doc-check`, do two things that the check itself will
+               NOT catch on its own:
+
+               - **Mirror every CLAUDE.md to its AGENTS.md.** For each
+                 `CLAUDE.md` you created or edited, `cp` it to the sibling
+                 `AGENTS.md` so the pair is byte-identical. The root
+                 `CLAUDE.md`/`AGENTS.md` pair is the one most often missed —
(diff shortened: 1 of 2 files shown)


litbot-9000:

Roasbeef:
@claude review this

PR #920, opened 2026-07-10
multi: Persist exit funding address across restarts

### LINKED ISSUE #893: Unable to exit due to funding_shortfall, despite address holding enough confirmed funds
On Signet, with btcwallet backend.

I ran `darepocli exit plan --outpoint 18cfdcb24e0d9f23f920da9788d075149d8e608667109cc9507648caba030a16:0`

I then funded the `funding_address` with the `recommended_total_funding_sat` (`eadde734631cffd8811f50c3104e4a3ceb93ac0ee51a903938549417607bcaa9`). However, even after multiple confirmations, running `darepocli exit plan` again still showed the same funding shortfall.

After restarting `darepod` and running the command again, i was shown another funding address.

I also funded this address, and everything worked as expected.

**What went wrong the first time?**

I then restarted darepod again, and was given a new address. Despite having 10,000 sat in the wallet, I was asked to deposit another 10,000 sat to exit the same vtxo.

**Funding addresses should persist throughout startups**

ellemouton:
Fixes #893.

## Problem

`darepocli exit plan` derives a backing-wallet `funding_address` for a VTXO
and reports the amount to send there. The address was held only in an
in-memory map (`Server.exitFundingAddresses`), so every `darepod` restart
started empty and derived a **brand-new** address for the same VTXO. A user
who funded the first address and restarted was then shown a different address
and asked to deposit a second time for the same exit (#893).

## Fix

Persist the per-outpoint funding address so it is stable across restarts.

- New `exit_funding_addresses` table shipped as its own additive migration `000011_exit_funding_addresses` (only CREATEs a new standalone table, so existing deployments gain it without a schema reset). Queries added to `db/sqlc/queries/unilateral_exit.sql`; generated
  code via `make sqlc`.
- `UnilateralExitPersistenceStore.FundingAddress(ctx, target, newAddress)`:
  SELECT the persisted address; on miss, derive via the existing
  `NewWalletAddress` path (outside the SQL write tx) and
  `INSERT ... ON CONFLICT DO NOTHING` with a re-read, so a concurrent caller
  that wins the race keeps its address.
- `exitPlanFundingAddress` now goes through the store; the in-memory
  `ExitFundingAddressBook` and `Server.exitFundingAddresses` field are removed.

## The "not credited until restart" symptom (issue's second complaint)

The issue also reports that confirmed funds sent to the address were not
credited toward the shortfall until a restart. That is **not a separate
neutrino-visibility bug** — it is a downstream effect of the same cross-restart
address churn this PR fixes, and is resolved here. Verified along two axes:

- **Code:** the exit funding address is derived via `walletcore.NewAddress ->
  btcwallet.NewAddress`, which already calls `chainClient.NotifyReceived`
  (`btcwallet@v0.18.0/wallet/wallet.go:3326`) — the identical live-neutrino
  rescan registration boarding scripts use (`wallet/import.go:486`). The
  shortfall read (`ListWalletUnspent -> ListUnspentWitness`) passes
  `maxConfs=9999999` and is a pure wtxmgr read with no sync gate. The server
  "retry list unspent while syncing" patch touches only the boarding path, not
  this one.
- **Empirical:** a live regtest-bitcoind + neutrino integration test derives
  the exit funding address post-sync, deposits to it, mines confirmations, and
  observes the UTXO credited **live via `ListUnspentWitness`, no restart**
  (passes in ~7s). Pre-fix, the deposit credited live *within a single process*
  too — the only failure was cross-restart, which is exactly what persistence
  addresses.

## Testing

- `TestExitFundingAddressStableAcrossRestart`: two independent stores over the
  same DB (simulated restart) return the **same** address for one outpoint
  without re-deriving; a different outpoint gets its own address.
- `make sqlc` / `make fmt-changed` / `make lint-changed-local` (0 issues) /
  `go test ./db/ ./unroll/` / `darepod` build / `commitmsg-lint` all green.
- A live btcwbackend+neutrino integration test (opt-in, boots its own regtest
  bitcoind) confirming live crediting is available on a separate branch; happy
  to fold it in if a Docker-dependent integration test is wanted here.

### DIFF
--- a/darepod/rpc_exit_plan.go
+++ b/darepod/rpc_exit_plan.go
@@ -487,9 +487,29 @@ func (s *Server) exitPlanFundingAddress(ctx context.Context, outpoint string,
 		return "", nil
 	}
 
-	return s.exitFundingAddresses.Address(
-		ctx, outpoint, s.NewWalletAddress,
-	)
+	target, err := parseOutpointString(outpoint)
+	if err != nil {
+		return "", fmt.Errorf("parse funding outpoint: %w", err)
+	}
+
+	// The funding address is persisted per target outpoint so it is stable
+	// across daemon restarts: a restart must not derive a fresh address and
+	// demand a second deposit for the same VTXO (darepo-client#893).
+	//
+	// Reuse the daemon's already-initialized unilateral-exit store; only
+	// build one in the rare window before startup wires s.ueStore. This
+	// avoids recreating the store and its logger on every call, e.g. across
+	// a batch exit-plan request that funds several outpoints.
+	store := s.ueStore
+	if store == nil {
+		dbStore := db.NewStore(
+			s.db.DB, s.db.Queries, s.db.Backend(),
+			s.subLogger(db.Subsystem),
+		)
+		store = dbStore.NewUnilateralExitStore(s.clk)
+	}
+
+	return store.FundingAddress(ctx, target, s.NewWalletAddress)
 }
 
 // ExitSummaryEntry is one in-progress exit's coarse contribution to the
--- a/darepod/server.go
+++ b/darepod/server.go
@@ -254,11 +254,6 @@ type Server struct {
 	// handlers that require wallet access select on this channel.
 	walletReady chan struct{}
 
-	// exitFundingAddresses caches backing-wallet funding addresses returned
-	// by GetExitPlan. The unroll package owns the address-book policy so
-	// RPC handlers only request an address for a target.
-	exitFundingAddresses unroll.ExitFundingAddressBook
-
 	// daemonReady is closed when all startup steps have
 	// completed: wallet initialized, mailbox transport
 	// connected, and wallet-dependent actors started. Test
--- a/db/migrations.go
+++ b/db/migrations.go
@@ -10,7 +10,7 @@ const (
 	// daemon.
 	//
 	// NOTE: This MUST be updated when a new migration is added.
-	LatestMigrationVersion uint = 10
+	LatestMigrationVersion uint = 11
 )
 
 // MigrationTarget is a functional option that can be passed to applyMigrations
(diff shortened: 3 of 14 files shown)


ellemouton:
Addressed the Gemini review:

- **Reuse `s.ueStore` instead of rebuilding the store per call** — done. `exitPlanFundingAddress` now uses the daemon's already-initialized `s.ueStore` and only constructs a store in the rare pre-startup window before `s.ueStore` is wired, avoiding the redundant store+logger allocation on every call (including across a batch exit-plan request). Folded into the commit; build/lint/`TestExitFundingAddressStableAcrossRestart` green.

@claude review this

ellemouton:
Thanks — thorough as always. Addressed the PR-introduced point; deferring the pre-existing one with rationale.

**2. Leaked HD gap address on a lost derive race — fixed.**
This regression was introduced here by moving the derive out of the old `ExitFundingAddressBook` lock, so I restored the invariant: added a `fundingMu sync.Mutex` on `UnilateralExitPersistenceStore` and serialized the check-derive-insert in `FundingAddress`. A second concurrent miss on the same store now waits, then observes the first caller's persisted row on re-read instead of deriving (and discarding) its own address — so `NewWalletAddress` is called at most once per target and the HD index no longer advances on a poll. The derive still runs outside any SQL write tx. `go test -race ./db/` green.

**1. Unsynchronized `s.ueStore` read — deferring, tracked separately.**
Agreed it's a data race by the memory model, but as you noted it's a *pre-existing* pattern: `s.ueStore` is already read unlocked from RPC goroutines at `rpc_exit_plan.go:550` and `rpc_server.go:4779`, and written once in `initUnrollSubsystem`. A correct fix has to convert *all* of those sites together (e.g. an `atomic.Pointer` or `sync.Once` accessor) — mixing guarded and unguarded access wouldn't remove the race. That's a focused hygiene change across the `s.ueStore` lifecycle rather than something to bolt onto this persistence PR, so I'd rather do it as its own change than expand scope here. This PR adds one read that follows the existing convention and lands in the same rare pre-startup window the fallback already targets.

Nits: `int32(target.Index)` matches the existing `unilateral_exit_jobs` convention (impossible for a real vout); leaving as-is for consistency, as you suggested.

@claude review this

PR #927, opened 2026-07-10
unroll: Floor proof-node confirmation-watch at commitment height

### LINKED ISSUE #884: Excessive neutrino log activity after unilateral exit

ellemouton:
Fixes #884. Combines what were previously two stacked PRs (the bounded-lookback stopgap #911 and the commitment-height fix) into one, since the lookback is now just the fallback within the same logic. Companion server PR: lightninglabs/darepo#666.

## Problem

A unilateral exit registers a confirmation watch per proof-graph node. Using a
genesis (height 1) hint made neutrino rescan **every block from tip to genesis**
when a proof tx never confirmed — flooding logs (800 kB → 80 MB in minutes) and
never terminating, since the v3/TRUC exit tx is rejected by signet peers
(darepo-client#884).

## Fix

Floor the hint at `min(commitment_height)` across the VTXO's ancestry fragments.
Nothing in a proof graph can confirm before its commitment tx, so the
commitment's on-chain confirmation height is the **tightest sound lower bound**.

- **Proto:** new `commitment_height` on `arkrpc.AncestryPath`.
- **Ingest:** `CommitmentHeight` on `lib/types.Ancestry`, populated at the single
  `AncestryFromRPC` choke-point.
- **DB:** persisted on `vtxo_ancestry_paths` via a **forward `ALTER ... ADD
  COLUMN` migration (`000012`)** — existing deployments gain the column without
  a schema reset (`LatestMigrationVersion` → 12).
- **unroll:** `proofNodeConfHeightHint` floors at `min(commitmentHeight)`.

The server supplies the height (already resolved as `RoundRow.ConfirmationHeight`,
previously discarded before the wire) in darepo#666.

## Fallback (why this is safe to merge before the server PR)

When no fragment carries a commitment height — legacy rows (column defaults to
0), empty-ancestry incoming VTXOs, or any producer that didn't set it — the hint
falls back to a **bounded lookback below the current tip**
(`proofNodeHeightHintLookback`, 10k blocks). That caps the rescan window without
depending on chain data. Until the server populates the field, behaviour is
exactly this lookback (the former #911 stopgap). The age-exceeds-lookback `WarnS`
now fires only on the fallback path.

## Testing

- `TestMinCommitmentHeight`, `TestProofNodeHeightHintUsesCommitmentHeight`,
  `TestProofNodeHeightHintFallsBackWithoutCommitmentHeight`, plus the existing
  bounded-lookback tests.
- `make rpc` / `sqlc` / `fmt-changed` / `lint-changed-local` (0 issues) /
  `unit pkg=unroll db vtxo lib/types` / `check-migration-version` /
  `commitmsg-lint`: all green.

### DIFF
--- a/arkrpc/indexer.pb.go
+++ b/arkrpc/indexer.pb.go
@@ -1493,9 +1493,13 @@ type AncestryPath struct {
 	// tree_depth is the depth of the served leaf within this fragment's
 	// tree. Worst-case unilateral-exit timing for the produced VTXO is
 	// max(tree_depth) across all AncestryPaths.
-	TreeDepth     uint32 `protobuf:"varint,4,opt,name=tree_depth,json=treeDepth,proto3" json:"tree_depth,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	TreeDepth uint32 `protobuf:"varint,4,opt,name=tree_depth,json=treeDepth,proto3" json:"tree_depth,omitempty"`
+	// commitment_height is the on-chain confirmation height of the commitment
+	// tx anchoring this fragment. Zero means unknown (legacy/unconfirmed),
+	// in which case the client falls back to a bounded lookback floor.
+	CommitmentHeight int32 `protobuf:"varint,5,opt,name=commitment_height,json=commitmentHeight,proto3" json:"commitment_height,omitempty"`
+	unknownFields    protoimpl.UnknownFields
+	sizeCache        protoimpl.SizeCache
 }
 
 func (x *AncestryPath) Reset() {
@@ -1552,6 +1556,13 @@ func (x *AncestryPath) GetInputIndices() []uint32 {
 func (x *AncestryPath) GetTreeDepth() uint32 {
 	if x != nil {
 		return x.TreeDepth
+	}
+	return 0
+}
+
+func (x *AncestryPath) GetCommitmentHeight() int32 {
+	if x != nil {
+		return x.CommitmentHeight
 	}
 	return 0
 }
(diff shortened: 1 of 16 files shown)


ellemouton:
@claude review this

ellemouton:
Addressed the two soundness edge cases from the review:

**1. Mixed known/unknown fragments** — fixed. Renamed `minCommitmentHeight` → `commitmentHeightFloor` and made it **all-or-nothing**: it returns the min only when *every* fragment carries a known (>0) height, and returns 0 (defer to the lookback fallback) if *any* fragment is unknown. A min over only the known fragments wasn't a sound bound for an unknown fragment's ancestor — now the floor is provable for all fragments or we don't use it. Test updated (the former "zeros ignored → 600" case is now "any unknown → 0").

**2. Fallback + uninitialized height → genesis floor** — documented. Added a comment in `proofNodeConfHeightHint` noting the invariant: proof watches are only registered while handling Start/Resume/HeightUpdated, all of which set `b.pending.Height` first, so `currentHeightHint()` is non-zero on the real path and the `proofNodeHeightHint(0)==1` genesis case isn't reachable in practice.

**Nit** — added a one-liner that the floor is per-target (global), so `txid` is only used for the fallback breadcrumb.

Re the Gemini migration comment: as you noted it's stale — the column is a forward migration `000012`, `000004` is untouched.

`unit pkg=unroll` / `lint-changed-local` (0) / `commitmsg-lint`: green. @claude review this

ellemouton:
Fixed the doc drift — the only outstanding item from your re-review. `unroll/CLAUDE.md` was stale (my earlier edit got reverted before commit); both `unroll/CLAUDE.md` and `unroll/AGENTS.md` now carry the identical `commitmentHeightFloor()` all-or-nothing description, and the 80-col run-on line is rewrapped. Verified the committed files are byte-identical.

(Noted separately: `swapwallet/CLAUDE.md`/`AGENTS.md` have a pre-existing divergence on `main`, unrelated to this PR and not touched here.)

Issue #923, opened 2026-07-10
swapwallet: FAILED activity rows have empty failure_reason on expiry


darioAnongba:
## Symptom

An in-swap RECV that expires flips its `activity` (canonical store-backed) row to `ENTRY_STATUS_FAILED` but leaves `failure_reason` empty. The user cannot tell from the feed why the entry failed. The machine-readable `failure_code` is correctly set to `ENTRY_FAILURE_CODE_EXPIRED`, but the human-readable supplement is blank.

## Repro (signet, confirmed live)

1. `recv --amt <sat>` to create a receive invoice.
2. Let the invoice expire without paying it; the backing swap reaches `SWAP_STATE_EXPIRED`.
3. `list` the activity feed: the RECV row shows `status=FAILED`, `failure_code=EXPIRED`, but `failure_reason=""`.

## Expected vs actual

- Expected: a terminal-failure row carries a clear reason, e.g. `failure_reason="invoice expired"`.
- Actual: `failure_reason=""` for expired/refunded/generic-failed terminal states whenever the swap FSM did not attach a reason string.

## Root cause

`swapwallet/normalize.go` `swapEntryFromSummary` sets `FailureReason` solely from `failureReasonFromTerminal(s.GetTerminalReason())`. For an expired swap the terminal reason is empty: `sdk/swaps/out_swap.go` `ReceiveSession.TerminalReason()` returns `interventionReason`, which is only populated for `NeedsIntervention`, never for `Expired`. The wallet layer already classifies the failure into `EntryFailureCode` (`failureCodeFromSwapState`) but has no code->reason fallback, so the reason is left blank even though the code is set. Both the live monitor path (`monitor.go`) and the derive/backfill read path (`history.go`) build rows through `swapEntryFromSummary`, so both surfaces show the gap.

## Proposed fix

In `swapEntryFromSummary`, classify the failure once and, when the FSM supplied no terminal reason, fall back to a stable default keyed on the failure code: EXPIRED->"invoice expired", REFUNDED->"refunded", NEEDS_INTERVENTION->"needs intervention", FAILED->"failed" (and TIMED_OUT->"timed_out" to match the deadline overlay). Non-failure rows (unspecified code, blank terminal reason) stay empty, preserving the `failure_reason != ""` means-failed convention. Reusing the existing `EntryFailureCode` taxonomy keeps reason and code consistent.

[...]

Issue #922, opened 2026-07-10
swapwallet: pending in-swap RECV missing from activity feed until terminal


darioAnongba:
## Symptom

A pending in-swap RECV does not appear in the `activity` feed. It surfaces in `swap show`/`swap list`, and `recv` returns a PENDING `ENTRY_KIND_RECV` entry (id == payment_hash), but `darepocli activity` and `activity --pending --kind recv` show nothing until the swap goes terminal. On signet the row only appeared once the swap moved EXPIRED -> FAILED.

## Repro (signet)

1. `darepocli recv --amt 50000` -> returns a PENDING RECV entry (id == payment_hash).
2. `darepocli swap show <id>` / `swap list` -> track it fine.
3. `darepocli activity` / `activity --pending --kind recv` -> EMPTY.
4. Feed only populates once the swap reaches a terminal state.

## Expected vs actual

- Expected: the pending RECV row is visible in the store-backed activity feed the instant `recv` returns, exactly like a pending SEND is.
- Actual: the RECV row reaches the store only when the async swap monitor projects a state transition; until then the feed is empty for that receive.

## Root cause

`receiver.Recv` (swapwallet/recv.go) dispatches `SwapService.StartReceive`, builds the pending `WalletEntry` via `swapEntryFromSummary(...)`, and returns it — but never writes it to the canonical activity store. The only writer for that row is the swap monitor's async path: monitor.go `fanOutSwapUpdate` -> `projectAndEmit`, driven off `SubscribeSwaps`. So the store-backed read path (`activity`) has no RECV row until the monitor fans the first transition.

The SEND path does not have this gap: router.go `sendInvoiceIntent` eagerly calls `r.runtime.project(context.WithoutCancel(ctx), entry)` on dispatch (store-only, since the monitor owns the live emit). The onchain-send and credit-send paths project on dispatch too. RECV is the only in-swap dispatch that skips the eager store write.

## Proposed fix

Mirror the SEND path: after `StartReceive` succeeds and the entry is built, call `r.runtime.project(context.WithoutCancel(ctx), entry)` (store-only `project`, NOT `projectAndEmit`, because the swap monitor owns the swap-backed row's live emit — emitting here too would double-fan). Off the RPC context so a client disconnect cannot cancel the write of an already-accepted receive.

The sub-dust credit-receive branch (`recvCredit`) returns earlier and is projected by the credit projector poll loop, so the eager write placed after `StartReceive` does not affect it and does not double-project.

[...]

PR #928, opened 2026-07-10
darepocli: accept snake_case --session_id on ark oor get

### LINKED ISSUE #900: darepocli: 'ark oor get' rejects --session_id (JSON/logs use session_id)
Split out from #613 (sub-defect #3).

`ark oor get` registers only `--session-id` (kebab; `cmd/darepocli/darepoclicommands/cmd_oor.go`) and there is no global flag normalization, so `--session_id` errors `unknown flag` — even though `ark oor list`'s JSON output and the daemon logs name the field `session_id` (proto `session_id`). The schema registry (`schema_registry_ark_observable.go`) also uses the kebab spelling.

**Fix (pick one):** register a `--session_id` alias on the flag, or set a global underscore→dash `SetGlobalNormalizationFunc` in `NewRootCmd` (`root.go`) so snake spellings are accepted everywhere, and align the schema registry.

ellemouton:
Fixes #900.

`ark oor get` registered only the kebab flag `--session-id`, so passing
`--session_id` errored with "unknown flag" — even though the RPC field,
`ark oor list` JSON output, and the daemon logs all name it `session_id`. A
user copying that field name straight from the output hit a dead end.

Install a flag-name normalization on the command that folds snake_case onto
the canonical kebab spelling, so `--session_id` and `--session-id` resolve to
the same flag. Scoped to this command, whose flags (and the inherited global
flags) are all defined in kebab case, so it only adds an alias and never
renames an underscore-defined flag. The schema registry keeps its existing
kebab spelling.

## Testing

- Added `TestOORGetAcceptsSnakeCaseSessionID` (both spellings resolve to the
  same value).
- `darepoclicommands` package passes; `fmt-changed`, `lint-changed-local`, and
  `commitmsg-lint` clean.

---

Note: this PR originally also bundled fixes for #922 and #923, but those are
already addressed by @darioAnongba's #925 with an equivalent approach, so those
two commits were dropped here. This PR now covers only #900.

### DIFF
--- a/cmd/darepocli/darepoclicommands/cmd_oor.go
+++ b/cmd/darepocli/darepoclicommands/cmd_oor.go
@@ -3,9 +3,11 @@ package darepoclicommands
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/lightninglabs/darepo-client/daemonrpc"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"google.golang.org/protobuf/proto"
 )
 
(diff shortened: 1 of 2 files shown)


ellemouton:
@claude review

Issue #865, opened 2026-07-06
activity: internal OOR send legs leak into activity log when swap correlation fails


litbot-9000:
## Description

Internal OOR (out-of-round) VTXO movements from swap operations (Lightning receive/send) can appear as standalone SEND entries in the activity log when the OOR session ID fails to correlate with a known swap summary.

## Reproduction

1. Board sats onchain
2. Receive sats over Lightning
3. List activity — an unexpected SEND entry appears with a `ledger-N` ID

Example output:

```
LAST UPDATE              KIND     STATUS    AMOUNT  FEE  PHASE      ID         NOTE
2026-07-06 20:08:00.000  SEND     COMPLETE  -1000   0    confirmed  ledger-3
2026-07-06 20:07:59.000  SEND     PENDING   500     0    settling   e4283a...
2026-07-06 18:32:22.000  RECV     PENDING   3000    0    settling   b11868...  Memo
2026-07-06 18:32:05.000  DEPOSIT  COMPLETE  50000   0    confirmed  30476a...:0
```

The `ledger-3` entry is the OOR send leg of the Lightning receive's internal plumbing — not a user-initiated payment.

## Root cause

`projectOORLedgerActivity` in `swapwallet/history.go` hides internal OOR legs only when the session ID matches a known swap's `FundingSessionId`, `ClaimSessionId`, or `RefundSessionId`. When the swap summary is missing or session IDs don't match, the raw ledger row surfaces as a visible `ledger-N` SEND entry.

The fallback to `ledger-N` ID format happens in `walletEntryFromLedgerRow` when `ChainTxid` is NULL (because `handleVTXOSent` doesn't set one).

## Existing test coverage

`TestHistoryKeepsZeroDeltaOORSendWithoutSwap` in `swapwallet/history_test.go` documents this exact pattern but does not treat it as a defect.

## Expected behavior

Internal OOR legs that are part of a swap operation should always be hidden from the user-facing activity log, regardless of whether the swap summary is available at query time.

bhandras:
Confirmed on current `main` (`42de79d8e422aac64d748177ae8015777866e303`) with the session-keyed receive-claim shape reported here.

The concrete path is narrower than the original “missing swap summary” hypothesis:

1. A completed receive swap claims its vHTLC through an outgoing OOR session.
2. `oor.sessionBehavior.queueVTXOSent` records the consumed vHTLC as `vtxo_sent`, keyed by the OOR session and without a chain txid.
[...]

PR #929, opened 2026-07-10
swapwallet: reconcile raw-OOR SEND/RECV activity live

### LINKED ISSUE #903: swapwallet: raw ark-oor SEND/RECV activity completes only at restart, not live
Split from #613 (its remaining residual after sub-defects #1/#2 were fixed by #595 and #613 was closed).

Raw out-of-round sends/receives (`ark send oor` / `ark oor receive`) are neither swap-backed nor credit-backed, so they have no live activity projector: the swap monitor and credit poll own SEND/RECV, and the C2 reconciler deliberately excludes SEND/RECV (`reconcilerKinds = {DEPOSIT, EXIT}`, `swapwallet/reconciler.go`). Their activity row therefore lands its PENDING→COMPLETE transition into the canonical store only at the next startup backfill (`reprojectActivity(nil)` in `resumeAll`), not live during an uninterrupted run.

Note: the everyday walletdk `send --offchain` path (Lightning via StartPay, or credit) IS live-projected; this affects the raw `ark oor` daemonrpc/CLI path specifically.

Options: add a raw-OOR-aware live projector; include the OOR kinds in the reconciler's re-derive (carefully, to avoid contending with the monitor on swap-backed SEND/RECV); or document it as a known raw-CLI limitation.

Out of scope of the (now-complete) canonical-activity-log epic #776, whose four children C1–C5 are done.

ellemouton:
Fixes #903.

A raw out-of-round send/receive (`ark send oor` / `ark oor receive`) is neither
swap-backed nor credit-backed, so it has no live activity projector: the swap
monitor and credit poll own SEND/RECV, and the reconciler deliberately excluded
those kinds (`reconcilerKinds = {DEPOSIT, EXIT}`). Its ledger-derived activity
row therefore flipped PENDING → COMPLETE in the canonical store only at the next
startup backfill (`reprojectActivity(nil)` in `resumeAll`), not live during an
uninterrupted run — so `activity` showed nothing for a raw OOR until restart.

## Fix

Add `SEND` and `RECV` to the reconciler's re-derive kinds so a raw-OOR row lands
within one reconcile pass (60s cadence).

### Why re-deriving the swap/credit-backed SEND/RECV rows too is safe

This is the subtlety the issue flagged ("carefully, to avoid contending with the
monitor"):

- **No double-emit / no wasted writes.** `ProjectEntry` change-suppression makes
  an unchanged row a no-op — no duplicate transition event, nothing fanned to
  subscribers. So the reconciler never re-emits a row the monitor already
  advanced; the swap DB is the shared source of truth both read from, so the
  derived state converges rather than regresses.
- **Authoritative, unlike the credit poll.** The credit poll must never emit for
  a mixed pay because it sees only the credit leg. The reconciler instead
  projects the fully-merged derived row — the same one `List` and the startup
  backfill produce — so any transition it does land is the authoritative one.

The everyday `send`/`recv --offchain` (Lightning/credit) path is unaffected: it
is already live-projected by the monitor; this only closes the gap for the raw
`ark oor` daemonrpc/CLI path.

## Testing

- Added `TestReconcileProjectsRawOORLive`: a recorded raw-OOR send + receive
  (no swap correlation) land as COMPLETE in the store after one reconcile pass.
- Full `swapwallet` package passes; `fmt-changed`, `lint-changed-local`, and
  `commitmsg-lint` clean.

### DIFF
--- a/swapwallet/projector.go
+++ b/swapwallet/projector.go
@@ -102,6 +102,26 @@ func (r *Runtime) projectEmitLocked(ctx context.Context,
 	return seq, err
 }
 
+// projectDerivedPage re-projects one page of derived entries into the canonical
+// store, shared by the full-history (reprojectActivity) and bounded-window
+// (reprojectRecentActivity) reconcile passes. projectEmitLocked serializes the
+// project+emit so a reconciled transition (a confirmed deposit, a forfeited
+// leave) reaches live subscribers in seq order; a change-suppressed no-op fans
+// out nothing. A per-entry failure is best-effort — skipped and retried on the
+// next pass — and the pending record is cleared only after a durable project,
+// so a partial page never strands or corrupts a row.
+func (r *Runtime) projectDerivedPage(ctx context.Context,
+	entries []*walletdkrpc.WalletEntry) {
+
+	for _, entry := range entries {
+		if _, err := r.projectEmitLocked(ctx, entry); err != nil {
+			continue
+		}
+
+		r.clearProjectedTerminalExit(entry)
+	}
+}
+
 // reprojectActivity pages the derive-on-read feed and projects every row into
 // the canonical activity store, returning the number of rows projected. Because
 // ProjectEntry is idempotent on canonical_id and suppresses no-op changes,
@@ -111,14 +131,15 @@ func (r *Runtime) projectEmitLocked(ctx context.Context,
 // reconciler. Returns (0, nil) when no store is wired.
 //
 // kinds optionally restricts the derived feed: the startup backfill passes nil
-// to seed every kind, while the reconciler passes only the backfill-only
-// producers (DEPOSIT, EXIT) so it neither re-projects nor contends with the
-// monitor-owned SEND/RECV rows.
+// to seed every kind, while the reconciler passes only the low-volume
+// DEPOSIT/EXIT producers here and reconciles the high-volume SEND/RECV kinds
+// separately over a bounded window (reprojectRecentActivity).
 //
-// A wallet-local EXIT's pending record is cleared only here, and only after its
-// terminal row is durably projected — never from the read/derive path — so a
-// failed projection or a row that paged out is retried on a later pass instead
-// of being stranded PENDING in the store.
+// A wallet-local EXIT's pending record is cleared only from here and
+// reprojectRecentActivity (via projectDerivedPage), and only after its terminal
+// row is durably projected — never from the read/derive path — so a failed
+// projection or a row that paged out is retried on a later pass instead of
+// being stranded PENDING in the store.
 func (r *Runtime) reprojectActivity(ctx context.Context,
 	kinds []walletdkrpc.EntryKind) (int, error) {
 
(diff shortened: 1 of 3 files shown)


ellemouton:
Thanks @gemini-code-assist — the O(N)-per-tick concern is valid and I've addressed it in the latest push.

**On the specific suggestions:**
- *Pending-only reconciliation* doesn't fit this bug: a raw OOR has no live projector, so its row is **never written to the store as PENDING during a run** in the first place — there's nothing pending to query. It would also miss a raw OOR created *and* completed within a single run (exactly the #903 case).
- *Track in a pending map*: raw OOR is created in `darepod` (`SendOOR`/`NewReceiveScript`), which `swapwallet` never observes, so there's no creation hook to seed a pending map — and the ledger row lands directly as `recorded`/COMPLETE, so there's no pending phase to track anyway.
- *Time-windowed scan*: this is the one that fits, and it's what I implemented.

**Change:** `DEPOSIT`/`EXIT` stay on the full-history pass (rare → cheap). `SEND`/`RECV` are now reconciled via a new `reprojectRecentActivity` that re-derives and projects only the most recent `rawOORReconcileWindow` (100) rows. This bounds the per-tick `ProjectEntry` DB-read work at **O(window)** instead of O(N). A raw OOR settles immediately and sorts to the top by `updated_at`, so the window reliably catches it; older SEND/RECV rows are terminal and immutable, and a restart's full backfill is the backstop. Added `TestReprojectRecentActivityBoundsWindow` to lock the bounding in.

One honest residual: the *derive read* still does a full `ListSwaps` scan per pass (needed for the swap↔OOR correlations that prevent internal-leg leaks, and `ListSwaps` has no limit param today). Fully bounding that needs a `ListSwaps` limit / incremental cursor — noted as a follow-up rather than folded into this fix.

ellemouton:
@claude review this

ellemouton:
Thanks @claude — great trace, and confirming the recency/ordering premise end-to-end is exactly the part I most wanted a second pair of eyes on. Addressed the notes in the latest push:

- **RECV/SEND coupling (latent trap):** added a comment on `rawOORReconcileKinds` noting that SEND must stay paired with RECV, since `deriveActivity` gates ledger collection on `DEPOSIT || EXIT || SEND` (not RECV) — reconciling RECV alone would silently derive nothing.
- **Duplication:** extracted the shared per-page projection loop into `projectDerivedPage`, now used by both `reprojectActivity` and `reprojectRecentActivity`.
- **`ListSwaps` O(N) residual:** filed as a tracking follow-up — #931 — since, as you note, it carries the same unbounded-growth shape one layer down. Left out of this PR to keep it scoped (bounding it correctly needs a `ListSwaps` limit/cursor while preserving the #865 correlation set).
- **Redundant ledger read per tick:** acknowledged; acceptable given the kind-filter split between the full and bounded passes.

Tests + fmt + lint green; commit amended.

ellemouton:
Thanks @claude — good catch, that's a real regression from the extraction. Fixed: the stranded `reprojectActivity` doc is moved back directly above `func reprojectActivity` (and refreshed so its `kinds` paragraph reflects the new DEPOSIT/EXIT-here + bounded-SEND/RECV split), leaving only the `projectDerivedPage` doc above `projectDerivedPage`. Every function again has a comment starting with its own name. Logic unchanged; tests + fmt + lint green; amended.

PR #932, opened 2026-07-10
swaps: cancel recovery when refund spend seen before index

### LINKED ISSUE #874: Flaky test: TestPaySessionRefundsWhenRefundLocktimePassesBeforeClaim under unit-cover
## Summary

`TestPaySessionRefundsWhenRefundLocktimePassesBeforeClaim`
(`sdk/swaps`) is flaky under the `unit-cover` CI job. It fails
intermittently while passing in every other unit job of the same run
(`unit-race`, `unit tags="test_sqlite"`, `unit tags="test_postgres"`,
`Compile + test (host)`).

## Observed failure

```
--- FAIL: TestPaySessionRefundsWhenRefundLocktimePassesBeforeClaim (1.36s)
    in_swap_test.go:1803:
        Error:      Not equal:
                    expected: 1
                    actual  : 0
        Test:       TestPaySessionRefundsWhenRefundLocktimePassesBeforeClaim
FAIL    github.com/lightninglabs/darepo-client/sdk/swaps
```

The failing assertion is `require.Equal(t, 1, daemonConn.cancelCalls)`
(`sdk/swaps/in_swap_test.go:1803`): the expected `cancel` call on the
daemon connection had not yet been recorded when the assertion ran. This
is an async-ordering/timing race in the FSM-driven refund path, not a
correctness regression.

First seen on: https://github.com/lightninglabs/darepo-client/actions/runs/28884469821/job/85681024155
(the change under test only touched `swapwallet/`; the `sdk/swaps` code
was identical to `main`).

## Why it looks like a `-cover` timing flake

- Passed in the 4 other unit jobs of the same CI run; only `unit-cover`
  failed.
- Passes locally 10x under `-cover` and 5x plain
  (`go test ./sdk/swaps/ -run TestPaySessionRefundsWhenRefundLocktimePassesBeforeClaim -cover -count=10`).
- The `-cover` build's added instrumentation shifts goroutine timing,
  and this test races a refund locktime against a claim.
- This test family has prior stabilization history: commit `7ea81252`
  ("test: Stabilize postgres and refund resume tests").

## Suggested fix direction

Make the assertion on `daemonConn.cancelCalls` (and any peer
`sendCustomCalls` / `armRecoveryCalls` counters it races with)
synchronize on the FSM reaching its terminal/refunded state — e.g. wait
for the state transition / `require.Eventually` on the counter — rather
than reading the counter immediately after driving the event.

ellemouton:
Fixes #874.

## What was flaky

`TestPaySessionRefundsWhenRefundLocktimePassesBeforeClaim` (`sdk/swaps`)
intermittently failed the final `require.Equal(t, 1, daemonConn.cancelCalls)`
(saw `0`), only in the `unit-cover` job.

## Root cause — not a timing race

The FSM refund path is fully synchronous, so the `-cover` async-ordering
theory in the issue doesn't hold. The real cause is a per-key data bug in the
test fixtures:

The mocks built the refund destination x-only key with
`clientPriv.PubKey().X().Bytes()`. `PublicKey.X()` returns a `*big.Int`, and
`big.Int.Bytes()` **strips leading zero bytes**, so ~1/256 of randomly
generated keys produced a `<32`-byte slice (measured ≈ 19/5000). The daemon
mock then failed `schnorr.ParsePubKey` on that slice and never indexed the
refund output, so the refund completed through `markRefunded` (spent vHTLC
observed) instead of `markRefundOutputIndexed`. Each CI job draws a fresh
random key, so the bad draw happened to land in `unit-cover` that run and
passed everywhere else — nothing to do with coverage instrumentation.

## The real bug this surfaced

`markRefunded`, unlike its siblings `markRefundOutputIndexed` and
`markRefundSessionCompleted`, transitioned to `Refunded` **without cancelling
the recovery** armed at funding time. This is a genuine production bug: the
refund can legitimately complete via the spent-vHTLC observation when the
refund output has not been indexed yet, leaving a **dangling armed recovery
row** behind. This PR cancels it there too.

## Changes

- `in_swap.go` / `recovery.go`: `markRefunded` cancels the armed recovery
  before going terminal (new `recoveryReasonRefundSpendObserved`).
- Fixtures: build x-only keys with `schnorr.SerializePubKey` (always
  zero-padded to 32 bytes) so the tests are deterministic — all 23
  `.X().Bytes()` sites replaced; none remain in the repo.
- New regression test
  `TestPaySessionRefundCancelsRecoveryWhenSpendObservedBeforeIndex` forces the
  spend-before-index path (new `skipLiveOnCustom` mock knob) and asserts the
  recovery is cancelled. Verified it fails without the production fix
  (`expected 1, actual 0`) and passes with it.

## Verification

Full `sdk/swaps` suite passes; refund tests pass 10× under `-cover`;
`go build`/`vet` clean; `make lint-changed-local` reports 0 issues.

### DIFF
--- a/sdk/swaps/in_swap.go
+++ b/sdk/swaps/in_swap.go
@@ -1616,6 +1616,25 @@ func (s *paySession) markRefunded(ctx context.Context,
 		slog.String("spender", spentVHTLC.SpentByTxID),
 	)
 
+	// The refund spend is on-chain, so the daemon-owned recovery armed at
+	// funding time is no longer needed. Cancel it before recording the
+	// terminal state, mirroring markRefundOutputIndexed and
+	// markRefundSessionCompleted so a refund that completes via the spent
+	// vHTLC observation (e.g. when the refund output has not been indexed
+	// yet) does not leave a dangling armed recovery row behind.
+	//
+	// This is the unilateral refund-without-receiver (timeout) path, so the
+	// spend is not a cooperative settlement: pass an empty cooperative txid
+	// like every other cancel call site. A non-empty value would have to be
+	// a canonical txid or the daemon rejects the cancel, which our
+	// retryable wrapping would then retry forever.
+	if err := cancelVHTLCRecovery(
+		ctx, s.client.daemon, s.refundRecoveryID,
+		recoveryReasonRefundSpendObserved, "",
+	); err != nil {
+		return newRetryableActionError(err)
+	}
+
 	return s.mutateAndPersist(ctx, func() error {
 		if spentVHTLC.Outpoint != "" {
 			s.vhtlcOutpoint = spentVHTLC.Outpoint
--- a/sdk/swaps/in_swap_test.go
+++ b/sdk/swaps/in_swap_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/ecdsa"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil/v2"
 	"github.com/btcsuite/btcd/chaincfg/v2"
 	"github.com/btcsuite/btcd/psbt/v2"
@@ -762,8 +763,10 @@ func TestPaySessionRefundsFundedVHTLCOnTimeout(t *testing.T) {
 			AmountSat: testInSwapAmountSat,
 		},
 		receiveInfo: &ReceiveInfo{
-			PubKeyXOnly: clientPriv.PubKey().X().Bytes(),
-			PkScript:    refundScript,
+			PubKeyXOnly: schnorr.SerializePubKey(
+				clientPriv.PubKey(),
+			),
+			PkScript: refundScript,
 		},
 		spendOnCustom: true,
 	}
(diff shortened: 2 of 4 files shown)


ellemouton:
@claude review this

ellemouton:
Addressed the substantive review point (thanks @claude): `markRefunded`
originally passed `spentVHTLC.SpentByTxID` as the `cooperative_txid`, which
was the only one of the 11 `cancelVHTLCRecovery` call sites to pass a
non-empty value. Two problems with that:

- **Semantics:** this is the unilateral refund-without-receiver (timeout)
  path, so the spend isn't a cooperative settlement — the field is documented
  as "the cooperative transaction that won."
- **Robustness:** the daemon runs `cooperative_txid` through
  `chainhash.NewHashFromStr`, which rejects anything that isn't a canonical
  txid with `InvalidArgument`. Our `newRetryableActionError` wrapping would
  then retry that cancel forever, stranding the session in `RefundInitiated`.
  `txidBytesFromString("")` returns `(nil, nil)`, which is exactly why the
  other 10 callers pass `""`.

Now passes `""` to match every sibling, with a comment explaining why, and the
regression test asserts `GetCooperativeTxid()` is empty.

The other two notes I left as-is: the `markRefundSessionCompleted` error-wrap
inconsistency is pre-existing and out of scope, and the fixture comment is
already covered by the commit message.

PR #933, opened 2026-07-10
swapwallet: surface settlement txid/height on completed leave EXIT

### LINKED ISSUE #924: walletdk: completed cooperative-leave EXIT missing settling txid/height
## Symptom
A COMPLETED cooperative-leave / on-chain-send EXIT activity entry shows the right `vtxo_outpoint` (the consumed input) and phase `confirmed`, but `progress.txid == ""` and `confirmation_height == 0`. A user cannot reconcile the settled leave against the chain even though the leave round's commitment tx confirmed on-chain.

## Repro (signet, confirmed live)
```
darepocli send <addr> --onchain --amt 100000     # or: exit --outpoint <op>
# wait for the leave round to confirm
darepocli activity --kind exit
# -> status COMPLETE, phase confirmed, vtxo_outpoint set,
#    but progress.txid == "" and confirmation_height == 0
```

## Expected vs actual
Expected: the completed EXIT carries the leave round's on-chain commitment txid and its confirmation height (like the unilateral-exit path carries the sweep txid). Actual: both are empty/zero.

## Root cause
The completed row is the runtime-local pending leave row (keyed by `send_job_id`, created in `router.go sendOnchainIntent`), flipped to COMPLETE on the derive/reconcile path when its consumed VTXO shows up FORFEITED:
- `swapwallet/history.go applyCooperativeLeaveForfeited` sets status/phase only; it never sets `Progress.Txid`/`ConfirmationHeight`.
- `swapwallet/history.go collectForfeitedVTXOOutpoints` collapses each forfeited VTXO to a bare outpoint set, discarding all other fields.

More fundamentally, the settling txid is not reachable over any current RPC:
- `ListVTXOs`/`descriptorToProto` exposes only the VTXO's origin `commitment_txid` (the round that CREATED it, not the leave round). `spent_by_txid` is OOR-only and unset for round forfeits.
- The leave's ledger row (`EventVTXOSent`, `handleVTXOSent`) records no `chain_txid`/`confirmation_height`, classifies as `type=round`, and is dropped by `classifyLedgerRow`.

The daemon does have the data: on ForfeitConfirmed (`round/actor.go`) it holds the new commitment txid + block height; `rounds` stores `commitment_txid`+`confirmation_height` and `vtxos` stores `forfeit_round_id`. It is simply never surfaced.

## Proposed fix
Daemon: add `settlement_txid` (int32 `settlement_height`) to `daemonrpc.VTXO`, populated for FORFEITED VTXOs from `rounds.commitment_txid`/`confirmation_height` joined via `vtxos.forfeit_round_id` in the light by-status list query. Client: thread those through `collectForfeitedVTXOSettlements` -> `decorateCooperativeLeaveEntry` -> `applyCooperativeLeaveForfeited` onto `progress.txid`/`confirmation_height`. Empty values leave today's behavior unchanged (still flips to COMPLETE, just no txid), so old daemons don't regress. No walletdkrpc proto change — `WalletEntryProgress` already carries and persists both fields.

ellemouton:
## Problem

A COMPLETED cooperative-leave / on-chain-send EXIT activity row showed the
right `vtxo_outpoint` and phase `confirmed`, but `progress.txid == ""` and
`confirmation_height == 0`. A user could not reconcile a settled leave against
the chain, even though the leave round's commitment tx confirmed. (The
unilateral-exit path already carries its sweep txid.)

The completed row is the runtime-local pending leave row flipped to COMPLETE
when its consumed VTXO shows up FORFEITED. That flip set status/phase only, and
the settling txid — the commitment tx of the round that *forfeited* the VTXO —
was not reachable over any RPC: `VTXO.commitment_txid` names the round that
*created* the VTXO, and `spent_by_txid` is OOR-only.

## Fix

The daemon already has the data: a forfeited VTXO records `forfeit_round_id`,
and `rounds` records that round's `commitment_txid` + `confirmation_height`.

1. **daemonrpc** — add `settlement_txid` (14) + `settlement_height` (15) to
   `message VTXO`, set only for FORFEITED VTXOs.
2. **db** — `ListVTXOsByStatus` LEFT JOINs `rounds` on `forfeit_round_id` (one
   query, no N+1) and carries the settling txid/height onto
   `vtxo.Descriptor`. The join columns are NULL for every VTXO whose forfeit
   round is unknown, so the descriptor keeps zero values there.
3. **darepod** — `descriptorToProto` emits the fields, guarded on a non-zero
   hash so non-forfeited/old-daemon VTXOs stay empty.
4. **swapwallet** — thread the settlement through
   `collectForfeitedVTXOSettlements → decorateExitEntry →
   decorateCooperativeLeaveEntry → applyCooperativeLeaveForfeited`, stamping
   `progress.txid` / `confirmation_height` when the txid is non-empty.

## Old-daemon parity

Every new field degrades to empty/zero. An empty txid still flips the row to
COMPLETE, just without on-chain coordinates — asserted in tests — so an old
daemon that does not send the fields does not regress.

## Tests

- **swapwallet**: `TestHistoryStampsSettlementOnForfeitedLeave` +
  empty-fields parity assertion on the existing forfeit test.
- **darepod**: `TestDescriptorToProtoSettlement` (forfeited / live /
  old-daemon).
- **db**: `TestVTXOPersistenceStoreListVTXOsByStatusSettlement` (join surfaces
  the settling round's txid+height; light path too; live VTXO stays zero).

`make fmt-changed`, `make lint-changed-local`, and `make commitmsg-lint` pass;
full db / darepod / swapwallet suites green.

Fixes #924

### DIFF
--- a/daemonrpc/daemon.proto
+++ b/daemonrpc/daemon.proto
@@ -708,6 +708,21 @@ message VTXO {
     // expiry_info reports the wallet/VTXO layer's current expiry posture for
     // this VTXO when the daemon has enough timing context to classify it.
     VTXOExpiryInfo expiry_info = 13;
+
+    // settlement reports where this VTXO was forfeited on-chain (the round
+    // commitment tx of the leave/cooperative-forfeit round). It is present
+    // only for FORFEITED VTXOs whose forfeit round is known, and unset
+    // otherwise, so absence is explicit rather than a zero-value sentinel.
+    VTXOSettlement settlement = 14;
+}
+
+message VTXOSettlement {
+    // txid is the hex-encoded txid of the round commitment tx that forfeited
+    // the VTXO (the leave/cooperative-forfeit round).
+    string txid = 1;
+
+    // height is the block height at which txid confirmed.
+    int32 height = 2;
 }
 
 message ListVTXOsRequest {
--- a/vtxo/interfaces.go
+++ b/vtxo/interfaces.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lightninglabs/darepo-client/baselib/protofsm"
 	"github.com/lightninglabs/darepo-client/lib/types"
 	"github.com/lightninglabs/darepo-client/round"
+	fn "github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/lightningnetwork/lnd/input"
 	"github.com/lightningnetwork/lnd/keychain"
 )
@@ -376,6 +377,14 @@ type Descriptor struct {
 	// CommitmentTxID is the txid of the commitment transaction.
 	CommitmentTxID chainhash.Hash
 
+	// Settlement, when present, carries the on-chain coordinates of the
+	// round commitment transaction that forfeited this VTXO (the leave /
+	// cooperative-forfeit round). Unlike CommitmentTxID, which anchors the
+	// round that CREATED this VTXO, this is the round that CONSUMED it. It
+	// is Some only for FORFEITED VTXOs whose forfeit round row is known,
+	// and None otherwise.
+	Settlement fn.Option[Settlement]
+
 	// BatchExpiry is the absolute block height at which the batch expires
 	// (operator can sweep via the batch-level timelock).
 	BatchExpiry int32
@@ -415,6 +424,18 @@ type Descriptor struct {
 	// construction exists, add the field to IncomingVTXOEvent and thread it
 	// here so the receive path mirrors the operator's value.
 	ConstructionVersion arkrpc.ConstructionVersion
+}
+
+// Settlement identifies the round commitment transaction that forfeited a
+// VTXO and the height at which it confirmed. It is carried as an optional on
+// the Descriptor (Some only for FORFEITED VTXOs whose forfeit round is known)
+// so absence is modelled explicitly rather than via a zero-hash sentinel.
+type Settlement struct {
+	// TxID is the txid of the forfeit round's commitment transaction.
+	TxID chainhash.Hash
+
+	// Height is the block height at which TxID confirmed.
+	Height int32
 }
 
 // MaxTreeDepth returns the largest TreeDepth across the Descriptor's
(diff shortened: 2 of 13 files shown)


ellemouton:
@claude review this

daemonrpc/daemon.proto
```diff
@@ -708,6 +708,15 @@ message VTXO {
...
     VTXOExpiryInfo expiry_info = 13;
+
+    // settlement_txid is the hex-encoded txid of the round commitment tx that
+    // forfeited this VTXO (the leave/cooperative-forfeit round). Set only for
+    // FORFEITED VTXOs; empty otherwise.
+    string settlement_txid = 14;
+
+    // settlement_height is the block height at which settlement_txid confirmed.
+    // Set only for FORFEITED VTXOs; 0 otherwise.
+    int32 settlement_height = 15;
```
ellemouton: why not use "option" then?

vtxo/interfaces.go
```diff
@@ -376,6 +376,19 @@ type Descriptor struct {
...
+	// CommitmentTxID, which anchors the round that CREATED this VTXO, this
+	// is the round that CONSUMED it. It is populated only for FORFEITED
+	// VTXOs whose forfeit round row is known; it is the zero hash
+	// otherwise.
+	SettlementTxID chainhash.Hash
+
+	// SettlementHeight is the block height at which SettlementTxID
+	// confirmed. It is populated alongside SettlementTxID for FORFEITED
+	// VTXOs and is 0 otherwise.
+	SettlementHeight int32
```
ellemouton: fn.Option?

daemonrpc/daemon.proto
ellemouton: Good call — done. Replaced the two scalar fields with a `VTXOSettlement` sub-message (`txid` + `height`) referenced as `VTXOSettlement settlement = 14;`. In proto3 a message field carries explicit presence, so a non-forfeited VTXO leaves it unset instead of shipping a zero-value sentinel — and it mirrors the Go-side `fn.Option[Settlement]` 1:1.

vtxo/interfaces.go
ellemouton: Done — the two fields are now a single `Settlement fn.Option[Settlement]` (with a `Settlement{TxID, Height}` type), following the `ConfInfo fn.Option[ConfInfo]` precedent in `round/interfaces.go` since the txid and height come from one round row and travel together. `None` for non-forfeited VTXOs; the darepod/db/swapwallet layers and tests were updated to match.

PR #934, opened 2026-07-10
db: Make pending-intent test verification use sqlc queries

ellemouton:
## Problem

`main` is red on `test_postgres`. The send-failure tests added in
[#915](https://github.com/lightninglabs/darepo-client/pull/915)
(`TestFailForfeitIntentsMarksSendFailed`, `TestOrphanSweepKeepsFailedSendRecord`,
`TestRepersistFailedSendIntentRearmsReplay`) verified `pending_intents` state
with raw `baseDB.QueryRowContext` calls using SQLite `?` placeholders:

```go
`SELECT status, failure_reason, failure_code FROM pending_intents WHERE intent_id = ?`
```

The pgx stdlib driver does not rewrite `?` → `$N`, so on postgres `intent_id = ?`
parses `?` as a dangling operator and errors with `syntax error at end of input
(SQLSTATE 42601)`. The tests passed on sqlite/unit-cover and failed only on
`test_postgres`. It also violated the repo rule *"never write raw SQL in Go."*

## Fix

Rather than patch the placeholders, move the reads into the sqlc CRUD layer and
have the tests call the generated methods — no raw SQL, and portable by
construction (sqlc emits the correct placeholder per dialect):

- Add `GetPendingIntentByID`, `CountPendingIntentAnchorsByIntentID`, and
  `CountPendingSendIntentsByIntentID` to `db/sqlc/queries/pending_intents.sql`.
- Route the six raw verification queries in `round_store_test.go` through
  `baseDB.Queries.*`; drop the raw SQL entirely.

## Testing

- `make unit pkg=db tags="test_postgres" case=<each of the three>` — all pass.
- `make unit pkg=db` (sqlite) — pass.
- `make lint-changed-local` — 0 issues; `commitmsg-lint` — OK.

### DIFF
--- a/db/round_store_test.go
+++ b/db/round_store_test.go
@@ -1774,33 +1774,23 @@ func TestFailForfeitIntentsMarksSendFailed(t *testing.T) {
 
 	// The failed intent's header is retained (not deleted) with the reason
 	// and typed code, so the failure is a durable, correlatable record.
-	var (
-		status     string
-		failReason sql.NullString
-		failCode   int64
-	)
-	row := baseDB.QueryRowContext(ctx,
-		`SELECT status, failure_reason, failure_code
-		 FROM pending_intents WHERE intent_id = ?`,
-		failing.ID[:],
-	)
-	require.NoError(t, row.Scan(&status, &failReason, &failCode))
-	require.Equal(t, "failed", status)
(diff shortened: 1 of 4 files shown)


ellemouton:
updating to just update the crud instead to not directly need sql queries in the tests

Issue #938, opened 2026-07-13
In-ark payment fails: script not registered for principal


Liongrass:
I generated an invoice in `darepo-client` and tried to pay it from another darepod wallet.

The payment "hung" on the side of the sender, with these logs repeating:

```
2026-07-13 13:47:48.463 [DBG] SWAP: Using local in-swap vHTLC metadata for refund err="get indexed vtxo by script: rpc error: code = Internal desc = indexer query failed: rpc error: code = Unauthenticated desc = script not registered for principal" hash=03f9dc1a245efb8288d172ffef1dd22a3a0f6cdfc76cd35ae178fd9560bf7fa8 outpoint=d0fd434f430eb5072fecb7ef97397695f3444a0c84a30950b01c0b0ab799daa0:0
2026-07-13 13:47:48.521 [DBG] SWAP: Cooperative in-swap refund not yet available hash=03f9dc1a245efb8288d172ffef1dd22a3a0f6cdfc76cd35ae178fd9560bf7fa8 outpoint=d0fd434f430eb5072fecb7ef97397695f3444a0c84a30950b01c0b0ab799daa0:0
```

On the side of the receiver the logs looked similar:

```
2026-07-13 20:55:00.703 [DBG] SWAP: Unable to query vHTLC state err="get indexed vtxo by script: rpc error: code = Internal desc = indexer query failed: rpc error: code = Unauthenticated desc = script not registered for principal" pk_script=512023543d343a084da621fc152cf5307901f26dc3a17c7d2b73dda15a7ea9a9e3d6
```

Once I stopped the receiver client the sender was unable to get its balance back:

```
2026-07-13 13:57:55.818 [DBG] SWAP: Using local in-swap vHTLC metadata for refund err="get indexed vtxo by script: rpc error: code = Internal desc = indexer query failed: rpc error: code = Unauthenticated desc = script not registered for principal" hash=03f9dc1a245efb8288d172ffef1dd22a3a0f6cdfc76cd35ae178fd9560bf7fa8 outpoint=d0fd434f430eb5072fecb7ef97397695f3444a0c84a30950b01c0b0ab799daa0:0
[...]

PR #939, opened 2026-07-13
sdk/swaps: fail in-swap fast when funding OOR is rejected

ellemouton:
## What this fixes

Partial fix for #938 (in-Ark swap wedge). This is the **client-side
robustness half**: it converts an opaque, eventual two-sided wedge into a
fast, clearly-attributed failure. It does **not** change why the operator
rejects a same-Ark funding OOR — that's the separate server/completion track
(see "Scope" below).

## Root cause (payer side)

The pay (in-swap) flow marks the vHTLC "funded from local OOR metadata" as
soon as the funding OOR is **submitted**, not when the operator **accepts**
it, and it deliberately tolerates the authoritative indexer rejecting the
vHTLC lookup with `script not registered for principal` (expected for a
same-Ark receiver whose script isn't registered under this principal).

But the funding OOR can be **rejected asynchronously after submit**: the OOR
session transitions to `Failed` and the funding input is released back to the
wallet, so **no vHTLC is ever created**. In the #938 logs:

```
13:46:53.292  DRPD  OOR transfer submitted
13:46:53.295  SWAP  "In-swap vHTLC funded from local OOR metadata"   ← optimistic
13:46:53.352  SWAP  Pay refund recovery armed
13:46:53.712  OORC  OutboxErrorEvent → state Failed                  ← operator rejected
13:46:53.714  VTXO  57cc7e4d…:0 SpendReleased → LiveState            ← input returned, unspent
```

The payer never noticed the `Failed`: it polled for a claim preimage that
could never arrive until the invoice deadline, then tried to refund a vHTLC
that doesn't exist. The receiver waited for the same non-existent vHTLC. Both
sides wedged. The indexer's "unregistered" rejection can't distinguish a
*rejected* funding OOR from a *funded-but-unqueryable* one — but the daemon's
durable **OOR session status** can.

## The change

In the funding/claim wait loops (`waitForFundedVHTLC`, `waitForClaimPreimage`),
poll the funding OOR session via `GetOORSession` — the same signal the
cooperative-refund path already uses. If it reports **`FAILED`**:
- cancel the optimistically-armed refund recovery, and
- fail the swap with the operator's `FailureReason` surfaced,

instead of polling to the deadline and attempting to refund a non-existent
vHTLC.

Only an explicit `FAILED` status fails the swap. A missing session, a still-
`PENDING` session, or a transient query error all preserve the existing
behaviour, so a healthy swap is never affected (and the intentional
"proceed on local metadata when the indexer rejects" path for in-Ark is
untouched).

**Funds are not at risk**: a rejected funding OOR releases the input back to
the wallet (unspent), so there is nothing to refund — this only turns an
eventual, opaque wedge into a fast, attributed failure.

## Test

`TestPaySessionFailsFastWhenFundingOORRejected` models the exact scenario:
funding OOR submitted, indexer rejects the vHTLC lookup as unregistered, and
the daemon's OOR session reports `FAILED`. It asserts the swap ends
`PayStateFailed` with the operator's reason surfaced, **never attempts a
refund** (`sendCustomCalls == 0`), and cancels the armed recovery.

Verified this is a genuine repro: with the guard disabled the test **hangs
and times out** (the wedge); with the guard it fails in ~0.1s.

`make lint-changed-local` → 0 issues; `make commitmsg-lint` → OK; full
`sdk/swaps` suite green.

## Scope (explicitly out)

- **Why the operator rejects a same-Ark funding OOR** — opaque client-side
  (the `OutboxErrorEvent` carries no reason; `GetOORSession` does, which this
  PR now surfaces). Root-causing that needs the operator/`arkd` side and a
  reproduction that captures the rejection reason. Tracked as the #938
  server/completion track.
- The **receiver** path already terminates at the invoice deadline (it is
  deadline-bounded, not an infinite hang) and has no signal for the payer's
  funding failure, so it is unchanged here.

Refs #938.

### DIFF
--- a/sdk/swaps/in_swap.go
+++ b/sdk/swaps/in_swap.go
@@ -771,6 +771,13 @@ func (s *paySession) waitForFundedVHTLC(ctx context.Context) error {
 			return err
 		}
 
+		// A funding OOR the operator rejected means no vHTLC was
+		// created; fail fast with the reason rather than waiting for
+		// funding that can never land.
+		if err := s.checkFundingRejected(ctx); err != nil {
+			return err
+		}
+
 		preimage, spentVHTLC, err :=
 			s.client.waitForInSwapClaimObservation(
 				ctx, s.cfg.PaymentHash, s.vhtlcPkScript,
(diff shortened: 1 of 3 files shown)


ellemouton:
@claude review this

ellemouton:
Thanks for the thorough trace. Addressed note 3 (added a comment on `failFundingRejected` noting the `PayStateFailed` transition is intentionally left to the FSM `fail` handler). On note 1: agreed — keeping the guard for all pay swaps is intentional so Lightning funding rejections also fail fast; the extra call is a local, poll-bounded RPC. Note 2 (2 call sites): leaving as-is.

ellemouton:
Also addressed review note 2: extracted the guard into a single `checkFundingRejected` helper used at both wait-loop sites, removing the duplicated block. (Tried extending it to the refund path too, but reverted — a legitimately-refunding swap's funding OOR is COMPLETED/reaped, never FAILED, so it'd be dead code and it collided with the refund-session reconciliation.)

ellemouton:
Superseded by #943, which includes this fail-fast commit as its first commit plus the typed OOR_REJECT_INPUT_NOT_SPENDABLE code, retryable classification, and best-height hint. Consolidating there per review.

Issue #673, opened 2026-06-02
unroll: third-party fee-sponsored unilateral exits


Roasbeef:
## Motivation

In this issue, we sketch a sponsored unilateral exit: a way for a third party
to front all the on-chain fees of an unroll in exchange for a cut, where they
only get paid if they actually drive the whole exit to completion. This builds
directly on the externally-driven PSBT unroll from #672.

The premise that makes this work is the asymmetry of where a client's money
lives. A client deep in Ark has plenty of off-chain value but typically zero
confirmed on-chain UTXOs, which is exactly why they can't CPFP their own
unilateral exit: the branch (tree/checkpoint/ark txs) and the final sweep all
need fees, and the fees have to come from on-chain coins the client doesn't
have. So the natural split is that a Sponsor with real UTXOs fronts the on-chain
fees, and the client (Alice) pays a premium out of value she does have, which
can ride a different rail entirely.

Seen that way, the Sponsor is just the external broadcaster we designed in #672
with an incentive bolted on. They run the external loop: pull the ready branch
nodes, CPFP each off its ephemeral anchor with their own fee input,
`submitpackage`, watch for confs, and finally drive the sweep. The new surface
here is purely the payment: how the Sponsor gets paid, and how that payment is
bound to finishing the entire job.

## Where the swap settles: at the sweep, all-or-nothing

We settle the whole thing once, at the sweep, not per branch level. Two reasons.

First, the only signature in the entire unroll that Alice produces fresh is the
final timeout-path sweep (`StandardVTXOTemplate` leaf 1, `<owner> OP_CHECKSIG
<delay> OP_CSV OP_DROP`, assembled by `TimeoutWitness` in
`lib/arkscript/spend_helpers.go`). Every branch node (`recovery.Node`,
`lib/recovery/proof.go`) is operator-pre-signed and already sits in the proof
Alice holds, so there's no live signature to bind a payment to and no secret to
extract from one. The seam where any binding can live is the sweep, full stop.

Second, settling once at the sweep gives the "only paid if they unroll the
entire thing" property for free. The sweep can't exist until the target VTXO is
on-chain, which can't happen until the whole branch confirmed, which is the
expensive work the Sponsor fronted. All-or-nothing at the tail means the Sponsor
never collects on a half-finished job, so they have every reason to finish (sunk
branch fees) and none to abandon midway. Per-level micro-payments would invert
that incentive and cost on-chain bytes per level, the opposite of what we want.

[...]

PR #943, opened 2026-07-14
multi: retry in-Ark funding when input not yet spendable on operator (#938)

ellemouton:
## What

Makes an in-Ark (same-Ark) pay swap survive the case where the payer's vHTLC
**funding OOR is rejected by the operator because the funding input VTXO is not
yet spendable on the operator's side**, instead of wedging both parties.

Supersedes #939 — its fail-fast commit is included here as the first commit.
**Companion operator-side PR: lightninglabs/lumos#673** (typed reject code +
best-height hint + the non-terminal re-validation that lets the retry actually
complete). **Both halves are required**; see the note under "What this PR does".

## Root cause (#938)

An in-Ark payment wedged both parties, each looping on `script not registered
for principal`. That string was a **red herring** — the indexer's registration
check only fires as a no-rows fallback, and there were no rows because the
**vHTLC was never created**.

The real cause, from the payer daemon's own OOR session record
(`wavecli ark oor get`):

```
OOR_REJECT_UNSPECIFIED: vtxo 57cc…:0 not spendable
```

The funding **input** — a VTXO the payer created via a round ~2 min earlier and
which its own wallet held solidly `live` — was rejected by the operator as "not
spendable." The operator (lumosd) promotes a round VTXO `pending → live` only
after the round commitment tx reaches its confirmation target; the payer,
running neutrino (no mempool), had already observed that confirmation in a block
and treated the VTXO as spendable. On signet's ~10-minute blocks, spending a
freshly minted round VTXO lands in the window where the client is ahead of the
operator's chain view → the funding OOR is rejected → (pre-fix) both sides hang.

It is transient: once the operator's chain view catches up, the identical submit
succeeds.

## What this PR does

1. **`sdk/swaps` (from #939):** the in-swap FSM consults the funding OOR session
   status and fails fast with the operator's reason instead of polling until the
   invoice deadline. This is what surfaced the real reason.
2. **`rpc/oorpb`:** adds a typed `OOR_REJECT_INPUT_NOT_SPENDABLE` reject code and
   a `server_best_height` field on `SubmitPackageRejection` (previously arrived
   as the indistinguishable `OOR_REJECT_UNSPECIFIED`).
3. **`oor`:** classifies the code into a typed, non-terminal `ErrInputNotSpendable`
   (mirroring `ErrUserBalanceExceeded`), carrying `ServerBestHeight`.
4. **`waved`:** drives the rejection through the OOR FSM as a **retryable**
   `OutboxErrorEvent`, so the existing submit re-drive resubmits after a short,
   named delay (`oorInputNotSpendableRetryDelay`) until the operator's chain view
   catches up — self-healing the race. All other typed rejects stay terminal.
   For a swap, the retry is bounded by the invoice deadline.

`server_best_height` is an advisory hint that the operator is merely behind on
chain sync, so the retry is a considered reattempt rather than a blind one.

> **Requires the operator-side fix — lightninglabs/lumos#673.** Before that PR
> the operator drove this rejection to a *terminal* session state and, because
> the OOR session id is the deterministic Ark txid, replayed the cached verdict
> for every retry without re-reading the now-live input — so this client retry
> could never succeed on its own. #673 keeps the transient rejection
> non-terminal so a re-submit re-validates. Deploy the operator first and ship
> the two together.

## Testing

- `oor`: unit test for the new classification + `ServerBestHeight` passthrough.
- The full reproduction **and** end-to-end recovery (create a round VTXO,
  withhold the operator's confirmation, fund an OOR → typed reject → daemon retry
  → confirm → success) land as a systest + a full-daemon itest in lumos#673.

Refs lightninglabs/wavelength#938. Supersedes #939.

### DIFF
--- a/oor/errors.go
+++ b/oor/errors.go
@@ -93,6 +93,42 @@ func (e *ErrUserBalanceExceeded) Is(target error) bool {
 	return ok
 }
 
+// ErrInputNotSpendable is the client-facing typed error returned when the
+// operator rejects an OOR submit because one of the input VTXOs was not in a
+// spendable state on the operator (neither live nor in-flight) at validation
+// time.
+//
+// Like ErrUserBalanceExceeded and unlike ErrOutputPolicyViolation, this is NOT
+// terminal for the same input set: the most common cause is that the input's
+// round commitment transaction has not yet reached the operator's confirmation
+// target even though the client already observed the confirmation, so the same
+// submit succeeds once the operator's chain view catches up. Callers should
+// wait and retry rather than restructuring the transfer.
+//
+// ServerBestHeight carries the operator's best block height at rejection time
+// (0 if the operator did not populate it). A caller that knows the height at
+// which its input confirmed can treat ServerBestHeight below that height as
+// proof the operator is merely behind on chain sync, making the retry provably
+// safe rather than a blind reattempt.
+type ErrInputNotSpendable struct {
+	Reason           string
+	ServerBestHeight uint32
+}
+
+// Error returns a human-readable description of the rejection cause.
+func (e *ErrInputNotSpendable) Error() string {
+	return fmt.Sprintf("oor input not spendable on operator "+
+		"(server_best_height=%d): %s", e.ServerBestHeight, e.Reason)
+}
+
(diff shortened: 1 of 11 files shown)


ellemouton:
@claude review this

PR #949, opened 2026-07-15
swapclientserver: Wait for swapd reconnects

### LINKED ISSUE #877: itest: swapruntime client daemon races swapd startup (connection refused on first CreateInSwap)
## Summary

When a swapruntime client daemon (`clientdarepod`) is started with its swap-server address pointing at a swapd instance that is **not yet running**, its gRPC channel to swapd never recovers once swapd comes up: the first `CreateInSwap` RPC fails with `connection refused`.

Surfaced while wiring swapdk-server's wallet-runtime itests into CI (lightninglabs/swapdk-server#199). Two tests fail reproducibly:
- `TestSwapClientDaemonPayCompletesInBackground`
- `TestSwapClientDaemonPayRetryWaitsForRefund`

## Reproduction / order

The itest sequence is:
1. `StartSwapRuntimeClientDaemon(name, swapListenAddr)` — client daemon comes up configured to reach swapd at `swapListenAddr`, **while swapd is still down**.
2. `FundSwapRuntimeClientDaemon(...)` — funding (swapd still down).
3. `StartSwapDaemon(SwapListenAddr: swapListenAddr)` — swapd now starts and its listener accepts (verified: a TCP dial-readiness probe added on the swapd side in swapdk-server passes).
4. `StartPay` → client daemon → `CreateInSwap` → **`rpc error: code = Unavailable ... dial tcp <swapListenAddr>: connect: connection refused`**.

## Analysis

swapd's own listener is confirmed up by step 3 (a swapdk-side readiness probe on `SwapListenAddr` now waits for accept). The failing dial is from the **client daemon's** channel to swapd. The channel appears to enter a transient-failure/backoff state during the window swapd is down and does not re-resolve/reconnect by the time `CreateInSwap` is issued, so the RPC returns the stale refusal.

## Suggested fix (client-daemon side)

Some combination of:
- Establish the swapd connection lazily / only on first use with proper reconnect (reset backoff), or use `WaitForReady` on the swapd-bound RPCs, or
- Provide a harness hook to wait until the client daemon can actually reach swapd (a lightweight probe RPC through the channel) before the test proceeds.

## Status in swapdk-server

The two tests are temporarily `t.Skip`ped in swapdk-server#199 with a pointer to this issue so the rest of the wallet-runtime suite (the walletdk tests) can run in CI. Re-enable once fixed here and the submodule is bumped.

bhandras:
## Summary

Fixes #877.

The swapruntime client daemon can create its daemon-owned gRPC channel to swapd before swapd is listening. If the first swap RPC lands while the channel is still in gRPC's transient failure/backoff state, the call can fail with the stale `connection refused` from startup.

This PR adds `grpc.WaitForReady(true)` as a default call option on the daemon's swapd gRPC dial options. That keeps swapd-bound daemon RPCs waiting for the channel to become ready until the caller context expires, while leaving REST transport and direct SDK-created gRPC clients unchanged.

It also adds a regression test that:

- creates the daemon swap server clients before any swap server is listening,
- starts `CreateInSwap`,
- verifies it does not fail before the server starts,
- starts a fake gRPC swap server on the same address,
- verifies the original `CreateInSwap` completes.

## Validation

- `make fmt-changed`
- `make lint-changed-local`

### DIFF
--- a/swapclientserver/service.go
+++ b/swapclientserver/service.go
@@ -872,6 +872,16 @@ func (d *daemonAuthOnlyInvoiceCreator) CreateInvoiceWithKeyRouteHintPaths(
 func swapServerDialOptions(cfg *waved.SwapConfig, addr string,
 	clientCerts clientTLSCertProvider) ([]grpc.DialOption, error) {
 
+	// The daemon may start before its configured swap server listener is
+	// accepting connections. WaitForReady prevents that stale startup
+	// refusal from failing the first swap RPC; user RPCs remain bounded by
+	// their request context, and background resume RPCs are bounded by the
+	// daemon root context so they keep waiting until swapd returns or the
+	// daemon shuts down.
+	waitForReadyOpt := grpc.WithDefaultCallOptions(
+		grpc.WaitForReady(true),
+	)
+
 	switch {
 	case cfg.ServerTLSCertPath != "":
 		tlsCfg, err := swapServerTLSConfig(
@@ -885,6 +895,7 @@ func swapServerDialOptions(cfg *waved.SwapConfig, addr string,
 			grpc.WithTransportCredentials(
 				credentials.NewTLS(tlsCfg),
 			),
+			waitForReadyOpt,
 		}, nil
 
 	case useInsecureSwapServerTransport(cfg, addr):
@@ -892,6 +903,7 @@ func swapServerDialOptions(cfg *waved.SwapConfig, addr string,
 			grpc.WithTransportCredentials(
 				insecure.NewCredentials(),
 			),
+			waitForReadyOpt,
 		}, nil
 
 	default:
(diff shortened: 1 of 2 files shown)


bhandras:
@claude please review this PR. Context: this fixes #877 by setting wait-for-ready only on the swapruntime daemon's swapd gRPC dial options, plus a regression where the daemon client starts before the fake swap server listener exists and the first `CreateInSwap` waits until the server starts.

bhandras:
@claude thanks, addressed the review feedback and pushed an updated branch.

Changes made:
- Added a comment above the daemon swapd `WaitForReady` option explaining why it exists, that it applies to the shared swapd channel, and how user RPCs vs background resume RPCs are bounded by their contexts.
- Simplified the first regression-test `select` so any early `CreateInSwap` return fails in one place and includes the returned error.

Validation rerun:
- `make fmt-changed`
- `make lint-changed-local`

Please re-review with this context.

PR #950, opened 2026-07-15
swapwallet: preserve credit-owned activity context

bhandras:
## Summary

Preserve accurate wallet activity for fully credit-funded payments and credit
receives.

This is the client-side companion to
[swapdk-server #228](https://github.com/lightninglabs/swapdk-server/pull/228),
which settles invoices owned by the same swap server directly in its credit
ledger.

## Production symptom

During a manual one-satoshi credit payment, the wallet initially projected the
correct completed activity:

```text
KIND  STATUS    AMOUNT  FEE  PHASE
SEND  COMPLETE  -1      0    confirmed
```

The generic SDK swap row records Ark funding rather than invoice principal.
For a fully credit-funded payment that amount is zero. The live swap monitor,
and independently the periodic history reconciler, could therefore replace
the correct credit-owned row with a zero-value terminal row.

Credit receives had a second context problem. `Receive` returned a rich
pending entry, but did not project it before terminal credit state arrived.
The completed row could consequently lose the receive memo, original invoice
request, and payment hash.

### Preserve credit activity ownership

Credit-only swap summaries are now recognized by their settlement type. The
generic swap monitor and derived-history reconciliation both skip these rows,
leaving the durable credit projector as the sole owner of their wallet
activity lifecycle.

This avoids both observed races:

- an immediate overwrite from the live swap subscription; and
- a delayed overwrite from the periodic reconciliation loop.

### Persist receive context

A newly created credit receive is now tracked and projected immediately. This
persists its operation ID, positive amount, memo, invoice request, and payment
hash before the asynchronous terminal update.

Terminal projections do not need to repeat immutable request context. Activity
upserts therefore retain the stored note and request when an update supplies
an empty value. The in-memory change detector mirrors the SQL upsert semantics
so it does not emit a false lifecycle change.

## Commit structure

1. `swapwallet: preserve credit activity ownership`
   protects live monitoring and periodic reconciliation.
2. `swapwallet: persist credit receive context`
   projects pending receives and preserves immutable context.

Each commit is independently focused and carries its corresponding regression
coverage.

## Validation

- `make sqlc-check`
- `make unit tags='swapruntime wavewalletrpc'`
- `make lint-local`
- focused monitor, reconciler, receive, and activity-store tests
- swapdk-server `TestWaveWalletDKOneSatCreditPaymentMatrix`

The server integration matrix gives Alice and Bob one external Lightning
receive each, runs two one-satoshi internal transfers in each direction, then
has both users pay an external one-satoshi Lightning invoice. It verifies
terminal status, signed amounts, fees, receive memo, invoice request, and
payment hash.

## Dependency

The code is useful independently as activity-projection hardening, but the
end-to-end same-server payment path is introduced by swapdk-server #228. That
PR should update its client gitlink to the merge commit from this PR before it
merges.

### DIFF
--- a/db/activity_store.go
+++ b/db/activity_store.go
@@ -316,9 +316,8 @@ func (s *ActivityPersistenceStore) PullEvents(ctx context.Context, cursor int64,
 // changesRow reports whether the projection would alter the current-state row
 // e, i.e. whether it represents a genuine lifecycle transition. It mirrors the
 // UpsertActivityEntry semantics: scalar lifecycle columns overwrite directly,
-// while the settlement/correlation handles are COALESCEd (a nil projection
-// value preserves the stored one, so it is a change only when non-nil AND
-// different). request_json and entry_json are deliberately NOT compared:
+// while an empty note and the settlement/correlation handles preserve their
+// stored values. request_json and entry_json are deliberately NOT compared:
 // protojson output is not byte-stable, so comparing it would report spurious
 // changes; the request is immutable per operation anyway.
 func (p ActivityProjection) changesRow(e sqlc.ActivityEntry) bool {
@@ -327,12 +326,14 @@ func (p ActivityProjection) changesRow(e sqlc.ActivityEntry) bool {
 		p.AmountSat != e.AmountSat ||
 		p.FeeSat != e.FeeSat ||
 		p.Counterparty != e.Counterparty ||
-		p.Note != e.Note ||
 		p.Phase != e.Phase ||
 		p.PhaseLabel != e.PhaseLabel ||
 		p.FailureCode != e.FailureCode ||
 		p.FailureReason != e.FailureReason ||
 		p.VtxoOutpoint != e.VtxoOutpoint {
+		return true
+	}
+	if p.Note != "" && p.Note != e.Note {
 		return true
 	}
 
--- a/db/sqlc/queries/activity_log.sql
+++ b/db/sqlc/queries/activity_log.sql
@@ -28,7 +28,7 @@ ON CONFLICT (canonical_id) DO UPDATE SET
     amount_sat     = EXCLUDED.amount_sat,
     fee_sat        = EXCLUDED.fee_sat,
     counterparty   = EXCLUDED.counterparty,
-    note           = EXCLUDED.note,
+    note           = COALESCE(NULLIF(EXCLUDED.note, ''), activity_entries.note),
     phase          = EXCLUDED.phase,
     phase_label    = EXCLUDED.phase_label,
     failure_code   = EXCLUDED.failure_code,
@@ -40,7 +40,7 @@ ON CONFLICT (canonical_id) DO UPDATE SET
     swap_session_id = COALESCE(EXCLUDED.swap_session_id, activity_entries.swap_session_id),
     ledger_txid     = COALESCE(EXCLUDED.ledger_txid, activity_entries.ledger_txid),
     boarding_addr   = COALESCE(EXCLUDED.boarding_addr, activity_entries.boarding_addr),
-    request_json    = EXCLUDED.request_json,
+    request_json    = COALESCE(NULLIF(EXCLUDED.request_json, ''), activity_entries.request_json),
     updated_at_unix = EXCLUDED.updated_at_unix;
 
 -- name: AppendActivityEvent :one
--- a/swapwallet/credit_projector.go
+++ b/swapwallet/credit_projector.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/lightninglabs/wavelength/credit"
+	"github.com/lightninglabs/wavelength/rpc/swapclientrpc"
 	"github.com/lightninglabs/wavelength/rpc/wavewalletrpc"
 )
 
@@ -24,6 +25,17 @@ const creditProjectInterval = 5 * time.Second
 // credit-backed rows, matching the pending entries the router and receiver
 // emit so the projected terminal update lands on the same row.
 const creditCounterparty = "credit"
+
+// creditProjectorOwnsSwapSummary reports whether the durable credit operation,
+// rather than the generic swap monitor or history reconciler, owns a swap's
+// wallet activity row. Credit-only SDK swaps persist a zero Ark funding amount,
+// while the credit operation retains the invoice principal. The server emits
+// CREDIT only for a fully credit-funded invoice; partial credit funding emits
+// MIXED and remains owned by the generic swap lifecycle.
+func creditProjectorOwnsSwapSummary(summary *swapclientrpc.SwapSummary) bool {
+	return summary != nil && summary.GetSettlementType() == swapclientrpc.
+		SwapSettlementType_SWAP_SETTLEMENT_TYPE_CREDIT
+}
 
 // startCreditProjectorLoop spawns the background goroutine that projects
 // durable credit-operation terminal state onto wallet rows. It is a no-op when
--- a/swapwallet/recv.go
+++ b/swapwallet/recv.go
@@ -196,6 +196,8 @@ func (r *receiver) recvCredit(ctx context.Context,
 	entry := creditReceiveEntry(
 		req, start.OpID, start.Invoice, paymentHashHex, amt,
 	)
+	r.runtime.trackPendingEntryWithoutTimeout(entry)
+	r.runtime.projectAndEmit(context.WithoutCancel(ctx), entry)
 
 	return &wavewalletrpc.RecvResponse{
 		Invoice: start.Invoice,
--- a/db/activity_store_test.go
+++ b/db/activity_store_test.go
@@ -51,6 +51,7 @@ func sampleProjection(id string) ActivityProjection {
 			0xbb,
 			0xcc,
 		},
+		RequestJSON:   `{"lightningInvoice":{"invoice":"lnbc1"}}`,
 		EntryJSON:     `{"id":"` + id + `"}`,
 		CreatedAtUnix: 100,
 		UpdatedAtUnix: 100,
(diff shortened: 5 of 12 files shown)


bhandras:
@claude review

Please review this two-commit client companion to
https://github.com/lightninglabs/swapdk-server/pull/228.

The first commit makes the durable credit projector the sole owner of activity
for `CREDIT` settlement summaries. Both the live generic swap monitor and the
periodic derived-history reconciler otherwise overwrite a fully credit-funded
one-satoshi SEND with the SDK row's zero Ark funding amount.

The second commit projects a credit receive immediately and preserves its
stored memo and invoice request when later terminal projections omit immutable
context. The in-memory change detector mirrors the SQL upsert behavior.

Please focus on activity ownership, whether skipping credit-only swap summaries
can hide any lifecycle that the credit projector does not cover, and the
empty-value preservation semantics. SQLC generation, the full tagged unit
suite, and all 71 configured linters pass locally. The server integration
matrix also passes repeated one-satoshi internal payments in both directions
plus external Lightning receives and sends for both wallets.

swapwallet/recv.go
```diff
@@ -196,6 +196,8 @@ func (r *receiver) recvCredit(ctx context.Context,
 	entry := creditReceiveEntry(
 		req, start.OpID, start.Invoice, paymentHashHex, amt,
 	)
+	r.runtime.trackPendingEntryWithoutTimeout(entry)
+	r.runtime.projectAndEmit(context.WithoutCancel(ctx), entry)
```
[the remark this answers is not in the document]
bhandras: Thanks. I’m leaving the anchored context unchanged here because this is the established durability contract for already-accepted wallet activity writes: the same `context.WithoutCancel` pattern is used by the send, receive, and admin projection paths so an RPC disconnect cannot cancel persistence.

Adding a timeout only to credit receives would make their durability semantics inconsistent with the other accepted operations. If these writes need a bound, it should be applied uniformly inside the projection/store layer with an explicit retry or failure policy. The full client lint and tagged unit suite pass with the existing pattern.

bhandras:
@claude review

Confirmed and made the cross-boundary invariant explicit.

On the server, `QuoteInSwap`/`CreateInSwap` return `CREDIT` only when credits cover the full invoice or credit is mandatory. Partial credit coverage always returns `MIXED`. On the client, those same quote fields set `CreditOnly=true` when `MustUseCredit` is true or applied credit plus the planned top-up covers the full principal. Therefore `settlement_type=CREDIT` implies projector ownership, while mixed payments remain owned by the generic swap lifecycle.

I folded two small clarifications into the activity-ownership commit:

- the helper comment now records the `CREDIT` versus `MIXED` server contract;
- `TestRouterSendInvoiceHandsCreditPayToRegistry` now asserts that a server `CREDIT` quote is handed to the registry with `CreditOnly=true`.

The focused router, monitor, and reconciler tests pass after the change. The companion swapdk-server PR has also been updated to pin the rewritten client head.

Gemini’s `WithoutCancel` suggestion is intentionally not applied. As your review notes, this is the package-wide durability pattern for already-accepted activity writes; a timeout should be designed uniformly in the projection/store layer rather than only for credit receives.

bhandras:
@claude review

One final mechanical update landed after the invariant clarification: the two commits are rebased onto current `main` (`f4f59a30`) and the first commit was formatted with the exact CI command, `make fmt-changed base=origin/main`.

No behavior changed beyond the already-reviewed comment and `CreditOnly` assertion. `make fmt-changed-check`, both commit messages, the focused router/monitor/reconciler/receive tests, and the activity-store test now pass on the rebased head. Please treat this as the final head for your verdict.

swapwallet/credit_projector.go
```diff
@@ -25,6 +26,17 @@ const creditProjectInterval = 5 * time.Second
...
 
+// creditProjectorOwnsSwapSummary reports whether the durable credit operation,
+// rather than the generic swap monitor or history reconciler, owns a swap's
+// wallet activity row. Credit-only SDK swaps persist a zero Ark funding amount,
+// while the credit operation retains the invoice principal. The server emits
+// CREDIT only for a fully credit-funded invoice; partial credit funding emits
+// MIXED and remains owned by the generic swap lifecycle.
+func creditProjectorOwnsSwapSummary(summary *swapclientrpc.SwapSummary) bool {
+	return summary != nil && summary.GetSettlementType() == swapclientrpc.
+		SwapSettlementType_SWAP_SETTLEMENT_TYPE_CREDIT
```
sputn1ck, line 37: [P1] Settlement rail is not proof of activity ownership

The public SwapClientService.StartPay RPC accepts max_credit_sat and can create a CREDIT summary without going through the local CreditRegistry. Monitor and history then suppress that summary, but the credit projector has no matching operation to emit, so the payment disappears from wallet activity. Please gate suppression on an explicit local owner marker or a matching credit operation rather than settlement type alone. The new monitor test currently wires no registry while expecting suppression, which captures this broken case.

db/sqlc/queries/activity_log.sql
```diff
@@ -40,7 +40,7 @@ ON CONFLICT (canonical_id) DO UPDATE SET
     swap_session_id = COALESCE(EXCLUDED.swap_session_id, activity_entries.swap_session_id),
     ledger_txid     = COALESCE(EXCLUDED.ledger_txid, activity_entries.ledger_txid),
     boarding_addr   = COALESCE(EXCLUDED.boarding_addr, activity_entries.boarding_addr),
-    request_json    = EXCLUDED.request_json,
+    request_json    = COALESCE(NULLIF(EXCLUDED.request_json, ''), activity_entries.request_json),
```
sputn1ck, line 43: [P2] Preserve immutable context in events too

This correctly preserves request_json in activity_entries, but ProjectEntry still appends the sparse terminal EntryJSON to activity_events and projectAndEmit emits that same sparse entry. Live and replayed terminal updates therefore omit memo, invoice request, and the receive payment hash even though List returns the enriched current row. Please append and emit the effective merged snapshot, and cover both live and resumable subscriptions.

db/activity_store.go
```diff
@@ -316,9 +316,8 @@ func (s *ActivityPersistenceStore) PullEvents(ctx context.Context, cursor int64,
 // changesRow reports whether the projection would alter the current-state row
 // e, i.e. whether it represents a genuine lifecycle transition. It mirrors the
 // UpsertActivityEntry semantics: scalar lifecycle columns overwrite directly,
-// while the settlement/correlation handles are COALESCEd (a nil projection
-// value preserves the stored one, so it is a change only when non-nil AND
-// different). request_json and entry_json are deliberately NOT compared:
+// while an empty note and the settlement/correlation handles preserve their
+// stored values. request_json and entry_json are deliberately NOT compared:
```
sputn1ck, line 320: [P2] Allow request-only enrichment

Because changesRow deliberately ignores RequestJSON, a rich projection that differs from an existing sparse row only by invoice request returns before the COALESCE upsert can enrich it. This can happen when a sparse credit projection wins the race and the receive memo is empty. Please compare the semantic request value or otherwise force immutable-context enrichment, with an empty-memo regression test.

sputn1ck (commented):
I reviewed the final head after it merged. The ownership and eager-context direction is correct, but the inline findings should be handled in a follow-up.

Two related correctness paths are on lines that were not changed by this PR:

- [P1] The registry resumes a credit-pay child before admission returns, while the router writes the initial PENDING row only after Ask completes. A fast internal settlement can therefore project COMPLETE first and then be overwritten by the stale PENDING write. The current upsert accepts the backward transition, and the projector may already have memoized COMPLETE, leaving activity permanently pending: https://github.com/lightninglabs/wavelength/blob/493cdc3d05af9e8e532e073d6d75cffce2784504/swapwallet/router.go#L299-L324
- [P2] pollCreditOps records the state and clears pending tracking before projectAndEmit, whose write error is discarded. One transient store failure makes subsequent polls skip the unchanged terminal state. This is more serious now that monitor/history intentionally suppress the CREDIT summary fallback: https://github.com/lightninglabs/wavelength/blob/493cdc3d05af9e8e532e073d6d75cffce2784504/swapwallet/credit_projector.go#L121-L150

db/sqlc/queries/activity_log.sql
bhandras: Addressed in follow-up #954. The projector now reads the current durable row, decodes it, and merges immutable note, request, creation, and correlation context into an effective WalletEntry before calling ProjectEntry. That effective snapshot is used for both the current-row upsert and activity_events append, and is also the entry emitted live. Tests assert that both live and resumable event streams retain memo, invoice request, and receive payment hash after a sparse terminal projection.

db/activity_store.go
bhandras: Addressed in follow-up #954. Request JSON is now compared semantically in changesRow, so request-only enrichment cannot be suppressed even when memo is empty. The regression test starts with an empty-memo row, adds only the request, verifies a second event is appended, and confirms the stored request survives.

swapwallet/credit_projector.go
bhandras: Addressed in follow-up #954. Generic swap activity is now suppressed only when the summary is CREDIT-settled and its payment hash is present in this wallet’s explicit local ownership set. Ownership is marked after successful credit admission and reconstructed from durable credit-only operations on restart. A public StartPay CREDIT summary without a local CreditRegistry operation therefore remains visible through the generic monitor. The focused monitor test covers both unowned and locally owned CREDIT summaries.

bhandras:
The two review-body findings are also addressed in follow-up #954. The activity upsert now rejects a terminal-to-PENDING transition atomically, covering the fast internal-settlement race where COMPLETE can arrive before the router’s admission projection. Credit polling now updates its memoized state and clears pending tracking only after projectAndEmit succeeds, so a transient store failure is retried. All inline findings have been answered at their original threads.

PR #952, opened 2026-07-15
swapclientserver: scope reconnect waiting

### LINKED ISSUE #951: swapwallet: Balance hangs when optional swap server is unavailable
## Summary

`WalletService.Balance` now waits until its caller context expires when the configured swap server is unavailable. This regressed when #949 added `grpc.WaitForReady(true)` as a default call option for every daemon-owned swap-server RPC.

The wallet's Ark balance is available and the swap-server credit lookup is intentionally optional, but the global wait-for-ready policy prevents that optional lookup from failing quickly. Callers therefore receive `DeadlineExceeded` instead of the Ark balance response.

This surfaced in swapdk-server PR #228:

- failing check: https://github.com/lightninglabs/swapdk-server/actions/runs/29421273788/job/87372171211
- failing test: `TestWalletBalanceSurfacesUnilateralExit`
- failure: `wallet_balance_test.go:28: rpc error: code = DeadlineExceeded desc = context deadline exceeded`

The test fails on its first `WalletService.Balance` call after a successful boarding round. It has not requested a unilateral exit yet.

## Root cause

`swapwallet.Service.fetchBalance` first obtains the daemon balance with `GetBalance`. It then performs a best-effort credit lookup:

```go
credits, err := s.deps.SwapService.ListCredits(ctx, ...)
if err != nil {
    return resp, nil
}
```

Before #949, an unavailable swap server made `ListCredits` fail quickly and the method returned `resp` without credit fields.

Commit `26e880fe` from #949 adds this option in `swapServerDialOptions`:

```go
grpc.WithDefaultCallOptions(grpc.WaitForReady(true))
```

Because it is a channel-wide default, the optional `ListCredits` call now waits until the `Balance` request context expires. Although `fetchBalance` then returns the already-built response, the gRPC request context has expired and the client receives `DeadlineExceeded`.

The original #949 goal is valid for first-use and resumed swap operations that should survive a temporary swapd startup or reconnect gap. The regression comes from applying that policy to every RPC, including optional read enrichment.

## Reproduction and isolation

Run from swapdk-server:

```bash
PAYMENTSERVICE_BIN=/path/to/paymentservice \
ITEST_LND_IMAGE_REUSE=1 \
make itest-swapruntime \
  icase=TestWalletBalanceSurfacesUnilateralExit
```

I isolated the client versions while keeping swapdk-server at `origin/master` (`81d13d64`):

| Client revision | Result |
| --- | --- |
| `7b317dd4`, the older server-pinned client before #949 | Pass, 39.238s |
| `f4f59a30`, current `main` and the base of #950 | Fail, 52.553s |
| `493cdc3d`, #950 head | Fail, 54.678s |

The full #228 branch also reproduces locally in 54.660s, matching CI. This shows the failure is present on #950's base and was not introduced by #950's activity changes.

## Suggested direction

Avoid making wait-for-ready the default for every swap-server method. Apply it selectively to operations that must wait through startup or reconnect, such as swap creation and background resume calls.

A narrowly bounded context around the optional credit lookup could protect `Balance`, but selective wait-for-ready semantics are preferable because other optional/read-only calls can otherwise acquire the same deadline-sized stall.

## Acceptance criteria

- The late-server regression from #877 remains fixed: the first applicable swap operation waits for swapd and succeeds after it becomes available.
- `WalletService.Balance` returns Ark balance fields promptly while swapd is unavailable.
- `WalletService.Status`, which composes `fetchBalance`, also remains responsive while swapd is unavailable.
- Optional credit fields may be omitted when swapd cannot be reached.
- `TestWalletBalanceSurfacesUnilateralExit` passes against current wavelength `main` when exercised through swapdk-server's swapruntime integration harness.
- Unit coverage distinguishes wait-for-ready swap operations from best-effort wallet reads.

bhandras:
## Problem

The daemon configures one gRPC channel for swapd. PR #949 made that entire channel use `WaitForReady(true)` so a state-changing swap RPC can survive swapd starting or reconnecting after the daemon.

That channel-wide policy also changed optional read-only calls. In particular, `swapwallet.Service.fetchBalance` first obtains the Ark balance and then calls `ListCredits` as an optional enrichment. The wallet deliberately ignores a failed credit read and can return the Ark balance alone. With the global wait-for-ready option, however, an unavailable swapd makes `ListCredits` consume the parent request's full deadline. The already-computed balance can no longer be returned because the wallet RPC context has expired.

This surfaced in swapdk-server PR #228 as `TestWalletBalanceSurfacesUnilateralExit` failing its first balance request with `DeadlineExceeded`.

## Changes

- Replace the channel-wide default call option with a unary client interceptor.
- Apply wait-for-ready to RPCs that create or advance swap and credit protocol state.
- Keep `QuoteInSwap`, `ListCredits`, mailbox calls, and unknown methods fail-fast.
- Preserve the reconnect behavior from #949 for `CreateInSwap` and the other durable protocol operations.

## Policy

Wait for swapd to become ready:

- `RequestChannelId`
- `CreateInSwap`
- `CreateCredit`
- `RedeemCredit`
- `AuthorizeInSwapRefund`
- `AcknowledgeOutSwapHtlc`
- `SignInSwapForfeit`
- `SubmitOutSwapForfeitSignature`

Fail fast when swapd is unavailable:

- `QuoteInSwap`
- `ListCredits`
- mailbox RPCs
- unknown future RPCs, until their retry semantics are chosen explicitly

## Tests

- Added an unavailable-server regression proving `ListCredits` returns `Unavailable` before its context expires.
- Added a table test that locks down the method-level reconnect policy.
- Preserved and reran the #949 regression proving `CreateInSwap` waits through a late server start.
- Reran the exact downstream failure against swapdk-server PR #228:

  `TestWalletBalanceSurfacesUnilateralExit` passed in 37.2 seconds.

- `go test -tags='swapruntime' ./swapclientserver -count=1`
- `make fmt-changed-check base=origin/main`
- `make lint-local`

Fixes #951.

### DIFF
--- a/swapclientserver/service.go
+++ b/swapclientserver/service.go
@@ -30,6 +30,7 @@ import (
 	sdkark "github.com/lightninglabs/wavelength/sdk/ark"
 	"github.com/lightninglabs/wavelength/sdk/swaps"
 	"github.com/lightninglabs/wavelength/serverconn"
+	"github.com/lightninglabs/wavelength/swaprpc"
 	"github.com/lightninglabs/wavelength/vtxo"
 	"github.com/lightninglabs/wavelength/waved"
 	"github.com/lightningnetwork/lnd/invoices"
@@ -872,14 +873,11 @@ func (d *daemonAuthOnlyInvoiceCreator) CreateInvoiceWithKeyRouteHintPaths(
 func swapServerDialOptions(cfg *waved.SwapConfig, addr string,
 	clientCerts clientTLSCertProvider) ([]grpc.DialOption, error) {
 
-	// The daemon may start before its configured swap server listener is
-	// accepting connections. WaitForReady prevents that stale startup
-	// refusal from failing the first swap RPC; user RPCs remain bounded by
-	// their request context, and background resume RPCs are bounded by the
-	// daemon root context so they keep waiting until swapd returns or the
-	// daemon shuts down.
-	waitForReadyOpt := grpc.WithDefaultCallOptions(
-		grpc.WaitForReady(true),
+	// Swap operations wait for the configured server to reconnect, while
+	// best-effort reads fail fast so they cannot consume an enclosing
+	// wallet RPC's entire deadline.
+	waitForReadyInterceptorOpt := grpc.WithChainUnaryInterceptor(
+		swapServerWaitForReadyInterceptor,
 	)
 
 	switch {
@@ -895,7 +893,7 @@ func swapServerDialOptions(cfg *waved.SwapConfig, addr string,
 			grpc.WithTransportCredentials(
 				credentials.NewTLS(tlsCfg),
 			),
-			waitForReadyOpt,
+			waitForReadyInterceptorOpt,
 		}, nil
 
 	case useInsecureSwapServerTransport(cfg, addr):
@@ -903,7 +901,7 @@ func swapServerDialOptions(cfg *waved.SwapConfig, addr string,
 			grpc.WithTransportCredentials(
 				insecure.NewCredentials(),
 			),
-			waitForReadyOpt,
+			waitForReadyInterceptorOpt,
 		}, nil
 
 	default:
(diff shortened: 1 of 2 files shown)


bhandras:
@claude Please review this follow-up with the regression context from #951 and #949 in mind.

The key invariant is that durable/state-changing swap RPCs must still wait through a late swapd start or reconnect, while optional reads such as `ListCredits` must fail fast so callers can return partial results without losing their own deadline. The existing late-`CreateInSwap` regression remains green, the new unavailable-`ListCredits` regression is green, and swapdk-server's previously failing `TestWalletBalanceSurfacesUnilateralExit` passes against this branch.

bhandras:
@claude Thanks, all three non-blocking suggestions are folded into the original commit:

- renamed the dial option to `waitForReadyInterceptorOpt`;
- restored the distinction between request-context-bounded user operations and daemon-root-context-bounded background resumes; and
- documented the fail-fast default for future RPCs and added an explicit unknown future swap-method test.

After folding, the focused package tests, formatter check, commit-message lint, and local lint all pass. I also reran swapdk-server's exact downstream `TestWalletBalanceSurfacesUnilateralExit` regression against the new SHA; it passed in 39.593 seconds. Please re-review the final form.

Issue #956, opened 2026-07-15
vtxo,walletdk: expiry auto-refresh queues but never joins a round


sputn1ck:
## Summary

Expiry-driven VTXO refresh is not complete today. A running WalletDK client
detects that a VTXO is approaching expiry and queues a local refresh intent,
but it never asks the round FSM to join the operator. A passive wallet can
therefore leave the intent in `PendingRoundAssembly` until the VTXO reaches its
critical threshold and starts unilateral exit.

The simplest WalletDK and mobile users cannot repair this themselves because
the typed WalletDK API has no refresh operation. Go callers can escape through
`Client.ArkRPC()`, but they must know to call both `RefreshVTXOs` and
`JoinNextRound`. The CLI hides this two-step sequence; WalletDK does not.

Auto-refresh should require no application action. The operator should signal
the client when its VTXOs enter the configured free-refresh window, and the
client should atomically refresh the complete eligible set in one round.

## Current behavior

The background path is split across four steps:

1. Each live VTXO actor subscribes to block epochs in
   `vtxo.(*VTXOActor).subscribeBlockEpochs`.
2. `vtxo.(*LiveState).handleBlockEpoch` calls `ExpiryConfig.CheckExpiry`. At
   `ExpiryStatusNeedsRefresh`, it emits a `ForfeitRequest` and moves the VTXO to
   `PendingForfeitState`.
3. The VTXO actor relays a `round.RefreshVTXORequest`. The round actor converts
   it into a forfeit-plus-replacement `IntentPackage` and adds it to an
   `Idle`/`PendingRoundAssembly` FSM.
4. The handler returns without emitting `IntentRequested`. Only that separate
   event produces `JoinRoundRequest`, so the operator never sees a passive
   auto-refresh.

`PendingForfeitState` continues watching blocks. If nothing else triggers the
assembling round, it eventually treats the stalled refresh as a forfeit timeout
and escalates to unilateral exit.

`EagerRoundJoin` does not close the gap. It currently applies to newly
confirmed boarding inputs and cooperative leaves, not expiry refreshes.

## WalletDK impact

The high-level `wavewalletdk.Client` and the mobile wrapper expose no refresh
method. Today the only explicit Go workaround is the advanced daemon surface:

```go
ark := client.ArkRPC()

_, err := ark.RefreshVTXOs(ctx, refreshRequest)
if err != nil {
    return err
}

[...]

PR #957, opened 2026-07-15
round: Join expiry refreshes automatically

sputn1ck:
## Summary

- Schedule expiry-refresh registration from the round actor itself, so passive wallets do not depend on a later host RPC.
- Reset a 500 ms registration quiet window as each refresh arrives, keeping co-expiring VTXOs in one paid round.
- Cover the single-refresh trigger, stale timeout handling, and a two-VTXO batch that emits exactly one cumulative `JoinRoundRequest`.

## Root cause

The expiry monitor already moved the VTXO to `PendingForfeit` and queued its paid replacement intent, but the round FSM remained in `PendingRoundAssembly`. Only a later host-triggered registration event emitted `JoinRoundRequest`, so a passive wallet could stall until critical expiry.

Registering every VTXO immediately is also unsafe: co-expiring VTXOs would create competing client FSMs, and same-client registrations replace one another on the server. The short, resettable timeout lets the block-epoch burst finish assembling before the client emits one cumulative registration.

This change closes the queue-to-join gap for the existing ordinary paid refresh path. It does not add or rely on server-side free-refresh signaling, cohort selection, fee waivers, or the pull requests linked from #956.

## Validation

- `go test ./round ./vtxo`
- `go test ./...`
- `make build`
- `make lint-changed-local`
- [Lumos #676](https://github.com/lightninglabs/lumos/pull/676) exercises the exact refresh boundary without `RefreshVTXOs`, `JoinNextRound`, or a harness registration hook.

Related: #956

### DIFF
--- a/round/actor.go
+++ b/round/actor.go
@@ -69,6 +69,11 @@ type RefreshVTXORequest struct {
 	// Amount is the VTXO value in satoshis.
 	Amount int64
 
+	// TriggerRegistration causes the round actor to emit IntentRequested
+	// after adding this refresh to the assembling round. Expiry-driven
+	// refreshes set this so they do not depend on a later host RPC.
+	TriggerRegistration bool
+
 	// OperatorFee is an advisory-only hint under the seal-time fee
 	// handshake (#270). The VTXO actor's RefreshFeeQuoter fills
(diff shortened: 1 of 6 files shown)


sputn1ck:
@claude please review

PR #958, opened 2026-07-15
multi: bound transient OOR submit-reject retries + USER_BALANCE

ellemouton:
## What

Follow-up to #938 / #943. Bounds the transient OOR submit-reject retry and
extends retryable recovery to `OOR_REJECT_USER_BALANCE`.

## Why

The daemon retries a transient OOR submit rejection by re-driving the submit
through the OOR FSM. That retry was **unbounded** — it re-drove every 15s
(`oorTransientRejectRetryDelay`) forever on a plain send. It was safe only
because `INPUT_NOT_SPENDABLE` self-resolves in ~1 block; a genuinely-stuck input
(a dropped/reorged commitment tx, an abandoned round) would spin resubmits
forever. And `USER_BALANCE` — documented as "hold and retry later; a same-shape
resubmit can succeed once the recipient's balance drops" — was kept terminal
precisely because auto-retrying it unboundedly would spin against a maybe-never
event.

## What this PR does

1. **Configurable max-duration bound.** `OORConfig.MaxTransientSubmitRetry`
   (default `1h`; `0` disables the bound) caps how long the
   `AwaitingSubmitAccepted` state keeps re-driving before failing the session
   terminally and releasing the pre-point-of-no-return input locks. The window
   start is persisted in the outgoing snapshot (`Version 4→5`; a pre-v5 snapshot
   decodes the absent record to `0`, a fresh window) so the bound survives a
   daemon restart. The transition reads a clock injected via the FSM
   `Environment` rather than `time.Now()`, so it stays deterministic under test;
   the clock and the cap are threaded from `waved`'s `OORConfig` through the
   registry and session-actor construction paths.
2. **`USER_BALANCE` is now bounded-retryable.** `oorRejectRetry` classifies both
   `OOR_REJECT_INPUT_NOT_SPENDABLE` and `OOR_REJECT_USER_BALANCE` as transient,
   each capped by the same window. `LineageTooLarge` / `OutputPolicy` stay
   terminal (they require the client to restructure the transfer into a new
   session).

## Companion

Paired with an operator-side lumos PR that keeps `USER_BALANCE` non-terminal so
the retry re-validates (the #938 server fix already does this for
`INPUT_NOT_SPENDABLE`). Both are needed for `USER_BALANCE` recovery end to end;
this client change is safe to merge first (an un-updated operator simply keeps
failing the retry terminally, same as today).

## Testing

- FSM (`oor`): within-budget reschedule carries the window forward; over-budget
  → terminal fail + input release; zero-budget → unbounded (legacy).
- Snapshot round-trip: v5 persists/restores the window start; a v4 snapshot
  restores `0` without error.
- `oorRejectRetry`: both transient codes retryable, others terminal.
- Config validation: `0` → default, negative rejected.
- `go build`, `go test ./oor/... ./waved/...`, and `custom-gcl` lint (0 issues).

### DIFF
--- a/oor/CLAUDE.md
+++ b/oor/CLAUDE.md
@@ -72,8 +72,10 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/oor.<Sym
   actor's DB transaction; both phase-1 hint resolution and phase-2
   authoritative metadata lookup go through durable `serverconn` query
   messages and return as fresh events.
-- Snapshots are versioned per direction (`OutgoingSnapshot.Version = 4`,
-  `IncomingSnapshot.Version = 1`); restore rejects a zero version.
+- Snapshots are versioned per direction (`OutgoingSnapshot.Version = 5`,
+  `IncomingSnapshot.Version = 1`); restore rejects a zero version. Outgoing
+  v5 adds the `FirstRejectUnixNanos` record (bounded transient submit-reject
+  retry window); a pre-v5 snapshot decodes it to 0 (a fresh window).
 - `StartTransferRequest.IdempotencyKey` dedup relies on a partial UNIQUE
   index on `oor_session_registry` (at most one live-or-completed row per
   key); a failed session never blocks a keyed retry.
--- a/oor/interfaces.go
+++ b/oor/interfaces.go
@@ -3,9 +3,11 @@ package oor
 import (
 	"encoding/hex"
 	"fmt"
+	"time"
 
 	"github.com/btcsuite/btcd/chainhash/v2"
 	"github.com/lightninglabs/wavelength/baselib/protofsm"
+	"github.com/lightningnetwork/lnd/clock"
 )
 
 // StateTransition is a type alias for the verbose protofsm.StateTransition type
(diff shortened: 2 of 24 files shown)


ellemouton:
Operator-side companion: **lightninglabs/lumos#677** (makes `USER_BALANCE` non-terminal on the operator so this client's bounded retry re-validates). Merge this client PR first, then #677 re-points its submodule to merged `main`.

ellemouton:
@claude review

ellemouton:
@claude review — addressed both findings from the last review:

1. **Wrong key in the validation error/test** — the message and the test assertion now reference the actual key `oor.maxsubmitretry` (was `oor.maxtransientsubmitretry`), so an operator grepping the error finds it.
2. **`now` before the nil guard** — reordered so the (guarded) cap is read before `env.Now()`, with a comment noting both reads are nil-safe for directly-constructed test envs.

The design note (USER_BALANCE re-driving ~240×/1h against the operator) is the intended, bounded + configurable trade-off — left as-is, with the `0`-disables escape hatch and the documented default rationale. Folded into the single commit; CI re-running.

PR #895, opened 2026-07-09
multi: reorg-aware chain observation (chainsource + backends)

ellemouton:
## Summary

PR 1 of the six-PR basic-v1 reorg-safety stack tracked by
[lightninglabs/lumos#454](https://github.com/lightninglabs/lumos/issues/454).

This PR is the **chain-observation layer**. It makes confirmation and spend
observation reversible end to end:

```
Observed → Reorged → Observed → Done
```

For `txconfirm`:

```
TxConfirmed → TxReorged → TxConfirmed → TxFinalized
```

The first positive observation is no longer terminal. Higher layers can keep a
fact provisional, react when it leaves the best chain, and observe it again
after reconfirmation.

## What is included

- **`chainsource`**: multi-shot confirmation/spend actors, reorg and done
  events, ordered delivery, and height-based finality synthesis for transports
  that do not provide native `Done`. Transport boundaries explicitly defer a
  buffered `Done` until the positive confirmation/spend identity has crossed
  first.
- **`txconfirm`**: reversible confirmation notifications, a distinct terminal
  `TxFinalized` state, and correct terminal-only replay for late subscribers.
- **LND/lndclient backend**: preserve reorg-capable observation instead of
  terminating after the first positive event. Today this transport supplies a
  payload-less reorg ping, so `chainsource` synthesizes finality from block
  height.
- **lwwallet/Esplora**: detect same-height, shorter, deeper, and mid-catch-up
  replacements, then forward disconnected/connected blocks in order.
- **Neutrino**: forward confirmation/spend reorg and done notifications.
- **Wallet subscriber compatibility**: adapt the boarding-sweep `txconfirm`
  subscriber to the new four-event lifecycle. Without this, `TxReorged` and
  `TxFinalized` fall through the old mapper as failure-shaped notifications.
  Broader boarding spend-watch recovery belongs to Client PR 2.
- **Harness and systests**: real chainsource and txconfirm reorg round trips.

## Native LND transport completion

This PR remains compatible with today's released lnd/lndclient by synthesizing
terminal `Done` at the `chainsource` boundary. The proper native transport path
is being completed in parallel:

- [lightningnetwork/lnd#10943](https://github.com/lightningnetwork/lnd/pull/10943)
  carries confirmation re-org depth and explicit confirmation/spend `Done` on
  the `ChainNotifier` gRPC streams.
- [lightninglabs/lndclient#283](https://github.com/lightninglabs/lndclient/pull/283)
  exposes those native fields/events through lndclient.

[...]

Issue #626, opened 2026-05-29
swaps: support vHTLC recovery across identity key rotation


sputn1ck:
## Problem

Swap vHTLC recovery currently assumes that the client-side vHTLC signing key is the daemon identity key at a fixed locator (`KeyFamilyNodeKey`, index `0`). That survives normal restarts with the same wallet seed/backend, but it is weak under identity-key rotation or wallet/backend migration.

From the current swap key audit:

- The SDK uses `DaemonConn.IdentityPubKey` as the client vHTLC key for both receive and pay swaps.
- Swap session rows persist the public keys used to build the vHTLC policy, so script reconstruction can survive a restart.
- Recovery still arms the daemon with a fixed identity-key signer locator instead of a persisted per-swap signing identity.

That means an old swap can retain enough public data to rebuild the vHTLC script, but recovery may fail if the daemon's current identity key no longer matches the key that created the old script.

## Desired behavior

`darepo-client` should support recovery of old vHTLCs after client identity-key rotation.

A possible shape:

1. Persist the recovery signing identity used at swap creation/funding time.
   - Store the public key that is already part of the vHTLC policy.
   - Store enough signing identity/version/locator metadata to know which signer should be used later.

2. Make recovery use the persisted signing identity.
   - Do not assume the current daemon identity key is the old swap key.
   - If the old key is still available, sign with it.
   - If the key rotated through a supported mechanism, use the defined key-history/delegation path.
   - If recovery cannot sign because the old key is unavailable, fail with a clear actionable reason.

3. Coordinate with indexer auth.
   - Script/outpoint lookup and recovery flows should still be able to query old vHTLCs created under an old client identity.
   - This may require a durable identity history or delegation model in the indexer/auth layer.

4. Add regression coverage.
   - Create/fund a vHTLC under identity key A.
   - Simulate restart/rotation to identity key B.
   - Verify recovery uses key A or an explicit rotation mapping, not the current identity key by accident.
   - Verify old-script lookup remains authorized under the chosen indexer auth model.

## Related

- lightninglabs/darepo#498 was opened first in the root repo, but this client recovery work belongs here.
- Related symptom/context: lightninglabs/darepo-client#538.
- Related operator-key binding work: lightninglabs/darepo-client#519.

PR #964, opened 2026-07-16
multi: Add configurable reorg recovery foundation

ellemouton:
## What this PR does

Teaches existing client subsystems that confirmed chain facts can later
disappear. It does **not** enable one-confirmation VTXO spending — it
establishes the recovery behaviour that sits underneath that feature.

It makes these operations recoverable:

- Unilateral exit and unroll transactions.
- Boarding-input sweeps.
- External-spend and broadcast observations.
- Restart after a reorg.

Instead of treating "once confirmed" as permanent, the client persists the
evidence and rechecks it. If the fact disappears, the subsystem returns to the
appropriate earlier state.

A key invariant: **broadcast ambiguity cannot resurrect a VTXO.** A VTXO is
relived only with positive evidence that its spending transaction was
definitely not broadcast.

It also introduces the configurable reorg-safety horizon `H` (default 30);
terminal chain finality is computed as `H + 1`. `reorgsafetydepth=0` resolves to
the safe default; the horizon is capped at 144 (the deepest reversible watch
window supported by every current backend).

## Review focus

- Does each recovery checkpoint survive restart correctly?
- What precise evidence permits a VTXO to be relived?
- Are the new durable messages TLV-compatible with rolling upgrades?
- Is the distinction between *seen*, *confirmed*, and *final* applied
  consistently?
- Does every ambiguous broadcast state fail closed?

## Out of scope

- Does not expose VTXOs provisionally.
- Does not create the lineage authority (that is #980).
- Defines a later protocol version but does not advertise it.

---
_Full design & diagrams: `REORG_SAFETY_PR_REVIEW_GUIDE.md` §4.1. Overall
architecture and PR stack: [lumos#454](https://github.com/lightninglabs/lumos/issues/454)._

### DIFF
--- a/waved/config.go
+++ b/waved/config.go
@@ -176,6 +176,16 @@ const (
 	// creating an excessive number of simultaneous cryptographic jobs.
 	MaxSigningWorkers = 64
 
+	// DefaultReorgSafetyDepth is the deepest chain replacement the daemon
+	// keeps provisional and promises to recover from once reorg-safe v2 is
+	// enabled. Terminal finality begins one confirmation later.
+	DefaultReorgSafetyDepth uint32 = 30
+
+	// MaxReorgSafetyDepth is the largest policy horizon supported by every
+	// current chain backend. LND's notifier retains reversible watches for
+	// at most 144 disconnected blocks.
+	MaxReorgSafetyDepth uint32 = 144
+
 	// DefaultWalletType is the default wallet backend. The "lwwallet"
 	// backend uses an in-process lightweight wallet backed by
 	// btcwallet and Esplora, requiring no external lnd node.
@@ -337,6 +347,13 @@ type Config struct {
 	// parallel. Zero selects a backend-aware default and one preserves the
 	// original serial behavior.
 	SigningWorkers int `mapstructure:"signingworkers"`
+
+	// ReorgSafetyDepth is the deepest chain replacement for which evidence
+	// remains provisional and recoverable. Terminal finality begins at
+	// ReorgSafetyDepth+1 confirmations. Zero uses
+	// DefaultReorgSafetyDepth. This policy is not advertised until the full
+	// v2 safety capability and depth negotiation are enabled.
+	ReorgSafetyDepth uint32 `mapstructure:"reorgsafetydepth"`
 
 	// RegistrationTimeout is the maximum wall-clock duration to
 	// wait for the server's RoundJoined admission watermark after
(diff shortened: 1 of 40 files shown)


ellemouton:
Rebased the three recovery commits onto the rewritten #895 head `233abdd6`; the PR diff is clean again and contains only Client PR 2. No recovery code changed during the rebase. Post-rebase verification: `go test -race ./unroll ./wallet ./waved ./chainsource -count=1` and `git diff --check` passed. Fresh CI is running on `f27fde5d`.

ellemouton:
@claude please review this PR — Client PR 2 of the reorg-safety stack (configurable reorg recovery foundation), now rebased directly onto `main` after #895 merged. Focus on the unilateral-exit and boarding-sweep reorg/restart recovery correctness, the fail-closed-on-missing-evidence behavior, and the configurable `reorgsafetydepth` (default 30) / terminal-depth wiring.

ellemouton:
@claude review this

ellemouton:
@claude review this — the release-blocking upgrade finding is fixed in current head `ce345672`. Started legacy checkpoints with no relive-safety TLV now decode fail-closed as unsafe, while current started checkpoints encode an explicit boolean so a future authoritative clear remains distinguishable. The regression test removes the new record from a started checkpoint and pins `ReliveUnsafe=true`; `go test -race ./unroll` passes.

The remaining notes are value-safe liveness boundaries, not release blockers for this stack: authoritative-negative clearing is deliberately deferred, and unavailable chain evidence deliberately stalls rather than guesses. The boarding retry/coalescing items should stay follow-up work rather than expanding this reorg foundation. Please re-review the current head.

ellemouton:
@claude review this

ellemouton:
@claude review this — final head is now rebased onto current main, including the round-status follow-up required by Lumos master.

ellemouton:
@claude review this — current restacked head `604b856d` after rebasing the full stack onto the latest upstream base. Please verify prior findings remain addressed and flag any new blocking issues.

waved/config.go
```diff
@@ -1361,6 +1391,21 @@ func (c *Config) validateWalletConfig() error {
 	return nil
 }
 
+// reorgSafetyDepth resolves the daemon's configured recovery horizon.
+func (c *Config) reorgSafetyDepth() uint32 {
+	if c.ReorgSafetyDepth == 0 {
```
Roasbeef, line 1446: At the client layer, zero means use the 30-block default. Lumos #626 gives an explicit zero a disable/immediate-finality meaning. We can't advertise v2 with two meanings for the same knob: one deployment config can otherwise produce 31-conf client finality and first-conf server finality. This is fine as disabled foundation, but capability negotiation/config validation needs to carry the same nonzero H on both sides and reject a mismatch before activation.

Roasbeef (commented):
In this review, I'm treating #964 as the client-side recovery foundation. The H/H+1 arithmetic and reversible chainsource lifecycle look sound as disabled substrate. The remaining comment is a cross-stack contract issue that needs to be closed before the capability is advertised.

ellemouton:
@claude the reorg-review feedback on this stack has been addressed and **folded into the branch** (force-pushed, no `fixup!` commits). Could you go through **Roasbeef's review comments on this PR** and assess whether each one is now resolved by the current code, flagging any that are **not** addressed?

Context for cross-PR fixes: some review points were folded into existing commits here, while genuinely new work (e.g. the consumer-edge redrive after MarkForfeited, and the F-H1 received-VTXO→commitment binding + OOR base-input terminator) landed on the tip PR **#990 `reorg-lineage-producers`**. The full per-item changelog + mapping is in the epic status comment: https://github.com/lightninglabs/lumos/issues/454#issuecomment-5109293790

waved/config.go
ellemouton: Reconciled on the server side: lumos #626 now also resolves an explicit zero to the default (`0 → DefaultReorgSafetyDepth`), so both layers give zero the same meaning. No client change needed — the client already resolved zero to the default.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #972, opened 2026-07-16
oor+db: accept same-commitment multi-leaf VTXO ancestry

### LINKED ISSUE #969: Balance drops to zero after successful offchain send
I dispatched a Lightning payment to an external node, which successfully completed.

However, my balance was then shown as zero.

In the activity log I can see my successful offchain send (2101 sats), and above, another unrecognized transfer over my remaining balance.

```
LAST UPDATE              KIND     STATUS    AMOUNT  FEE  PHASE      ID                                                                      NOTE
2026-07-16 14:10:48.000  SEND     COMPLETE  -3438   0    confirmed  ledger-18                                                               
2026-07-16 14:10:52.000  SEND     COMPLETE  -2101   1    confirmed  15281e1fdc2a461ca05898a62ba20e6cd032dad4b17f27978f14a7568e9ca88f
```

I noticed this in the logs and have attached the full logs below.

```
2026-07-16 14:11:20.088 [WRN] OORC: Session commit failed err="invalid incoming VTXO ancestry: fragment 1 carries duplicate commitment txid 0b17fc13289ed8d98348debe62683333a05500c0c03840192893301f2ced299f" session_id=57a717405398fb056570c4278777d4e1a8d196b545062de7a8a3a9d26dae61d3
2026-07-16 14:11:20.089 [WRN] OORC: Durable actor Tell message failed err="invalid incoming VTXO ancestry: fragment 1 carries duplicate commitment txid 0b17fc13289ed8d98348debe62683333a05500c0c03840192893301f2ced299f" actor_id=oor-session-57a717405398fb056570c4278777d4e1a8d196b545062de7a8a3a9d26dae61d3 delivery_id=019f6cc4-93b5-7c10-963e-75ce4d75b464 msg_type=DriveEventRequest attempts=4 effective_attempts=5
```

There are no vtxos listed in `wavecli ark vtxos list`

[2026-07-16_waved.log](https://github.com/user-attachments/files/30102790/2026-07-16_waved.log)

Roasbeef:
In this PR, we fix #969, where a successful offchain send zeroed out the
sender's balance: the change VTXO's incoming session was rejected with
`invalid incoming VTXO ancestry: fragment 1 carries duplicate commitment
txid`, rolled back on every retry, and the coin was never materialized.

The root cause is a client/server contract drift. Since darepo `4af6edb3`
(multi-tree ancestry in `combineVirtualLineage`), the indexer emits one
`AncestryPath` per (commitment, batch-tree path): an OOR spend whose
inputs sit at *different leaves of the same commitment tree* legitimately
carries several fragments anchored at the same commitment txid, one
root-to-leaf path per leaf. The client, however, still enforced
one-fragment-per-commitment at three layers: the receive-side cross-check
(`oor.validateIncomingAncestry`), the persistence codec
(`db.upsertAncestryPaths`), and a schema-level
`UNIQUE(vtxo_outpoint, commitment_txid)` on `vtxo_ancestry_paths`. Any
incoming transfer whose inputs shared a round hit the wall at commit
time.

Fragment identity is now the (commitment txid, tree path) pair, computed
by the new `db.AncestryFragmentKey` helper and shared by all validators.
Migration 000014 rebuilds `vtxo_ancestry_paths` without the UNIQUE
constraint (SQLite can't drop a table constraint in place; one shared
`.sql` serves both backends via the BLOB→BYTEA replacement layer). The
sweep over remaining consumers found one silent-drop variant, also fixed
here: `waved.recoveryAncestry` deduped the synthesized vHTLC recovery
target's ancestry by commitment alone, dropping the second leaf's exit
path. Everything else already tolerates the multi-leaf shape: the unroll
proof assembler dedupes byte-identical overlapping nodes and rejects only
conflicting ones, the fraud watch model keys by outpoint, and exit
feasibility just counts fragments.

changes. A server-side systest exercising the same-commitment multi-leaf
receive end-to-end will follow in a darepo PR that re-pins the client
submodule to this branch.


Roasbeef:
Pushed three follow-up commits out of an adversarial review pass plus end-to-end validation on the server side:

- `oor: reject conflicting duplicate lineage nodes at receive` — with per-commitment uniqueness gone, a malicious operator could pad the ancestry with a near-duplicate path (same txid, different bytes, e.g. a stripped signature on a shared prefix node) that persisted cleanly and only failed proof assembly at unilateral-exit time. The receive gate now enforces the proof assembler's same-txid-must-be-byte-identical rule across fragments, so that shape fails fast with a typed error at the boundary. Honest multi-leaf fragments share byte-identical prefix nodes, so nothing legitimate is rejected.
- `unroll: dedupe overlapping ancestry in the exit-cost estimate` — feasibility used to demand one concurrent wallet input per raw fragment and sized every fragment's transactions independently; overlapping same-commitment fragments now dedupe by txid (and CPFP parents by root txid), so a padded ancestry can no longer block an affordable exit.
- `db+waved: document fragment-key scope, pin RPC-seam regression` — documents that `AncestryFragmentKey` deliberately excludes `InputIndices`, and adds a test driving a synthesized two-path same-commitment indexer response through the full metadata resolution path.

Cross-repo validation: a darepo systest (`TestOORSameCommitmentMultiLeafAncestryE2E`, PR to follow) builds the #969 topology live — one round mints two leaves, each hops through a 1-in/1-out self-send, then both hop coins are co-spent with change. Against this branch the change and the recipient coin both materialize with two same-commitment fragments (test passes in ~27s); against `main` the change never materializes and the log fills with `carries duplicate commitment txid`, reproducing the issue exactly.

ellemouton:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
Addressed the review feedback (both automated passes), folded in as fixups so the commit structure is unchanged:

1. **`treePathVBytes` test-only** — moved into `feasibility_test.go` with a note that it survives only as a test oracle; production sizing routes through the dedup loop in `recoveryEstimate`.
2. **Receive-gate scope** (observations 2+3) — `checkFragmentNodeConsistency` now documents that it is deliberately scoped to operator-supplied ancestry fragments and is best-effort (skips unserializable nodes), NOT a pre-assembly dry-run: the assembler's conflict map additionally spans client-resolved OOR-chain nodes and stays the authoritative check at exit time.
3. **Postgres migration confirmed** — ran the db suite under `-tags test_postgres` (real dockerized Postgres): `TestMigrationSteps` walks the full up/down ladder including 000014's table rebuild through the BLOB→BYTEA replacement layer, and `TestVTXOPersistenceStoreSameCommitmentMultiLeafRoundTrip` persists and loads the same-commitment multi-leaf shape on Postgres. All green.
4. **`MaxAncestryPaths` nit** — no change, per the second pass's correction: the constant exists (`vtxo/incoming_ancestry.go`) and is the relevant indexer-response cap.

Also on the branch since the last review: `multi: stop treating txconfirm finality as an unroll failure`. The lumos itest matrix on the systest PR (lightninglabs/lumos#680) exposed drift from the reorg-observe train: every EnsureConfirmedReq subscriber now receives the full lifecycle, but the unroll adapter mapped the new TxFinalized (and best-effort TxReorged) to a terminal TxFailedMsg — so the moment a backend synthesized finality for a proof node, the whole unroll job failed. TxFinalized now replays as an idempotent confirmation, TxReorged is dropped (superseded by the next reliable event), and the previously-failing multi-input OOR itests pass against this head.

PR #973, opened 2026-07-16
multi: allow ancestry fragments to share a commitment txid

### LINKED ISSUE #969: Balance drops to zero after successful offchain send

ellemouton:
Fixes #969.

## Symptom

A user sent a 2101-sat Lightning payment that succeeded, but their balance
then dropped to **zero**. The activity log showed the expected `-2101` send
plus an "unrecognized" `-3438` debit, and the daemon logged:

```
[WRN] OORC: Session commit failed err="invalid incoming VTXO ancestry:
fragment 1 carries duplicate commitment txid 0b17fc13..."
```

## Root cause

The send's coin-selection picked **two VTXOs that both descend from the same
commitment tx** (`0b17fc13…`), totalling 3438 sats, to pay 2101 (→ 1337
change). The OOR Ark tx spent both; the **self-change VTXO's ancestry**
therefore legitimately carries **two fragments with the same commitment txid
but different leaves** — each input VTXO sits at a different leaf of that one
commitment tree, so each needs its own root→leaf path.

Several client layers assumed *one ancestry fragment per commitment tx* and
either rejected or silently dropped the second fragment. The self-change
materialization was rejected outright, so the 1337 change was never credited
and the balance fell to zero (the `-3438` entry is the OOR input debit; the
missing `+1337` change credit is the lost balance).

An audit of every ancestry consumer found the wrong invariant in exactly four
places (everything else — unroll proof assembly, fraud watch, the TLV codec,
the DB read path — already iterates per-fragment and keys by tree-node/outpoint
hashes, so it is safe):

## Fix

Relax the invariant to key on the **(commitment txid, ark input index)** pair
rather than the commitment alone. A given input descends from a given
commitment through exactly one leaf, so a repeated pair is redundant/malformed,
but distinct leaves of one commitment (different inputs) are legitimate.

- **`oor` `validateIncomingAncestry`** — rejects only a repeated
  `(commitment, input)` pair; accepts distinct leaves of one commitment. This
  is the correct invariant at the incoming boundary, where all fragments of a
  single OOR tx share one input-index frame of reference.
- **`db`** — drop the `UNIQUE(vtxo_outpoint_hash, vtxo_outpoint_index,
  commitment_txid)` schema constraint via a table rebuild (migration `000014`;
  SQLite has no `DROP CONSTRAINT`, and the rebuild is dialect-portable). Row
  identity is already the `PRIMARY KEY (…, path_order)`. `upsertAncestryPaths`
  drops its mirror dedup: the codec also persists synthesized vHTLC
  recovery-target descriptors that aggregate fragments across roots (whose
  per-root input indices aren't comparable), so cross-fragment uniqueness can't
  live at this layer — it stays at the incoming boundary above.
- **`waved` `recoveryAncestry`** — no longer de-duplicates root fragments by
  commitment txid. That silently dropped a divergent leaf and left the
  synthesized vHTLC recovery target unable to prove that input on-chain (a
  latent fund-safety bug at unilateral-exit time). Roots are already deduped by
  outpoint upstream, so concatenation carries no true duplicates.

This is client-only: the operator/indexer already produces correct per-leaf
fragments (the client received two, then rejected them).

## Testing

- **Reproduction (unit):** `TestValidateIncomingAncestryInputCoverage/
  two_inputs_share_a_commitment` fails on `main` with the exact issue error and
  passes with the fix.
- `validateIncomingAncestry` still rejects a genuinely redundant duplicate
  fragment; `upsertAncestryPaths` persists shared-commitment fragments;
  `recoveryAncestry` preserves every leaf (regression for the silent drop).
- Full `oor` + `db` suites pass (sqlite); `make sqlc` regenerates the
  consolidated schema (the `UNIQUE` is gone).

## Notes

- **Fix-forward only.** This prevents new occurrences; recovering change VTXOs
  that were already stranded before the fix (like the reporter's 1337 sats) is
  tracked separately. In principle a fixed client that re-queries the indexer
  and re-materializes should recover it on its own — to be confirmed.
- A real-daemon end-to-end test (two same-commitment inputs → send → assert the
  change materializes) is a good follow-up; the current systest fixture seeds a
  single VTXO, so the reproduction here is at the unit/validation layer.

### DIFF
--- a/db/ancestry_codec.go
+++ b/db/ancestry_codec.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/btcsuite/btcd/chainhash/v2"
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/lightninglabs/neutrino/cache"
 	"github.com/lightninglabs/neutrino/cache/lru"
@@ -135,15 +134,17 @@ func (c *ancestryTreeCache) getOrDecode(treePath []byte) (*tree.Tree, error) {
 // matches the round-create path where the VTXO manager has not yet
 // filled in finalized lineage.
 //
-// The function performs three defensive checks before any row is
-// inserted: the slice is at most maxAncestryRowsPerVTXO long, every
-// CommitmentTxID is pairwise distinct, and (implicitly, by deriving
-// path_order from the loop index) path_order is a contiguous run
-// starting at 0. These constraints are also enforced at the schema
-// level (UNIQUE on commitment_txid, CHECK on path_order), but
-// surfacing them in Go means a future caller bypassing
-// BuildIncomingVTXODescriptor sees a clear error rather than a
-// constraint violation hidden behind sqlc plumbing.
+// The function bounds the slice at maxAncestryRowsPerVTXO and derives
+// path_order from the loop index (a contiguous run starting at 0); row
+// identity is the PRIMARY KEY (..., path_order). It does NOT enforce any
+// cross-fragment commitment/input uniqueness: distinct leaves of one
+// commitment tree share a commitment_txid (wavelength#969), and a
+// synthesized recovery-target descriptor aggregates fragments from several
+// roots whose per-root input indices are not comparable, so both are
+// legitimate here. The structural invariant that each input descends from a
+// commitment through exactly one leaf is enforced upstream at the incoming
+// boundary (validateIncomingAncestry), the only producer for which
+// per-fragment input indices share a single frame of reference.
 func upsertAncestryPaths(ctx context.Context, q ancestryStore,
 	outpointHash []byte, outpointIndex int32,
 	ancestry []vtxo.Ancestry) error {
@@ -151,15 +152,6 @@ func upsertAncestryPaths(ctx context.Context, q ancestryStore,
 	if len(ancestry) > maxAncestryRowsPerVTXO {
 		return fmt.Errorf("ancestry has %d rows, max %d", len(ancestry),
 			maxAncestryRowsPerVTXO)
-	}
-
-	seen := make(map[chainhash.Hash]struct{}, len(ancestry))
-	for i, a := range ancestry {
-		if _, dup := seen[a.CommitmentTxID]; dup {
-			return fmt.Errorf("ancestry[%d] duplicate "+
-				"commitment_txid %s", i, a.CommitmentTxID)
-		}
-		seen[a.CommitmentTxID] = struct{}{}
 	}
 
 	err := q.DeleteVTXOAncestryPaths(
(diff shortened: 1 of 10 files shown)


ellemouton:
@claude review this

ellemouton:
Alternative schema strategy in #979: same code fix, but it drops the `UNIQUE` by editing the original `000004` migration in place instead of adding a rebuild migration — simpler, but only takes effect on a fresh DB (assumes we're OK resetting pre-release client databases). Pick whichever we prefer; I'll close the other.

ellemouton:
Closing in favor of #972, which is more complete. Same root cause and same migration-000014 rebuild, but #972: (1) keys fragment identity on `(commitment, tree-path)` via a shared `AncestryFragmentKey` helper used across all layers (rather than my `(commitment, input-index)` proxy, which forced dropping the db-codec check); (2) adds an adversarial conflicting-node check; and (3) importantly, also fixes an `unroll/feasibility.go` over-count for overlapping same-commitment fragments that this PR missed (over-reserving fees/wallet-inputs could refuse an affordable exit). Recommend we take #972.

PR #979, opened 2026-07-17
multi: allow ancestry fragments to share a commitment txid (DB-reset variant)

ellemouton:
Alternative to #973 for **#969** — same code fix, simpler schema change.

## Context

Both this PR and #973 fix the same bug: an OOR Ark tx that spends two input
VTXOs from the **same round's commitment tx** produces a VTXO whose ancestry
carries two fragments with the same commitment txid (different leaves), which
several layers wrongly rejected — stranding the change and dropping the balance
to zero. See #973 for the full root-cause write-up.

## What differs from #973

The **code fix is identical** (`oor` validation, `db` codec, `waved` vHTLC
recovery — all keyed on the `(commitment, ark input index)` pair instead of the
commitment alone). Only the **schema change** differs:

- **#973** adds a forward migration (`000014`) that rebuilds the
  `vtxo_ancestry_paths` table to drop the `UNIQUE(…, commitment_txid)`
  constraint. This **preserves existing databases** (SQLite has no
  `DROP CONSTRAINT`, hence the rebuild), at the cost of a new migration.
- **This PR** assumes existing (pre-release) client databases **can be reset**,
  so it simply **edits the original `000004` migration in place** to remove the
  `UNIQUE` line. No new migration, no table rebuild, no `LatestMigrationVersion`
  bump.

Trade-off: minimal and clean, but it **only takes effect on a fresh database**.
An already-migrated client keeps the old constraint (golang-migrate tracks
applied versions by number, not content) until its DB is recreated — so this is
only viable if we're comfortable wiping all client DBs.

## Why the constraint was wrong (recap)

`commitment_txid` is the **batch/round root** txid — deliberately *shared* by
every VTXO minted in that round. Requiring it unique per ancestry row forbade
the legitimate "two coins from one round spent together" case. Row identity is
already the `PRIMARY KEY (…, path_order)`; the structural per-leaf check lives
in `validateIncomingAncestry` at the incoming boundary.

## Testing

Same regression tests as #973: `validateIncomingAncestry` accepts two
same-commitment fragments (and still rejects a redundant one);
`upsertAncestryPaths` persists shared-commitment fragments; `recoveryAncestry`
preserves every leaf. `make sqlc` regenerates the consolidated schema (the
`UNIQUE` is gone); full `oor` + `db` suites pass, including the migration tests
against the edited `000004`.

Pick whichever schema strategy we prefer; I'll close the other.

### DIFF
--- a/db/ancestry_codec.go
+++ b/db/ancestry_codec.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/btcsuite/btcd/chainhash/v2"
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/lightninglabs/neutrino/cache"
 	"github.com/lightninglabs/neutrino/cache/lru"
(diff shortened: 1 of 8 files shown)


ellemouton:
Closing in favor of #972 (see #973 for the rationale). #972 uses the safe rebuild migration, so the DB-reset shortcut this PR took isn't needed.

PR #987, opened 2026-07-17
multi: Preview refresh fees and gate refresh on consent

### LINKED ISSUE #986: wavecli: ark vtxos refresh gives no fee visibility or warning before charging
## Summary
`wavecli ark vtxos refresh` queues a refresh that incurs a real operator fee, but the command never estimates, previews, or warns about that fee, and there is no confirmation step. Even `--dry_run` returns no fee. The user only learns the amount *after* it's charged, via fee history. We should surface the fee in the refresh path (at minimum under `--dry_run`).

## Current behavior (verified on `main` @ `e3868c99`)
- `RefreshVTXOsResponse` (`waverpc/daemon.proto`) has only `queued_outpoints` and `status` (`"queued"` | `"preview"`) — **no fee field**.
- `--dry_run` returns `status: "preview"` + the outpoints — still no fee.
- `vtxosRefresh` (`cmd/wavecli/waveclicommands/cmd_vtxos.go`) makes **no `EstimateFee` call** and prints only the JSON response. There is **no confirmation prompt or `--yes`** on refresh (unlike `vtxos leave`, whose `--all` prompt warns about moving funds on-chain — and even that isn't a fee warning).
- The binding fee is set by the operator at seal time and **auto-accepted** against the client cap `MaxOperatorFeeSat` (default 1,000,000 sat / 0.01 BTC, fail-closed) with no per-refresh confirmation.
- The charged amount only becomes visible afterward via `GetFeeHistory` (a `FeeTypeRefresh` ledger entry).

The fee is non-trivial: a refresh is priced as **on-chain share + liquidity (cost-of-capital, ∝ δ, δ_min-floored) + margin**.

## Existing partial mitigation
A timing-aware estimate exists, but only as a separate, manual command: `wavecli ark fees estimate --amount=<sats> --boarding=false --remaining-blocks=<δ>`. The user has to look up each VTXO's amount and remaining blocks themselves — it's keyed on amount+δ, not on an outpoint, and it isn't wired into the refresh flow.

## Impact
A user (or an agent) can run `ark vtxos refresh` and be charged — up to the 0.01 BTC cap — with no heads-up. This is a poor UX, undercuts "fees are transparent," and blocks honest product copy: because the CLI neither shows nor warns about the fee, we can't claim a user is informed before a manual refresh. (Interim honest messaging: "a small, capped fee may apply, visible in your activity afterward.")

## Proposed changes
1. **Surface the fee under `--dry_run`.** It already validates without queuing — return an estimated fee there: total + per-outpoint breakdown (liquidity / on-chain share / margin), reusing `EstimateFee`.
2. **Make the estimate outpoint-aware.** The daemon already knows the selected VTXOs (explicit outpoints, or all live VTXOs for `--all`); it can look up each one's amount + remaining blocks and estimate automatically — removing the manual `amount`/`remaining-blocks` step.
3. **Add fee fields to `RefreshVTXOsResponse`** (at least on the dry-run preview): estimated total + breakdown.
4. **(Optional) Print the estimate on a real refresh**, and/or gate large refreshes behind a confirmation consistent with the `leave --all` pattern — non-interactive-safe (require `--yes`/`--dry_run` when stdin isn't a TTY, so agents aren't blocked).

## Acceptance criteria
- `ark vtxos refresh --dry_run` returns an itemized estimated fee for the selected VTXO(s) with no manual amount/δ input.
- The estimate is clearly labeled advisory (binding fee is set at seal time and may differ).
- Non-interactive/agent invocation is not broken by a blocking prompt.

## Additional considerations
- **Free-late-refresh awareness:** once the free-late-refresh feature lands (companion PRs lightninglabs/wavelength#953 / lightninglabs/lumos#675), the estimate should reflect the waiver (show ~0 for a would-be-free late, complete-cohort refresh) rather than over-quoting the paid price — the current `EstimateFee` is amount+δ only and has no cohort/window awareness. Related pricing decision: lightninglabs/lumos#603.
- Not necessarily launch-blocking, but it gates the "informed before you refresh" claim.

## References
- CLI: `cmd/wavecli/waveclicommands/cmd_vtxos.go` (`vtxosRefresh`)
- RPC: `waverpc/daemon.proto` (`RefreshVTXOs`, `RefreshVTXOsResponse`, `EstimateFee`, `GetFeeHistory`)
- Companion free-late-refresh: lightninglabs/wavelength#953, lightninglabs/lumos#675
- Pricing-basis decision: lightninglabs/lumos#603

levmi:
In this PR, we fix #986, where `ark vtxos refresh` queued a fee-incurring refresh with no estimate, warning, or consent anywhere in the flow — the binding seal-time quote is auto-accepted against `maxoperatorfeesat`, so the user only learned the charge afterwards via the fee history.

A dry-run refresh now returns an itemized advisory fee estimate for the selected VTXOs. The daemon resolves each target's amount and remaining lifetime from its own store and chain tip (no more manual `--amount` / `--remaining-blocks` input), fetches the operator's liquidity / on-chain share / margin breakdown through the existing `EstimateFee` proxy (deduped and validated per quote), and applies the free-late-refresh waiver locally — the operator's `EstimateFee` prices every refresh as paid, so a selection fully inside the advertised window previews with an explicit zero total while the per-outpoint rows keep the ordinary paid quote to show what the waiver saves (all-or-nothing, mirroring the seal-time `isFreeRefresh` rule from lightninglabs/lumos#675). Estimate failures degrade all-or-nothing to an `estimate_error` field rather than failing the preview, and the total uses explicit proto presence so a degraded estimate is absent on the wire, never a zero an agent could misread as free.

A real refresh is now gated on consent on every surface. The CLI mirrors the `leave --all` pattern: an interactive TTY shows the estimate and prompts (on stderr, preserving the stdout-JSON stream split — the pre-existing leave prompt moves to stderr in its own commit for symmetry), `--yes` skips the gate for scripted use, and non-interactive stdin refuses with an actionable `INVALID_ARGS` error instead of blocking on a prompt an agent cannot answer. The MCP tool — which calls the daemon directly and never ran the CLI gate — enforces the same contract through a new `yes` argument with an immediate, actionable error, and its description now names the fee and the preview-first flow, matching what the schema registry promises.

The dry-run path also becomes an honest validity probe: it moves ahead of the wallet-ready gate for parity with the `LeaveVTXOs` H-5 ordering (whose comment already claimed this symmetry), unknown and non-live explicit outpoints now fail `InvalidArgument` instead of echoing back as a plausible preview, repeated outpoints are collapsed so they can neither double-count in the estimate nor register doomed duplicate forfeit pairs, and an empty `--all` preview honestly reports `"preview"`. The real refresh path is unchanged beyond the shared selection canonicalization, and a new wallet-actor test pins its `--all` expansion.

Downstream surfaces are unaffected: `swapdk-server` reads only `queued_outpoints` + `status` from refresh responses, and the SDK passes the proto through verbatim. Version skew is handled in both directions — a new CLI against a pre-feature daemon warns that no estimate came back, and old clients simply ignore the new fields. A follow-up issue against lumos for a waiver-aware / outpoint-keyed server-side estimate will be filed once this lands.

See each commit message for a detailed description of the individual changes.

## Validation

- `make unit pkg=./waved` (also with `tags="test_sqlite"`)
- `make unit pkg=./cmd/wavecli/waveclicommands`
- `make unit pkg=./waverpc`
- `go test -race ./waved/ -run TestRefresh -count=1` and `go test -race ./cmd/wavecli/waveclicommands/ -count=1`
- `make fmt-changed-check && make tidy-module-check && make sample-conf-check`
- `make rpc` (regenerated output committed; tree clean on re-run)
- Every commit builds and vets standalone (verified in a throwaway worktree)
- Real-binary smoke: `--help`, `schema ark.vtxos.refresh`, non-TTY refusal exits 2 with the `INVALID_ARGS` envelope before any RPC reaches the daemon, `--yes` proceeds straight to dispatch

fix #986

### DIFF
--- a/.claude/skills/waved.md
+++ b/.claude/skills/waved.md
@@ -128,7 +128,10 @@ surfaces the raw waverpc methods underlying them.
 # Raw VTXO inventory + lifecycle
 wavecli ark vtxos list --no-tls
 wavecli ark vtxos list --status live --min_amount 10000 --no-tls
-wavecli ark vtxos refresh --all --no-tls
+# A real refresh is fee-gated: preview with --dry_run, consent with
+# --yes (required on non-interactive stdin).
+wavecli ark vtxos refresh --all --dry_run --no-tls
+wavecli ark vtxos refresh --all --yes --no-tls
 
 # Raw transaction history (the wallet-shaped feed is `activity`)
 wavecli ark listtransactions --no-tls
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -103,7 +103,7 @@ wavecli ark vtxos list --ndjson | jq '.amount_sat'
 # Refresh — see BUGS_FOUND.md bug-1/bug-2; the refresh path is
 # currently not landing on the operator.
 wavecli ark vtxos refresh --outpoint txid:0 --dry_run
-wavecli ark vtxos refresh --outpoint txid:0
+wavecli ark vtxos refresh --outpoint txid:0 --yes
 ```
 
 ### Send operations
--- a/cmd/wavecli/waveclicommands/AGENTS.md
+++ b/cmd/wavecli/waveclicommands/AGENTS.md
@@ -152,6 +152,16 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/cmd/wave
 - `recovery escalate` refuses to run on non-interactive stdin unless
   `--yes` is passed — it never blocks on a y/N prompt an agent can't
   answer.
+- `ark vtxos refresh` is gated on fee consent: a real refresh fetches
+  the dry-run estimate and prompts with it on a TTY, and refuses on
+  non-interactive stdin without `--yes` (same posture as `leave --all`
+  and `recovery escalate`). The MCP tool enforces the same contract
+  through its `yes` argument — no prompt exists there, so a bare real
+  refresh returns an immediate actionable error. `--dry_run` previews
+  the itemized advisory estimate and never prompts. A failed estimate
+  degrades to a "still charged the seal-time fee" warning and its
+  total is absent on the wire (explicit proto presence) — it never
+  blocks the flow and is never rendered as a zero fee.
 
 ## Deep Docs
 
--- a/cmd/wavecli/waveclicommands/CLAUDE.md
+++ b/cmd/wavecli/waveclicommands/CLAUDE.md
@@ -152,6 +152,16 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/cmd/wave
 - `recovery escalate` refuses to run on non-interactive stdin unless
   `--yes` is passed — it never blocks on a y/N prompt an agent can't
   answer.
+- `ark vtxos refresh` is gated on fee consent: a real refresh fetches
+  the dry-run estimate and prompts with it on a TTY, and refuses on
+  non-interactive stdin without `--yes` (same posture as `leave --all`
+  and `recovery escalate`). The MCP tool enforces the same contract
+  through its `yes` argument — no prompt exists there, so a bare real
+  refresh returns an immediate actionable error. `--dry_run` previews
+  the itemized advisory estimate and never prompts. A failed estimate
+  degrades to a "still charged the seal-time fee" warning and its
+  total is absent on the wire (explicit proto presence) — it never
+  blocks the flow and is never rendered as a zero fee.
 
 ## Deep Docs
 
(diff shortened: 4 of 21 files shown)


levmi:
@claude review this PR

levmi:
Addressed the review findings in two follow-up commits:

- **#2** (`--all` + `--outpoint` silently dropping outpoints): b2c79024 extracts a shared `buildRefreshVTXOsRequest` builder (mirroring `buildLeaveVTXOsRequest`) with the mutual-exclusion guard, and routes the MCP tool through it so both surfaces get the same guard.
- **#3** (components not verified against `total_fee_sat`): bb825776 extends `refreshQuoteSane` to require the itemization to reconcile with the total (each component individually bounded by `MaxSatoshi` first, so the sum is wrap-free).

Per **#1**, the accumulation overflow flagged by Gemini is not reachable (the loop invariant keeps `total ≤ MaxSatoshi` at the top of each iteration and each vetted quote is `≤ MaxSatoshi`), so no change there. #4 left as-is (dedup keeps the common case cheap).

Roasbeef (approved):
LGTM 🎳

PR #995, opened 2026-07-18
sdk/swaps+swaprpc: settle credit-attach receives over the server-funded in-ark leg

Roasbeef:
In this PR, we teach the receive session to settle a credit-attach receive over the swap-server-funded in-ark leg from swapdk-server#233 (as revised on that issue). It stacks on #991: that PR made a credit-bound session hard-reject any in-ark event, since the legacy direct p2p shape can never carry the server's custodial credit; this one relaxes that rejection into full triplet validation for the new credit-shaped events, so the payment actually settles instead of failing clean.

On the wire, `InArkHtlcEvent` grows `requested_amount_sat` and `attached_credit_sat`, mirroring `OutSwapHtlcEvent`: a credit-shaped event carries the padded vHTLC amount in `amount_sat` together with the invoice amount and the credit portion. `RequestChannelIdRequest` grows `supports_in_ark_credit`, which this client now always advertises; the server records it on the receive intent and only publishes credit-shaped events to receivers that set it, so deployed clients without this code never see an event shape they would reject.

The receiver validates the triplet exactly like the out-swap acceptance path (requested must match the invoice, attached credit must match the plan, amount must equal the padded vHTLC) and terminal-fails on any mismatch. The rail split also moves the ACK decision off the settlement type alone: the legacy p2p rail keeps skipping the out-swap server ACK (the sender funded before the event was published), while a credit-shaped session sends it, because the swap server funds the padded vHTLC only after the receiver durably accepted the event. The claim policy is unchanged in shape: the event's sender key (the swap server's funding key on this rail) fills the sender slot and the Ark operator stays in the server slot, so funding validation and the claim path work as they do for every other padded vHTLC.

The sender side needs no changes at all: sub-dust pays were already forced onto the credit rail with its auto top-up, and that is exactly the rail the server settles internally on the other side (lightninglabs/swapdk-server#234).

Depends on #991; the first two commits are that PR's branch.

### DIFF
--- a/credit/messages.go
+++ b/credit/messages.go
@@ -336,6 +336,16 @@ type CreditOpSummary struct {
 	// WalletEntry transition. Always false for receive and redeem ops.
 	CreditOnly bool
 
+	// OORSessionID is the delegated Ark top-up OOR session id, when the
+	// operation funded credit through an OOR transfer. It lets the wallet
+	// activity view correlate the raw ledger row of the top-up transfer
+	// back to the credit operation it funded.
+	OORSessionID string
+
+	// TopupSat is the Ark top-up amount the operation funded to cover a
+	// pay shortfall, when known.
+	TopupSat int64
+
 	// LastError is the terminal failure reason, when failed.
 	LastError string
 }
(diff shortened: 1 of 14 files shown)


Roasbeef:
GitHub closed this un-reopenably when the stacked base branch was deleted at merge time (the head was also force-pushed during the rebase onto main). Continued in #996 with identical content rebased onto the merged #991.

PR #1013, opened 2026-07-21
release: Publish wasm runtime and mobile bindings on release

### LINKED ISSUE #1012: release: consider including wasm and mobile binding assets in releases
Would it be possible to include the wasm and mobile binding artifacts in the release process? The release build (`make release` / `make docker-release`) currently covers `waved` and `wavecli`, but a couple of asset sets that downstream consumers (including wavelength-sdk) rely on aren't part of it yet:

- **Browser wasm runtime**: `wavewalletdk.wasm`, `wavewalletdk.wasm.gz`, `wasm_exec.js`, and the sqlite worker assets. These are currently only built locally via `make wasm-wallet`.
- **Mobile bindings**: `Wavewalletdk.aar` and `Wavewalletdk.xcframework`. CI already builds these in `mobile-bindings.yml`, but only on `mobile-v*` tags, and the outputs are expiring workflow artifacts rather than release assets.

Right now anyone consuming these has to build them by hand or pull them from CI runs, which makes it hard to point the SDK at a stable set of release assets.

#713 added the gomobile bindings and left build/publish ownership as an open question; this issue proposes resolving that as part of the release process.

One possible approach: the wasm build is pure Go with no CGO, so it could probably slot into `scripts/release.sh` and the signed manifest without much trouble. The mobile bindings depend on host toolchains (NDK, Xcode), so it might make more sense to keep those in CI, trigger them on `v*` tags too, and attach the outputs to the GitHub release outside the manifest. Open to other ideas if there's a better way to structure this.

Roasbeef:
## Motivation

Closes #1012. The release build covers `waved` and `wavecli`, but the two
asset sets that `wavelength-sdk` depends on aren't published anywhere stable:
the browser wasm runtime is built only via `make wasm-wallet` locally, and the
gomobile bindings live as expiring CI artifacts under a separate `mobile-v*`
tag. Downstream consumers have had to build them by hand or scrape a CI run.

This PR wires both into the release process, along the split the issue
sketched: the wasm runtime goes to the hosted bucket the SDK already reads
from, and the mobile bindings become GitHub release assets on `v*` tags.

## What's in here

**wasm runtime → hosted bucket (`make wasm-publish`).** The SDK's web app
fetches the runtime at `<base>/<version>/<file>`, where the version segment is
its `RUNTIME_MANIFEST_VERSION` and must equal the release tag. Rather than
plumbing a GCP key into CI (the bucket has no CI writer wired), the new target
uploads with the release manager's own `gcloud` credentials at tag time:

```shell
make wasm-publish tag=<TAG> bucket=gs://<runtime-assets-bucket>
```

It builds via `wasm-wallet`, then uploads exactly the manifest file set to
`gs://<bucket>/<TAG>/`. The file list in `scripts/publish-wasm-assets.sh` is
the source of truth on the publishing side and tracks the SDK's
`RUNTIME_ASSET_FILES`. Assets go up as raw bytes with no content-encoding
transform, since `wavewalletdk.wasm.gz` is a pre-gzipped payload the consumer
stores verbatim.

**mobile bindings → GitHub release assets.** The Mobile Bindings workflow now
builds the `.aar`/`.xcframework` on `v*` tags too (not just `mobile-v*`), and a
new `publish` job attaches `Wavewalletdk.aar` and `Wavewalletdk.xcframework.tar.gz`
to the release for the tag. It creates the release when one doesn't exist yet,
so a tag push alone is enough, but stays compatible with a release manager
cutting the signed release out of band: the create is skipped when the release
already exists and the upload clobbers, so re-runs are idempotent. rc/alpha/beta
tags publish as pre-releases. Only the `publish` job carries `contents: write`.

**docs.** `docs/release.md` gains a section on both flows, and the mobile guide
picks up a pointer plus the `damobile` → `wavelength-mobile` repo rename.

## Notes

- The wasm publish is deliberately manual/local, per the decision to use local
  gcloud auth at the final tag rather than adding CI GCP credentials.
- A companion PR on `wavelength-mobile` repoints its fetch scripts at the new
  release assets and catches up the stale `darepo-client`/`darepod`/`Walletdk`
  naming.

## Testing

- `make wasm-publish` dry-run and guards verified; the built `bin/wasm/` set
  matches the script's file list exactly.
- `mobile-bindings.yml` passes YAML validation.
- All three commits pass `make commitmsg-lint`.

### DIFF
--- a/.github/workflows/mobile-bindings.yml
+++ b/.github/workflows/mobile-bindings.yml
@@ -2,12 +2,19 @@ name: Mobile Bindings
 
 # The mobile binding package compiles on every relevant PR (a fast, host-tagged
 # build), but the expensive gomobile .aar / .xcframework builds (which cross
-# compile the whole embedded daemon for every ABI) run only when a
-# `mobile-v*` tag is pushed, so routine merges do not trigger them.
+# compile the whole embedded daemon for every ABI) run only on a tag push, so
+# routine merges do not trigger them. Two tag namespaces drive the heavy build:
+# a `mobile-v*` tag for a mobile-only cut (artifacts only), and a `v*` release
+# tag, which additionally attaches the built `.aar` / `.xcframework` to the
+# GitHub release for that tag (see the `publish` job).
 on:
   push:
     tags:
       - "mobile-v*"
+      # Anchored to a digit so only real release tags (v0.1.0, v1.2.3-rc1)
+      # trigger the heavy build and the release-creating publish job, not any
+      # stray `v`-prefixed tag (e.g. `vendor-x`).
+      - "v[0-9]*"
   pull_request:
     paths:
       - "sdk/wavewalletdk/**"
@@ -54,11 +61,11 @@ jobs:
         run: go test -tags="${MOBILE_TAGS}" ./sdk/wavewalletdk/mobile/...
 
   ########################################################################
-  # Android .aar — tag builds only.
+  # Android .aar — tag builds only (mobile-v* or v*).
   ########################################################################
   android:
     name: Build Android .aar
-    if: startsWith(github.ref, 'refs/tags/mobile-v')
+    if: startsWith(github.ref, 'refs/tags/mobile-v') || startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
       - name: Git checkout
@@ -105,11 +112,11 @@ jobs:
           if-no-files-found: error
 
   ########################################################################
-  # iOS .xcframework — tag builds only (macOS runner for Xcode).
+  # iOS .xcframework — tag builds only, mobile-v* or v* (macOS for Xcode).
   ########################################################################
   ios:
     name: Build iOS .xcframework
-    if: startsWith(github.ref, 'refs/tags/mobile-v')
+    if: startsWith(github.ref, 'refs/tags/mobile-v') || startsWith(github.ref, 'refs/tags/v')
     runs-on: macos-latest
     steps:
       - name: Git checkout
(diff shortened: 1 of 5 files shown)


Roasbeef:
Addressed the review feedback plus a follow-up hardening pass (fixups squashed into the original three commits, force-pushed).

**From the Gemini review:**
- **Content-Type on upload** — every asset now uploads with an explicit `--content-type`: `.wasm` → `application/wasm` (covers both `wavewalletdk.wasm` *and* `sqlite3.wasm`), `.js` → `text/javascript`. One deliberate nuance: `wavewalletdk.wasm.gz` is labeled `application/gzip`, **not** `application/wasm` — tagging pre-gzipped bytes as wasm without `Content-Encoding: gzip` (which we intentionally don't set, since the SDK stores it verbatim) would hand a browser a broken module.
- **Validate before building** — `make wasm-publish` now checks `tag`/`bucket` (and that `bucket` is a `gs://` URI) *before* `$(MAKE) wasm-wallet`, so a missing/malformed arg fails instantly instead of after the multi-minute wasm build.

**Additional hardening (from a second review pass on the two untested paths):**
- **Immutable-prefix overwrite guard** — the script refuses to overwrite an already-published `<bucket>/<version>/` unless `FORCE=1`, so an accidental re-run can't silently replace bytes clients cache forever.
- **Post-publish verification** — after uploading, it re-lists the prefix and asserts all 8 files landed, so a mid-loop failure can't leave a partial, 404-serving version.
- **Auth/bucket pre-flight** — a cheap `gcloud storage ls` up front fails fast on an expired session or unreachable bucket instead of mid-upload.
- **`v*` trigger tightened to `v[0-9]*`** — now that the workflow *creates* a GitHub release, an accidental `v`-prefixed tag (e.g. `vendor-x`) no longer spawns one.
- **Provisional release notes** — the CI `publish` job no longer uses `--generate-notes` (which reads as a finished changelog). Since it usually wins the race against the manual signed-manifest flow, it seeds an explicitly-provisional note, and `docs/release.md` now tells the release manager to finalize the title/notes and attach the signed `waved`/`wavecli` manifest to that same release.

Issue #997, opened 2026-07-20
wavecli: CLI usability audit + suggestions (developers, vibe coders, agents)


levmi:
## Summary

A structured usability audit of `wavecli` was run against modern CLI design best practices, evaluating three audiences: **human developers**, **vibe coders** (people who paste commands and errors into LLMs), and **AI agents** driving the CLI programmatically (Claude Code, Codex, MCP clients). This master issue carries the consolidated findings and a prioritized, source-verified improvement plan. Individual workstreams below are intended to become linked sub-issues as they're picked up.

**Method + provenance:** best-practices research (clig.dev, 12-factor CLI, agent-ergonomics literature) → full audit of `cmd/wavecli` at `main` (87f228f6) → improvement plans drafted independently by two frontier models (Claude Fable 5 and GPT-5.6-Sol), then cross-examined over three adversarial debate rounds until consensus. Every feasibility claim was verified against source; file references are relative to `cmd/wavecli/waveclicommands/` unless noted.

**Headline:** the audit found an unusually strong base — structured error envelopes with semantic exit codes on every failure path, TTY-gated prompts, argv-free secrets, `EmitUnpopulated` proto-JSON. There is **no missing pillar; there is drift, gaps at the seams, and a closing window**: the command surface becomes LLM training data and cached agent context the day it launches, so the few justified breaking changes are cheap now and near-impossible later.

## Strategic themes

1. **Spend the pre-launch breaking window on the surface, now.** Two defects get an order of magnitude more expensive post-launch: the flag-casing split (`--max_fee` vs `--sweep-all`) and the global `--json`-means-*input* collision (today `--json '{...}'` is a raw request payload — an LLM-suggested `--json` can silently no-op or, worse, alter a send). These are the only proposed breaks; everything else is additive.
2. **Make the interface's self-descriptions true, and keep them true mechanically.** The schema registry and MCP metadata drift from the real cobra tree in eight confirmed classes. A schema that teaches agents wrong flags is worse than no schema. Fix once, then a CI parity test prevents all future drift.
3. **Finish the error and non-interactive contracts at the edges users hit first.** Remediation + retryability for daemon-down/missing-cred errors, a distinct approval-required exit class, dry-run success exiting 0 (today: exit 10 with an error-shaped envelope), a global `--no-input` guarantee, every prompt on stderr.
[...]

PR #1023, opened 2026-07-21
wavecli: Separate JSON input from output

darioAnongba:
Addresses P0-1 of #997.

## What changed

- make global `--json` a boolean machine-output flag
- move raw proto request input to `--request-json`, including generated `dev` RPCs
- make `activity` and `activity inspect` honor global `--json`
- keep activity table bytes identical on terminals and pipes; add only a one-line stderr hint for an unformatted pipe
- make `ark sweep list` parse raw request JSON instead of silently ignoring it
- rename schema metadata from `json_input` to `request_json_input`
- update CLI documentation and regression tests

## Concrete impact

Before, the conventionally machine-output-looking `--json` flag was raw input, so agent-generated commands could fail or alter requests. `ark sweep list --json ...` was worse: it silently ignored the payload. After this change, output and input directions are unambiguous and raw sweep-list requests either work or fail loudly.

This is intentionally breaking before launch for scripts that used `--json '{...}'`; the replacement is a spelling-only migration to `--request-json '{...}'`.

## Validation

- `make unit pkg=./cmd/wavecli/...`
- `go test ./cmd/wavecli/... -race`
- `make lint-changed-local`

### DIFF
--- a/cmd/wavecli/waveclicommands/root.go
+++ b/cmd/wavecli/waveclicommands/root.go
@@ -102,10 +102,15 @@ func newRootCmd(devMode bool) *cobra.Command {
 		"maximum duration for each daemon RPC; 0 disables the deadline",
 	)
 
+	pf.Bool(
+		"json", false, "emit machine-readable JSON output (raw "+
+			"request input uses --request-json)",
+	)
+
 	pf.String(
-		"json", "", "raw JSON request payload (maps directly to "+
-			"the RPC request proto); when set, bespoke flags "+
-			"are ignored",
+		"request-json", "", "raw JSON request payload (maps "+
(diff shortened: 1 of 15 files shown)


cmd/wavecli/waveclicommands/root.go
```diff
@@ -60,6 +61,16 @@ func newRootCmd(devMode bool) *cobra.Command {
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
+	cmd.SetFlagErrorFunc(func(_ *cobra.Command, err error) error {
```
darioAnongba: no need for such backwards compat additions.

Issue #1041, opened 2026-07-22
recv --offchain below the 1000-sat dust limit hangs ~30s then fails with DEADLINE_EXCEEDED — credit receive never completes on signet


levmi:
## Summary

On public signet, any offchain receive below the operator dust limit (1000 sats) fails after a ~30-second hang with `DEADLINE_EXCEEDED` from the `CreateCredit` RPC. Exactly 1000 sats succeeds instantly and repeatably; 999 and below never does. Per `docs/credit_system.md`, sub-dust receives are supposed to route to the credit subsystem precisely so that small amounts work ("They let a wallet transact sub-dust amounts without changing the operator dust limit") — instead the credit path is non-functional against the signet operator, and the failure surfaces as an opaque client-side timeout rather than an actionable error.

## Environment

- **OS / arch**: macOS, Darwin 24.6.0, arm64
- **Go**: go1.26.5 darwin/arm64
- **Daemon**: version=0.1.0 (tag `v0.1.0`, ff510b11), built with `make build-wavewalletrpc` (wavewalletrpc + swapruntime)
- **Network**: signet (public endpoints, auto-resolved)
- **Wallet backend**: lwwallet → default signet Esplora (`mempool-signet.testnet.lightningcluster.com`, HTTP 200)

## Config

No `waved.conf` — flags only, fresh wallet:

```
waved --network=signet --wallet.type=lwwallet --datadir=<datadir> \
  --rpc.listenaddr=localhost:12029 --rpc.gateway.listenaddr=localhost:12031 \
  --wallet.password_file=<pwfile>
```

## Failure sequence

CLI (identical result for amt = 1, 5, 10, 50, 100, 250, 500, 600, 999):

```
$ wavecli --network=signet --datadir=<datadir> --rpcserver=localhost:12029 \
    recv --offchain --amt 999
… ~30 s …
{"error": {"code": "DEADLINE_EXCEEDED", "message": "context deadline exceeded",
 "details": "recv invoice: start credit receive: create receive credit: create credit: CreateCredit RPC"}}
```

Control (same wallet, moments later):

```
$ … recv --offchain --amt 1000
{"invoice": "lntbs10u1p4xzr9ypp5…", …}          # instant, repeatable
```

[...]

PR #1045, opened 2026-07-22
Gracefully degrade sub-dust credit receive failures (#1041)

ellemouton:
## Problem

A sub-dust offchain receive (`recv --offchain --amt <dust>`) routes to the
operator credit subsystem. When the swap server does not complete the credit
request, the receive hung ~30s, then failed with an opaque `DEADLINE_EXCEEDED`,
left no daemon log, and recorded no activity row (#1041).

## What this does (client-side graceful degradation)

- **Faster, dedicated timeout** — a separate `ReceiveAdmitTimeout` (default 15s)
  for the interactive receive-admission path, instead of reusing the 30s general
  `AdmitTimeout`.
- **Typed, actionable error** — the failure maps to a gRPC `Unavailable` status
  carrying the new `CREDIT_RECEIVE_UNAVAILABLE` reason, with a message telling
  the user to receive at least the dust amount instead of leaking the raw
  deadline code.
- **Visible attempt** — a FAILED activity row is projected and the failure is
  logged at the routing site.
- **SDK reconstruction** — `wavewalletdk` turns the new reason back into an
  `errors.Is`-able `ErrCreditReceiveUnavailable` sentinel; the round-trip
  contract test covers it.

## Scope

This is the client half: it degrades gracefully so the user gets a fast, typed,
visible failure. The server-side root cause (the swap server not completing
`CreateCredit`) is addressed in the companion PR
lightninglabs/swapdk-server#242.

## Testing

- New `TestRecvCreditFailureIsTypedAndRecorded` reproduces the failure and locks
  all three guarantees (typed `Unavailable`, no leaked deadline, FAILED row).
- `swapwallet`, `credit`, and `sdk/wavewalletdk` suites green; `fmt` and local
  lint clean.
- Validated end to end against live signet.

Addresses #1041.

### DIFF
--- a/credit/registry.go
+++ b/credit/registry.go
@@ -91,11 +91,18 @@ func NewRegistry(cfg RegistryConfig) (*Registry, error) {
 	if cfg.PollInterval <= 0 {
 		cfg.PollInterval = DefaultPollInterval
 	}
+	if cfg.ReceiveAdmitTimeout <= 0 {
+		// Preserve the original AdmitTimeout override for embedders
+		// that configured the receive CreateCredit call before the
+		// dedicated timeout existed.
+		if cfg.AdmitTimeout > 0 {
+			cfg.ReceiveAdmitTimeout = cfg.AdmitTimeout
+		} else {
+			cfg.ReceiveAdmitTimeout = DefaultReceiveAdmitTimeout
+		}
+	}
 	if cfg.AdmitTimeout <= 0 {
 		cfg.AdmitTimeout = DefaultAdmitTimeout
-	}
-	if cfg.ReceiveAdmitTimeout <= 0 {
-		cfg.ReceiveAdmitTimeout = DefaultReceiveAdmitTimeout
 	}
 
 	earmark := &atomic.Pointer[EarmarkFunc]{}
@@ -406,10 +413,10 @@ func (b *registryBehavior) admit(ctx context.Context, msg CreditMsg,
 //
 // This is the one server round-trip an admission makes on the supervisor
 // goroutine, which serializes every admission, resume, and reap. It is bounded
-// by AdmitTimeout so one slow or hung receive cannot head-of-line-block the
-// whole subsystem; on timeout the synchronous caller retries under the same
-// stable op key, and CreateCredit is idempotent, so the retry reuses the same
-// invoice.
+// by ReceiveAdmitTimeout so one slow or hung receive cannot
+// head-of-line-block the whole subsystem; on timeout the synchronous caller
+// retries under the same stable op key, and CreateCredit is idempotent, so the
+// retry reuses the same invoice.
 func (b *registryBehavior) createReceiveInvoice(ctx context.Context,
 	rec *db.CreditOperationRecord, memo string) error {
 
--- a/credit/registry.go
+++ b/credit/registry.go
@@ -94,6 +94,9 @@ func NewRegistry(cfg RegistryConfig) (*Registry, error) {
 	if cfg.AdmitTimeout <= 0 {
 		cfg.AdmitTimeout = DefaultAdmitTimeout
 	}
+	if cfg.ReceiveAdmitTimeout <= 0 {
+		cfg.ReceiveAdmitTimeout = DefaultReceiveAdmitTimeout
+	}
 
 	earmark := &atomic.Pointer[EarmarkFunc]{}
 	if cfg.AutoRedeem.EarmarkedSat != nil {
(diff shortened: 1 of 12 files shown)


swapwallet/recv.go
```diff
@@ -180,7 +202,29 @@ func (r *receiver) recvCredit(ctx context.Context,
 		},
 	).Await(ctx).Unpack()
 	if err != nil {
```
Roasbeef, line 206: We're classifying every `Ask(...).Await(ctx)` failure as a credit-server outage here, but that result also carries the RPC caller's own `context.Canceled` or `context.DeadlineExceeded`. In that case we'll return `Unavailable` and, since `projectFailedCreditReceive` strips cancellation, persist a FAILED row for a receive the caller merely abandoned. I reproduced this with a canceled context: the error loses `context.Canceled` and one row is projected. Can we check `ctx.Err()` first, return the caller cancellation unchanged, and only project/map a completed registry failure?

credit/registry.go
```diff
@@ -415,14 +418,24 @@ func (b *registryBehavior) createReceiveInvoice(ctx context.Context,
 		return fmt.Errorf("get identity pubkey: %w", err)
 	}
 
-	callCtx, cancel := context.WithTimeout(ctx, b.cfg.AdmitTimeout)
+	callCtx, cancel := context.WithTimeout(ctx, b.cfg.ReceiveAdmitTimeout)
```
Roasbeef, line 428: This leaves the exported `RegistryConfig.AdmitTimeout` with no runtime consumer. Before this PR, an embedder could set that field to bound this exact `CreateCredit`; after the switch, an unset `ReceiveAdmitTimeout` always becomes 15 seconds and the existing override is silently ignored. Can we preserve an explicitly supplied `AdmitTimeout` as the compatibility fallback when the new field is zero, then use 15 seconds only when both are unset? A small config test will lock this down.

Roasbeef (changes requested):
I found two correctness issues that we should fix before this lands. Details are inline.

Roasbeef:
I pushed the two review fixes as standalone commits on
`fix/pr1045-review-findings`:

- [`40e8892a`](https://github.com/lightninglabs/wavelength/commit/40e8892a8265cf6e91e2bd0f30308365fc7dee1f) preserves caller cancellation and avoids recording a FAILED row for an abandoned receive.
- [`6131ddb4`](https://github.com/lightninglabs/wavelength/commit/6131ddb4d7f30a2917dfbb91492bc02505596def) preserves existing `AdmitTimeout` overrides while keeping the new 15-second default when both timeout fields are unset.

They apply cleanly on top of `eca73156`. The focused `credit`, `swapwallet`,
and `wavewalletdk` suites pass, and the changed-code linter reports zero
issues.

swapwallet/recv.go
```diff
@@ -180,7 +202,37 @@ func (r *receiver) recvCredit(ctx context.Context,
...
-		return nil, fmt.Errorf("start credit receive: %w", err)
+		// Preserve caller-owned cancellation. A canceled RPC says
+		// nothing about the operator's credit subsystem, and it must
+		// not leave a durable FAILED row for an attempt the caller
+		// abandoned.
+		if ctxErr := ctx.Err(); ctxErr != nil {
```
bhandras, line 214: This ensures that the call context signals shutdown or cancellation, but on shutdown the registry itself may report back some error but we'd still blame the operator below. Its more of a nit but imo worth fixing.

swapwallet/errors_grpc.go
```diff
@@ -70,6 +70,16 @@ var sentinelMappings = []sentinelMapping{
 		ErrBalanceLimitExceeded, codes.FailedPrecondition,
 		wavewalletrpc.ReasonBalanceLimitExceeded,
 	},
+
+	// A sub-dust receive whose credit request the swap server never
+	// completes is Unavailable (a function of operator/server state, not a
+	// malformed argument), reusing the swap-backend-unavailable reason so
```
bhandras: nit: The agent flagged this as stale and it is right, we now have the specific credit receive is unavailable.

bhandras (approved):
LGTM, pending 2 nits 🎉

swapwallet/recv.go
```diff
@@ -210,6 +262,35 @@ func (r *receiver) recvCredit(ctx context.Context,
...
+	entry := &wavewalletrpc.WalletEntry{
+		Id:            id,
+		Kind:          wavewalletrpc.EntryKind_ENTRY_KIND_RECV,
+		Status:        wavewalletrpc.EntryStatus_ENTRY_STATUS_FAILED,
+		AmountSat:     int64(amt),
+		Counterparty:  "credit",
+		CreatedAtUnix: now,
+		UpdatedAtUnix: now,
+		Note:          req.GetMemo(),
+		FailureReason: "credit receive unavailable",
```
bhandras, line 291: nit: one more thing the agent marked is that we could set the failure code here.

PR #1046, opened 2026-07-23
Fail closed when the wasm wallet database can't persist to OPFS

jamaljsr:
## Summary

A second browser tab can't acquire the wallet database's exclusive OPFS handles, and the wasm SQLite worker used to treat that open failure as a cue to fall back to an in-memory database. The daemon then booted against a throwaway store and died much later inside migrations with an unrelated-looking `SQLITE_CANTOPEN`, while any writes that had already succeeded were doomed to vanish on page close. That's the raw trace behind lightninglabs/wavelength-sdk#48.

This makes the open fail closed instead: `require_persistent` on the DSN turns a missing persistent OPFS VFS into a real locked-database error at open time, so there's no silent in-memory downgrade and no confusing failure later in migrations.

## Changes

- Set `require_persistent=true` on both wasm SQLite DSNs: the generic db store (`db/sqlite_open_wasm.go`) and btcwallet's OPFS-backed walletdb (`lwwallet/walletdb_wasm.go`).
- Widen the open-retry predicate to `SQLITE_BUSY` and `database is locked` alongside the existing `SQLITE_CANTOPEN`, since contention now surfaces as an open failure rather than an in-memory fallback. The helpers are renamed `isWASM*RetryableOpen` to match.
- The retry budget still absorbs the brief handle-release race right after a reload. A lock holder that never goes away exhausts the retries and returns the locked error to the caller, where the SDK maps it to a typed `wallet_locked`.

## Testing

- Built the daemon to wasm and exercised the two-tab flow in the browser against a regtest stack: a second tab surfaces a locked-database error at open time instead of booting on an in-memory store, and a single tab (including a reload after the previous runtime releases its handles) still opens normally.

## Related

- Addresses the daemon half of lightninglabs/wavelength-sdk#48. The SDK PR adds the fast cross-tab detection (a Web Locks pre-check) and the typed `wallet_locked` mapping, and will cross-link here once it's open.

### DIFF
--- a/db/sqlite_open_wasm.go
+++ b/db/sqlite_open_wasm.go
@@ -27,6 +27,14 @@ func openSQLiteDatabase(cfg SQLiteOpenConfig) (*SQLiteOpenResult, error) {
 	values.Set("vfs", wasmSQLiteVFS)
 	values.Set("mode", "rwc")
 
+	// A wallet-grade database silently degrading to the in-memory VFS
+	// would lose every write on page close, so fail closed when no
+	// persistent OPFS VFS can be opened. The most common trigger is
+	// another tab of the same origin holding the exclusive OPFS handles;
+	// failing here surfaces that as a clear locked-database error instead
+	// of a later migration failure against a throwaway database.
+	values.Set("require_persistent", "true")
+
 	pragmas := make([]string, 0, len(cfg.Pragmas)+1)
 	for _, pragma := range cfg.Pragmas {
 		switch strings.ToLower(pragma.Name) {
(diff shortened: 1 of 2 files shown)


jamaljsr:
Complemented by the SDK-side fast-path guard in lightninglabs/wavelength-sdk#56, which detects the second tab via a Web Lock before it reaches this daemon path.

litbot-9000:
@sputn1ck: review reminder

litbot-9000:
@sputn1ck: review reminder

litbot-9000:
@sputn1ck: review reminder

litbot-9000:
@sputn1ck: review reminder

litbot-9000:
@sputn1ck: review reminder

litbot-9000:
@sputn1ck: review reminder

litbot-9000:
@sputn1ck: review reminder

PR #1047, opened 2026-07-23
swap: expose derived Lightning routing terms

bhandras:
## What it does

Exposes the fee terms derived for Ark-to-Lightning payments without adding a
third caller-controlled fee knob.

Callers continue to provide:

- `max_fee_sat`: the maximum total fee accepted for the payment;
- `max_credit_sat`: the existing cap on credit use.

Quote and create responses report the service fee, current Lightning route
estimate, and funded routing allowance accepted as part of the total fee.
These are read-only admission terms. The wallet or CLI does not supply a
separate routing budget.

## Why

The total fee cap expresses how much the caller is willing to pay, while the
returned decomposition explains what the accepted fee funds:

```text
caller policy:    max total fee
accepted terms:   service fee + funded routing allowance
route preview:    current Lightning route estimate
```

`max_credit_sat` remains unchanged and independent of this fee breakdown.

## Quote and creation semantics

`QuotePay` is a side-effect-free preview, not a binding quote token.
`StartPay` evaluates the invoice again and accepts the payment only when the
resulting total fee remains within `max_fee_sat`.

Once `StartPay` creates the pay session, the service fee and funded routing
allowance returned by creation are persisted for retries and recovery. The
allowance is a fixed component of the accepted total fee, not a separate
caller-controlled or refundable balance.

Ordinary quotes continue to send the wallet account key even when
`max_credit_sat` is zero. This preserves mandatory-credit detection for
sub-dust invoices and keeps quote and create settlement selection consistent.

## Validation and compatibility

- The response fields are additive; existing callers can ignore them.
- The request contract does not gain a routing-budget field or CLI flag.
- Preview and create validation require
  `service fee + routing allowance == total fee`.
- Every returned amount is checked before conversion to durable signed
  `BIGINT` storage.
- Legacy rows or older responses with a non-zero total fee and no decomposition
  are normalized conservatively as `service fee = 0` and
  `routing allowance = total fee`.
- Live options responses must provide an exact decomposition; compatibility
  normalization is limited to legacy interfaces and durable legacy rows.
- Options-capable quote clients report a clear configuration error when the
  wallet daemon needed for account identity is unavailable.
- Restored sessions normalize legacy rows, then reject overflowed or
  non-reconciling durable fee terms before resuming.
- `StartPay` response coverage asserts the final accepted split reaches callers.
- The existing public `Wait` recovery path remains covered by the funding-grace
  retry test.

## Changes

- Adds service-fee, route-estimate, and funded-allowance fields to quote and
  create responses.
- Persists accepted service fee and routing allowance with pay sessions.
- Restores and validates accepted terms after restart instead of deriving them again.
- Exposes the same read-only breakdown through swap-client summaries and
  prepared wallet sends.
- Reports final accepted terms after `StartPay` rather than treating the
  earlier preview as frozen.
- Names the all-available-credit sentinel used by wallet preparation.
- Uses one consistent description for the client-funded allowance across the
  public protobuf surfaces.

## Commit structure

1. Expose derived routing terms in wire responses.
2. Persist accepted routing terms with pay sessions.
3. Retain and validate the terms through creation and recovery.
4. Report accepted terms through the local swap service.
5. Show derived terms in prepared wallet sends.

Generated protobuf and sqlc output stays with the schema change that owns it.
Every commit is below 800 changed lines.

## Verification

- `make rpc`
- `make tidy-module-check`
- `make lint-local`
- `make commitmsg-lint range='origin/main..HEAD'`
- `go test ./sdk/swaps`
- `go test -tags='wavewalletrpc swapruntime' ./swapwallet`

### DIFF
--- a/rpc/swapclientrpc/swap_client.pb.go
+++ b/rpc/swapclientrpc/swap_client.pb.go
@@ -472,8 +472,15 @@ type QuotePayResponse struct {
 	// larger than that cap.
 	ExceedsMaxFee bool         `protobuf:"varint,7,opt,name=exceeds_max_fee,json=exceedsMaxFee,proto3" json:"exceeds_max_fee,omitempty"`
 	CreditQuote   *CreditQuote `protobuf:"bytes,8,opt,name=credit_quote,json=creditQuote,proto3" json:"credit_quote,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	// server_fee_sat is the service fee retained by the swap server.
+	ServerFeeSat uint64 `protobuf:"varint,9,opt,name=server_fee_sat,json=serverFeeSat,proto3" json:"server_fee_sat,omitempty"`
+	// estimated_routing_fee_sat is the current whole-satoshi route estimate.
+	EstimatedRoutingFeeSat uint64 `protobuf:"varint,10,opt,name=estimated_routing_fee_sat,json=estimatedRoutingFeeSat,proto3" json:"estimated_routing_fee_sat,omitempty"`
+	// routing_fee_budget_sat is the client-funded Lightning routing allowance
+	// that would be charged into the vHTLC as part of fee_sat.
+	RoutingFeeBudgetSat uint64 `protobuf:"varint,11,opt,name=routing_fee_budget_sat,json=routingFeeBudgetSat,proto3" json:"routing_fee_budget_sat,omitempty"`
+	unknownFields       protoimpl.UnknownFields
+	sizeCache           protoimpl.SizeCache
 }
 
 func (x *QuotePayResponse) Reset() {
@@ -560,6 +567,27 @@ func (x *QuotePayResponse) GetCreditQuote() *CreditQuote {
 		return x.CreditQuote
 	}
 	return nil
+}
+
+func (x *QuotePayResponse) GetServerFeeSat() uint64 {
+	if x != nil {
+		return x.ServerFeeSat
+	}
+	return 0
+}
+
+func (x *QuotePayResponse) GetEstimatedRoutingFeeSat() uint64 {
+	if x != nil {
+		return x.EstimatedRoutingFeeSat
+	}
+	return 0
+}
+
+func (x *QuotePayResponse) GetRoutingFeeBudgetSat() uint64 {
+	if x != nil {
+		return x.RoutingFeeBudgetSat
+	}
+	return 0
 }
 
 type StartPayRequest struct {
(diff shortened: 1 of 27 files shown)


bhandras:
@claude review

bhandras:
Addressed both findings and folded the fixes into the original commits:

- `validateInSwapQuote` now rejects server-returned `server_fee_sat` and `routing_fee_budget_sat` values above `math.MaxInt64` before either component can reach durable BIGINT storage. The existing mismatch table covers both boundaries.
- `paySessionFromRow` now restores `routingFeeBudgetSat` on the live session as well as `cfg`, and the persistence/resume test asserts the restored value.

Validated with the full `sdk/swaps` test package, changed-file lint, formatting, sqlc regeneration, module tidy, schema consistency, and commit-message lint. All rewritten commits are signed.

@claude review

bhandras:
The remaining unit-race failure was a pre-existing timing assumption in `TestPaySessionResumeFundingGraceEventuallyRetries`: the test canceled reconciliation after 5 ms and then expected SQLite persistence to have completed. Under `-race`, cancellation could land after the mocked send but before the store write. The new standalone commit drives exactly one resumed funding reconciliation synchronously, preserving the same grace-period and durable-session assertions without changing production behavior. It passed 50 consecutive focused race runs, the full `sdk/swaps` race suite, the normal suite, and commit-message lint.

bhandras:
@claude review

Please re-review the rewritten branch from scratch. The product contract changed materially since the prior review: callers now retain only the existing total fee and credit caps; there is no caller-controlled routing budget, request field, wallet option, or CLI flag. The service-derived route estimate and funded allowance are read-only quote/create terms, and accepted terms remain durable across restart. Please verify the request/response API boundaries, generated protos, persistence and recovery behavior, overflow checks, and tests against that final model.

bhandras:
Addressed the remaining test-coverage suggestion and folded it into `swaps: persist accepted routing terms`: the resume test now asserts both `ServerFeeSat` and `RoutingFeeBudgetSat` on the restored session config. While verifying it, `make fmt-check` exposed two existing branch-local style differences; both formatter-only changes were folded into their owning commits. Focused SDK and tagged wallet tests, `fmt-check`, and commit-message lint pass. @claude review the refreshed head please.

bhandras:
@claude please re-review the rewritten branch from scratch. The local Opus/Fable pass led to four contract hardenings: preview and create now validate that service fee plus routing allowance exactly equals total fee; legacy zero/zero decompositions normalize conservatively; ordinary zero-credit quotes retain the account key so mandatory-credit detection stays consistent; and the public Wait-based recovery test is restored. RPC generation, module checks, full local lint, commit-message lint, and the focused SDK/RPC/wallet suites pass. There is still no caller-controlled routing-budget field.

bhandras:
@claude please review the refreshed head. The synchronize review found two non-blocking defensive gaps and both are now folded into their owning commits: options-capable and compatibility quote paths share a guarded account-key helper, so a missing daemon returns `daemon is not configured` instead of panicking; and restored legacy pay sessions normalize the absent fee decomposition just like durable summaries. The nil-daemon path has a regression test. Exact fee-split reconciliation remains validated on both preview and create response paths. RPC generation, tidy, changed-file formatting, full local lint, commit-message lint, and the focused SDK/RPC/wallet tests pass.

bhandras:
@claude please review the final head `37610b46`. Compatibility normalization now runs only on legacy `WithCredits`/base interfaces and durable legacy rows. The current options contract no longer normalizes a zero/zero decomposition, so an options-capable server that omits the split fails exact reconciliation. Focused tests cover both sides: a legacy interface with a non-zero total is normalized to routing-funded, while a current options response with the same omission is rejected. SDK, tagged RPC/wallet tests, full local lint, formatting, and commit-message checks pass; all five commits remain below 800 changed lines.

bhandras:
@claude please re-review the refreshed head `26b1398d`. The latest suggestions are folded into their owning commits: `StartPay` now directly asserts the final accepted service-fee/routing-budget split; restored pay rows normalize only the legacy zero/zero shape and then run exact overflow/reconciliation validation; the new columns have migration comments; and create-path fee-term errors carry the same in-swap quote context as previews. A corrupt-row regression proves restore fails before resuming. SDK, tagged local service/wallet tests, formatting, full lint, and commit-message checks pass; all five commits remain below 800 changed lines.

bhandras:
Closing after design discussion: the fee decomposition (`server_fee_sat`, `estimated_routing_fee_sat`, and `routing_fee_budget_sat`) is server pricing policy rather than caller contract. The caller controls one knob (`max_fee_sat`), and both fee components share identical vHTLC and refund semantics, so the split has no client-side decision value. Exposing it would freeze internal pricing policy into the public API and require split-identity validation, durable display-only state, and legacy compatibility behavior that protect no funds.

The economics fix now remains entirely server-side. These fields would be additive, so a concrete wallet UX requirement can reintroduce them later without compatibility cost.

PR #1055, opened 2026-07-28
baselib/actor: decay the idle mailbox poll instead of polling at a fixed rate

### LINKED ISSUE #721: baselib/actor: adaptive idle polling or enqueue wake for PeekNextMailboxMessage
CORRECTED after review: the durable mailbox consume loop is ALREADY wake-driven. An empty peek blocks on select{poll ticker, m.wake, ctx} and DurableMailbox.Send fires the wake on every same-process enqueue (baselib/actor/durable_mailbox.go Receive loop), so hot actors are event-driven today. The original framing of this issue (add an enqueue wake) was wrong.

What remains is much smaller: the fallback poll ticker fires at PollInterval for every resident actor regardless of traffic, which the stress ledger shows as ~35k empty-peek read commits per 1000 payments (resident actors x poll rate x run duration). Reads are WAL-concurrent and an earlier profile sized this at ~2-3% CPU, so the only candidate lever is decaying the ticker for long-idle actors (fast when recently active, slow when dry; the ticker exists as a crash/cross-process fallback, so it must never stop entirely).

Low priority; close if the decay complexity is not worth 2-3% CPU.

Roasbeef:
Fixes #721.

### What

Replaces the durable mailbox's fixed poll `Ticker` with a backoff that widens while the mailbox stays empty. The wait starts at `PollInterval`, roughly doubles per consecutive empty poll, and is capped by a new `MaxPollInterval` defaulting to 30s. Any wake or successfully claimed message snaps it back to the floor.

### Why #721 was reopened

The issue was closed on the reasoning that the residual fallback ticker was worth "~2-3% CPU, not worth the complexity". That estimate was measured on SQLite — the issue body says so explicitly: *"Reads are WAL-concurrent."* On SQLite an empty peek is a WAL read with no network, TLS, auth or isolation machinery, and 2-3% is entirely believable.

On Postgres the identical empty peek is a TLS round trip, a `BEGIN ISOLATION LEVEL SERIALIZABLE`, the anti-join `UPDATE`, and a `COMMIT` carrying SSI predicate-lock bookkeeping. Measured on the signet lumosd deployment:

```
25s CPU profile
  67.5% cum  wavelength/db.TransactionExecutor.ExecTx
               <- 99.8% from actordelivery.(*Store).LeaseNextMessage
  16.4% cum  lumos/db.TransactionExecutor.ExecTx
               <- 94.2% from db.MailboxEnvelopeStore.pullIteration

goroutines (4029 total)
  1286 @ durable_mailbox.go:465   the empty-peek select this PR changes
   290 @ MailboxEnvelopeStore.pullIteration via clientconn ingressLoop

Postgres, 40s window
  xact_commit   2,035/s
  tup_updated       9/s   -> >99% of those transactions changed nothing

cgroup, 30s sample
  nr_throttled/nr_periods  281/300  -> 93.7% of CFS periods throttled
  usage 497m against a 500m limit
```

So roughly 84% of the daemon's CPU, not 2-3%. A ~30x miss, entirely from the backend swap.

That starvation is what made Lightning pay-ins into Wavelength take 25s to 2m14s against a flat ~17s for pay-outs: the slow direction needs three serialized indexer round trips through a process frozen ~87ms out of every 100ms.

### This does not cost delivery latency

Worth stating plainly because it is the obvious objection. The fallback poll is not the delivery path.

`NewDurableMailbox` registers a post-commit `RegisterMailboxWake` callback with stores that implement `MailboxWakeRegistrar` (`db/actordelivery/store_impl.go:1724` asserts `*Store` does), and that callback rouses the *exact* mailbox a committed transaction enqueued into, regardless of how far its backoff has decayed. The pre-commit wake inside `Send` is already documented in-tree as "a harmless empty poll" for the folded-transaction case. So a same-process enqueue is exactly as prompt as before.

What the ceiling *does* bound is discovery of an enqueue this process did not make, since `RegisterMailboxWake` is same-process only. `MaxPollInterval` is therefore the worst-case cross-process/cross-replica delivery latency. lumosd is single-replica today so this is latent, but it is documented at the config field so a future multi-replica deployment knows the knob exists and what it costs.

### Decisions worth flagging to a reviewer

**The lease-error path decays too**, via the same helper with identical reset semantics. A store failing every claim is precisely when retrying once per second per actor does the most harm — a wedged database gets another thousand doomed transactions a second. A transient blip costs at most one extra doubling.

**`Ticker` → `Timer`**, since a Ticker cannot express a varying wait. `baselib/go.mod` is at `go 1.25.11` (this code lives in the separate `baselib` module, not the root one), past the 1.23 boundary where `Reset` became synchronizing and clears any value staged before the call, so re-arming needs no manual drain.

**One behavior change beyond the decay.** The old code held one Ticker for the loop's life, so a tick could accumulate during processing and grant a free immediate re-poll afterwards. The timer is now armed only when about to wait, so that free poll is gone. It removes empty polls rather than adding latency — the loop already re-polls unconditionally at the top of each iteration.

**Normalization is required, not a nicety.** A zero `PollInterval` previously panicked in `NewTicker`; `Timer.Reset(0)` would instead tight-spin the store, which is worse. A ceiling below the floor is raised to the floor, degrading to the constant cadence this replaces.

**A third wait site.** The leaseless mutation-failed throttle also routes through the shared helper. Its behavior is unchanged — it is only reachable right after a successful claim, which resets the backoff, so it stays a flat `PollInterval` throttle.

### Expected win, honestly bounded

Not 30x. Every idle stretch pays 1+2+4+8+16s of ramp before reaching the ceiling, and every wake *and* every claim restarts it, so a mailbox with real traffic sees little reduction and only long-idle actors approach the asymptote. Against the signet numbers, the idle-poll component itself should fall ~30x (1,286/s to ~43/s), but process-wide commits should land near 800/s — roughly 2.5x — because ~750/s of the measured 2,035/s was never idle polling. CPU should improve more than that ratio suggests given the throttling.

### Not covered here

Two sibling loops share this shape and are deliberately left alone here:

- **`clientconn` ingress pull loops** (`lumos/db.MailboxEnvelopeStore.pullIteration`, ~16% of the measured CPU, 290 goroutines). A second independent 1Hz loop that #721 never scoped. Stacked follow-up in lumos, deliberately after this pattern settles rather than inventing two decay implementations in parallel.
- **`baselib/actor/outbox_publisher.go:151`** — same `time.NewTicker(cfg.PollInterval)` anti-pattern, and its doc comment describes the same "polling is only the fallback for missed wakes" model. Blast radius is far smaller: roughly one publisher loop per process rather than one per resident actor, so ~1 tx/s against the mailbox path's ~1,286/s. Out of scope here, named so it is not forgotten.

### Testing

`go test ./baselib/actor/...` passes, including under `-race` and at `-count=5`. `gofmt` clean, `make lint-changed-local` reports 0 issues, `sg scan` clean on changed lines.

Assertions were validated against five mutations of the implementation — a no-op `decay()`, a removed ceiling clamp, a removed wake reset, a removed claim reset, and a timer that stops re-arming at the ceiling. Every mutant is killed, so the tests fail for the reasons they claim.

Timing assertions are deliberately one-sided (a poll did not arrive *earlier* than allowed) rather than clock-injected: driving the durability fallback off a test clock means a frozen clock silently disables crash-recovery polling, and `internal/actortest/ledger_e2e_test.go:30-37` already documents that injecting a frozen clock here breaks the mailbox. A loaded CI box can only stretch an observed gap, never compress one.


Roasbeef:
@claude review this

Roasbeef:
Thanks — all three observations are fair. Actioned two, declining one, and the review also caught a factual error in my own commit message that I have corrected.

### The go version was wrong, and you had it right

The review says `baselib/go.mod` is at `go 1.25.11`. It is, and my commit message and PR description both said "go 1.26" — I had read the **root** `go.mod`, not the module that actually builds this code. `baselib/` is a separate module:

```
$ grep -m1 '^go ' go.mod          # root
go 1.26.0
$ grep -m1 '^go ' baselib/go.mod  # the module that owns baselib/actor
go 1.25.11
```

The conclusion is unaffected — 1.25.11 is still past the 1.23 boundary where `Reset` became synchronizing — but the cited fact was wrong, and someone later checking whether this code still satisfies the timer-semantics constraint would have looked at the wrong file. Corrected in both the commit message (force-pushed, content byte-identical, only the message changed) and the PR body. The in-code comment was already correct since it says "Go 1.23+" and points at the go directive without naming a version.

### 1. `outbox_publisher.go` — added to the follow-up list

Agreed, and it is the same pattern for the same reason. Added to the "Not covered here" section explicitly rather than leaving it implicit, with the reason it is lower priority: roughly one publisher loop per process versus one per resident actor, so ~1 tx/s against the mailbox path's ~1,286/s. Same anti-pattern, three orders of magnitude less of it.

Not folding it into this PR — it is a separate loop with a separate lifecycle, and keeping this diff to the mailbox path is what makes the mutation testing meaningful.

### 2. The initial timer arming — declining, with reasoning

You are right that `time.NewTimer(backoff.current())` at loop start contradicts the `waitIdle` comment's "armed only here". The suggested `t := time.NewTimer(0); if !t.Stop() { <-t.C }` idiom is exactly the manual-drain pattern that the Go 1.23 semantics made unnecessary, and reintroducing it here would be a small trap for the next reader — it reads as though a drain is required, which is the thing this change specifically does not need.

Since the cost is one live timer per receive loop until the first wait, and `Reset` clears any stray fire, I would rather fix the comment than the code. I have left both as-is for now; if a reviewer prefers, the cheaper edit is to reword the comment to "armed for each wait here" rather than restructure the construction.

### 3. `decay()`'s first branch — leaving it

Agreed it is dead under the maintained invariant. Leaving it deliberately: it is what makes `decay()` correct in isolation rather than only correct given its caller, and the `rapid` property test exercises it directly by constructing backoffs at the ceiling. Cheap insurance against a future caller that sets `cur` outside the invariant.

### On the corroboration

Worth noting for anyone reading later: this PR also went through a separate adversarial review pass, and the two found different things. That is the second time this session the two passes have not subsumed each other — earlier, on lightninglabs/lumos#720, this reviewer caught a CI-breaking 80-column violation that the deeper pass missed entirely despite being explicitly told to run the repo's linter. Worth knowing that "the deep reviewer reported clean" and "the tooling was actually run" are different claims.

Issue #1054, opened 2026-07-28
mailbox: add a non-durable control-plane message class (cancel, shed, backpressure)


Roasbeef:
## Summary

The mailbox protocol has no way to send a message that is only worth delivering
*right now*. Every envelope is durable by construction, and that single gap is
behind two separate problems we've hit in production.

`mailboxrpcserver/server.go:93` (lumos) makes this unconditional:

```go
_, err := s.store.Append(ctx, req.Envelope)
```

Every `Send` is a durable append into the recipient's mailbox. There is no
best-effort class, no TTL, no "deliver if the client is attached, otherwise
drop". Mailboxes are also bounded (`ErrMailboxFull`), so a burst of
short-lived messages doesn't just cost writes, it can evict or block real ones.

### 1. No cancel primitive

`mailbox/pb/mailbox.proto:13-16` defines the full set of kinds:

```proto
KIND_UNSPECIFIED = 0;
KIND_REQUEST     = 1;
KIND_RESPONSE    = 2;
KIND_EVENT       = 3;
```

When a caller abandons a request (deadline fires), nothing tells the operator to
stop. The work runs to completion and the response arrives with no waiter left.
On signet this produced 6,581 late responses in a 30h window, 100% of them
`arkrpc.IndexerService/ListVTXOsByScripts`, every correlation ID distinct — pure
queueing of abandoned in-flight work, confirmed not a replay wedge. Under CPU
starvation (the box was CFS-throttled 93.7% of periods) every deadline fired, so
every retry piled more doomed work onto a server that couldn't answer the first
attempt. See lightninglabs/lumos#718 and lightninglabs/swapdk-server#265 for the
incident this fed.

The client-side half is fixed: stable per-logical-request idempotency keys and a
bounded outstanding set (`MaxInFlightUnary`). That stops retries from *minting*
new work. It cannot reach in and stop work already running.

### 2. Shed responses can't be answered without amplifying writes

`lumos/indexer/operator.go` sheds correctly but can only *say so* about once per
client per second, because saying so costs a durable write. The
`!allowed && !notify` branch returns `nil` silently, with the tradeoff spelled
out in its own comment:

> Building and sending another response costs a durable mailbox write, so we drop
> this one silently rather than let a flood amplify into the shared write path.

That's the right call given the primitives available, and it leaves the client
unable to distinguish "shed" from "slow", so it burns its full deadline and
retries. Which is precisely the loop the shedding exists to break.

## Why these are one issue

[...]

PR #1056, opened 2026-07-28
multi: mint one idempotency key per logical mailbox RPC

Roasbeef:
In this PR, we stop client retries from amplifying into unbounded work on the
operator. The trigger was a `GetIndexedVTXOByPkScript` on signet that took over
67 seconds to answer a SQL query that runs in about 2ms, sitting next to 6,581
responses in a 30 hour window that arrived with no waiter left to receive them.
All 6,581 were `ListVTXOsByScripts`, and every correlation ID was distinct,
which rules out a replay wedge since nothing was being redelivered. That's the
signature of pure queueing: requests the caller had already given up on, which
the operator was still grinding through one at a time.

## Where it comes from

`mailbox/rpc/options.go` has always documented `IdempotencyKey` as "stable per
semantic operation (not per send attempt)". The contract was there. Nothing ever
exercised it. The only place in production that builds a `mailboxrpc.RPCOptions`
is `firstOpt` in `indexer/client.go`, and with no options passed it returns the
zero value, so the field reaches the transport empty and `UnaryFacade.SendRPC`
mints a fresh random key to fill it. Every unary mailbox RPC the daemon has ever
sent looked brand new to the operator, including the eight sequential re-issues
`retryRecoveryIndexerRPC` makes of a single scan. Recovery walks the entire
recovery window one script at a time, so it's the heaviest indexer client we
have and the first one the per-client limiter sheds, and it was answering that
shed by handing over eight distinct keys for one logical query, against a
limiter that had just asked it to slow down.

## What we do about it

The new `mailbox/rpc/retry.go` gives us `NewIdempotencyKey`, `Retry`,
`RetryWithKey`, `IsShedError` and `RetryPolicy`. `Retry` mints the key once, at
the logical-call boundary, and threads the same `RPCOptions` into every attempt,
so the operator sees one unit of work where it used to see N. The key stays
random rather than becoming a digest of the request payload, since two callers
issuing byte-identical concurrent reads are still two logical requests and a
digest would fold them into one. The loop also tightens what counts as
retryable: only an explicit `ResourceExhausted` is re-issued, only after a
jittered exponential backoff (so a fleet shed by the same token bucket doesn't
wake together and recreate the burst), and a context that's already cancelled or
past its deadline never re-sends at all.

`retryRecoveryIndexerRPC` then moves onto that helper, which is mostly a matter
of threading `RPCOptions` down through each call site, and drops its flat 150ms
delay for the jittered backoff.

Review caught one recovery call that was left behind:
`ResolveIncomingMetadataFromIndexerWithLimits`, which recovery reaches for every
OOR event it replays, was the only indexer call in the flow still sent
single-shot while its four siblings in the same file retried. It pages through
`ListVTXOsByScripts`, which was 100% of the late responses on signet, so a
single shed there failed the whole `InitWallet` and the natural response to that
is to re-run recovery from the top, re-minting a key for every script in the
window. That's the amplification this branch removes, performed by hand. The
retry seam goes inside the page loop rather than around it, since each page
carries its own cursor and is therefore its own logical query. The live receive
path keeps its single-shot behavior on purpose: it has a caller blocked on the
other end of a receive, so failing fast beats holding that caller through a
backoff.

Third, `ConnectorConfig.MaxInFlightUnary` (default 256) bounds how many unary
RPCs one client can have outstanding at once. This is the piece that addresses
the queueing directly rather than the key churn: `RpcMeta_Kind` carries only
`REQUEST`, `RESPONSE` and `EVENT`, so a caller whose deadline fires has no way
to tell the operator to stop, and the operator runs the abandoned request to
completion and answers a waiter that's long gone. Capping the outstanding set is
the only bound available to us without a protocol change. A send past the cap is
refused locally with `ResourceExhausted` before it touches the wire, which is
the same code a shedding operator returns, so the retry path above backs off on
it without having to learn where it came from.

## What we left alone

Two things that look like the same shape and aren't. The durable indexer path
already derives stable keys from query-parameter bytes, and is in fact the model
this copies. The poll loops in `sdk/swaps/out_swap.go` and
`credit/transitions.go` aren't retries at all: each poll is a genuinely new read
looking for a state change, so a shared key there would be a correctness bug,
not a fix.

## What this doesn't fix

Worth being blunt about the edges here, since several of them are real.

There's no cancel primitive, and we didn't invent one. A real cancel needs a new
`RpcMeta_Kind` enum value, lumos-side ingress handling by correlation ID, and a
version bump negotiated through `ArkVersionPolicy`. That's cross-repo protocol
work, tracked in #1054. This PR ships the bounded client-side mitigation
instead.

The in-flight cap is soft. Concurrent senders read `WaiterCount()` before any of
them registers, so the set can overshoot by one burst. Making it exact would
mean holding the registry lock across the whole send. It stops unbounded growth;
it does not police a precise ceiling.

The 256 default is a judgment call, not a measured number. It's configurable,
and worth watching on signet.

Stable correlation IDs across retries are new behavior. `SendRPC` defaults the
correlation ID to the idempotency key, so a retry now reuses the abandoned
attempt's correlation ID, which is what lets a late answer land on the retry's
waiter instead of being acked into the void. That's safe because retries of one
logical request are sequential by construction, so the shared ID never spans two
live waiters. A future caller that retries _concurrently_ under the same key
would have two waiters contending for one correlation ID, so flagging it here.

This fixes amplification, not starvation. The server was CFS-throttled 93.7% of
periods, and deadlines will keep firing until #721 lands. Expect the tombstones
to drop substantially, not to zero.

There's a remaining gap on the single-shot callers. `rpc_swap_lookup.go`,
`rpc_vtxo_expiry.go`, `incoming_ancestry_fetcher.go` and
`rpc_custom_refresh_metadata.go` still mint a key per send, as does the live
receive path through `incoming_metadata.go`. That's correct today since nothing
retries them in-process, but an external client re-driving
`GetIndexedVTXOByPkScript` after its own 1s timeout is indistinguishable from a
new request. Closing that means carrying a caller-supplied idempotency token
across the `waverpc` public API boundary, which isn't a call to make
unilaterally.

Two smaller things we know about and chose not to code around. A response to
attempt N that arrives after attempt N's waiter is gone gets buffered, and the
retry's `RegisterWaiter` then completes immediately from that buffer. For an
idempotent read re-issued with the same cursor that's the right answer to the
right question, and if the buffered envelope happens to be the shed itself it
costs one attempt out of the budget before the next real send. It's
self-limiting and the shared key means the operator dedups it either way.
Separately, a cancellation that lands mid-backoff now returns the last
`ResourceExhausted` rather than the context error, where the old loop wrapped
`ctx.Err()`. No caller distinguishes the two today, so changing the error
identity here would add risk without buying anything.

Finally, one test assertion changed rather than being removed, and reviewers
should look at it directly: `TestRetryRecoveryIndexerRPCStopsOnContextCancel`
went from `attempts == 1` to `require.Zero(t, attempts)`. It's a strictly
stronger assertion tracking a deliberate behavior change, since a caller with an
already-dead context now stops before the first send instead of after it.

One more thing on the operator side: lumos is only partially fixed there. Its
indexer replies `ResourceExhausted` only for the first shed per client per
window, and the `!allowed && !notify` branch returns nil silently because every
response costs a durable mailbox write. That residual silent drop is the same
loop seen from the other end, and it's tracked in #1054 with a cheap interim (a
retry-after hint) in flight separately.

See each commit message for the incremental detail.


Roasbeef:
Pushed two follow-up commits after an adversarial review pass. Notes on what
moved and what didn't, so reviewers can go straight at the parts I argued with.

The one real finding was a call the branch had missed.
`ResolveIncomingMetadataFromIndexerWithLimits` is reached from seed recovery for
every OOR event it replays, and it was the only indexer call left in that flow
still sent single-shot while its four siblings in `wallet_recovery.go` all
retried under a stable key. It pages through `ListVTXOsByScripts`, which was
100% of the late responses we saw on signet, so one shed page failed the entire
`InitWallet`. The operator's response to that is to re-run recovery from the
top, which re-mints a key for every script in the window, so the gap recreated
by hand exactly the amplification this branch exists to remove. Fixed in
`da725cf`, with the retry seam inside the page loop rather than around it since
each page carries its own cursor and is its own logical query. Two tests cover
it, and I mutation-tested both: dropping the `opts` threading and bypassing the
seam each kill `TestResolveIncomingMetadataRecoveryRetriesShedPage`.

The live OOR receive path deliberately keeps its single-shot behavior rather
than inheriting recovery's retries. It has a caller blocked on the other end of
a receive, so failing fast beats holding that caller through a backoff, and
widening retry coverage on a money path isn't something to do as a side effect
of a recovery fix. `TestResolveIncomingMetadataLiveReceiveStaysSingleShot` pins
that so it can't drift.

`7dc2ff0` is cleanup: `Retry` now delegates to `RetryWithKey("")` instead of
duplicating the minting, and `DefaultRetryPolicy` is gone since it had no
callers and `normalize()` already gives a zero `RetryPolicy` the same defaults.
It also writes down two things that were previously true only by convention:
`RetryWithKey` assumes its key isn't already in flight elsewhere, and
`MaxInFlightUnary` is a per-connector cap, so one subsystem saturating it
fails unrelated unary RPCs daemon-wide until they drain.

Three things I pushed back on. A late response to attempt N can be buffered and
then immediately satisfy attempt N+1's `RegisterWaiter`; for an idempotent read
re-issued with the same cursor that's the right answer to the right question,
and if the buffered envelope is the shed itself it costs one attempt out of the
budget. Self-limiting, and the shared key means the operator dedups it either
way, so it's documented in the description rather than coded around. A
cancellation landing mid-backoff now returns the last `ResourceExhausted` where
the old loop wrapped `ctx.Err()`; I checked and no caller in the recovery path
branches on cancellation, so changing error identity mid-branch buys nothing and
risks something. And the per-refusal `WarnS` in `SendRPC` stays, because it's
the only signal that the cap is biting at all.

I also checked `incoming_ancestry_fetcher.go` for the same shape and it
dissolved: it's wired only into the live receive handler at `server.go:4579`,
the recovery handler (`recoveryOORHandler`) has no ancestry fetcher at all, and
its documented failure mode is persisting without ancestry rather than failing
the receive. Left alone.

@claude review this

Roasbeef:
Went through all three. Took the first, left the other two, reasoning below.

**(1) `MaxInFlightUnary` doc scope.** Confirmed and fixed in `f14b4c1`.
`RegisterWaiter` has exactly two callers, both in `unary_facade.go`, so the
registry holds live unary waiters and nothing else, and `admitUnary` is only
reached from `SendRPC`. The durable egress handlers never touch either. The
comment now says so, and says why the asymmetry is right rather than an
oversight: a durable response that outlives its waiter falls through to durable
route dispatch and gets redelivered, whereas a live unary response with no
waiter is acked into the void. The discarding is what the cap is bounding, so
gating durable egress on it would cost throughput and buy nothing.

**(2) The prune scan under the registry lock on every send.** Real, and I'm
leaving it. It's O(n) in the waiter set, which the cap itself bounds at 256, so
the worst case is a 256-entry map walk under an uncontended mutex on a path that
is about to do a network round trip. Making it cheaper means either not pruning
(which reintroduces the wedge the prune exists to prevent, since a TTL-lapsed
entry would inflate the count and lock the client out of sending) or keeping a
separate live counter, which is more state to get wrong for a saving that
doesn't show up until unary throughput is orders of magnitude past recovery's
sequential walk. Worth revisiting if that ever happens.

**(3) Per-attempt context discarded in the recovery closures.** Correct
observation, and it's latent rather than live: `Retry` forwards the caller's
context to every attempt unchanged, so the captured outer `ctx` and the passed
one are the same value today. I'm not threading it through, because doing so
means changing `retryRecoveryIndexerRPC`'s callback signature and the
`indexerPageCall` seam and all six call sites, which is a lot of churn for a
provable no-op on a branch that's already the base for stacked work. The reason
it's tolerable to leave is that anyone adding a per-attempt sub-context to
`Retry` has to touch `Retry` itself, and the `_ context.Context` parameters
sitting in these closures are exactly what they'd trip over. If that ever lands
and the trip-over doesn't happen, this becomes a real bug, so it's worth
knowing about.

On the sandbox note: `go build`, `go test -race`, and `make lint-changed-local`
all run clean locally on the touched packages, and CI is green on the branch.

PR #1057, opened 2026-07-28
db: run read-only Postgres transactions at REPEATABLE READ

Roasbeef:
In this PR, we relax the transaction isolation level for read-only Postgres
transactions from `SERIALIZABLE` down to a read-only `REPEATABLE READ`, and
leave writers exactly where they are.

Today `BaseDB.BeginTx` hardcodes `sql.LevelSerializable` for every transaction
we open, reads included. Under Postgres' serializable snapshot isolation that
isn't free for a reader: a read-only `SERIALIZABLE` transaction still takes
`SIRead` predicate locks and still joins the serialization conflict graph, so it
can both eat a `40001` itself and act as the pivot that aborts a concurrent
writer. On a daemon as read heavy as ours, that's pure cost. A signet
measurement put us at 2,035 transaction commits per second with 99.6% of them
changing no rows at all, against a Postgres box sitting at 794m of a one core
limit largely on predicate-lock bookkeeping.

For a concrete version of that cost, the lumos client-connection ingress
long-poll is a read path paying the SSI toll for nothing at all. Each
`pullIteration` in `db/mailbox_store.go` opens a `ReadTxOption()` transaction
around `PullMailboxEnvelopes` and only ever runs a SELECT. The loop is
wake-driven rather than a fixed-rate poll, since it selects on a notify channel
that `Append` fires, so the poll timer is a fallback and not the mechanism. The
trouble was that the fallback interval was one second against a five second
wait window, and the knob for it was entirely unwired: `WithPullPollInterval`
had no production callers, there was no config field behind it, and the store
options builder never emitted it. So one second is flatly what ran on signet,
with no deployment able to have tuned it.

Worth stating the ordering, since the constant is moving. Before that fix, each
five second long-poll cost five read transactions, on the order of 290 per
second across the parked ingress loops in NobleSun's signet profile. Raising
the fallback above the wait window takes it to one per RPC, roughly 58 per
second. That change reduces how many of these transactions exist and this PR
reduces what each one costs, so the two compose rather than overlap, and
neither owns the combined win.

The first commit is the whole functional change, and it's deliberately
self-contained and independently landable. It follows
lightningnetwork/lnd#10997. The isolation choice funnels through a small
`txIsolationLevel` helper that gates on the `BackendType` that `BaseDB` already
carries, so SQLite is untouched. There's nothing to win there anyway, since
SQLite only ever admits a single writer and is effectively serializable already.

Worth calling out that we open these transactions as `READ ONLY` as well as at
the relaxed level, and the flag is doing at least as much work as the level is.
Postgres only skips predicate-lock acquisition for a transaction that is
genuinely declared read only, so the level on its own would buy us nothing. We
were already passing that flag through, which is why this lands as two lines.

The second commit surfaces the `ConstraintName` and `Detail` fields of a
`pgconn.PgError` in the message our mapped error types render. `PgError.Error`
prints only the severity, the message and the SQLSTATE, so both fields were
going on the floor before they reached a log line. The detail is the only thing
that tells two very different `40001` aborts apart: a true SSI abort carries a
reason code naming the transaction's role in the conflict graph, while an
ordinary write-write conflict on the same row carries no detail at all. Now that
readers no longer take predicate locks, that's the signal that says whether a
given write path still leans on SSI or would be equally happy at `REPEATABLE
READ`. The constraint name matters for a different reason, covered below.

The third commit is `docs/postgres_isolation.md`, which writes down the policy,
the operator caveats, and the audit that decided where the boundary sits.

changes.

## Why writers stay at SERIALIZABLE

This one is the actual deliverable rather than a shortfall. The task was scoped
to go all the way and move writers too, and the audit came back saying we
shouldn't. Both halves of that judgement matter.

The upside is small. Writers are a rounding error in this workload. If 99.6% of
transactions change nothing, then relaxing the remaining 0.4% can't recover much
beyond what the read-only change already takes off the table. The read side was
the entire problem.

The risk is not small. An audit of all 64 write closures found fourteen sites of
concern. Eleven of those genuinely depend on the serialization graph and would
break if we relaxed the level, and several of them sit in fund-critical paths.
For calibration, the comparable audit in lnd found four such shapes across about
340 closures, so even counting the eleven alone our hit rate is far denser. The
full table lives in `docs/postgres_isolation.md` rather than here, classified
into three shapes: read-check-then-write, a lost creation race that surfaces as
a `23505`, and write skew.

Two structural blockers compound it. Under `REPEATABLE READ` the same-row
`40001` becomes the only conflict signal we have left, but
`TxAwareActorDeliveryStore.ExecTx` has no retry loop at all, and
`TransactionExecutor.ExecTx` skips its own whenever it joins an ambient actor
transaction, which is the normal case for the ledger, audit, credit and activity
stores. So the one signal that would have to absorb the change isn't retried on
the dominant code path. Separately, a lost creation race that SSI reports as a
retryable `40001` can become a `23505` unique violation under `REPEATABLE READ`,
and a retry loop correctly refuses to retry that.

The doc also inventories the six partial unique indexes in the schema, which is
useful well beyond this change. An `ON CONFLICT` target that doesn't match a
partial index predicate never arbitrates against that index, so the upsert
quietly degrades into a plain insert and you find out about it later.

Unlike lnd#10999 we don't add a configuration knob for the write level. A knob
whose non-default setting silently enables about a dozen known anomalies is a
loaded gun with a safety catch rather than a feature. It should become
configurable once those paths are hardened, and not before.

To be explicit about this, since the audit table reads like a bug list: this PR
doesn't break any of those sites. Writers keep talking to each other through the
full conflict graph exactly as before, because Postgres tracks dependencies only
among serializable transactions, and a read-only `REPEATABLE READ` transaction
was never able to protect a write in a different transaction anyway. The table
is a record of what would have to be fixed before writers could move, not of
what this change breaks.

The original draft went further and said that none of the audited sites is a
live bug today, on the grounds that SSI is masking all of them. An adversarial
pass over the table caught that, and it turns out to be half wrong in a way
worth knowing about. SSI promotes a lost creation race to a retryable `40001`
only when the losing transaction read the contested key before inserting it,
since that read is what leaves the predicate lock the conflict graph is built
from. Three of the fourteen sites insert blind, so there's no dependency for
SSI to find and they already lose with a bare `23505` at `SERIALIZABLE` today.
Those three are reachable now, independently of this work, and relaxing the
level would take nothing away from them. That's why the audit table now splits
shape B into read-check and blind, and why the count above is eleven rather than
fourteen. Both halves of the mechanism are pinned against a real Postgres in
`TestPostgresConflictShapes` rather than left as an argument. The remaining
eleven are genuinely masked today.

## What we're giving up

Snapshot isolation is not serializability, and this is a real if modest
weakening for readers. A read-only `REPEATABLE READ` transaction is no longer
guaranteed to observe a state corresponding to some serial ordering of the
writers running alongside it, so the read-only transaction anomaly described by
Fekete and O'Neil is once again permitted. The argument in the code comment is
that our read paths only ever consume a point-in-time view and never depended on
being ordered against writers in other transactions, and that a read feeding a
later write in a separate transaction was never protected across that boundary
at any isolation level. That claim is worth a reviewer checking rather than
taking at face value, since it's what the whole change rests on.

There's an operator consequence too, written up in the doc. A `REPEATABLE READ`
transaction pins its snapshot for its entire lifetime, and Postgres can't vacuum
row versions still visible to an open snapshot. Some of our read transactions
are long lived: the ancestry resolver walks a tree, and several round and VTXO
listings iterate large result sets. A slow one now delays cleanup and can bloat
tables, and it sits `idle in transaction` while the daemon computes between
queries, so `idle_in_transaction_session_timeout` and `statement_timeout` need
to be generous enough to cover a full pass. None of these hold a transaction
open across a network call or an actor send, so their duration is bounded by
database and CPU work rather than by a remote peer.

## What was and wasn't verified locally

The backend gate and the level selection are covered by a plain unit test that
runs everywhere. Everything that asserts on Postgres' own behaviour sits behind
the `test_postgres` build tag and wants a Docker Postgres, and those have been
run against a real fixture rather than left as an aspiration. They cover that
the requested options survive the pgx stdlib driver and reach the server as a
read-only `REPEATABLE READ` transaction, that the server rejects a write inside
one, that an SSI pivot abort carries a reason code, that a `REPEATABLE READ`
same-row conflict still raises a retryable `40001` with no detail, that a
`REPEATABLE READ` lost creation race raises a non-retryable `23505` instead,
and that under `SERIALIZABLE` the same race raises a `23505` when the loser
inserts blind but a `40001` with a pivot reason code when it reads first. The
isolation assertion reads back `SHOW transaction_isolation` and `SHOW
transaction_read_only`, so it's the server's own view of the transaction rather
than a restatement of what we asked for.

What isn't verified here: the signet numbers are a reported measurement rather
than something reproducible from this branch, and the write-path audit was
produced by reading code rather than by running anything against a live
database. The claim that no read path depends on cross-transaction
serializability is reasoned from the call sites in the same way, and isn't
something a test can pin down.

## Cross references

The retry gap named above is partly closed by #1053, which adds a retry loop to
the actor commit transaction. lightninglabs/lumos#718 and
lightninglabs/swapdk-server#265 are the incident this is meant to take pressure
off.

### DIFF
--- a/db/CLAUDE.md
+++ b/db/CLAUDE.md
@@ -216,3 +216,9 @@ when adding one.
 ## Deep Docs
 
 - [ARCHITECTURE.md](../ARCHITECTURE.md) — System-wide package map.
+- [docs/postgres_isolation.md](../docs/postgres_isolation.md) — Isolation
+  policy: read-only Postgres transactions run at `REPEATABLE READ` with
+  `READ ONLY` (no `SIRead` predicate locks, never a 40001), writers stay
+  `SERIALIZABLE`. Also holds the write-path snapshot-isolation audit and the
+  inventory of the six partial unique indexes that any new `ON CONFLICT`
+  target has to be checked against.
--- a/db/CLAUDE.md
+++ b/db/CLAUDE.md
@@ -221,4 +221,6 @@ when adding one.
   `READ ONLY` (no `SIRead` predicate locks, never a 40001), writers stay
   `SERIALIZABLE`. Also holds the write-path snapshot-isolation audit and the
   inventory of the six partial unique indexes that any new `ON CONFLICT`
-  target has to be checked against.
+  target has to be checked against, along with the caveat that a conflict
+  target can also miss a plain unique constraint declared inline in a
+  `CREATE TABLE`.
--- a/db/interfaces.go
+++ b/db/interfaces.go
@@ -367,13 +367,57 @@ type BaseDB struct {
 	*sqlc.Queries
 }
 
+// txIsolationLevel returns the isolation level a transaction against the given
+// backend should be opened with.
+//
+// Read-write transactions always run at SERIALIZABLE, which is the level the
+// storage layer has always promised writers. Read-only transactions on Postgres
+// are instead opened at REPEATABLE READ, which in Postgres is snapshot
+// isolation: the transaction reads from a single consistent snapshot for its
+// whole lifetime, taken when its first statement runs rather than at BEGIN.
+//
+// That is a real, if modest, weakening. Snapshot isolation is not
+// serializability, so the reader is no longer guaranteed to observe a state
+// that corresponds to some serial ordering of the writers running alongside it.
+// The read-only transaction anomaly described by Fekete and O'Neil is once
+// again permitted. We accept that because our read paths only ever consume a
+// point-in-time view of the database and never depended on being ordered
+// against writers in other transactions. A read that feeds a later write in a
+// separate transaction was never protected across that boundary at any
+// isolation level.
+//
+// In exchange, a read-only REPEATABLE READ transaction takes no part in
+// Postgres' serializable snapshot isolation conflict graph. It acquires no
+// SIRead predicate locks, is not itself subject to SSI serialization failures,
+// and can no longer cause a concurrent writer to be aborted as a pivot. Since
+// the daemon is extremely read heavy, this removes a large amount of needless
+// abort pressure from the system.
+//
+// SQLite is always effectively serializable because it only ever admits a
+// single writer, so there is nothing to gain there and we leave it alone.
+func txIsolationLevel(backend sqlc.BackendType,
+	readOnly bool) sql.IsolationLevel {
+
+	if readOnly && backend == sqlc.BackendTypePostgres {
+		return sql.LevelRepeatableRead
+	}
+
+	return sql.LevelSerializable
+}
+
 // BeginTx wraps the normal sql specific BeginTx method with the TxOptions
 // interface. This interface is then mapped to the concrete sql tx options
 // struct.
 func (s *BaseDB) BeginTx(ctx context.Context, opts TxOptions) (*sql.Tx, error) {
+	readOnly := opts.ReadOnly()
+
+	// The read-only flag is not just advisory here: Postgres only skips
+	// predicate lock acquisition for a transaction that is actually
+	// declared READ ONLY, so the flag is what makes the relaxed isolation
+	// level worth anything.
 	sqlOptions := sql.TxOptions{
-		ReadOnly:  opts.ReadOnly(),
-		Isolation: sql.LevelSerializable,
+		ReadOnly:  readOnly,
+		Isolation: txIsolationLevel(s.Backend(), readOnly),
 	}
 
 	return s.DB.BeginTx(ctx, &sqlOptions)
--- a/db/interfaces_postgres_test.go
+++ b/db/interfaces_postgres_test.go
@@ -1,0 +1,88 @@
+//go:build test_postgres
+
+package db
+
+import (
+	"database/sql"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// txSessionState reads back the isolation level and read-only access mode that
+// Postgres actually applied to the given transaction. Asserting on the server's
+// own view is the only way to prove that the requested sql.TxOptions survived
+// the trip through the pgx stdlib driver and into the BEGIN statement.
+func txSessionState(t *testing.T, tx *sql.Tx) (string, string) {
+	t.Helper()
+
+	var isolation string
+	err := tx.QueryRow("SHOW transaction_isolation").Scan(&isolation)
+	require.NoError(t, err)
+
+	var readOnly string
+	err = tx.QueryRow("SHOW transaction_read_only").Scan(&readOnly)
+	require.NoError(t, err)
+
+	return isolation, readOnly
+}
+
+// TestPostgresBeginTxIsolation asserts that read-only transactions against
+// Postgres are opened as a read-only REPEATABLE READ tx, while writers stay
+// SERIALIZABLE. The read-only flag matters as much as the level, because
+// Postgres only skips SIRead predicate lock acquisition for a transaction that
+// is genuinely declared read only.
+//
+// The subtests deliberately share one store. Each store costs a docker
+// container, and the fixture derives its port from a docker port binding that
+// is not always populated under load.
+func TestPostgresBeginTxIsolation(t *testing.T) {
+	t.Parallel()
+
+	ctx := t.Context()
+	store := NewTestPostgresDB(t)
+
+	t.Run("read-only", func(t *testing.T) {
+		tx, err := store.BeginTx(ctx, ReadTxOption())
+		require.NoError(t, err)
+		defer func() {
+			require.NoError(t, tx.Rollback())
+		}()
+
+		isolation, readOnly := txSessionState(t, tx)
+		require.Equal(t, "repeatable read", isolation)
+		require.Equal(t, "on", readOnly)
+	})
+
+	t.Run("read-write", func(t *testing.T) {
+		tx, err := store.BeginTx(ctx, WriteTxOption())
+		require.NoError(t, err)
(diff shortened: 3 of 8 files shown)


Roasbeef:
Ran an adversarial pass over this before asking for a bot review. Summary of what moved and what didn't.

**What changed as a result.** The audit table had four errors, all now fixed in two follow-up commits.

The big one: the doc claimed every site in the table is currently masked by SSI and so isn't a live bug today. That's only true for the sites that read the contested key before inserting it. SSI promotes a creation race to a retryable `40001` only when the loser holds a predicate lock from a prior read, and three of the sites insert blind, so they already lose with a bare `23505` at `SERIALIZABLE`. Those are reachable today and relaxing the level would take nothing away from them. Rather than argue about it I pinned both halves against a real Postgres in `TestPostgresConflictShapes`, and the blind-race subtest fails if you write it to the old claim. Shape B is now split into read-check and blind, and the headline count is restated as fourteen sites of concern of which eleven genuinely depend on SSI.

The table was also missing a row: `UpsertSession` targets `session_id` while `oor_session_registry` carries a partial unique index on the idempotency key, so two sessions racing on one key have different session IDs, the `DO UPDATE` never fires, and the loser gets a `23505`. The dedup probe runs in its own transaction, so the write is blind. Same shape as the `UpsertOperation` and `CreatePendingBoardingSweep` rows.

`UpsertBinding` was classified A only, but `oor_vtxo_bindings` declares `UNIQUE (session_id, output_index, link_kind)` on top of its primary key and the conflict target doesn't cover it, so it's B as well. That constraint is inline in the `CREATE TABLE` rather than a partial index, which means the partial index inventory structurally can't catch it; the inventory now says so.

Finally the `UpsertPendingIntent` row pointed at the wrong read. The `status <> 'failed'` guard is same-row with the write it races, so `REPEATABLE READ` still catches it as an ordinary write-write conflict. The real exposure is the `NOT EXISTS` anti-join over `pending_intent_anchors`. Also corrected a claim that the UTXO audit insert is targetless; it names its target, and is safe because that target matches a total index.

**What I pushed back on.** The suggestion to fold the duplicated v4/v5 `errors.As` pairs in `PgErrorDetail` and `PgErrorConstraint` into a shared unexported `pgErrorFields` helper. It'd save about ten lines, but both are exported and each reads fine standing alone, and the "walks the error chain four times" argument is about an error path that only runs when a transaction has already aborted. Not worth churning reviewed code for.

**What was checked and held.** That `READ ONLY` plus `REPEATABLE READ` survives the pgx v5 stdlib driver and reaches the server (asserted server-side via `SHOW transaction_isolation` / `SHOW transaction_read_only`, and it matches what `TxOptions.beginSQL` emits); that the backend gate can't leak RR onto SQLite, including the `BackendTypeUnknown` case falling to SERIALIZABLE; that no `ReadTxOption()` closure writes, directly or transitively; and that no read path depends on cross-transaction serializability, in particular that every polling loop reopens its transaction per attempt rather than pinning a snapshot across polls. The six partial unique index predicates and the "64 write closures" count both check out verbatim.

**One latent thing I did not touch,** since it's pre-existing and out of scope here: the ambient-join in `TransactionExecutor.ExecTx` ignores `txOptions` by design, so a store calling `ExecTx(..., WriteTxOption(), ...)` from inside an actor read closure would silently join a `READ ONLY` transaction and fail at runtime on Postgres while succeeding on SQLite. The code comment acknowledges the design. Worth a guard eventually, but not in this PR.

@claude review this

Roasbeef:
Thanks, the `Unwrap` finding is a good catch and it's fixed in 6b1e50d.

You're right that it's exactly the thing that undercut commit 2. I verified it the way you'd want: the mapped error really was a dead end for `errors.As`, so `PgErrorConstraint(mapped)` and `PgErrorDetail(mapped)` both returned empty while the rendered log line looked perfectly fine, which is why nothing noticed. `ErrSQLUniqueConstraintViolation` was indeed the only one of the five mapped types without an `Unwrap`.

The test gap you identified was the real problem, so I fixed that too rather than just the method. `TestPostgresConflictShapes` now asserts `PgErrorConstraint(mapped)` and `PgErrorDetail(mapped)` alongside the existing raw-error checks. I confirmed the assertion genuinely discriminates by disabling the new `Unwrap` and re-running: it fails on `Should NOT be empty` at the mapped-constraint assertion, and passes with the method restored. Extraction from the mapped error is the path that actually gets exercised downstream of `ExecTx`, so that's the one worth pinning.

On #2, agreed, and thanks for pulling the precedents. The rule is about production query paths going through sqlc for type safety and dialect portability, and these probes are the opposite case: they exist to construct specific conflict shapes at hand-picked isolation levels, which is precisely the control that generated queries would take away. Leaving them.

On the ambient-join note, agreed that the blast radius is marginally larger now that readers are genuinely `READ ONLY` server-side rather than nominally so. Still out of scope here, but your framing of the guard as cheap insurance is right and I'll carry it into the follow-up rather than let it evaporate.

On reader retry, that's the right reading and worth having on the record: a read-only `REPEATABLE READ` transaction takes no predicate locks and cannot write, so there's no way for it to raise a `40001`, and the absence of a retry loop on the relaxed path costs nothing. The retry gap in the audit is entirely writer-side, and writers stay `SERIALIZABLE` in this PR.

PR #1058, opened 2026-07-28
serverconn: keep inbound RPC dispatch out of the ingress write tx

Roasbeef:
In this PR, we stop the ingress loop from holding an open write transaction
across a network round trip to the operator. This is the follow-up that #1053
names in its "what this does not fix" section.

`runFoldedDispatch` (`serverconn/ingress.go`) opened a write transaction and
called `dispatchBatch` inside it. That's right for the durable half of a pulled
batch, where each dispatcher is an actor `Tell` whose enqueue joins the ambient
transaction and has to commit atomically with the pull cursor. It's wrong for
the mux-bridged half, where the dispatcher serves an inbound `KIND_REQUEST`
through the local handler and then puts the `KIND_RESPONSE` back on the wire
with `Edge.Send`. Nothing in that path touches the local durable mailbox, so
all it contributes to the transaction is its own latency.

That latency isn't theoretical on either backend. Production opens SQLite with
`_txlock=immediate`, so an open write transaction holds the one database-wide
write lock and every other writer in the process stalls behind it for as long
as the operator takes to answer, with no conflict required. On Postgres,
`db.BaseDB.BeginTx` pins SERIALIZABLE, so the same window becomes an SSI
conflict window wide enough to produce 40001 aborts, and it's a leading source
of the aborts behind lightninglabs/lumos#718 and the fund loss in
lightninglabs/swapdk-server#265. The operator side hit the identical shape in
lumos and measured 30 to 60 second stalls, which is why its `clientconn`
ingress deliberately dispatches outside the checkpoint transaction. The SQLite
half is the one more people feel, since SQLite is the default deployment and it
needs no conflict at all to bite.

There's a second interaction with #1053 worth spelling out. #1053 gives the
actor commit transaction a serialization-retry loop, and because
`dispatchBatch` sits inside that closure, the retry loop currently replays the
network send up to `numRetries` times per turn. This PR is what stops that.

## Splitting on the terminus

The connector can't infer which dispatchers do IO, because an
`EnvelopeDispatcher` is an opaque closure. So the wiring layer declares it
through the new `ConnectorConfig.NonTxRoutes` set, and
`waved.Server.buildRPCDispatchers` fills it in. Every route it bridges to the
local `ServeMux` resolves to `handleInboundRPC`, and those are precisely the
ones that must not run under the fold. The event routes the `EventRouter`
contributes are untouched, since they terminate in a durable actor `Tell` whose
enqueue belongs with the cursor.

An envelope is only hoisted when three conditions hold together, and each rules
out a different way of hoisting durable work by mistake (`isNonTxRequest`). Its
route has to be listed in `NonTxRoutes`, because that's the only place that
knows what the closure does. It has to be a `KIND_REQUEST`, so that a
`KIND_EVENT` or the no-waiter `KIND_RESPONSE` fallback, both of which resolve
to a durable `Tell`, can never leave the fold. And a dispatcher has to actually
be registered for the route, so the hoisted path never has to reproduce
`dispatchBatch`'s skip-warn for an unroutable envelope. Each condition fails
closed, so the worst a missing mark can do is leave today's behavior in place.

The kind condition needs one thing on the other side of it, which the review
below caught. The sender picks the kind, and `handleInboundRPC` never reads it:
it checks that `Rpc` and `Body` are present and then serves the envelope through
`ServeRPC` regardless. So an envelope stamped `KIND_EVENT` on a marked route, or
a `KIND_RESPONSE` on one whose waiter is gone, would fail the gate, fall into the
durable partition, and resolve right back to the same mux dispatcher with the
transaction open. `dispatchBatch` now skips any non-request envelope arriving on
a marked route, the way it already skips one the table cannot route. It skips
rather than erring on purpose: a dispatch error isn't permanent, so returning one
would back the loop off and re-pull the same envelope forever.

## Ordering, and why validation moved

The hoisted dispatch runs before the commit, not after it, and that ordering is
load-bearing. A crash between a hoisted send and the commit leaves the cursor
where it was, the batch is re-pulled, and the request is served again; the
operator demultiplexes the duplicate `KIND_RESPONSE` by correlation ID exactly
as it already does for the legacy non-transactional dispatch path. Committing
first and sending after would invert this into at-most-once, where a crash in
the window advances the cursor past a request that was never answered and the
caller only ever sees its own RPC deadline. That trades a benign duplicate for
a silently lost request, which is the wrong direction on
`docs/RPC_MAILBOX_CONTRACT.md`.

Because a hoisted send can't be taken back, `validateInboundEnvelope` also
moves from a lazy check on the durable partition inside `dispatchBatch` to a
sweep over the whole pulled batch up front. That part is required rather than
cosmetic: finding a version mismatch at seq 5 after having already answered seq
1 would be a new exposure. One permanently incompatible envelope anywhere in
the batch now stops the loop before any of the batch is acted on.

## What I'm not confident about

Five things, and I'd rather list them than have this read as more finished than
it is.

`NonTxRoutes` is unsafe by omission. A future route added to `Dispatchers`
without a matching `NonTxRoutes` entry silently reintroduces the bug. What's
there today is a helper in `buildRPCDispatchers` that writes both maps in one
call, doc comments on both the config field and the gate, and an invariant in
`serverconn/CLAUDE.md`, but none of that is compile-enforced. The alternative
is to make the dispatch map's value a struct with an explicit `NonTx bool` so
that registration can't forget, which costs a change to the config field's
shape and touches every registration site including
`EventRouter.AsDispatcherMap`. I went with the set because it keeps the diff on
the ingress path instead of churning every call site, but I don't think that's
obviously the right call and I'd take the struct if a reviewer prefers it.

Only `waverpc.DaemonService/GetInfo` is hoisted today, and it's a read, so the
crash-duplicate costs nothing right now. It stops costing nothing the moment
the TODO in `buildRPCDispatchers` lands `WalletService.SignVTXO` or
`RoundService.SubmitNonces`: correlation-ID dedup on the operator makes a
duplicate *response* harmless, it does not make a second *signature* harmless.
There's a note saying so at the gate itself, on the theory that whoever wires
the next route in will be reading `isNonTxRequest` rather than this PR.

No `make systest` run. The change is exercised by unit tests against a modeled
writer lock, not against real SQLite or real Postgres.

I did not audit the `ServeRPC` handlers for whether any of them do their own
database writes. It's tempting to argue that none can, since under the old code
such a handler would have deadlocked against the ingress writer lock, but the
absence of a reported deadlock isn't evidence and I don't want to lean on it.

Batch-partial failure is unchanged in class but now includes sends. If seq 1 is
served and seq 3 fails, the cursor doesn't move and seq 1 is served again on
the re-pull. That's the same at-least-once exposure the legacy dispatch path
has always had, now covering one more kind of step.

## Testing

Four tests in `serverconn/ingress_nontx_test.go`, plus coverage of the gate in
`serverconn/ingress_split_test.go`. The regression test asserts the
property directly rather than by timing: it backs the connector with a store
that takes a real mutex for the lifetime of `ExecTx`, modeling the single
global writer, and has both dispatchers report whether that lock was held when
they ran. The hoisted request must observe it free, the durable dispatch in the
same batch must still observe it held, and the recorded step order has to put
the serve before the commit. The other three cover a failed hoisted dispatch leaving the
cursor unmoved, one invalid envelope stopping the loop before anything in the
batch is delivered, and the mislabeled-kind skip described above.
`TestIsNonTxRequestRequiresAllConditions` walks the gate's three conditions one
at a time.

I mutation-tested both guards by hand rather than trusting that green is green.
Forcing `isNonTxRequest` to return false fails three tests, and disabling the
mislabeled-kind skip fails the new one with `[true true]`, meaning both
mislabeled envelopes ran the mux bridge with the modeled writer lock held. So
those assertions are load-bearing and not passing for the wrong reason.

`make unit pkg=serverconn`, `make unit pkg=waved`, `make lint-changed-local`
(0 issues) and `make commitmsg-lint` all pass. See the individual commit
messages for the incremental detail.


Roasbeef:
A Fable 5 reviewer went over this adversarially before I asked for a bot pass: invariant verification with concrete crash interleavings, a simplification pass, and a narrow hunt for other live instances of "IO inside an open write transaction" across `serverconn/`, `waved/` and `mailbox/`. Two commits came out of it, plus one I added after checking a third-order claim it made.

The real finding is that the kind guard left the other half of its own hole open. `isNonTxRequest` only hoists a `KIND_REQUEST`, which is what stops a durable `Tell` from escaping the fold, and that part is right. But the sender picks the kind and `handleInboundRPC` never reads it: it checks that `Rpc` and `Body` are present, then serves through `ServeRPC` and answers with `edge.Send`. So an envelope stamped `KIND_EVENT` on a marked route, or a `KIND_RESPONSE` on one whose waiter is gone, failed the gate, fell into the durable partition, and resolved right back to the same mux dispatcher with the transaction open. The exact stall this PR removes, still reachable, just by mislabeling the envelope. `dispatchBatch` now skips any non-request envelope arriving on a marked route. It skips rather than returning an error on purpose, and the reviewer was right to flag why: a dispatch error isn't permanent, so erring would back the loop off and re-pull the same envelope forever, trading a stall for a wedge. `TestDispatchBatchSkipsMislabeledNonTxKinds` covers it, and with the guard disabled it fails with `[true true]`, meaning both mislabeled envelopes ran the mux bridge with the modeled writer lock held.

Second, smaller: the `!ok` branch in `dispatchNonTxRequests` skipped silently. It's unreachable while the dispatch table is built once at wiring time, but it's the one place this path can advance the cursor past a request nobody answered, so it warns now instead of doing that quietly.

I also went and checked the reviewer's best secondary observation, that mis-marking a *durable* route as non-tx is worse than the stall it's meant to prevent, and it holds up. The hoisted `Tell` would commit ahead of the cursor, and the ingress path propagates no outbox ID, so `DurableMailbox.Send` falls through to `generateID()` and the re-pull after a crash in that window enqueues a second copy under a fresh UUIDv7 that `EnqueueMailboxMessage`'s `ON CONFLICT (id) DO NOTHING` can't collapse. Exactly-once local delivery becomes a duplicate the receiving actor can't detect. That's the actual reason the two maps have to be written together, so it's now on the `NonTxRoutes` doc comment rather than living in my head.

Two things I didn't take.

The reviewer looked at the `NonTx bool` struct alternative I raised in the description and came down against it for now, on the grounds that the struct only removes the maps-desync hazards that gate conditions 2 and 3 plus the `addMuxRoute` helper already neutralize, and can't remove the hazard that matters (marking a durable closure) because the closure stays opaque either way. I find that convincing enough to leave as is, but it's still the open decision on this PR and I'd rather a human settle it than have two agents agree with each other.

It also suggested having `splitIngressEnvelopes` return the resolved dispatcher alongside each hoisted envelope, which would delete the second map lookup and the impossible `!ok` branch outright. That's a fair simplification and it would moot the warn I just added. I went with the warn because the split's job is partitioning, and widening it to return pairs couples it to the dispatch table for the sake of one branch that's already unreachable. Preference call, not a strong one.

The variant hunt came back with exactly one live instance, and it's the mislabeled-kind path above. Otherwise `runFoldedDispatch` is the only `ExecTx` in the three trees, the event-router dispatchers inside the fold are DB-only (the wake-channel send in `DurableMailbox.Send` is non-blocking), and every egress `Edge.Send` site runs outside any transaction. It also resolved the unaudited-`ServeRPC`-handlers gap from the description without using the unsound "it would have deadlocked" argument: ingress reachability is gated by the `Dispatchers` map, only `GetInfo` is in it, and `RPCServer.GetInfo` does no DB writes. It does make chain-backend height queries, which is more network IO that used to sit inside the transaction.

Still open and unchanged: no `make systest` run locally, so real SQLite and real Postgres behavior under this path is unobserved. CI runs both, which covers some of it.

`make unit pkg=serverconn`, `make unit pkg=waved`, `-race` on the ingress tests, `make lint-changed-local` (0 issues) and `make commitmsg-lint` all pass.

@claude review this

Roasbeef:
Took two of the three, pushed back on one.

**The waiter-response asymmetry is real and is fixed** (`0123136`). You're right that the skip sat ahead of the kind switch, so a `KIND_RESPONSE` on a marked route was dropped even with a live waiter registered, and right that it's legacy-path-only because `splitIngressEnvelopes` peels waiter-backed responses off before the fold ever calls `dispatchBatch`. I didn't want to fix it with a comment. "Marked routes never carry waiter-backed responses" was an unstated invariant holding up a silent drop, and a silently dropped response that a caller is blocked on is the same failure mode this PR argues against in its ordering section, so leaving it unreachable-but-undefended read wrong. The skip now goes through `resolvesToNonTxDispatcher`, which asks the narrower question: would this envelope actually reach a dispatcher here? A `KIND_REQUEST` never (it's the kind the mark is about), a `KIND_RESPONSE` with a live waiter never (it's delivered in memory and breaks out). `TestDispatchBatchKeepsWaiterResponseOnNonTxRoute` drives `dispatchBatch` directly, which is the only path that can present one. Removing the carve-out fails it. The wait is bounded to a second on purpose, since without the bound the regression shows up as a package-wide hang rather than a failure.

**Took the wiring-test suggestion too** (`0f75cd2`), and I think it was the most useful thing in your review. `TestNonTxRoutesResolveToMuxBridge` builds the dispatch table and the event-router map from the same `Server` and asserts every marked route resolves to a dispatcher, and never to one the `EventRouter` contributed, comparing closures by code pointer since that's the only identity Go gives you. Marking `oorpb.OORMailboxService/SubmitPackage` fails it with the route named. That's aimed at the hazard that actually costs something rather than the desync one: a durable `Tell` hoisted out of the fold commits ahead of the cursor, and the re-pull after a crash in that window enqueues a second copy under a fresh UUIDv7 that `ON CONFLICT (id) DO NOTHING` can't collapse. It doesn't make a mis-mark impossible, but it moves the failure from production to CI, which is most of the value the struct shape would have bought.

**Pushed back on the double validation.** The inner `validateInboundEnvelope` in `dispatchBatch` isn't dead work, it's just redundant on the folded path. `dispatchBatch` is still called directly on the legacy non-`TxAware` store path (`ingress.go:178`), and that path has no up-front sweep, so removing the inner call would silently stop validating inbound versions there. Agreed it's not worth churning, but for the opposite reason to the one given: it's load-bearing, not merely cheap.

On the `NonTxRoutes` shape, that's now two reviewers landing on "keep the set", which I'm reading as weak evidence rather than strong, since neither of you is the person who has to live with the next route being added. Leaving it as the open question for a human.

Issue #955, opened 2026-07-15
ci: TestDurableAskConcurrentRequests flakes on Postgres


bhandras:
## Summary

`TestDurableAskConcurrentRequests` timed out in the Postgres unit matrix for wavelength #954 even though the PR does not modify the durable actor implementation or `internal/actortest`. The same CI run passed the changed Postgres `db` package and every other completed package.

## Failure

- Workflow run: https://github.com/lightninglabs/wavelength/actions/runs/29431960832
- Failed job: https://github.com/lightninglabs/wavelength/actions/runs/29431960832/job/87409014847
- PR: https://github.com/lightninglabs/wavelength/pull/954
- Tested merge SHA: `e7c939c57ce9d649279bf89e4eeb414eb8ec82bc`
- Runner: `lightninglabs-runnerdeploy-9v22g-4gksk`
- Matrix: `unit tags="test_postgres"`

Relevant output:

```text
--- FAIL: TestDurableAskConcurrentRequests (37.83s)
    test_postgres.go:29: Creating new Postgres DB for testing
    postgres_fixture.go:97: Connecting to Postgres fixture: ...localhost:32990/test?sslmode=disable
    e2e_test.go:1637: condition not met within timeout
FAIL
FAIL github.com/lightninglabs/wavelength/internal/actortest 47.104s
```

The timeout is `eventuallyWithOutboxPublish(... durableAskResponseTimeout ...)`, waiting for `senderBehavior.AskResponseCount() >= 5`. The helper explicitly forces an outbox publish on every polling iteration, so this failure means at least one concurrent response was still not observed before the deadline.

## Why this appears unrelated to #954

#954 changes activity projection and wallet-local credit ownership under `db/activity_store*`, `db/sqlc/activity_log*`, and `swapwallet/*`. It does not modify `internal/actortest`, the durable actor/outbox implementation, or its schemas. In the failed Postgres matrix, the changed `github.com/lightninglabs/wavelength/db` package passed in 173.921 seconds before the actor test timed out. The SQLite unit/system matrices and the durable mailbox model check also passed.

## Suggested investigation

1. Reproduce the Postgres case repeatedly with `-run TestDurableAskConcurrentRequests -count=N`, including concurrent package load.
2. On timeout, print the observed response count and correlation IDs rather than only `condition not met`.
3. Inspect pending/in-flight outbox rows and publisher errors for the missing correlation ID.
[...]

PR #1059, opened 2026-07-28
wallet: cap per-round boarding fanout to bound nonce signing (#858)

### LINKED ISSUE #858: Auto boarding can miss nonce timeout with 500 VTXOs on signet
## Summary

`darepod` can auto-replay a persisted boarding request that successfully joins a
signet round, but then misses the operator's VTXO nonce collection window when
the replay expands into 500 VTXO outputs.

In the reproduced run, the client accepted the seal-time quote, validated the
commitment transaction, then spent about 19 seconds generating MuSig2 nonces for
500 signer sessions. The server immediately failed the round with:

```text
ClientFailed: VTXO nonce collection timeout
```

The wallet stays in pending boarding state and no VTXOs materialize.

## Environment

- Repo: `lightninglabs/darepo-client`
- Local submodule checkout: `d86a09e789929dddba9b377b3bb8fd8084b496f5`
- `darepod getinfo` reports:
  - version: `0.0.2-alpha`
  - commit: `34d5ecb`
  - network: `signet`
  - wallet type: `lwwallet`
- Server:
  - `arkd-signet.testnet.lightningcluster.com:443`
  - `max_vtxo_amount`: `200000`
  - `max_user_balance`: `100000000`
  - `min_operator_fee`: `1000`

## Reproduction

Start the signet daemon with debug logging and unlock the existing wallet:

```shell
darepod --network=signet \
  --datadir="$HOME/.darepod-signet" \
  --debuglevel=debug \
  --rpc.listenaddr=localhost:10029 \
  --server.host=arkd-signet.testnet.lightningcluster.com:443 \
  --server.transport=grpc \
  --server.insecure=false \
  --swap.serveraddress=swapd-signet.testnet.lightningcluster.com:443 \
  --swap.servertransport=grpc \
  --swap.serverinsecure=false \
  --wallet.type=lwwallet \
  --wallet.esploraurl=https://mempool-signet.testnet.lightningcluster.com/api \
  --wallet.pollinterval=90s \
  --maxoperatorfeesat=50000 \
  --rpc.notls \
  --rpc.no-macaroons

darepocli --rpcserver=localhost:10029 --no-tls --no-macaroons unlock \
  --wallet_password_file=/tmp/darepod-signet-wallet-pass
```

The daemon then auto-replays a persisted board request.

## Observed log timeline

```text
15:07:40.264 Replaying persisted Board request after restart target_vtxo_count=0 live_anchor_count=4 live_intent_count=1 stale_intent_count=0
15:07:40.308 Boarding request accepted boarding_balance=401999632 vtxo_amount=100000000 vtxo_count=500 change_amount=301999632
15:07:49.701 Processing board request vtxo_count=500 boarding_input_count=4
15:07:55.060 Sending JoinRoundRequest to server boarding_requests=4 vtxo_requests=500 forfeit_requests=0 leave_requests=1
15:07:56.322 Intent admitted; awaiting seal-time quote round_id=019f377e-1ecd-75c0-a2d8-676980d7b8be boarding_intent_count=4 vtxo_intent_count=500
15:08:07.267 Received seal-time quote round_id=019f377e-1ecd-75c0-a2d8-676980d7b8be operator_fee_sat=1020
15:08:14.259 Commitment transaction validated successfully round_id=019f377e-1ecd-75c0-a2d8-676980d7b8be client_trees=500 vtxo_tree_count=4
15:08:14.259 Generating MuSig2 nonces for VTXO tree signing round_id=019f377e-1ecd-75c0-a2d8-676980d7b8be boarding_intent_count=4 vtxo_intent_count=500
15:08:33.187 Generated MuSig2 nonces, sending to server round_id=019f377e-1ecd-75c0-a2d8-676980d7b8be session_count=500 signer_key_count=500
15:08:33.187 outbox event index=0 type=*round.SubmitNoncesRequest
15:08:33.191 State transition from_state=NoncesSent to_state="ClientFailed: VTXO nonce collection timeout"
```

The relevant client path appears to be `round/transitions.go`, where
`CommitmentTxValidatedState` handles `GenerateNonces` synchronously by creating
one `tree.NewSignerSession` per requested VTXO and then calling
`session.GetNonces()` for each session before it emits `SubmitNoncesRequest`.

## Current wallet state after failure

```json
{
  "confirmed_sat": "0",
  "pending_in_sat": "401999632",
  "pending_out_sat": "0",
  "credit_available_sat": "0",
  "credit_reserved_sat": "0"
}
```

`darepocli ark vtxos list` returns an empty list.

## Expected behavior

Auto round joining should either complete for the fanout that the client chooses
or choose a smaller fanout that fits the operator's nonce collection deadline.

If the operator timeout is expected for large batches, the client should recover
by retrying with smaller batches or otherwise avoid repeatedly selecting a batch
shape that cannot finish.

## Possible fixes to consider

- Cap auto-boarding fanout based on the operator's nonce collection timeout or a
  locally measured nonce-generation budget.
- Generate nonce sessions in parallel if the wallet signer path supports that
  safely.
- Split recovered boarding intents into smaller rounds after a recoverable
  `VTXO nonce collection timeout`.
- Surface a clearer status to the user when automatic boarding fails after
  admission, since the visible balance remains pending.

ellemouton:
## Problem

Auto-boarding a large deposit can strand it. A confirmed boarding balance is
split into VTXO pieces sized by the operator's per-VTXO maximum, so under terms
like `MaxUserBalance / MaxVTXOAmount = 500` a single board mints ~500 outputs.
Boarding signs one MuSig2 nonce per VTXO — an O(N) step the operator collects
within a bounded round window — so that fanout overruns the window on signet.
The round fails **recoverable**, the deposit never converts to VTXOs, and the
board-intent replayer re-issues the **identical** 500-piece shape on every
retry (and every restart), so it re-fails forever. Liveness, not funds — the
deposit sits in `pending_in_sat` with an empty VTXO list.

## Fix

Cap the per-round fanout at a configurable `DefaultMaxVTXOsPerBoardRound` (128)
inside `clampBoardingAmount`: board at most the cap's worth of max-size VTXOs
and route the remainder to the **existing** change leave output, which re-boards
over successive rounds. Because it reuses the change-routing already there, the
split's `minByMax` lands ≤ cap automatically and value is conserved.

128 keeps per-round signing well inside a typical window even on the slowest
serial-signing backend (~5s at the ~38ms/nonce observed on signet) while still
boarding a large balance in a handful of rounds. Tunable via
`WithMaxVTXOsPerBoardRound` for a tighter/looser operator window.

## Testing

- New `TestClampBoardingAmountPerRoundCap`: the exact 500-piece signet scenario
  — asserts the uncapped path still forces 500 (the bug), the capped path mints
  128 with the remainder as change, conservation holds, the change re-boards
  again capped (convergence), and a non-binding cap leaves the split untouched.
- Extended the existing rapid fuzz to draw a per-round cap and re-assert the
  clamp invariants (conservation, `[floor, maxVTXO]` piece bounds) hold with it.
- Full `wallet` suite green; daemon build, `fmt-changed`, `lint-changed-local`
  clean.

## Scope

This is the proactive cap (root-cause). The operator doesn't advertise its
nonce-collection deadline, so the cap is a conservative local assumption; a
**reactive back-off** that shrinks the target on a nonce-collection-timeout
failure code (plumbed through `RoundFailedNotification`) is the natural
follow-up for operators with an unusually tight window.

Closes #858

### DIFF
--- a/wallet/board_limits.go
+++ b/wallet/board_limits.go
@@ -26,6 +26,27 @@ const minChangeFloor = btcutil.Amount(330)
 // minting oversized or unboundedly many outputs.
 const maxBoardOutputs = 1000
 
+// DefaultMaxVTXOsPerBoardRound bounds how many VTXO outputs a single boarding
+// round mints. Boarding signs one MuSig2 nonce per VTXO — an O(N) step the
+// operator collects within a bounded round window — so a balance whose
+// per-VTXO split (MaxUserBalance / MaxVTXOAmount) needs many hundreds of
+// pieces can overrun that window; the round fails "recoverable", the deposit
+// never converts, and every retry re-issues the same oversized shape and
+// re-fails, stranding the deposit (wavelength#858).
+//
+// 128 keeps the per-round signing work well inside a typical window even on
+// the slowest (serial-signing) backend (~5s at the ~38ms/nonce observed on
+// signet). The clipped remainder returns as a change leave output paying a
+// fresh boarding script, which re-confirms on-chain as a new boarding intent
+// ready to board again. That next board is driven the same way any board is:
+// automatically on the next confirmation under WithEagerRoundJoin, or by the
+// operator's/host's usual board trigger otherwise — so a large balance clears
+// over a handful of rounds rather than stranding on one oversized shape. It is
+// a deliberately conservative default because the operator does not advertise
+// its nonce-collection deadline; tune it with WithMaxVTXOsPerBoardRound for a
+// tighter or looser window.
+const DefaultMaxVTXOsPerBoardRound uint32 = 128
+
 var (
 	// ErrBoardingCapReached is returned when the operator's maximum
 	// user balance leaves no usable headroom to board any of the
@@ -102,7 +123,8 @@ type boardingClamp struct {
 // change output could neither be created in the batch transaction nor
 // re-boarded later.
 func clampBoardingAmount(total btcutil.Amount, targetCount uint32, maxVTXO,
-	headroom, floor btcutil.Amount) (*boardingClamp, error) {
+	headroom, floor btcutil.Amount,
+	perRoundVTXOCap uint32) (*boardingClamp, error) {
 
 	if total <= 0 {
 		return nil, fmt.Errorf("boarding balance must be positive")
@@ -117,6 +139,35 @@ func clampBoardingAmount(total btcutil.Amount, targetCount uint32, maxVTXO,
 		boardAmt = headroom
 		cappedByHeadroom = true
 	}
+
+	// Cap the per-round fanout so the client's O(N) nonce signing stays
+	// inside the operator's nonce-collection window. A balance whose split
+	// (MaxUserBalance / MaxVTXOAmount) would need more than perRoundVTXOCap
+	// pieces can overrun that window and strand the deposit
+	// (wavelength#858), so board at most the cap's worth of max-size VTXOs
+	// and let the remainder return as change to re-board next round. This
+	// reuses the change-routing below, and bounds minByMax at or under the
+	// cap. Zero disables the per-round cap; it needs a per-VTXO maximum to
+	// size a piece.
+	if perRoundVTXOCap > 0 && maxVTXO > 0 {
+		cap64 := int64(perRoundVTXOCap)
+		perRoundMax := cap64 * int64(maxVTXO)
+
+		// Clip only when the bound is trustworthy and spendable. The
+		// division check skips an overflowed multiply. Requiring
+		// perRoundMax >= floor keeps the clipped board above dust and
+		// preserves error attribution: on an absurdly small per-VTXO
+		// max (maxVTXO < floor) the clip is skipped, so the switch
+		// below reports ErrMaxVTXOBelowFloor, not a misleading
+		// below-floor error against the full balance.
+		if perRoundMax/cap64 == int64(maxVTXO) &&
+			perRoundMax >= int64(floor) &&
+			int64(boardAmt) > perRoundMax {
+
+			boardAmt = btcutil.Amount(perRoundMax)
+		}
+	}
+
 	switch {
 	case boardAmt <= 0:
 		// Headroom is zero or negative: the wallet is already at or
@@ -215,6 +266,24 @@ func clampBoardingAmount(total btcutil.Amount, targetCount uint32, maxVTXO,
 				count = uint32(maxByFloor)
 			}
 
+			// Cap the piece count at the per-round budget so the
+			// O(N) nonce signing stays inside the operator's
+			// collection window (wavelength#858). Clipping boardAmt
+			// alone does NOT bound the count: the window above
+			// honors the caller's targetCount up to maxByFloor, so
+			// a large requested count with a small floor could
+			// still mint far more than the cap's worth of
+			// sub-maximum pieces. The boardAmt clip guarantees
+			// perRoundMax = cap * maxVTXO, hence minByMax <= cap,
+			// so capping here never drops count below minByMax; and
+			// since count stays <= maxByFloor the (now fewer,
+			// larger) pieces remain within [floor, maxVTXO].
+			if perRoundVTXOCap > 0 &&
+				count > perRoundVTXOCap {
+
+				count = perRoundVTXOCap
+			}
+
 		// Infeasible even at the fewest pieces the maximum allows:
 		// boardAmt cannot be divided into whole [floor, maxVTXO]
 		// pieces. Board as many full maximum-size VTXOs as fit and
@@ -293,8 +362,19 @@ func (a *Ark) applyBoardingLimits(ctx context.Context, total btcutil.Amount,
 		floor = minChangeFloor
 	}
 
+	// Resolve the per-round VTXO cap, defaulting when unset so the
+	// per-VTXO-maximum-driven fanout is always bounded (wavelength#858).
+	// The cap engages only when the operator advertises a per-VTXO maximum;
+	// with no per-VTXO max the split is instead bounded by the caller's
+	// target count and the dust floor, which is not the #858 vector.
+	perRoundCap := a.maxVTXOsPerBoardRound
+	if perRoundCap == 0 {
+		perRoundCap = DefaultMaxVTXOsPerBoardRound
+	}
+
 	clamp, err := clampBoardingAmount(
 		total, targetCount, terms.MaxVTXOAmount, headroom, floor,
+		perRoundCap,
 	)
 	if err != nil {
 		return nil, nil, err
--- a/wallet/board_limits_test.go
+++ b/wallet/board_limits_test.go
@@ -211,7 +211,7 @@ func TestClampBoardingAmount(t *testing.T) {
 
 			clamp, err := clampBoardingAmount(
 				tc.total, tc.targetCount, tc.maxVTXO,
-				tc.headroom, dust,
+				tc.headroom, dust, 0,
 			)
 			if tc.wantErr != nil {
 				require.ErrorIs(t, err, tc.wantErr)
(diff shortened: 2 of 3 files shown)


ellemouton:

It fixes auto-boarding stranding a large deposit (#858): a boarding balance splits into VTXO pieces sized by the operator's per-VTXO max, so `MaxUserBalance / MaxVTXOAmount = 500` mints ~500 outputs in one round, and the O(N) per-VTXO nonce signing overruns the operator's round window — the round fails "recoverable", the deposit never converts, and the replayer re-issues the identical shape forever. The fix caps the per-round fanout (`DefaultMaxVTXOsPerBoardRound = 128`, configurable via `WithMaxVTXOsPerBoardRound`) by clipping `boardAmt` to `cap × maxVTXO` in `clampBoardingAmount` and routing the remainder to the existing change leave output, which re-boards over successive rounds.

Please pay particular attention to:
- **The cap clip in `clampBoardingAmount`** — is it placed correctly relative to the existing headroom clip and the sub-floor change-shift logic? Any interaction where the cap clip could push `boardAmt` below the floor, or leave a sub-dust change that the later floor-shift then mishandles?
- **Overflow guard** — `perRoundMax := cap64 * int64(maxVTXO)` with the `perRoundMax/cap64 == int64(maxVTXO)` division check: is that a sound int64-overflow guard, and is skipping the clip on overflow the right degradation?
- **Convergence** — does the clipped remainder genuinely re-board (via `boardingChangeLeave`) and shrink each round so a large balance completes in a bounded number of rounds rather than looping?
- **Default value** — is 128 a reasonable conservative default given the unknown operator nonce-collection window, and is the config knob wired the right way?
- **Scope** — I deferred the reactive back-off (shrink the target on a nonce-collection-timeout failure code) as a follow-up; flag if you think it belongs here.

ellemouton:
Thanks for the thorough review — addressed all three non-blocking items plus the test suggestion in the latest push:

- **(a) Error attribution** — added `perRoundMax >= floor` to the cap-clip guard. On an absurdly small per-VTXO max (`maxVTXO < floor`) the clip is now skipped, so the switch reports `ErrMaxVTXOBelowFloor` accurately instead of a misleading below-floor error against the full balance.
- **(b) Scope caveat** — reworded the `applyBoardingLimits` comment: the cap bounds the *per-VTXO-maximum-driven* fanout (the #858 vector); with no per-VTXO max the split is bounded by the caller's target count and the floor, not this cap.
- **(c) Style nit** — added the missing blank line before the `switch`.
- **Test** — added an explicit case where `total` sits just above `perRoundMax` so the change-shift fires on top of the cap clip (change pinned to the floor, count ≤ cap, value conserved).

Kept the reactive back-off out of scope as a follow-up, per your agreement.

Note on CI: the two red postgres jobs are pre-existing flakes unrelated to this boarding-only change — they failed on *different* unrelated tests each run (`TestDurableAskWithSpecialCorrelationIDs`, `TestConsiderRedeemInterlock`) while sqlite + `unit-race`/`cover`/`swapruntime` pass consistently (cf. #955).

ellemouton:
### Follow-up: make the per-round cap operator-advertised

This PR ships the per-round VTXO cap as a **client-side proactive guard** with a configurable local default (`DefaultMaxVTXOsPerBoardRound = 128`, tunable via `WithMaxVTXOsPerBoardRound`). 128 is a deliberately conservative *guess*, because the operator doesn't currently advertise its nonce-collection budget.

The natural follow-up is to make it **server-authoritative**: have the operator advertise a per-round board cap (e.g. `MaxBoardVTXOsPerRound`) in its `OperatorTerms`, and have the client prefer that over the local default — falling back to the default only for operators too old to advertise it. That removes the guess and caps boarding to the operator's *actual* rule, so the client never shapes a round-join it knows the server will reject.

The client-side seam already exists here: `applyBoardingLimits` fetches operator terms and clamps against `MaxVTXOAmount`/`MaxUserBalance`, and `clampBoardingAmount` now takes a `perRoundVTXOCap` parameter — so the client change is just sourcing `perRoundCap` from the advertised term (fallback to the default). The bulk is server-side (lumos computing + advertising the value + the wire-contract field).

Design note: the operator's true constraint is a nonce-collection *time window*, and the max count that fits depends on client signing throughput — so advertising a conservative *count* (server-picked from its round timeout) is the pragmatic, client-simple choice, rather than advertising the raw window and making each client estimate its own throughput.

Planned as a separate cross-repo PR pair (lumos terms + wavelength client), with a `waved.Config` knob for the cap folded in.

bhandras (approved):
LGTM 🎉

wallet/board_limits.go
```diff
@@ -117,6 +135,35 @@ func clampBoardingAmount(total btcutil.Amount, targetCount uint32, maxVTXO,
...
+	// Cap the per-round fanout so the client's O(N) nonce signing stays
+	// inside the operator's nonce-collection window. A balance whose split
+	// (MaxUserBalance / MaxVTXOAmount) would need more than perRoundVTXOCap
+	// pieces can overrun that window and strand the deposit
+	// (wavelength#858), so board at most the cap's worth of max-size VTXOs
+	// and let the remainder return as change to re-board next round. This
+	// reuses the change-routing below, and bounds minByMax at or under the
+	// cap. Zero disables the per-round cap; it needs a per-VTXO maximum to
+	// size a piece.
+	if perRoundVTXOCap > 0 && maxVTXO > 0 {
```
bhandras, line 152: afaict, this does not really cap the `targetCount` right? Looking a bit further i can actually see `count := targetCount`

wallet/board_limits_test.go
```diff
@@ -255,6 +255,90 @@ func TestClampBoardingAmount(t *testing.T) {
...
+		total, 0, maxVTXO, bigCap, floor, 0,
+	)
+	require.NoError(t, err)
+	require.Equal(t, uint32(500), uncapped.VTXOCount)
+	require.Equal(t, btcutil.Amount(0), uncapped.Change)
+
+	// Capped: board at most roundCap max-size VTXOs; the remainder returns
+	// as change to re-board next round.
+	capped, err := clampBoardingAmount(
+		total, 0, maxVTXO, bigCap, floor, roundCap,
```
bhandras, line 289: So to the above, here the targetCount is set to zero which masks the bug

wallet/board_limits.go
```diff
@@ -26,6 +26,23 @@ const minChangeFloor = btcutil.Amount(330)
...
+// operator collects within a bounded round window — so a balance whose
+// per-VTXO split (MaxUserBalance / MaxVTXOAmount) needs many hundreds of
+// pieces can overrun that window; the round fails "recoverable", the deposit
+// never converts, and every retry re-issues the same oversized shape and
+// re-fails, stranding the deposit (wavelength#858).
+//
+// 128 keeps the per-round signing work well inside a typical window even on
+// the slowest (serial-signing) backend (~5s at the ~38ms/nonce observed on
+// signet) while still boarding a large balance in a handful of rounds: the
+// clipped remainder returns as a change leave output that re-boards next
```
bhandras: nit: my agent marked that this is not automatic though as it only re-boards if EagerRoundJoin is set.

wallet/board_limits.go
ellemouton: Good catch — you're right, and this was a real bug, not just a naming thing. Clipping `boardAmt` only bounded the *lower* edge of the count window (`minByMax ≤ cap`); the feasible branch then still honored `targetCount` up to `maxByFloor`, so a large requested count with a small floor could mint far more than `cap` sub-maximum pieces — the exact O(N) fanout this is meant to prevent.

Fixed by also capping `count` at `perRoundVTXOCap` in the feasible branch (just pushed). The `boardAmt` clip guarantees `perRoundMax = cap × maxVTXO` hence `minByMax ≤ cap`, so capping never drops `count` below `minByMax`; and since the capped count stays `≤ maxByFloor`, every (now fewer, larger) piece remains within `[floor, maxVTXO]`. Both mechanisms are needed: the amount clip keeps pieces `≤ maxVTXO` at `count = cap`, the count cap keeps `VTXOCount ≤ cap`.

wallet/board_limits_test.go
ellemouton: Exactly — all the existing cases passed `targetCount=0` (→1, then raised to `minByMax`), so they only ever exercised the `boardAmt`-clip path and never the `targetCount`-driven one that was actually broken. Added a case with `targetCount=500` on a balance *below* `perRoundMax` (so the amount isn't even clipped): on the old code it mints 500 pieces; with the count cap it's bounded to 128, each still a valid `[floor, maxVTXO]` output with value fully conserved and no change. That case fails before the fix and passes after.

wallet/board_limits.go
ellemouton: Fixed the comment — it no longer claims automatic re-boarding. Reworded to: the clipped remainder returns as a change leave output paying a fresh boarding script that re-confirms as a new boarding intent, and `that next board is driven the same way any board is: automatically on the next confirmation under WithEagerRoundJoin, or by the operator's/host's usual board trigger otherwise`.

PR #1060, opened 2026-07-29
mailboxrpc: honor the operator's retry-after hint

Roasbeef:
In this PR, we teach the mailbox retry helper to wait as long as the operator
asks when it sheds a request, instead of falling back on a delay we picked
ourselves.

`Retry` already backs off on `ResourceExhausted` with jittered exponential
backoff, which is the right shape when we know nothing. But the operator does
know something: its rate limiter can say exactly when the next token arrives.
The lumos side now attaches that as a `google.rpc.RetryInfo` detail
(lightninglabs/lumos#726), and this reads it. A client that returns too early
gets shed again, and because the operator can only afford to answer roughly one
shed per second, that second shed is silent, so the client burns a full deadline
learning nothing. Honoring the hint is what turns an ongoing loop into a single
wasted round trip.

The hint rides on the status detail rather than an envelope header on purpose.
`DecodeErrorHeaders` already reconstructs the status, so `RetryAfter` reads it
off the error the retry loop is already holding. A bare header would never reach
here: the `call` closure returns only an `error`, and the transport drops the
rest of the headers, so we would have had to plumb them through the facade for
this one field.

## Treating the hint as untrusted

A retry-after is a number from someone else, so `backoffFor` does not take it at
face value. Three cases, each deliberately different rather than folded together:

An absurd but well-formed hint, say a hundred years, is clamped to
`RetryPolicy.MaxRetryAfter` (30s by default) instead of ignored. A hostile or
buggy operator gets to slow us down, which is its right, but not to park us
forever.

A malformed hint, meaning a duration outside the protobuf range or a detail we
cannot resolve, is treated as no hint at all and falls through to the computed
backoff. Garbage should not be honored, and it should not be clamped either,
because clamping garbage silently invents a number.

A non-positive hint is also treated as no hint. This one matters most: honoring
a zero would collapse the backoff into a hot loop aimed at a server that just
told us it is overloaded, which is worse than having no hint at all.

The clamped hint is used without jitter, which is a departure from the computed
path. Jitter exists to break up a fleet-wide schedule, and there is no fleet-wide
schedule here: the hint comes from the caller's own per-client bucket, so two
clients already get different values. Jittering downward would just return early
and defeat the point.

`TestRetryClampsHostileServerHint` holds the ceiling. Deleting the clamp makes it
hang past its timeout rather than fail an assertion, which is the honest
signature of what the bug would be.

## A rebase note

This began stacked on the idempotency-key work in #1056, now merged. Rebasing it
onto main hit one conflict: that branch deleted `DefaultRetryPolicy` as dead
code, and this branch had added `MaxRetryAfter` to it. We kept the deletion, and
dropped the one test assertion that called it. The property it checked, that the
default ceiling is applied, is already asserted two lines above through
`RetryPolicy{}.normalize()`, which is the documented zero-value contract and the
thing production actually goes through.

## Scope

This is the client half of a cheap interim, not a fix. The underlying problem is
that a shed response costs a durable mailbox write, which is why the operator
rate-limits how often it can afford to answer at all. Removing that constraint
needs a non-durable delivery class, tracked in #1054. Until then this makes the
answers we do get more useful, and it does nothing for the sheds that stay
silent.

`go test ./mailbox/... -race`, `make fmt-changed`, `make lint-changed-local`
(0 issues) and `make commitmsg-lint` are all clean.

### DIFF
--- a/mailbox/rpc/CLAUDE.md
+++ b/mailbox/rpc/CLAUDE.md
@@ -30,14 +30,19 @@ and server-side routing need without including any transport implementation.
   handing every attempt the same `RPCOptions.IdempotencyKey`. The key is
   minted once at the logical-call boundary (random, not a payload digest, so
   two identical concurrent reads stay distinct). Only `ResourceExhausted` is
-  retried, after a jittered exponential backoff; a cancelled or expired
+  retried, after the operator's retry-after hint when the shed response names
+  one and a jittered exponential backoff otherwise; a cancelled or expired
   context never re-sends.
-- `RetryPolicy` — `MaxAttempts` / `BaseDelay` / `MaxDelay`; the zero value
-  selects the package defaults (4 attempts, 200 ms base, 5 s cap), which
-  mirror `serverconn/mailboxpull`'s backoff shape.
+- `RetryPolicy` — `MaxAttempts` / `BaseDelay` / `MaxDelay` / `MaxRetryAfter`;
+  the zero value selects the package defaults (4 attempts, 200 ms base, 5 s
+  cap, 30 s retry-after ceiling). The first three mirror
+  `serverconn/mailboxpull`'s backoff shape.
 - `NewIdempotencyKey` — Mints one 16-byte hex key per logical request.
 - `IsShedError` — Reports whether an error (wrapped or not) is the operator's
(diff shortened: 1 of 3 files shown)


Roasbeef:
@claude review this

PR #1061, opened 2026-07-29
sdk/swaps: prevent duplicate in-swap vHTLC funding

bhandras:
## Problem

The pay-side swap FSM persists `FundingInitiated` before asking Waved to
fund the vHTLC. Waved deliberately continues an accepted OOR transfer after
the caller context ends, so an RPC deadline or process crash can lose the
response even though the transfer was accepted.

The SDK previously retried that request without an idempotency key after its
restart grace period. The retry could select different wallet inputs and fund
the same vHTLC a second time. Besides consuming extra liquidity, every extra
output must be recovered before the receiver can sweep it.

There is a second ambiguity at the funding cutoff. A response can be lost
before the SDK persists either the session ID or recipient outpoint. If the
SDK terminally expires that swap without first reconciling the accepted key,
the funded vHTLC is stranded and refund recovery is never armed. Conversely,
a normal keyed retry after expiry must not be allowed to admit a fresh,
already-expired transfer.

## Fix

This PR makes the funding intent and its recovery metadata durable across the
whole client/Waved boundary:

* derive one stable `in-swap-funding:<payment_hash>` idempotency key and use it
  for the initial submission, same-process retries, and restart recovery;
* recover the original recipient outpoint from Waved's durable OOR snapshot
  when a keyed replay finds an existing transfer;
* carry the swap's buffered funding cutoff into Waved as an admission
  deadline, enforced at the durable registry and child admission boundaries;
* add an `existing_only` reconciliation mode that never selects wallet inputs
  and never admits a new transfer;
* serialize a reconciliation miss through the registry actor and report an
  accepted-but-not-yet-committed admission as pending rather than absent;
* require both the session ID and exact recipient outpoint before persisting
  funding completion or arming refund recovery; and
* treat unavailable or incomplete reconciliation results as retryable.

Before the cutoff, the stable key may admit exactly one transfer. At or after
the cutoff, the SDK uses the same request in read-only mode:

* a committed winner returns its original outpoint and arms refund recovery;
* a pending admission remains retryable;
* an unavailable daemon remains retryable; and
* `NotFound` is authoritative because Waved will no longer admit that intent,
  so the swap can safely transition to `Expired`.

The resulting invariant is that one payment hash addresses at most one active
or completed OOR funding intent. A lost response can repeat the lookup across
the expiry boundary, but it cannot create another transfer or strand a known
winner without recovery.

## Commit structure

1. `waved: recover keyed OOR recipient outpoints`
   exposes durable recipient recovery and covers both the preflight and
   actor-race replay paths.
2. `sdk/swaps: key in-swap vHTLC funding`
   applies the stable key and fail-closed retry behavior to the pay FSM.
3. `waved: add deadline-bound OOR reconciliation`
   adds the admission deadline, read-only lookup, and pending handoff state.
4. `sdk/swaps: reconcile funding after expiry`
   routes the funding cutoff and hard-expiry timer through authoritative
   reconciliation.

## Testing

The accepted-response-lost regression now crosses the funding cutoff and a
restart ambiguity grace longer than the swap itself. It verifies that an
incomplete first reconciliation stays retryable, a later reconciliation
recovers the original outpoint, only one daemon-side intent was accepted, and
exactly one refund recovery job is armed.

The authoritative-miss regression resumes a `FundingInitiated` swap for which
no transfer was accepted. It verifies that the SDK makes one existing-only
lookup after expiry and persists `Expired` only after Waved returns
`NotFound`.

Additional validation:

* `go test ./oor ./sdk/ark ./sdk/swaps ./waved -count=1 -timeout=10m`
* race-enabled expiry reconciliation regressions in `sdk/swaps`
* race-enabled admission deadline and lookup tests in `oor`
* race-enabled existing-only and expired request tests in `waved`
* `make lint-changed-local base=origin/main`
* `make fmt-changed-check base=origin/main`
* `make commitmsg-lint range=origin/main..HEAD`
* `git diff --check origin/main...HEAD`

### DIFF
--- a/oor/actor_durable_message.go
+++ b/oor/actor_durable_message.go
@@ -30,6 +30,17 @@ const (
 	// startPayloadIdempotencyKeyType stores the optional caller-provided
 	// OOR send idempotency key.
 	startPayloadIdempotencyKeyType tlv.Type = 5
+
+	// Type 6 is deliberately unused. New backward-compatible records use
+	// odd TLV types so older readers can safely ignore them.
+
+	// startPayloadAdmissionDeadlineType stores the optional absolute
+	// admission deadline as Unix nanoseconds.
+	startPayloadAdmissionDeadlineType tlv.Type = 7
+)
+
+const (
+	lookupTransferIdempotencyKeyRecordType tlv.Type = 1
 )
 
 const (
@@ -139,11 +150,12 @@ const (
 )
 
 type startTransferPayload struct {
-	OperatorPubKey []byte
-	CSVDelay       uint32
-	Inputs         []*TransferInputSnapshot
-	Recipients     []recipientPayload
-	IdempotencyKey string
+	OperatorPubKey             []byte
+	CSVDelay                   uint32
+	Inputs                     []*TransferInputSnapshot
+	Recipients                 []recipientPayload
+	IdempotencyKey             string
+	AdmissionDeadlineUnixNanos int64
 }
 
 type recipientPayload struct {
@@ -173,6 +185,7 @@ func encodeStartTransferPayload(payload startTransferPayload) ([]byte, error) {
 	operatorKey := payload.OperatorPubKey
 	csvDelay := payload.CSVDelay
 	idempotencyKey := []byte(payload.IdempotencyKey)
+	admissionDeadline := uint64(payload.AdmissionDeadlineUnixNanos)
 
 	records := []tlv.Record{
 		tlv.MakePrimitiveRecord(
@@ -189,6 +202,9 @@ func encodeStartTransferPayload(payload startTransferPayload) ([]byte, error) {
 		),
 		tlv.MakePrimitiveRecord(
 			startPayloadIdempotencyKeyType, &idempotencyKey,
+		),
+		tlv.MakePrimitiveRecord(
+			startPayloadAdmissionDeadlineType, &admissionDeadline,
 		),
 	}
 
@@ -222,6 +238,7 @@ func decodeStartTransferPayloadWithLimits(raw []byte,
 		inputsRaw   []byte
 		recipients  []byte
 		idKey       []byte
+		deadline    uint64
 	)
 
 	records := []tlv.Record{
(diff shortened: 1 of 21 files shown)


bhandras:
@claude please re-review the updated four-commit series, with particular
attention to the expiry-boundary funds-safety invariant added in the last two
commits.

The new behavior is:

* Waved enforces the caller's funding cutoff at admission, after checking for
  an existing keyed winner but before accepting a fresh transfer.
* Post-cutoff SDK retries are read-only: winner metadata is recovered, pending
  or unavailable state remains retryable, and only an authoritative `NotFound`
  permits the swap to become `Expired`.
* The registry distinguishes an admission handed to a child but not yet
  committed, so reconciliation cannot mistake that handoff window for absence.
* Tests cover both accepted-response-lost recovery after expiry and a
  definitively absent transfer after expiry.

The branch is rebased onto current `main`; package tests, focused race tests,
changed-code lint, formatting, commit-message lint, and whitespace checks pass
locally.

bhandras:
@claude confirmed: there is no path that restores or instantiates a row-less
OOR child from durable mailbox contents. Boot restoration enumerates
non-terminal registry rows and calls `ensureChild`; the receptionist and
resident child maps are in-memory, and mailbox contents alone are not scanned
to recreate actors.

I folded the requested contract documentation into the Waved reconciliation
commit:

* `pendingOutgoingKeys` now states that it is a same-boot witness, the registry
  row is the durable witness, and a pre-first-commit crash leaves no resumable
  transfer;
* the skipped TLV type documents why type 6 cannot be used for this optional
  backward-compatible record; and
* recipient-outpoint recovery documents its positional empty-string contract.

The leading blank lines are retained because `gofmt` restores them before
comment-led blocks. Package tests, formatting, all 71 changed-code linters,
commit-message lint, and whitespace checks pass on the updated head. Please
give the final head a quick re-review.

sputn1ck (approved):
LGTM. The payment-scoped funding key, daemon-side admission deadline, existing-only reconciliation, and durable recipient-outpoint recovery close the lost-response safety gap for new attempts cleanly.

Non-blocking / optional follow-up: the upgrade boundary does not fully reconcile legacy `FundingInitiated` rows whose original OOR submission was unkeyed, or older rows with a session ID but no recipient outpoint. Also, a new SDK should ideally not rely on `existing_only` or admission-deadline semantics when connected to an older Waved that ignores those fields. Given the current small user population, I do not think this needs to block the PR: an atomic rollout (or Waved-first rollout) plus checking/draining known `FundingInitiated` swaps is reasonable. Before broader adoption, it would be worth adding a legacy-state migration/reconciliation path or daemon capability check, with an upgrade regression test.

PR #1066, opened 2026-07-29
txconfirm: release abandoned fee input on CPFP reselect

### LINKED ISSUE #664: txconfirm: CPFP fee-input reselect leaks the abandoned reservation/lease
## Summary

In `broadcastWithCPFP`, when the precise-vsize fee recompute forces a fee-input **reselect** to a different UTXO, the originally-reserved UTXO is left reserved and wallet-leased without being released. It stays locked for the parent's lifetime (or until the lease expiry) even though the CPFP child never spends it.

## Location

`txconfirm/broadcaster.go:735-747` (the reselect block inside `broadcastWithCPFP`).

## Detail

1. `selectFeeInput(totalFee+DustLimit)` picks UTXO **A** (`broadcaster.go:703`).
2. `reserveFeeInput(txid, A)` adds A to `state.UsedFeeOutpoints`/`UsedFeeInputs` and takes a wallet lease on A via `LeaseOutput` (`broadcaster.go:709`).
3. If the fee input's real script class is heavier than the change-script proxy, the package fee grows and the code reselects against the larger `totalFee+DustLimit` threshold, possibly landing on a different UTXO **B** (`broadcaster.go:735`).
4. If `B != A`, it sets `feeInput = B` and `reserveFeeInput(txid, B)` (`broadcaster.go:744-747`) — but never calls `releaseFeeOutpoint(txid, A)`.

A is now reserved + wallet-leased but unused (the child built at `broadcaster.go:754` only spends B).

## Impact

- Spurious UTXO **lock**, not fund loss: A is unavailable to other fee selections.
- Bounded, not permanent: A is released when the parent reaches a terminal state (eviction releases all the parent's reservations) or when `DefaultFeeInputLeaseExpiry` elapses.

## Fix (with caveat)

Release A before reserving B when they differ. **Caveat:** `selectFeeInput` at `:703` can legitimately return a *carried-over* reservation on a fee **bump** — TRUC package RBF requires the replacement child to double-spend the previous child's fee input (the "a parent may re-pick UTXOs from its own reserved set" invariant in `txconfirm/CLAUDE.md`). So the release must only apply when A was *not* part of a prior committed bump (e.g. on the initial submission, or when A is not in the carried-over set), otherwise it would drop the input RBF needs.

## Discovery

Found during the multi-agent review of the #509 checkpoint-broadcast-hardening change (darepo-client#663). Pre-existing; not introduced by that PR.

ellemouton:
Closes #664.

## Problem

In `broadcastWithCPFP`, when the precise per-input vsize recompute grows
the package fee enough to push the originally selected fee input's change
below the dust limit, the broadcaster reselects a larger fee input — but
it had already `reserveFeeInput`'d (and wallet-`LeaseOutput`'d) the
original before the reselect and never released it. The abandoned UTXO
stays reserved and wallet-leased for the parent's whole lifetime (or until
`DefaultFeeInputLeaseExpiry`), even though the CPFP child only ever spends
the reselected input. It's a spurious UTXO lock (not fund loss) that
starves other fee selections.

## Fix

Release the abandoned input's reservation + wallet lease
(`releaseFeeOutpoint`) when the reselect lands on a different outpoint.

**The caveat that makes this non-trivial:** `selectFeeInput` can
legitimately return a *carried-over* reservation on a fee **bump** — TRUC
package RBF requires the replacement child to double-spend the previous
child's fee input, so that input must stay reserved. The release therefore
skips a carried-over input: `feeOutpointReserved` records whether the input
was already reserved *before* the (idempotent) re-reservation, so a
freshly-selected-and-abandoned input is released while a carried-over one
is left intact.

## Testing

`TestCPFPReselectReleasesAbandonedFeeInput` drives the exact reselect path
(a P2PKH fee input heavier than the OP_TRUE change proxy grows the fee and
forces a reselect off the small UTXO) and asserts the abandoned input is
dropped from the in-memory reservation set **and** its wallet lease is
released, while the reselected input stays reserved. It fails on `main` and
passes with the fix. Full `txconfirm` suite + `lint-changed-local` green.

### DIFF
--- a/txconfirm/broadcaster.go
+++ b/txconfirm/broadcaster.go
@@ -479,6 +479,26 @@ func (b *CPFPBroadcaster) reserveFeeInput(ctx context.Context,
 	}
 }
 
+// feeOutpointReserved reports whether the parent txid already holds a
+// reservation on the given outpoint. It distinguishes a fee input carried over
+// from a prior committed bump (which TRUC package RBF requires the replacement
+// child to keep double-spending, so it must stay reserved) from one freshly
+// selected during the current submission (safe to release if abandoned). It
+// does not create a parent state entry, so a pre-reservation probe cannot leak
+// an empty shell.
+func (b *CPFPBroadcaster) feeOutpointReserved(parentTxid chainhash.Hash,
+	op wire.OutPoint) bool {
+
+	state, ok := b.parentStates[parentTxid]
+	if !ok || state.UsedFeeOutpoints == nil {
+		return false
+	}
+
+	_, held := state.UsedFeeOutpoints[op]
+
+	return held
+}
+
 // releaseWalletLease calls wallet.ReleaseOutput with the package-scoped
 // LockID, silently ignoring "unknown output" / "not leased" class
 // errors since those are a normal consequence of a wallet that is
@@ -889,6 +909,13 @@ func (b *CPFPBroadcaster) broadcastWithCPFP(ctx context.Context, height int32,
 			err)
 	}
 
+	// Record whether this input was carried over from a prior committed
+	// bump before we (re-)reserve it. If a later reselect abandons it, we
+	// only release inputs freshly selected this submission: a carried-over
+	// input must stay reserved because TRUC package RBF requires the
+	// replacement child to double-spend the previous child's fee input.
+	feeInputCarriedOver := b.feeOutpointReserved(txid, feeInput.Outpoint)
+
 	b.reserveFeeInput(ctx, txid, feeInput)
 
 	// If the chosen fee-input's actual script class differs from the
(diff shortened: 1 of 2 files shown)


ellemouton:

It closes #664: `broadcastWithCPFP` reserved + wallet-leased a fee input, then on a precise-vsize reselect switched to a larger input without ever releasing the original — leaking the reservation/lease for the parent's lifetime.

The interesting part is the **carried-over caveat**, so please scrutinise:

- **Correctness of `feeOutpointReserved` as the carried-over signal.** I capture it *before* `reserveFeeInput` (which is idempotent and re-adds the outpoint), so on a fee bump a carried-over input reads `true` and is NOT released, while a freshly-selected input reads `false` and IS released on reselect. Is capturing before the re-reservation the right discriminator, and are there paths where an input is carried-over yet not yet in `UsedFeeOutpoints` at that point?

- **TRUC RBF safety.** The whole reason to skip carried-over inputs is that package RBF requires the replacement child to double-spend the previous child's fee input. Does the skip fully preserve that? And is there any scenario where a *fresh* input we now release was actually needed?

- **No double-release / state-shell leak.** `releaseFeeOutpoint` also prunes an empty parent-state entry — on reselect we release the old then immediately `reserveFeeInput` the new; any risk of dropping then mis-recreating state, or of releasing an outpoint that a later fallback path releases again?

Repro test `TestCPFPReselectReleasesAbandonedFeeInput` fails on main, passes here.

Roasbeef (approved):
LGTM 🫛

PR #1068, opened 2026-07-30
unroll: cancel LND broadcast of a proof tx on terminal failure

### LINKED ISSUE #609: unroll: LND keeps rebroadcasting a known-rejecting proof tx every ~60s after unroll job is marked Failed
## Environment

- `darepo-client`: `dcbacc0a` (also reproducible on current main `32b33fb5`)
- Network: signet
- Wallet backend: `lnd`
- Ark server: `arkd-signet.testnet.lightningcluster.com:443`

## Summary

After a unilateral-exit unroll job transitions to `EXIT_JOB_STATUS_FAILED`, **LND keeps rebroadcasting the proof tx every ~60 s indefinitely**. The bitcoind / mempool rejects it with the same `min relay fee not met` error each retry. darepod has stopped monitoring the job, but the underlying broadcast persists in LND's queue.

## Expected Behavior

When darepod marks an unroll job `FAILED` for any reason, it should cancel the LND-side broadcast for that proof tx. Either:

1. Issue an explicit cancel via the LND wallet RPC.
2. On subsequent darepod startups, reconcile FAILED unroll jobs against LND's pending broadcasts and cancel any that survive.

## Actual Behavior

Following the cascade documented in #602 / #608 (sub-dust VTXO accepted for exit → proof tx rejected at min relay fee → job FAILED), the LND log shows the broadcast retry loop continuing well after darepod has stopped touching the job:

```
[ERR] LNDC: PublishTransaction failed
   txid=ee770706083ad4cf50c469242eeeb44a810f639c8a7a675bcbee986dbe6ea35a
   error="min relay fee not met"
[INF] LNDC: ... will retry in 60s ...
```

Observed at the same minute-aligned cadence over ~30 minutes of observation. Same txid, same error, every retry. No visible darepod-side activity during the loop.

Impact:
- Noise in LND logs.
- Slow drain on LND's broadcast queue / bandwidth.
- On a signet node, polite — but a production daemon with N stuck unroll jobs would generate N concurrent retry loops indefinitely.

## Reproduction

Run the cascade from #608 (initiate exit on a sub-dust VTXO, watch FAILED). Then tail the LND container's log:

```sh
docker logs -f darepod-signet-lnd 2>&1 | grep -i "PublishTransaction failed"
```

Confirm the same txid keeps appearing at ~60 s intervals after the unroll job is FAILED.

## Notes

Discovered during a swapdk testing sprint Batch C unilateral-exit pass. Companion to #602 (FAILED unroll removes VTXO from local view) and #608 (sub-dust acceptance kicks off the cascade in the first place).

ellemouton:
Closes #609.

> Depends on lightninglabs/lndclient#284 (`WalletKit.RemoveTransaction`), now **merged** — this PR bumps `lndclient` to the merged master commit (no `replace`).

## Problem

After a unilateral-exit unroll job goes `FAILED`, **LND keeps rebroadcasting the rejecting proof tx from its own wallet queue every ~60s indefinitely** — the wavelength daemon has stopped touching the job, but the transaction (a zero-fee anchor parent that can never relay alone) stays in LND's rebroadcast set. Nothing tells LND to drop it.

## Fix

Give the unroll actor a way to abandon its broadcast transactions on terminal failure:

1. **`go.mod`** — bump `lndclient` to the master commit adding `WalletKit.RemoveTransaction`.
2. **`chainbackends`** — `TxBroadcaster` gains `RemoveTransaction`; `LNDBackend` exposes it as the optional `chainsource.TxRemover` capability, wrapping lnd's `WalletKit.RemoveTransaction` RPC.
3. **`chainsource`** — a `TxRemover` optional-capability interface + a `RemoveTxRequest` routed through the chain-source actor. The handler forwards to a capable backend and treats a backend **without** the capability (a pure Esplora chain source) or an already-absent tx as a **no-op success** — so callers request removal uniformly without knowing the backend, and no `ChainBackend` implementer/fake had to change.
4. **`unroll`** — on terminal `PhaseFailed`, the actor removes its in-flight (broadcast-but-unconfirmed) proof/sweep txids via `RemoveTxRequest`. Confirmed txs are left alone (on-chain, never rebroadcast). The removal runs once, is best-effort, and detaches cancellation like the terminal registry handoff.

**Backend scope:** the LND backend is fully covered. `lwwallet`/`btcwbackend` sit on btcwallet, which also rebroadcasts — but since they don't implement `TxRemover`, they no-op today; a btcwallet removal path is a scoped follow-up.

## Testing

- `chainsource`: `TestChainSourceActorRemoveTx` — forwards to a capable backend, treats an ignorable "not found" error as success, and no-ops on a backend without the capability.
- `unroll`: `TestTerminalFailureRemovesAbandonedBroadcasts` — a terminal failure removes the in-flight proof tx.

`chainsource` + `chainbackends` + `unroll` suites, `go build`, `gofmt`, and `lint-changed-local` are all green.

### DIFF
--- a/chainbackends/lnd.go
+++ b/chainbackends/lnd.go
@@ -61,6 +61,13 @@ type TxBroadcaster interface {
 	// The label parameter is optional and may be used for wallet tracking.
 	PublishTransaction(ctx context.Context, tx *wire.MsgTx,
 		label string) error
+
+	// RemoveTransaction removes the transaction with the given hash, and
+	// any transactions that depend on it, from the wallet's store and its
+	// unconfirmed-rebroadcast set. It is used to abandon a transaction the
+	// caller no longer wants the wallet to keep rebroadcasting (e.g. a
+	// unilateral-exit proof tx whose job has terminally failed).
+	RemoveTransaction(ctx context.Context, txid chainhash.Hash) error
 }
 
 // PackageSubmitter atomically submits parent+child transaction packages.
@@ -236,6 +243,23 @@ func (b *LNDBackend) BroadcastTx(ctx context.Context, tx *wire.MsgTx,
 	b.logger(ctx).InfoS(ctx, "Transaction broadcast successfully",
 		btclog.Hex("txid", txHash[:]),
 	)
+
+	return nil
+}
+
+// RemoveTx removes a previously broadcast transaction from lnd's wallet so it
+// stops being rebroadcast. It satisfies the optional chainsource.TxRemover
+// capability; a caller that has abandoned a transaction (e.g. a terminally
+// failed unilateral-exit proof tx) uses it to stop lnd from perpetually
+// re-publishing a transaction that can never confirm.
+func (b *LNDBackend) RemoveTx(ctx context.Context, txid chainhash.Hash) error {
+	b.logger(ctx).InfoS(ctx, "Removing transaction from LND wallet",
+		btclog.Hex("txid", txid[:]),
+	)
+
+	if err := b.broadcaster.RemoveTransaction(ctx, txid); err != nil {
+		return fmt.Errorf("failed to remove transaction: %w", err)
+	}
 
 	return nil
 }
--- a/chainbackends/lnd_test.go
+++ b/chainbackends/lnd_test.go
@@ -93,6 +93,13 @@ func (s *stubBroadcaster) PublishTransaction(
 	return nil
 }
 
+func (s *stubBroadcaster) RemoveTransaction(
+	_ context.Context, _ chainhash.Hash,
+) error {
+
+	return nil
+}
+
 type stubPackageSubmitter struct {
 	result *btcjson.SubmitPackageResult
 	err    error
--- a/chainbackends/lndclient_adapters.go
+++ b/chainbackends/lndclient_adapters.go
@@ -49,6 +49,15 @@ func (b *LndClientTxBroadcaster) PublishTransaction(ctx context.Context,
 	return b.walletKit.PublishTransaction(ctx, tx, label)
 }
 
+// RemoveTransaction removes the transaction (and its descendants) from lnd's
+// wallet via the WalletKit RemoveTransaction RPC, stopping lnd from
+// rebroadcasting it.
+func (b *LndClientTxBroadcaster) RemoveTransaction(ctx context.Context,
+	txid chainhash.Hash) error {
+
+	return b.walletKit.RemoveTransaction(ctx, txid)
+}
+
 // Ensure LndClientTxBroadcaster implements TxBroadcaster.
 var _ TxBroadcaster = (*LndClientTxBroadcaster)(nil)
 
--- a/chainsource/backend.go
+++ b/chainsource/backend.go
@@ -134,6 +134,21 @@ type ChainBackend interface {
 	Stop() error
 }
 
+// TxRemover is an optional capability a ChainBackend may implement to remove a
+// previously broadcast transaction from the wallet, so the wallet stops
+// rebroadcasting it. A caller that has abandoned a transaction (e.g. a
+// unilateral-exit proof tx whose job terminally failed) uses it to stop a
+// full-node wallet from perpetually re-publishing a transaction that can never
+// confirm (wavelength#609). Backends without a rebroadcasting wallet (e.g. a
+// pure Esplora chain source) do not implement it; the chainsource actor treats
+// a RemoveTxRequest as a no-op for those.
+type TxRemover interface {
+	// RemoveTx removes the transaction with the given hash, and any
+	// transactions that depend on it, from the wallet's store and its
+	// unconfirmed-rebroadcast set.
+	RemoveTx(ctx context.Context, txid chainhash.Hash) error
+}
+
 // ConfRegistration encapsulates the channels and control functions for a
 // confirmation registration. This mirrors lnd's chainntnfs.ConfirmationEvent
 // structure but provides a backend-agnostic interface.
--- a/chainsource/broadcast_errors.go
+++ b/chainsource/broadcast_errors.go
@@ -37,7 +37,38 @@ var (
 		// Wallet-layer variants.
 		"output already spent",
 	}
+
+	// ignorableRemoveErrs lists error substrings returned when removing a
+	// transaction that the wallet does not (or no longer) knows about. The
+	// caller's goal -- the transaction is not in the wallet's rebroadcast
+	// set -- is already satisfied in that case, so the removal is treated
+	// as a success (wavelength#609).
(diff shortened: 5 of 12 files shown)


ellemouton:

It closes #609: after an unroll job fails terminally, LND keeps rebroadcasting the doomed proof tx from its own wallet queue forever. The fix threads a wallet `RemoveTransaction` down through a new optional `chainsource.TxRemover` capability and has the unroll actor drop its in-flight broadcasts on terminal `PhaseFailed`. (Depends on the now-merged lndclient#284.)

Please scrutinise:

- **Which txids get removed.** `removeAbandonedBroadcasts` removes `job.PlannerState.InFlightTxids` plus the sweep txid (only when it advanced past pending), and deliberately NOT `ConfirmedTxids`. Is "in-flight only" the right set — any risk of removing a tx that actually confirmed, or missing one LND still holds? The once-flag (`abandonedBroadcastsRemoved`) sits before the registry-handoff guard so it fires independent of the registry — correct placement?

- **The optional-capability design.** `TxRemover` is type-asserted in `handleRemoveTx`; a backend without it (Esplora) and an already-absent tx (`IsIgnorableRemoveError`: "not found"/"no such"/etc.) both resolve to a no-op success. Is that no-op-vs-error boundary right, and is the substring-based ignorable classification too broad/narrow?

- **Best-effort + lifecycle.** The removal Asks chainsource with a `context.WithoutCancel` detached context (mirroring the terminal registry handoff) and only logs on failure. Any concern doing a cross-actor Ask on the terminal path, or with the once-flag not retrying a transient chainsource failure?

- **Backend scope gap.** Only the LND backend implements `TxRemover`; `lwwallet`/`btcwbackend` (btcwallet, which also rebroadcasts) no-op today. I called this out as a scoped follow-up — agree that's acceptable for closing #609, or should the btcwallet path be in-scope?

Tests: `TestChainSourceActorRemoveTx` (handler: forward / ignorable / no-op) and `TestTerminalFailureRemovesAbandonedBroadcasts` (end-to-end).

ellemouton:
Thanks for the review — addressed all four points and force-pushed (the two fix sets are folded into their respective commits, so the history stays a clean 4-commit story).

**1. Blocking/un-deadlined `Ask` on the terminal path (the pre-merge concern).**
`removeAbandonedBroadcasts` no longer runs the `ChainSource.Ask` loop synchronously on the actor's `Receive` goroutine. The txid set is still gathered synchronously (cheap), but the removals now fire on a detached goroutine — mirroring txconfirm's async wallet-lease release — so a slow or hung lnd `RemoveTransaction` RPC can never sit in front of the terminal registry handoff or stall later messages for the target. Each op carries its own `context.WithTimeout(context.WithoutCancel(ctx), removeAbandonedBroadcastTimeout)` (new named const, 30s, documented) so a stuck backend bounds the goroutine's lifetime rather than leaking it. Cancellation stays detached so a caller disconnect can't suppress the cleanup.

**2. `handleRemoveTx` ignorable-branch formatting wart.** Collapsed to the clean `a.logger(ctx).DebugS(ctx, "RemoveTx returned ignorable error", …)` form and removed the split string literal.

**3. `"no such"` → `"no such transaction"`** in `ignorableRemoveErrs`, so the substring match is specific to the wallet's "unknown tx" phrasing rather than the broadest possible token.

**4. Confirmed txids never removed.** Added `TestAbandonedBroadcastRemovalSkipsConfirmedTxids`, which drives a job state carrying both an in-flight and a confirmed txid through the cleanup and asserts the in-flight one is removed while the confirmed one never is.

On the older-lnd note: pre-0.18 nodes return `Unimplemented` for `WalletKit.RemoveTransaction`. That is *not* in `ignorableRemoveErrs`, so it surfaces as a logged warning on the best-effort path (it doesn't affect the terminal outcome) rather than being silently swallowed — which is the behavior we want until the min-lnd floor moves. The btcwallet-backed removal path (`lwwallet`/`btcwbackend`) remains a scoped follow-up.

ellemouton:
@claude please review this PR

Roasbeef:
Pushed a commit on top of this. Digging through the terminal path, the cleanup doesn't actually fire on the case #609 describes.

`applyFailedEvent` strips the failing txid from `InFlightTxids` right before it stamps `FailReason`:

```go
job.PlannerState.InFlightTxids = removeHash(
        job.PlannerState.InFlightTxids, event.Txid,
)
...
job.FailReason = event.Reason
```

So by the time `removeAbandonedBroadcasts` reads planner state, the tx that killed the job is no longer in the list. And that's precisely the tx lnd is sitting on and rebroadcasting. Confirmed it by driving the real path (`setImmediateFailed` on the root proof tx, same shape as `TestProofTxFailureTransitionsToFailed`): job reaches `PhaseFailed`, removal set comes back empty.

Both tests here route around it. `TestTerminalFailureRemovesAbandonedBroadcasts` goes through the external-spend path, and the comment in it even calls out the choice ("A target spend, unlike a root-output spend, does not confirm the still-in-flight root"). The other one calls `removeAbandonedBroadcasts` directly with a hand-built `JobState`. So the suite is green over a different failure than the one we're fixing.

The fix tracks failed txids on the behavior and unions them into the removal set. In-memory only, since the registry only restores non-terminal records, so a terminal job never re-runs this after a restart. Keeps the checkpoint codec untouched. The new test drives the hard-failure path, and fails without the fix. Two small things while in there: the sweep txid was appended without being added to `seen`, harmless while it was the last entry but not once something follows it, and the final list is now sorted since map iteration order isn't stable.

Worth writing down somewhere in the code: the proof-tx hard-failure path is the _safest_ removal we have. An operator-signed proof node can't be rebuilt or replaced, so the failure is terminal by construction, and txconfirm has already evicted the entry (`TxStateFailed` is terminal) and released the fee-input lease via `broadcaster.Evict`. Nothing else in the system is still trying to land it. Which is what makes it funny that it was the one path not covered.

Two things I left alone that are worth a look before this goes in.

The external-spend path is where removal is actually racy. Nothing failed there, so txconfirm is still tracking those txs, and `CancelInterestReq` has no production caller anywhere (only txconfirm's own tests use it). For a tx still in `TxStateBroadcasting`, `retryBroadcastingParents` will re-publish it and put it right back in lnd's wallet. And since `RemoveTransaction` takes descendants with it, the CPFP child leaves the wallet and frees its fee input while txconfirm still holds the lease on it. If we want that path solid, we should cancel first, then remove.

Second, the comment on the detached goroutine claims the 30s timeout bounds a hung backend. Not quite: `ChainSourceActor.Receive` dispatches `handleRemoveTx` on `actorCtx` rather than the caller's ctx, so the deadline frees the unroll goroutine while the shared chainsource Receive stays blocked on the RPC until it returns on its own. Same is already true of `handleBroadcastTx`, so it's not new here, but the comment promises more than we get.

PR #1071, opened 2026-07-30
harness: add watch-only LND test nodes

bhandras:
## Problem

The Docker integration harness initializes every LND with `--noseedbackup`. That is convenient for ordinary test nodes, but it prevents integration tests from exercising the production watch-only plus remote-signer wallet shape. Tests could only mock the watch-only account-family boundary.

## Change

- Allow an LND container to remain at the wallet unlocker for explicit initialization.
- Add `StartAdditionalWatchOnlyLND`, which starts a separate signing LND, exports its account xpubs, initializes the returned node as watch-only, and connects it to the signer over the harness Docker network.
- Copy only the signer TLS certificate and admin macaroon into the watch-only node instead of mounting the signer data directory.
- Add a Docker-backed harness test that verifies both nodes share an identity, family `100` derives successfully, and an unimported family such as `9999` fails with LND’s real `watching-only` error.

Existing harness users retain the automatic `--noseedbackup` behavior.

## Validation

- `go test -tags=itest ./harness -run '^TestHarnessWatchOnlyLND$' -v -count=1 -timeout=10m`
- `go test ./harness -run '^$' -tags=itest`
- `make lint-changed-local`
- `make tidy-module-check`
- `git diff --check`

### DIFF
--- a/harness/harness.go
+++ b/harness/harness.go
@@ -38,6 +38,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	"github.com/lightninglabs/wavelength/chain"
 	lnrpc "github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnrpc/walletrpc"
 	"github.com/lightningnetwork/lnd/macaroons"
 	"github.com/ory/dockertest/v3"
 	"github.com/ory/dockertest/v3/docker"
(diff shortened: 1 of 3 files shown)


bhandras:
@claude review

bhandras:
@claude Please review this test-harness change, with particular attention to:

- whether the watch-only LND is initialized from the signer’s complete account set;
- whether the remote-signer TLS and macaroon handling is appropriately scoped for a test harness;
- whether existing `--noseedbackup` harness behavior remains unchanged;
- cleanup, naming, concurrency, and integration-test reliability.

Local validation includes the real Docker-backed `TestHarnessWatchOnlyLND`, which confirms shared signer/watch-only identity, successful family-100 derivation, and the expected `watching-only` failure for family 9999.

bhandras:
@claude Addressed the startup-race finding and folded it into the original harness commit.

`initWatchOnlyLND` now polls the LND state service until the fresh wallet reaches `NON_EXISTING` before invoking `InitWallet`. Each probe uses its own bounded context, so the 90-second startup wait cannot outlive and repeatedly reuse an expired RPC context.

Validation on the amended head `ea452c52`:

- Docker-backed `TestHarnessWatchOnlyLND` passes;
- changed-file lint passes with 0 issues;
- module-tidiness and commit-message checks pass;
- downstream hidden-primary swap integration test passes against this exact harness commit.

Please re-review the amended head and confirm whether any actionable issue remains.

PR #1076, opened 2026-07-31
lwwallet: resolve the wallet start block from the Esplora index

### LINKED ISSUE #1073: lwwallet: avoid sequential block scan with esplora client
I noticed this while testing out on signet. If I don't specify a birthday block and the wallet falls back to block zero, the `lwwallet` ends up fetching each block in the whole chain.

I understand this is because `EsploraChainService` is just an implementation of the `btcwallet/chain.Interface`, which only knows sequential scanning.
But since we're already talking to a fully-indexing chain service (Esplora), this could be bypassed for scanning wallet outputs completely, as the index should be able to provide that.

Roasbeef:
In this PR, we stop `lwwallet` from walking the entire chain on startup, and
instead ask the Esplora index directly where the wallet first shows up. Fixes
#1073.

The reported symptom was a wallet created without a birthday fetching every
block from genesis to the tip on signet. The mechanism is `btcwallet`'s
`locateBirthdayBlock`, which binary searches block timestamps for the seed
birthday within a +/-2h window. That search has no lower bound below genesis,
so a birthday older than the chain just walks off the bottom of the range and
returns block 0. It then persists that as _verified_, sets synced-to to 0, and
`recovery()` walks every height, followed by `Rescan()` walking them all over
again.

Measured on a 506 block regtest chain before the fix:

```
total 1636   block-height 619   block-header 507   block-raw 506
```

Roughly 3.2 Esplora requests per block of chain history. That extrapolates to
~800k requests on signet and ~3M on mainnet.

## Why a better birthday isn't enough

Three separate inputs land on block 0, and only one of them is a plumbing bug:

A `Config` with no `Birthday` at all, whose zero `time.Time` is January of year
1 (this is what the issue reporter hit, and it's why `waved`'s own
`wavecli create` path is fine: `WalletSeedFromMnemonic` passes a real
`cipherSeed.BirthdayTime()`). The SDK passkey path, which pins
`aezeed.BitcoinGenesisDate` into the mnemonic itself in `entropyToMnemonic`. And
an honest restore of a seed genuinely older than the chain.

That last one is the reason plumbing alone doesn't close this. A 2023 seed
restored onto testnet4 predates that chain's genesis entirely, and it's a
perfectly legitimate input. Even a correct birthday only _shortens_ the walk
rather than removing it, since the cost is O(chain-since-birthday) either way. A
seed from a year ago on mainnet is still ~52k blocks.

Worth noting this was never signet specific. `aezeed.BitcoinGenesisDate` is
`time.Unix(1231006505, 0)`, which is exactly mainnet's genesis timestamp, so
mainnet lands on block 0 too, just via the other branch of the search.

## Asking the index instead

We're already talking to a fully indexing chain source, so the "where did this
wallet first appear" question has a direct answer that costs nothing in chain
length.

Discovery derives each active key scope's receive and change scripts through
`btcwallet`'s own scoped key managers, which matters because that's what gets
each script the address type its scope's schema dictates (P2TR for BIP86, P2WPKH
for BIP84, nested P2SH for BIP49) rather than us reimplementing that mapping. It
probes `/scripthash/:h` for use under the configured gap limit, then takes the
earliest confirmed height across everything it found.

The result gets written into `waddrmgr` as a _verified_ birthday block before
`btcwallet` starts. That's the lever: `birthdaySanityCheck` returns a verified
block untouched and skips the timestamp search entirely, so `recovery()` and
`Rescan()` both begin where we put them.

A couple of details that are load-bearing. The stamp sits one block _below_ the
earliest appearance, because recovery resumes at synced-to plus one and would
otherwise skip the very block we're aiming at. And the write order in `waddrmgr`
matters: `PutSyncedTo` refuses a height whose predecessor's hash is absent, but
only once a birthday block exists, so we record synced-to first while
`FetchBirthdayBlock` still errors.

Receive branches are probed before change branches, and we skip the change
branches entirely when no receive address anywhere has been used. A change
output only exists if the wallet authored a spend, which requires it to have
held an output, which requires something to have paid a receive address first.
That halves the probe count in the case that matters most, a brand new wallet.

If discovery fails (unreachable index, a script with pathological history), it's
not fatal. The wallet still starts and `btcwallet` falls back to its own search,
just slowly.

Same 506 block chain after:

```
total 923   block-height 100   block-header 6   block-raw 6   scripthash 800
```

Block traffic goes 1638 -> ~12, and it's now flat in chain length. The 100
`block-height` calls are the `TipPoller`'s pre-existing reorg history seed, not
the scan. The 800 scripthash probes are the gap limit walk over both branches of
all four scopes, also constant.

## The script hash byte order bug

Building the systests turned up a separate pre-existing bug that was blocking
this, so it's the first commit.

`scriptHashHex` reversed the SHA256 digest before hex encoding, on the grounds
that this matched Electrum. It does match the Electrum _wire protocol_, whose
`blockchain.scripthash.*` methods reverse the digest. Esplora's REST API does
not: `/scripthash/:hash` wants it forward.

The failure mode is silent, which is what makes it nasty. Esplora answers an
unknown script hash with an empty result rather than an error, so every lookup
just came back with nothing found. That means `GetScriptUtxos` has been
returning no UTXOs the whole time, and with it the two chain backend paths built
on it: `checkConfByScript` and the script branch of confirmation hash lookup.

The existing test couldn't catch it because it built the expected request path
by calling `scriptHashHex`, so it agreed with the helper under any convention.
It now computes the expectation independently, and there's a fixed vector test
plus a systest that pins the encoding against a real electrs.

I found this the hard way: the first cut of the restore systest failed, and a
baseline run with the new code disabled showed restore working fine and scanning
all 508 heights. So the fix was losing funds, and the reversed digest was why
discovery saw every address as unused.

## The neutrino sibling

The advisor-review pass turned up the same shape one backend over, so there's a
fourth commit for it. `waved`'s env-seed startup passes a zero `time.Time` into
wallet creation, and a raw hex seed from `WAVED_LWWALLET_SEED` carries no
birthday to recover (unlike an aezeed mnemonic). For `lwwallet` that no longer
bites, since it resolves its own start height from the index. The
neutrino-backed `btcwallet` sibling has no index to consult, so it still pays
the full scan; compact filters make that slow rather than catastrophic.

I deliberately did **not** guess `time.Now()` there. A birthday that's too early
only costs scan time; one that's too late silently hides funds received before
it. Guessing would have been the second kind, and would have broken any CI flow
that funds a seed then restarts against a fresh data dir. Instead the operator
can state one via `WAVED_LWWALLET_SEED_BIRTHDAY` (calendar date or RFC3339), and
unset keeps today's genesis-anchored behaviour. A malformed date is an error
rather than a silent fall back to the same zero time, so a typo can't degrade
quietly into a full chain scan.

The three remaining zero-birthday call sites are open-only paths where a nil
seed means btcwallet never reaches the branch that reads the birthday. They're
commented as such so the value reads as deliberate.

## Testing

The three systests run against the Docker harness (bitcoind + electrs) and count
Esplora requests through a reverse proxy, so "does the wallet walk the whole
chain" is an assertion rather than an impression.

`TestLwWalletRestoreStartsAtFirstUse` is the one I care most about, since
resolving a start height from the index is only correct if it never lands
_above_ the wallet's own funds. It funds a wallet at a known height, throws the
database away, grows the chain, and restores from the same seed. The restored
wallet has to recover the full balance, and its scan has to begin at the funding
block: stamps 406 for a funding at 407, lowest height fetched 406.

Full systest suite is green (19 tests), as are all 63 unit test packages.

changes.

### DIFF
--- a/docs/daemon_cli_guide.md
+++ b/docs/daemon_cli_guide.md
@@ -161,10 +161,22 @@ prefix and dots replaced by underscores:
 |----------|-------------|
 | `WAVED_WALLET_PASSWORD` | Wallet password for create/unlock |
 | `WAVED_LWWALLET_SEED` | Hex-encoded raw seed (dev/CI only) |
+| `WAVED_LWWALLET_SEED_BIRTHDAY` | Creation date of that seed, `YYYY-MM-DD` or RFC3339 |
 | `WAVED_NETWORK` | Bitcoin network override |
 | `WAVED_WALLET_TYPE` | Wallet backend type |
 | `WAVED_WALLET_ESPLORAURL` | Esplora URL |
 | `WAVED_SERVER_HOST` | Operator server address |
+
+A raw seed carries no birthday, unlike an aezeed mnemonic, so a wallet
+created from `WAVED_LWWALLET_SEED` without `WAVED_LWWALLET_SEED_BIRTHDAY`
+starts scanning at genesis. That is slow but never misses funds. Set the
+birthday only when the seed is known to have no history before that date:
+too early only costs scan time, while too late silently hides earlier
+funds. A malformed date is rejected rather than treated as unset.
+
+The `lwwallet` backend usually resolves its own start height from the
+Esplora address index and ignores this, so it matters most for the
+neutrino-backed `btcwallet` backend, which has no index to consult.
 
 ## Initial Wallet Setup
 
--- a/lwwallet/CLAUDE.md
+++ b/lwwallet/CLAUDE.md
@@ -45,6 +45,16 @@ base logic with the neutrino-backed `btcwbackend` sibling via the extracted
 - `EsploraChainService` — `chain.Interface` adapter over `EsploraClient`,
   driven by a shared `TipPoller`. Feeds btcwallet's internal address-credit
   pipeline. Constructor: `NewEsploraChainService(esplora, tipPoller, logger)`.
+- Start-block discovery (`start_block.go`) — resolves where a new or
+  restored wallet begins scanning from the Esplora address index instead of
+  from the seed birthday, and writes it into `waddrmgr` as a *verified*
+  birthday block before `BtcWallet.Start()`. Entry point:
+  `Wallet.stampStartBlock`. Left to itself btcwallet binary searches block
+  timestamps (`locateBirthdayBlock`), which bottoms out at genesis for any
+  birthday older than the chain and then walks every block to the tip.
+  Discovery instead derives each active key scope's receive/change scripts
+  through btcwallet's own scoped key managers, probes `/scripthash/:h` under
+  the configured gap limit, and takes the earliest confirmed height.
 - `BoardingBackendAdapter` — Embeds `walletcore.BoardingBackendBase` for
   shared key derivation/script import; implements `wallet.BoardingBackend`
   and `wallet.OutputLeaser`. Queries Esplora directly for UTXOs (bypasses
@@ -79,6 +89,24 @@ base logic with the neutrino-backed `btcwbackend` sibling via the extracted
   (known limitation; acceptable for confirmation-target use cases).
 - LRU caches only hold immutable, hash-addressed data; a verified hash prevents
   a compromised Esplora endpoint from injecting arbitrary cache entries.
+- `scriptHashHex` hex-encodes the SHA256 digest in its natural byte order.
+  Esplora's REST API differs from the Electrum wire protocol here, which
+  reverses it, and a wrong order fails silently because the API answers an
+  unknown script hash with an empty result rather than an error.
+- Start-block discovery only ever stamps a wallet that has no birthday block
+  at all, and never moves the sync cursor backwards. Once btcwallet has
+  recorded a birthday block it owns the cursor.
+- A discovery failure is non-fatal: the wallet still starts and btcwallet
+  falls back to its own timestamp search, just slowly. Discovery refuses to
+  stamp rather than stamping on partial information, because a short walk
+  finds fewer used scripts, which raises the height it reports and moves the
+  stamp toward the tip: the direction that loses funds.
+- Start-block discovery trusts the Esplora index. That holds for a single
+  self-consistent instance, but a load-balanced public endpoint can serve a
+  current tip and a still-syncing script index from different backends, and
+  partial history is indistinguishable from an unused script. A restore in
+  that window stamps above its own funds permanently, since the stamp is
+  written verified. Point `EsploraURL` at an endpoint you control.
 - UTXO enumeration queries Esplora directly rather than btcwallet's internal
   UTXO set, because btcwallet does not credit-mark non-default scope outputs.
 - `Stop()` explicitly closes btcwallet's internal database to prevent resource
--- a/lwwallet/esplora.go
+++ b/lwwallet/esplora.go
@@ -177,16 +177,18 @@ type esploraBlock struct {
 	Timestamp int64 `json:"timestamp"`
 }
 
-// scriptHashHex computes the Esplora script hash for a pkScript. The hash
-// is SHA256(pkScript) with the bytes reversed and hex-encoded, matching
-// Electrum's script hash format used by Esplora.
+// scriptHashHex computes the Esplora script hash for a pkScript: the
+// hex-encoded SHA256 of the script, in the digest's natural byte order.
+//
+// This is deliberately NOT the Electrum convention. The Electrum wire
+// protocol (blockchain.scripthash.*) takes the digest byte-reversed, and
+// Esplora's REST API does not — /scripthash/:hash wants it forward.
+// Reversing here makes every lookup miss silently, since the API answers
+// an unknown script hash with an empty result rather than a 404. See
+// systest.TestEsploraScriptHashEncoding, which pins the byte order
+// against a real electrs instance rather than against this function.
 func scriptHashHex(pkScript []byte) string {
 	h := sha256.Sum256(pkScript)
-
-	// Reverse the hash bytes (Electrum convention).
-	for i, j := 0, len(h)-1; i < j; i, j = i+1, j-1 {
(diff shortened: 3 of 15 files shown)


Roasbeef:
@claude review this

Roasbeef:
Thanks, both reviews caught real things. Addressed all of it:

**Codex P2, `maxDiscoveryIndex` doesn't bound the batch (`start_block.go`)** — correct, and it was the more serious of the two. `usedBranchScripts` passed `gapLimit` straight through as the batch size, so with a large `RecoveryWindow` (an unbounded `uint32` from config) the very first batch would derive and probe the whole thing before `index < maxDiscoveryIndex` ever got consulted again. `make([][]byte, 0, count)` with a multi-billion count is an OOM, not a slow startup. Fixed in two places: the batch is now clamped to the remaining index budget (`min(gapLimit, maxDiscoveryIndex-index)`), and `discoveryGapLimit` clamps the configured window to the budget as well, since a window wider than the budget can't be honoured anyway.

**Claude #1, the scripthash systest asserted nothing** — also correct, and a bit embarrassing since I wrote that test specifically to pin the byte order. `fwd == byAddr || revd == byAddr` is vacuous: both encodings are computed locally, so one of them necessarily matches regardless of what `scriptHashHex` does. It now asserts the direction (`fwd == byAddr` and `revd == 0`), and more importantly runs the actual `lwwallet.NewEsploraClient` against the funded script and requires `stats.Used()`. That's the part that ties our encoding to electrs' rather than to a stub that agrees with us by construction. Reverting `esplora.go` now fails this test. There was also a leftover `t.Logf("chaincfg %s", ...)` whose comment claimed to report which encoding the client produces and did nothing of the sort; removed.

**Claude #2, silent truncation at the index cap** — added a `WarnS` with the scope, branch and index when a branch walk stops on the cap rather than on a full unused run.

**Claude #3, serial pagination in `earliestScriptHeight`** — leaving as is. The used-script set is bounded by the gap limit and this is once per wallet creation, so I'd rather not add concurrency against a rate-limited endpoint for a path that isn't hot. Noted if it ever shows up in a profile.

One thing I found in my own pass and documented rather than changed: discovery only reaches HD-derived scripts, so imported taproot scripts (boarding addresses) are outside any gap-limit walk, since they encode a script tree rather than a derivation path. That costs nothing, because `BoardingBackendAdapter.ListUnspent` enumerates boarding outputs by querying Esplora per imported address, so their visibility never depended on btcwallet's sync height. Spelled out in the `usedScripts` doc comment so the next person doesn't have to re-derive it.

Unit tests, full systest suite (19 tests) and local lint all green after the changes.

Roasbeef:
Ran a `/advisor-review` pass (Fable) as a closing gate. It found four real correctness issues, one of which I had introduced myself in the previous round, so this was worth doing.

**The clamp I added for the Codex OOM finding created a fund-loss bug.** `discoveryGapLimit` clamped to `min(recoveryWindow, 10_000)`, but btcwallet's `recovery()` uses the *unclamped* window as its horizon. With `recoverywindow=20000` and a first-used address at index 15 000, discovery sees 10 000 consecutive unused addresses and exits **normally** — no truncation warning, because `unusedRun` reaches `gapLimit` before the index guard fires — reports no history, and stamps `tip-6` as verified. btcwallet's own recovery would have found index 15 000. The stamp now sits above the funds, permanently, since a verified birthday block is never revisited.

Fixed by refusing rather than narrowing: `usedScripts` rejects up front with `errGapLimitTooWide` when the window exceeds one branch walk's budget, and the whole attempt aborts so btcwallet falls back to its own search.

**My "still safe" comment on the index cap was backwards.** I wrote that a truncated walk was fine because "a smaller discovered set can only lower the start height". It is the opposite: fewer used scripts raises the minimum height reported, moving the stamp toward the tip. That is the direction that loses funds. A capped walk now returns `errDiscoveryCapped` instead of warn-and-continue.

**The change-branch skip had a hole in its bootstrap argument.** I reasoned that a change output requires a spend, which requires a held output, which requires a receive address to have been paid. The bootstrap step is where it breaks: a wallet-owned output need not come from a derived receive address at all. aezeed is lnd-compatible, and an lnd wallet whose channels were opened from externally funded PSBTs holds its only on-chain coins in force-close sweeps, which lnd sends to *change* addresses. Such a seed has funds on internal branches with every external branch untouched. Skip removed, both branch sets always walked. Cost goes 400 -> 800 probes on a fresh wallet, which I took; PR body numbers updated.

**Load-balanced Esplora staleness** has no small code fix, so it is documented as a trust assumption in the `stampStartBlock` doc and `lwwallet/CLAUDE.md`: a public endpoint can serve a current tip and a still-syncing script index from different backends, and partial history is indistinguishable from an unused script. A restore in that window stamps above its funds permanently. Point `EsploraURL` at an endpoint you control.

Also fixed a test-quality finding: `TestDiscoveryCoversWalletAddresses` built its probe set *after* `BtcWallet.Start()` while production discovery runs before it, so a scope registered by `Start()` — exactly the drift the test guards against — would have been included and the test would have passed anyway.

Four new regression tests, each verified to fail without its fix. Re-verified through the same reviewer against the new head: refusal paths abort cleanly with no partial writes (both waddrmgr writes share one `walletdb.Update`), the `maxDiscoveryIndex` field fails closed if ever zero, and all four tests are load-bearing. Nothing new above LOW; took two of the three nits (folded the duplicate log into the returned error, added coverage of the production threshold constant).

Full systest suite green (19), 63 unit packages green, lint clean.

Out of scope and **not** fixed here, but worth a follow-up: `waved/server.go`'s env-seed create path passes `time.Time{}` into `startBtcwallet`, so the neutrino-backed sibling still has this exact shape. It is mitigated by compact filters rather than per-block HTTP, so it is slow rather than catastrophic, but it is the same bug one backend over.

sputn1ck (approved):
Tested this exact head through the server lwwallet daemon integration suite. The Esplora start-block path was exercised successfully; the observed failures were local harness process resets under concurrent Docker load, not client assertion failures.

PR #1077, opened 2026-07-31
lwwallet: speed up sync by bypassing the linear block scan

### LINKED ISSUE #1073: lwwallet: avoid sequential block scan with esplora client

guggero:
Fixes https://github.com/lightninglabs/wavelength/issues/1073 and is complementary to https://github.com/lightninglabs/wavelength/pull/1076.

Implements three different speedups:
 - On `TipPoller.Start()`, fetching the 99 recent block hashes synchronously blocks the whole wallet startup -> moved to background.
 - In `EsploraChainService.FilterBlocks`, since we're already talking to an address indexer, we can bypass the sequential `s.esplora.GetRawBlock()` calls by looking up the history of the addresses to scan.
 - On `EsploraChainService.Rescan()` we do the same by fetching the whole history of the addresses in question.

### DIFF
--- a/lwwallet/AGENTS.md
+++ b/lwwallet/AGENTS.md
@@ -45,6 +45,16 @@ base logic with the neutrino-backed `btcwbackend` sibling via the extracted
 - `EsploraChainService` — `chain.Interface` adapter over `EsploraClient`,
   driven by a shared `TipPoller`. Feeds btcwallet's internal address-credit
   pipeline. Constructor: `NewEsploraChainService(esplora, tipPoller, logger)`.
+- Start-block discovery (`start_block.go`) — resolves where a new or
+  restored wallet begins scanning from the Esplora address index instead of
+  from the seed birthday, and writes it into `waddrmgr` as a *verified*
+  birthday block before `BtcWallet.Start()`. Entry point:
+  `Wallet.stampStartBlock`. Left to itself btcwallet binary searches block
+  timestamps (`locateBirthdayBlock`), which bottoms out at genesis for any
+  birthday older than the chain and then walks every block to the tip.
+  Discovery instead derives each active key scope's receive/change scripts
+  through btcwallet's own scoped key managers, probes `/scripthash/:h` under
+  the configured gap limit, and takes the earliest confirmed height.
 - `BoardingBackendAdapter` — Embeds `walletcore.BoardingBackendBase` for
   shared key derivation/script import; implements `wallet.BoardingBackend`
   and `wallet.OutputLeaser`. Queries Esplora directly for UTXOs (bypasses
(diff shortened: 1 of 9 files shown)


sputn1ck:
Update after taking over the PR from @guggero. The branch is now organized as three atomic commits; Guggero remains the author of both implementation commits, with the follow-up fixes recorded as co-authored work.

## What this PR changes, and why

### 1. Seed the tip poller's reorg history without blocking wallet startup

Previously, `TipPoller.Start` synchronously walked roughly 100 headers before returning. Since wallet startup waits for the poller, that could add many seconds of serial Esplora requests on a mobile connection.

The history walk now runs under the poller's component-owned goroutine/context, so `Start` can publish the current tip and return promptly. The polling loop still does not begin until the complete retained hash ring has been seeded, which means reorg handling never observes a partial ring.

We also changed transient seed failures from "log and continue polling" to bounded exponential retries. This matters because entering the polling loop with a gap would make a later reorg look like internal corruption and disable reorg handling until the missing height aged out of the ring. Requests and retry timers use `runCtx`, so `Stop` cancels them promptly.

### 2. Use Esplora's address index for `FilterBlocks` and `Rescan`

The old path downloaded and parsed every block body in the recovery range. That defeats the purpose of using an indexed Esplora backend and is the dominant cost of restoring an old wallet.

The new path:

- queries paginated script histories for derived wallet addresses and watched outpoints;
- keeps only confirmed candidates in the requested height range;
- fetches and runs candidates through the existing wallet matching logic;
- validates each indexed height/hash against the canonical chain before advancing wallet state;
- deduplicates candidates and preserves parent-before-child transaction ordering;
- returns the lowest matching block from `FilterBlocks`; and
- keeps the per-height hash/header walk required by btcwallet, while avoiding raw block downloads on the indexed fast path.

The history pagination stop condition is deliberately strict: we stop only after the oldest page entry is *below* the minimum requested height. This ensures we fetch every page when more than one page of transactions shares the boundary height.

### 3. Fail safely when the index answer cannot be trusted

Both `FilterBlocks` and `Rescan` fall back to the original full-block scan when history is truncated at the safety cap, an indexed block identity conflicts with the canonical chain, history/transaction retrieval fails, or a watched outpoint's script cannot be represented as an address.

That last watched-outpoint case previously used `continue`, which silently skipped a potentially relevant spend. Routing it to the canonical scan preserves the old behavior instead.

Cancellation is not treated as an index failure: context cancellation is returned directly rather than starting an expensive fallback scan during shutdown.

### 4. Document the remaining index-trust boundary

A successful empty script-history response is indistinguishable from a lagging-but-otherwise-valid Esplora index. Failing closed on every empty answer would require scanning every height without wallet activity, negating the optimization entirely.

The package docs now state this invariant explicitly: recovery trusts the completeness of successful index responses, and a lagging index can cause `PutSyncedTo` to advance past omitted activity permanently. Wallet recovery should therefore use a trusted, fully indexed Esplora instance, ideally one under the operator's control.

### Test coverage added

The new tests cover:

- startup returning without waiting for the seed walk;
- polling waiting for a complete ring;
- a reorg racing startup;
- recovery after a transient seed failure;
- cancellation of in-flight seed requests and retry timers;
- fully spent addresses, unconfirmed entries, and out-of-range entries;
- multi-page histories, including more than one page at the minimum height;
- watched-outpoint spends;
- false index candidates and canonical block-hash conflicts;
- parent-before-child transaction ordering;
- history-cap and unsupported-address fallbacks in both scan paths;
- cancellation without fallback;
- indexed `Rescan` notification behavior; and
- a deliberately incomplete fake index documenting the successful-empty trust contract.

The fake Esplora uses the service `runCtx`, so the cancellation branches are exercised, and fast-path tests reject unexpected block-body requests.

## Validation

Passed locally:

- `make fmt-changed-check`
- `make lint-changed-local`
- `make tidy-module-check`
- `make build`
- `go test ./lwwallet -count=1`
- targeted `go test -race` for the new retry, reorg, fallback, pagination, index-contract, and cancellation cases
- `go test ./lwwallet` independently on each implementation commit

The remaining acknowledged follow-ups are performance-only: bounding/caching concurrent history requests and smoothing the `maxScriptTxPages` fallback cliff. Both currently preserve correctness by falling back to canonical block scanning.

PR #735, opened 2026-06-12
docs: nightly doc-gardening sweep 2026-06-12

litbot-9000:
Nightly automated doc-gardening sweep for 2026-06-12.

Workflow run: https://github.com/lightninglabs/darepo-client/actions/runs/0

## What changed

**New CLAUDE.md/AGENTS.md** (packages previously missing docs):
- `swaprpc` — External swap-server gRPC service definitions
- `mailbox/pb` — Mailbox wire protocol protobuf stubs
- `rpc/oorpb` — OOR transfer mailbox protocol stubs
- `rpc/swapclientrpc` — Daemon-side swap client service stubs
- `db/sqlc` — Main database type-safe query layer (17 migrations)
- `db/actordelivery/sqlc` — Actor delivery query layer
- `sdk/swaps/sqlc` — Swap client persistence query layer
- `serverconn/hellotestpb` — Test-only HelloService proto package

**Updated CLAUDE.md/AGENTS.md** (stale docs reflecting recent changes):
- `arkrpc`, `daemonrpc`, `rpc/roundpb` — Proto changes: removed global
  ForfeitScript/SweepKey/SweepDelay from ServerInfo; added 5 per-round
  keys to ClientBatchInfo (TreeCosignKey, ConnectorOperatorKey, SweepKey,
  SweepDelay, ForfeitKey)
- `chainsource` — Block epoch actor auto-reconnect with exponential backoff
- `oor` — New `normalizeCheckpointOwnerLeaves` operator-key-rotation fix
- `round` — Per-round sweep delay validation (moved from global init)
- `sdk/ark` — Multi-recipient OOR send (Recipients array)
- `sdk/swaps` — New `AcknowledgeOutSwapHTLC` swap server method
- `swapwallet` — Cooperative-leave EXIT row completion heuristic
- `db` — BoardingStore now embeds InternalKeyQuerier
- `internal/actortest` — outboxDeliveryTimeout increased to 30s

**ARCHITECTURE.md** — Added `swaprpc` to Layer 3 package table.

Please review the updated CLAUDE.md/AGENTS.md and ARCHITECTURE.md diffs.

PR #1083, opened 2026-08-03
vtxo: Make automatic refresh safe and cohort-aware

sputn1ck:
Addresses [Lumos #735](https://github.com/lightninglabs/lumos/issues/735).

## Problem

Expiry-driven refreshes were scheduled independently at each VTXO's
OOR-aware safety threshold. VTXOs from the same expiry batch could therefore
pay separate seal-time fees, and automatic rounds accepted any quote below the
broad global fee cap. Error and restart boundaries could also leave reserved
VTXOs stuck in `PendingForfeit` or `Forfeiting`.

## Changes

- Coordinate eligible same-expiry VTXOs into one canonical, bounded, atomic
  refresh cohort, even when their OOR depths produce staggered trigger heights.
  Preserve a sibling's later free-refresh window when it is still safe to wait.
- Mark expiry-driven outputs with local auto-refresh provenance, derive a fresh
  signing key for every replacement, preserve duplicate-policy outputs by
  source outpoint, and put the change marker on the output with the most
  headroom.
- Reject authoritative quotes that create a VTXO below the advertised minimum.
  Add optional absolute and proportional automatic-refresh fee limits while
  retaining the mandatory global fee cap for every round and reseal.
- Log accepted and rejected authoritative quotes, replace the misleading
  pre-seal zero-fee log, and apply a six-block retry cooldown while the VTXO is
  safe. Critical and expired paths bypass the cooldown.
- Fix the expiry RPC's missing `ChainDepth`, so reported thresholds match the
  safety calculation used by the actor.
- Release an entire cohort when the round handler rejects it before ownership
  transfer. Across restart, release `Forfeiting` only when storage proves the
  consuming round never reached its signature-handoff checkpoint; re-drive a
  terminal confirmation when a joined settlement proves the round confirmed.
  Missing or ambiguous evidence fails closed.
- Expose the new limits through `waved`, the embedded Go SDK, mobile JSON
  config, the sample config, and operator/client documentation.

The internal restart lookup can hydrate settlement data for `Forfeiting`
descriptors, but the public RPC remains unchanged: settlement is exposed only
for terminal `Forfeited` VTXOs.

## Validation

- `go test ./... -count=1`
- `go test -race ./round ./vtxo ./waved -count=1`
- `make lint-changed-local`
- `make tidy-module-check`
- `make sample-conf-check`
- `go test -tags='mobile wavewalletrpc swapruntime' ./sdk/wavewalletdk/mobile -count=1`

Companion server PR: [Lumos #737](https://github.com/lightninglabs/lumos/pull/737).
It adds cohort-aware server pricing, post-fee output floors, and the
incident-shaped integration test, and pins this commit through the client
submodule.


sputn1ck:
@claude please review

cmd/waved/main.go
```diff
@@ -278,6 +271,28 @@ func newRootCmd() *cobra.Command {
...
+			"automatic refresh maintenance; zero disables this "+
+			"additional cap",
+	)
+	f.Uint32(
+		"maxautorefreshfeerateppm", cfg.MaxAutoRefreshFeeRatePPM, "m"+
+			"aximum automatic-refresh operator fee as parts "+
+			"per million of refreshed value; zero disables "+
+			"this additional cap",
+	)
+}
```
bhandras, line 297: **Config shape: consider one budget curve instead of two stacked caps** (re the knobs introduced in bd3b288d)

The manual/automatic split itself is right — unattended maintenance deserves its own budget, and "I'll approve a large fee on a deliberate refresh but the daemon may never silently pay that" can't be expressed with the global cap alone. What I'd push back on is the *composition* of the two automatic knobs. They combine as

```
effective auto cap = min(maxautorefreshfeesat, value × maxautorefreshfeerateppm / 1e6)
```

which is the wrong shape against a fee of `fixed share + margin + per-input liquidity`:

- For small cohorts the ppm term collapses below the fixed component, so a low-value coin (e.g. a lone 1,000-sat VTXO) becomes **permanently unrefreshable** under a ppm cap that sounds generous (1% ⇒ fee must be ≤ 10 sats). It will then cooldown-retry until critical and unilaterally exit — costing more than the fee the policy refused.
- The absolute knob can't rescue it, because `min()` only tightens.
- Operationally, three flags all named `max…` make it genuinely unclear which one to set and which binds when.

Proposal: keep `maxoperatorfeesat` as the single mandatory hard ceiling, and replace the two automatic caps with one budget curve that matches the fee formula's shape:

```
auto budget(value) = max(autorefreshfeefloorsat,
                         value × autorefreshfeerateppm / 1e6)
…always clamped by maxoperatorfeesat
```

- **floor** answers "what may unattended maintenance always spend per round" — roughly a small multiple of the typical fixed share + margin. One flat allowance per *round* matches the server-side cohort pricing in the companion PR exactly, since the fixed components are now charged once per cohort.
- **rate** answers "what's economical at scale".
- Zero/zero keeps today's default (only the global applies).
- Validate `floor ≤ maxoperatorfeesat` at config load.
- Keep the floor operator-configured rather than derived from the server's fee estimate, so an operator can't loosen client budgets via inflated estimates.

On the obvious objection — `max(flat, proportional)` is the composition swapdk-server just deprecated for in-swap routing budgets: the problem there was an *uncapped* budget derivation whose proportional term scaled with the payment amount, i.e. unbounded exposure. Here the global cap remains a mandatory clamp above the whole expression, and the proportional term is against value the user is preserving out of their own coin. `max()` acting as a floor **under a hard ceiling** is the standard fee-limit shape (cf. LND's fee limits), not an open-ended budget.

Side benefit: with a sane floor, the "never-viable small VTXO drifts to an expensive unilateral exit" failure mode requires the user to have explicitly set the floor below the fixed fee, instead of being the accidental default of any low ppm setting.

This would ripple into the same files this PR already touches (waved config validation, SDK/mobile overlays, sample-waved.conf, docs), so it's cheapest to reshape now before the knob names ship.

sputn1ck:
Addressed the latest review feedback and rebased this PR onto current `main`.

Review changes:
- Replaced the two stacked automatic-refresh caps with one floor-plus-rate budget curve:
  `min(maxoperatorfeesat, max(autorefreshfeefloorsat, value * autorefreshfeerateppm / 1e6))`.
- Kept zero/zero as the backward-compatible global-cap-only policy.
- Added validation that the floor cannot exceed `maxoperatorfeesat`, and updated the CLI, sample config, SDK/mobile overlays, docs, and focused tests.
- The current head also snapshots the operator key once for the entire cohort and rolls back both the leader and siblings if the round handoff fails, with regression coverage.
- Adapted the PR's test actor refs to the new upstream `TryTell` interface after rebasing.

Local verification passed for the affected `vtxo`, `round`, `waved`, SDK, and mobile packages, plus changed-code lint, formatting, module tidiness, sample-config validation, and commit-message lint.

The latest CI failure was runner infrastructure rather than a code assertion: the `unit-race`, Linux ARMv7, and PostgreSQL jobs all reported that their self-hosted runner lost communication with GitHub. I have rerun only the failed jobs: https://github.com/lightninglabs/wavelength/actions/runs/30910217074

cmd/waved/main.go
bhandras: `7b204aae` fully addresses this — thanks! Verified at `ea6311bf`:

- one `max(floor, rate × value)` curve clamped by `maxoperatorfeesat`, computed in a single helper shared by enforcement and logging;
- `floor <= maxoperatorfeesat` rejected at config load (and the embedded path is covered, since `wavewalletdk` runs `daemonCfg.Validate()` before boot);
- fail-closed budget on non-positive automatic value;
- `TestAutoRefreshFeeBudgetDefaultsToGlobalCap` locking that zero/zero means *disabled*, not zero budget — the one subtle regression this refactor could have introduced;
- `TestAutoRefreshFeeBudgetUsesOneCurve` reproducing exactly the small-VTXO-rescue scenario from this thread, and the CLI guide now documents the floor's purpose.

Ran `go test ./round ./vtxo ./waved ./sdk/wavewalletdk/...` and the tagged mobile suite at head — all green. Resolving.

db/round_store.go
```diff
@@ -280,6 +281,44 @@ func NewRoundPersistenceStore(
 	}
 }
 
+// HasForfeitRoundCheckpoint reports whether roundID has a durable round row.
+// CommitState creates that row at InputSigSent, before any collected VTXO
+// forfeit signatures are submitted to the operator. Consequently, absence is
+// positive proof that a recovered Forfeiting VTXO never crossed the signature
+// handoff boundary and may safely be released. Malformed IDs and query errors
+// are returned so callers can fail closed.
+func (s *RoundPersistenceStore) HasForfeitRoundCheckpoint(ctx context.Context,
```
bhandras, line 290: This doc comment states the safety fence correctly, and I verified it holds today: in `finishForfeitCollection`, `RoundStore.CommitState` runs synchronously inside the transition (an error aborts it), and `SubmitVTXOForfeitSigsToServer` only leaves on the post-commit outbox — so "no round row ⟹ no forfeit signatures reached the operator" is sound.

But the invariant is only documented here at the *consumer*. At the *source*, `CommitState`'s own doc comment still says it "should be called at the point of no return **when the client has sent partial signatures**" — which reads as commit-*after*-send, the exact opposite of the ordering this release path now depends on. With `releasePreCheckpointForfeits` in the tree, a future refactor that reorders `finishForfeitCollection` (or submits signatures incrementally during collection) would silently turn the startup sweep into a release-a-leaked-signature hazard, with every existing test still green.

Three asks:
1. Fix the stale `CommitState` doc comment (commit happens *before* submission, and the release fence requires that).
2. Add a short invariant note in `finishForfeitCollection` pointing at this lookup and `releasePreCheckpointForfeits`.
3. Ideally, a regression test asserting the transition emits no `SubmitVTXOForfeitSigsToServer` when `CommitState` fails (the harness can observe the outbox), so the ordering is pinned by CI rather than by comments.

bhandras (commented):
Re-reviewed at `ea6311bf`. The fee-budget reshape in `7b204aae` resolves the config-shape thread completely (details there); commits 1–3 are content-identical to the previous head apart from `TryTell` rebase adaptations. Ran `go test ./round ./vtxo ./waved ./sdk/wavewalletdk/...` plus `-tags='mobile wavewalletrpc swapruntime' ./sdk/wavewalletdk/mobile` at head — all green.

Overall this is careful work: the fail-closed discipline in the restart paths, the token/generation-matched cohort rollbacks, and the atomic single-turn cohort registration all check out under scrutiny, and the test coverage (deterministic cap, single aggregate deadline, same-block adoption, relay-failure release, crash-gap reconcile) is strong.

Three items before merge from my side:

1. **Pin the commit-before-send ordering invariant** that `releasePreCheckpointForfeits` depends on — see the inline comment on `HasForfeitRoundCheckpoint`. This is the only substantive ask.
2. **Squash the fixup commits**: `vtxo: Apply canonical formatting` and `vtxo: Adapt test refs to backpressure` are both fixups — and the latter is inconsistent with the rebase, which *did* fold the equivalent round-harness `TryTell` adaptation into commit 1 while giving the vtxo ones their own commit.
3. **lumos#737 needs a re-pin**: it pins `50e35104`, which this force-push orphaned; re-pin to the final head before that PR merges.

Nit: the word-split flag-help strings (`"proport"+"ional"`, `"fixed"+" fee"`) survive the rename — reflow when convenient.

bhandras (approved):
LGTM 🎉

PR #1093, opened 2026-08-04
serverconn+round: never park the ingress path, make backpressure visible

Roasbeef:
In this PR, we fix the wedge behind the recurring "ark client goes deaf to round events" incidents (lightninglabs/lightning-infra#3675, two production occurrences). The serverconn ingress loop delivered server events with a blocking, deadline-free `Tell` into fixed-capacity in-memory mailboxes, while holding the SQLite global write transaction. Once a target mailbox filled, the sole ingress goroutine parked forever with the writer pinned. WAL mode meant readers kept answering RPC, so the pod looked healthy while every write in the process was frozen, and the only log line on that path was at trace level. The client silently dropped out of every round it had joined ("all clients dropped at seal time" on the operator side) until someone restarted it.

This is the client-side sibling of lightninglabs/lumos#734: an actor's receive goroutine (or here, the loop feeding all of them) must never block indefinitely on a peer's bounded mailbox. It builds on the `TryTell` primitive from #1082.

## What changes

The ingress path now dispatches through a new `TellWithoutParking` primitive that picks the right send per resolved target: `TryTell` for a bounded in-memory mailbox, ordinary transactional `Tell` for a durable one. A full mailbox becomes a *deferral* instead of a park: the fold commits the cursor at the deferred envelope's own `event_seq` (everything below it is fully handled, the undelivered tail is never acked), and the pull loop is the redrive. Redriven batches are clamped and a served watermark prevents re-answering hoisted non-transactional requests every cycle. Backpressure gets its own backoff with a 500ms ceiling that resets on progress, so a drained backlog clears in seconds rather than at the transport-failure schedule's 30s. Since the store retries its transaction body on retryable errors, a per-dispatch record of out-of-transaction deliveries keeps those retries from re-sending messages that already landed in a mailbox.

Beyond the ingress path, we close the two other ways the audit found to park the round actor's receive loop: the round-to-wallet Ask cycle gets a bounded `askWallet` (both the enqueue and the Await, 30s default), and `FSM.CurrentState` gets a context-aware variant, with the scanning call sites sharing one deadline per scan instead of paying a per-FSM budget N times.

## Observability

Both incidents produced zero evidence, so visibility is half this PR. Deferrals now log at error level (once per episode, re-logged every 30s with episode counters) and increment `waved_serverconn_ingress_dispatch_deferred_total`; two new gauges split "the ingress loop is alive" (last poll) from "dispatch is landing events" (last event). The metrics README gains an alerting section that maps each expression to the failure it detects; the deferral rate is the primary wedge signal, since a deaf client keeps polling and poll staleness alone can never see this failure.

## What this does not claim

One residual is documented in `serverconn/doc.go` rather than fixed: all sixteen inbound routes share one strictly ordered cursor, so a stalled target still delays events for the other routes until it drains. What this PR removes is the writer pin and the silence. Decoupling the lanes (a per-target deferral queue with its own cursor) is follow-up work.

## Testing

The regression tests wire a real actor system, the real event router, and the existing writer-lock harness, and each fails on the pre-fix code in seconds, one per mechanism: the storm test observes the loop parked inside the write transaction (the exact production signature), the redrive test counts unbounded hoisted re-serves, the recovery test times the drain on production-shaped config (63s pre-fix, under 4s post), and the replay test counts duplicate in-memory deliveries under transaction retries.

On top of the example-based tests, two further layers pin the invariants. Property tests (rapid) drive the real fold, cursor arithmetic, and backoff across randomized batches, deferral points, and store-retry injections; each property was validated non-vacuous by mutating the production code and watching it fail. And the P model suite gains a machine for this pipeline with six spec monitors (no acked loss, no tx-scoped duplicate, bounded at-least-once, served-once-per-incarnation, writer never parks, eventual drain); each pre-fix bug is a one-field config toggle the checker catches, and crash/restart nondeterminism is part of the schedule space.

The diff went through an adversarial review pass, a verification pass, and a complexity-focused design review; all findings are addressed in-line or documented as residuals. The design review drove one API simplification: the delivery helper is two exported names (`TellWithoutParking` + `NonParkingTeller` returning `fn.Result[bool]`), with the capability-query surface deleted rather than exported.

See the individual commit messages for the incremental story.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -87,7 +87,7 @@ package may import from a higher layer.
 
 | Package | Purpose |
 |---------|---------|
-| [`p-models`](p-models/) | Executable P formal models and Go conformance bridge for distributed-systems properties (durable mailbox, Read/Commit fence) |
+| [`p-models`](p-models/) | Executable P formal models and Go conformance bridge for distributed-systems properties (durable mailbox, Read/Commit fence, ingress deferral and redrive) |
 | [`p-models/durableactor/bridge`](p-models/durableactor/bridge/) | Go conformance harness: replays P model mailbox traces against the real `db/actordelivery` store |
 | [`harness`](harness/) | Docker-based Bitcoin/LND integration test environment |
 | [`systest`](systest/) | System-level end-to-end tests |
--- a/baselib/actor/bounded_delivery.go
+++ b/baselib/actor/bounded_delivery.go
@@ -1,0 +1,165 @@
+package actor
+
+import (
+	"context"
+
+	"github.com/lightningnetwork/lnd/fn/v2"
+)
+
+// NonParkingTeller is implemented by every reference in this package and is
+// how a reference answers the question "deliver this without ever parking me"
+// for the target it actually resolves at send time.
+//
+// The interface exists because the reference, not the caller, knows two
+// things the decision depends on: what kind of mailbox the concrete target
+// has, and which target a send-time-resolving reference (a Router, or a
+// transform wrapping one) picks for this particular message. A Router key
+// can hold both a bounded and a durable actor at once; any answer computed
+// before selection can disagree with the target the message is handed to,
+// and picking the wrong primitive for a durable target is the expensive
+// direction (see TellWithoutParking). Making the reference decide and
+// deliver in one step is what keeps the answer and the delivery consistent.
+type NonParkingTeller[M Message] interface {
+	// TellWithoutParkingTo delivers msg to whichever target this
+	// reference resolves for it, choosing Tell or TryTell by that
+	// target's own mailbox. The result carries whether the delivery
+	// escaped the caller's transaction: true means a successful TryTell
+	// hand-off into an in-memory mailbox, which is already visible to
+	// the receiving actor and cannot be rolled back, so a caller whose
+	// transaction may run more than once must not send it a second
+	// time. False means nothing escaped: a Tell enqueue joins whatever
+	// transaction the caller's context carries and is undone with it, a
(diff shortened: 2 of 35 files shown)


Roasbeef:
@claude review this

Roasbeef:
@claude review this

Roasbeef:
Both items addressed in fef8edff:

1. **Deferral log level split by episode age.** The opening line of an episode is now `WarnS` (a brief deferral is ordinary backpressure the deferred counter already captures, and the repo rule is right that error level is for internal bugs), and the interval re-logs escalate to `ErrorS`: an episode still open a full `deferralLogInterval` later means the target has stopped draining, which is exactly the stuck-receive-loop bug class this path exists to expose. The decision is recorded in the doc comment next to the call, so the next reader hits the reasoning rather than the lint rule. The deferred-counter rate stays the paging signal either way.

2. **The no-ctx `CurrentState` semantic change is now stated in the protofsm commit body**, with the reasoning for why the exposure is theoretical for oor/txconfirm/unroll: those machines are driven by `AskEvent` from the same goroutine that issues the state queries, so a transition cannot be in flight when a query runs, and every caller already handled the reply half's one second timeout, so the error path is not new. One related note: txconfirm converting a state-read error into `TxStateFailed` is a pre-existing issue reachable before this PR (via `ErrStateMachineShutdown` and the old reply-half timeout) and is tracked separately in #1095.

Thanks for tracing the clamp/watermark disjoint-window interaction independently; that is the subtlest piece of the cursor arithmetic and it now has three independent confirmations (adversarial review, the P model's non-vacuity mutations, and this pass).

PR #1100, opened 2026-08-05
waved: Reserve exact managed OOR inputs

sputn1ck:
## Problem

Outpoint-only `custom_inputs` resolve local VTXO descriptors, but previously bypassed the wallet manager reservation path. The OOR transfer could succeed and then fail during local completion because `SpendCompletedEvent` was delivered to a VTXO that was still in `LiveState`.

This also prevented a safe one-off consolidation of the sub-minimum VTXOs described in lightninglabs/lumos#735.

## Fix

- Treat an all-outpoint-only `custom_inputs` list as exact managed wallet inputs.
- Reserve those exact outpoints through the existing durable `SpendingState` path; do not substitute other wallet liquidity.
- Keep explicitly described custom-policy inputs on their existing path.
- Reject mixed managed/custom input modes and duplicate outpoints before submission.
- Preserve input order, ordinary recipient fanout, and the existing target/change-floor validation.
- Report durably locked exact inputs as retryable liquidity after restart.

No protobuf or CLI surface change is required; the existing full JSON request is sufficient.

## Compatibility notes

- Outpoint-only inputs now intentionally require live, locally managed VTXOs. Callers spending non-standard inputs must provide their explicit policy/spend description.
- Custom-input shape and outpoint validation now runs during `dry_run` instead of returning a preview for malformed input.
- An over-funded exact-input set may still create wallet change when the residual meets the operator floor, matching the ordinary wallet-selected OOR path.

## Tests

- Regression coverage for an exact 744 + 744 sat input set, including `SpendingState` reservation and successful transition to `SpentState`.
- RPC coverage proving exact managed inputs retain multi-recipient fanout and use the requested outpoints.
- Restart-shape coverage proving durable `Spending`, `PendingForfeit`, and `Forfeiting` inputs return retryable liquidity errors.
- Wallet forwarding, explicit-custom fanout rejection, and mixed-mode rejection coverage.
- `go test ./vtxo ./wallet ./waved -count=1`
- `make lint-changed-local`
- `make fmt-changed-check`
- `make tidy-module-check`

### DIFF
--- a/lib/actormsg/vtxo_admission.go
+++ b/lib/actormsg/vtxo_admission.go
@@ -20,10 +20,9 @@ import (
 // (wallet → vtxo → round → wallet).
 
 // SelectAndReserveSpendRequest asks the VTXO manager to select VTXOs covering
-// a target amount and atomically reserve them for an OOR spend. The manager
-// runs largest-first coin selection, then Asks each selected VTXO actor to
-// process SpendReserveEvent. If any reservation fails, already-reserved
-// VTXOs are rolled back.
+// a target amount, or use explicitly named managed VTXOs, and reserve them for
+// an OOR spend. If any reservation fails, already-reserved VTXOs are rolled
+// back.
 type SelectAndReserveSpendRequest struct {
 	actor.BaseMessage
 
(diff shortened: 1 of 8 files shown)


sputn1ck:
@claude please review

PR #1103, opened 2026-08-05
swaps: authorize credit account requests

bhandras:
## What it does

- defines a canonical, method-bound proof for account-scoped credit requests
- signs short-lived request digests with the daemon identity key
- passes the account key named by each request into the signer and rejects any key that does not match the wallet identity
- attaches a fresh nonce and signature in both gRPC and REST swap transports
- rejects missing local signing support before a protected request leaves the client

## Protocol

The signature commits to the exact RPC method, deterministic request bytes, account public key, expiry, and a 32-byte random nonce. The authorization field is excluded from its own request digest. Proofs expire after one minute by default and signers reject expiries beyond five minutes.

The signer receives the account key extracted from the same canonical payload it signs. The daemon refuses to sign unless that key matches its identity key, preventing a caller from asking one wallet to authorize another wallet's account.

## Compatibility

Ordinary Ark-to-Lightning quote and create requests remain unsigned when no credit account is supplied. Lightning-to-Ark route requests are account-scoped because the route may inspect or reserve receive credit, so receive callers must use a connection configured with an identity signer. The exported constructors document this distinction.

## Testing

- `go test ./swaprpc ./sdk/swaps ./sdk/ark ./waved`
- `make fmt-changed-check`
- `make lint-changed-local workers=4`

The changed-file lint pass runs all 71 configured linters and reports zero issues.

### DIFF
--- a/cmd/wavecli/waveclicommands/devrpc/registry_generated.go
+++ b/cmd/wavecli/waveclicommands/devrpc/registry_generated.go
@@ -339,6 +339,13 @@ func generatedRegistry() []serviceSpec {
 					Output:   "waverpc.SignOutSwapHtlcAckResponse",
 					Comments: "SignOutSwapHtlcAck signs the accepted terms of an out-swap vHTLC with\nthe daemon identity key.",
 				},
+				{
+					Name:     "SignCreditAccountAuthorization",
+					Aliases:  []string{"sign-credit-account-authorization"},
+					Input:    "waverpc.SignCreditAccountAuthorizationRequest",
+					Output:   "waverpc.SignCreditAccountAuthorizationResponse",
+					Comments: "SignCreditAccountAuthorization signs one canonical swap credit-account\nrequest digest with the daemon identity key.",
+				},
 			},
 		},
 		{
--- a/rpc/restclient/clients.go
+++ b/rpc/restclient/clients.go
@@ -446,6 +446,20 @@ func (c *DaemonServiceClient) SignOutSwapHtlcAck(ctx context.Context,
 	return out, err
 }
 
+// SignCreditAccountAuthorization signs one swap credit-account request.
+func (c *DaemonServiceClient) SignCreditAccountAuthorization(
+	ctx context.Context, in *waverpc.SignCreditAccountAuthorizationRequest,
+	_ ...grpc.CallOption) (*waverpc.SignCreditAccountAuthorizationResponse,
+	error) {
+
+	out := new(waverpc.SignCreditAccountAuthorizationResponse)
+	err := c.client.Post(
+		ctx, "/v1/daemon/sign-credit-account-authorization", in, out,
+	)
+
+	return out, err
+}
+
 // GetIndexedVTXOByPkScript looks up one indexed VTXO by pkScript.
 func (c *DaemonServiceClient) GetIndexedVTXOByPkScript(ctx context.Context,
 	in *waverpc.GetIndexedVTXOByPkScriptRequest, _ ...grpc.CallOption) (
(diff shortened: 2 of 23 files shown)


bhandras:

Please focus on the protocol and client-side authorization boundary:
- the canonical digest must bind the exact RPC method and all request fields except the authorization itself
- account identity, expiry, and nonce must be included in the signed BIP-340 tagged digest
- gRPC and REST paths must attach fresh proofs for every protected request
- ordinary in-swap quote/create calls without a credit account must remain compatible
- daemon signing permissions and error handling must fail closed

The downstream verifier persists nonces before handler execution, so please also flag any client behavior that could accidentally reuse a nonce or sign a different payload from the one transmitted.

bhandras:
@claude please re-review the current head (`3eed78b0`).

The first review has been addressed as follows:

- `RequestChannelID` remains intentionally authenticated for every receive. Receive route registration is account-scoped because it may inspect or reserve receive credit; allowing an unsigned request would preserve the account-freezing path this series closes. The gRPC/REST constructor and method docs now state which operations require a signer, and a regression test proves an unsigned receive fails before transport.
- In the production receive flow, `client_vhtlc_pubkey` is the daemon identity key returned by `IdentityPubKey`; the separate payment-scoped receive-auth key is not passed in that field. The coupling is now explicit and fail-closed anyway: the canonical digest's account key is passed into `CreditAccountAuthorizationSigner`, and the daemon rejects signing unless it exactly matches its identity key. The daemon signing RPC carries and validates that account key, and tests cover a mismatch.
- protobuf requests are cloned with `proto.Clone`; no generated message state is copied by value.
- formatting, focused tests, the swapruntime transport test, all 71 changed-file linters, and commit-message lint pass locally.

Please focus the second pass on whether the explicit account-key signer contract closes the concern without weakening receive authorization, and flag any remaining merge blockers.

bhandras:
claude: **Final review of the credit-account authorization client changes — no merge blockers.**

Reviewed final PR head `3eed78b0`, merged as `cd74140e79154c591a96a554199e1bcd7b0b76d3`.

### Prior findings resolved

- `RequestChannelId` intentionally requires authorization for every receive because it reads account-scoped credit state. Constructor documentation now states that receive and credit-account operations require a signer. Tests prove unsigned `ListCredits` and `RequestChannelId` fail before transport.
- The signer contract is explicit and fail closed. `authorizeCreditAccountRequest` passes the request account key into the signer. The wallet daemon signs only when that key equals its identity key and rejects foreign keys.
- The production receive path uses the daemon identity as `client_vhtlc_pubkey`, so the account key in the signed digest matches the key transmitted in `RequestChannelId`.
- The unsigned request copy uses `proto.Clone`; protobuf state is not copied by value.

### Payload and replay integrity

- The digest is computed from the same request that is sent, with only `account_authorization` cleared before deterministic marshaling.
- Each call creates a fresh random nonce and expiry. Retries produce a new proof rather than replaying the previous nonce.
- The signature commits to the account key, exact request payload, method, expiry, and nonce.
- Ordinary `CreateInSwap` and `QuoteInSwap` requests without an account key remain unsigned.
- The authenticated REST transport carries the same proof, and signer access uses the expected swap write permission.

### Server dependency

The server must reconstruct the same method-and-payload digest, verify the BIP-340 signature against the account key carried by the request, and consume the nonce durably before entering the handler. swapdk-server PR #293 implements those requirements.

### Non-blocking follow-up

The package documentation said `swaprpc` contained only generated Go files, but this PR added hand-written authorization code and tests. Updating that documentation is useful housekeeping, not a merge blocker.

**Verdict: safe and compatible. The account-key signer contract closes the earlier concern and strengthens receive authorization by failing closed on identity divergence.**

PR #1111, opened 2026-08-06
roundpb: reject VTXO trees that share a child

ellemouton:
`TreeFromProto` validated three things about each child reference: the index is greater than its parent's, it is in range, and the output index is within the parent's output count. It never checked that a node is named as a child only **once**.

## The comment was the give-away

The existing comment claimed the pre-order check already covered this:

> *This prevents cycles (self-refs, mutual refs, back-edges) and **diamond DAGs (shared children)** in a single check.*

The first half is true, the second is not. `childIdx > i` rules out cycles, but two parents at indices 0 and 1 can both name child 5 and both satisfy it. What decodes is a DAG, not a tree.

## Why decode is the right place to fix it

The tree is walked recursively in many places. A shared child sits on more than one root-to-node path, so each walk re-visits its whole subtree once per path that reaches it — a chain of shared children turns a linear-sized message into exponential work.

Fixing it at the decode boundary means the DAG never becomes a `tree.Node` graph at all, which makes every downstream walk safe without auditing each one for re-visit behaviour. That is the argument for a single guard here over N guards spread across the consumers.

The check is a `parentOf` map recording which parent claimed each child, so a second claim can name both nodes in the error.

## The tests were already there, asserting the wrong thing

Two existing tests covered this shape and **asserted it was accepted**, documenting the gap as known:

- `TestTreeFromProtoDiamondDAG` — *"Diamond detection would require tracking parent counts. This test documents the current behavior."*
- `TestTreeFromProtoDiamondDoubleVisit` — went further and counted the consequence, asserting the shared child is visited **twice** in a depth-first walk.

Both now assert the decode is refused. The double-visit the second one demonstrated is the amplification in miniature, so inverting them is better than deleting them: the shape they describe is exactly what the guard exists to reject.

I also corrected the package `CLAUDE.md` invariant, which documented the pre-order check as covering this.

## Testing

`make unit pkg=rpc/roundpb` passes, as do `lib/tree` and `round` — the packages that consume the decoded tree. `make lint-changed-local` 0 issues, `make commitmsg-lint` OK.

### DIFF
--- a/rpc/roundpb/convert.go
+++ b/rpc/roundpb/convert.go
@@ -394,6 +394,11 @@ func TreeFromProto(pt *VTXOTree,
 	//    node's output count. Without this, downstream code that
 	//    accesses Outputs[outIdx] would panic.
 	//
+	// The bounds check runs before the single-parent check so a wild
+	// index is reported as out of range rather than as a sharing
+	// violation, and so parentOf never records an index that was
+	// never valid in the first place.
+	//
 	// parentOf records the parent that claimed each child, so a
 	// second claim can name both of them.
 	parentOf := make(map[uint32]int, len(pt.Nodes))
@@ -406,6 +411,11 @@ func TreeFromProto(pt *VTXOTree,
 					"back-reference)", i, childIdx)
 			}
 
+			if int(childIdx) >= len(goNodes) {
+				return nil, fmt.Errorf("node[%d] child index "+
+					"%d out of range", i, childIdx)
+			}
+
 			if prev, dup := parentOf[childIdx]; dup {
 				return nil, fmt.Errorf("node[%d] child index "+
 					"%d is already a child of node[%d]; "+
@@ -413,11 +423,6 @@ func TreeFromProto(pt *VTXOTree,
 					childIdx, prev)
 			}
 			parentOf[childIdx] = i
-
-			if int(childIdx) >= len(goNodes) {
-				return nil, fmt.Errorf("node[%d] child index "+
-					"%d out of range", i, childIdx)
-			}
 
 			if int(outIdx) >= len(goNodes[i].Outputs) {
 				return nil, fmt.Errorf("node[%d] child output "+
@@ -428,6 +433,25 @@ func TreeFromProto(pt *VTXOTree,
 
 			goNodes[i].Children[outIdx] = goNodes[childIdx]
 		}
+	}
+
+	// The three checks above give "every node has at most one
+	// parent", which describes a forest, not a tree: nothing so far
+	// requires a node to be reachable from index 0. A sender could
+	// therefore pad the message with nodes no walk from Root will
+	// ever reach, and every one of them would still be deserialized
+	// and, far more expensively, run through ComputeFinalKey below
+	// (a MuSig2 aggregation with a sort and a copy per node).
+	//
+	// Since every node other than the root must be claimed exactly
+	// once, counting the claims pins the decoded shape to a single
+	// connected tree rooted at index 0 and lets the loop below skip
+	// work nobody can reach. flattenNode always emits exactly this
+	// shape, so no well-formed producer is affected.
+	if len(parentOf) != len(pt.Nodes)-1 {
+		return nil, fmt.Errorf("tree has %d nodes but only %d are "+
+			"claimed as children; unreachable nodes or "+
+			"multiple roots", len(pt.Nodes), len(parentOf))
 	}
 
 	// Compute FinalKey for each node now that we have the
--- a/rpc/roundpb/convert.go
+++ b/rpc/roundpb/convert.go
@@ -372,18 +372,32 @@ func TreeFromProto(pt *VTXOTree,
 		goNodes[i] = node
 	}
 
-	// Wire up children references. We enforce two structural
+	// Wire up children references. We enforce three structural
 	// invariants:
 	//
 	// 1. Pre-order invariant: childIdx > i. Since flattenNode
 	//    serializes in pre-order, children always have higher
-	//    indices than parents. This prevents cycles (self-refs,
-	//    mutual refs, back-edges) and diamond DAGs (shared
-	//    children) in a single check.
+	//    indices than parents. This prevents cycles: self-refs,
+	//    mutual refs and back-edges.
 	//
-	// 2. Output index bounds: outIdx must be within the parent
+	// 2. Single parent: no node may be named as a child twice.
+	//    The pre-order check alone does NOT give this, contrary to
+	//    what this comment used to claim. Two parents at indices 0
+	//    and 1 can both name child 5 and both satisfy childIdx > i,
+	//    which decodes a DAG rather than a tree. Every recursive
+	//    walk over the result then re-visits the shared subtree once
+	//    per path that reaches it, so a decoder that accepts a
+	//    shared child hands the rest of the system exponential work
+	//    from a linear-sized message.
+	//
+	// 3. Output index bounds: outIdx must be within the parent
 	//    node's output count. Without this, downstream code that
 	//    accesses Outputs[outIdx] would panic.
+	//
+	// parentOf records the parent that claimed each child, so a
+	// second claim can name both of them.
+	parentOf := make(map[uint32]int, len(pt.Nodes))
+
 	for i, pn := range pt.Nodes {
 		for outIdx, childIdx := range pn.Children {
 			if childIdx <= uint32(i) {
@@ -391,6 +405,14 @@ func TreeFromProto(pt *VTXOTree,
 					"%d must be > parent index (cycle or "+
 					"back-reference)", i, childIdx)
 			}
+
+			if prev, dup := parentOf[childIdx]; dup {
+				return nil, fmt.Errorf("node[%d] child index "+
+					"%d is already a child of node[%d]; "+
+					"tree must not share children", i,
+					childIdx, prev)
+			}
+			parentOf[childIdx] = i
 
 			if int(childIdx) >= len(goNodes) {
 				return nil, fmt.Errorf("node[%d] child index "+
--- a/arkrpc/AGENTS.md
+++ b/arkrpc/AGENTS.md
@@ -34,5 +34,17 @@ Proto source: `arkrpc/ark.proto`, `arkrpc/indexer.proto`.
 - **Never edit generated code** — regenerate via `make rpc`.
 - Conversion round-trip: `TreePathFromTree(t)` → `TreePathToTree(pb)` must
   reproduce the original tree (excluding derived `FinalKey` fields).
+- `TreePathToTree` sits on the untrusted indexer receive path
+  (`vtxo.AncestryFromRPC` → `AncestryPathToTree`), so it validates the
+  decoded shape rather than trusting it: pre-order child indices (no
+  cycles), single parent (no node named as a child twice), full
+  reachability from index 0 (one connected tree, not a forest), index
+  bounds, and a node cap (`DefaultMaxTreePathNodes`, overridable via
+  `WithMaxTreePathNodes`). Single-parent is the load-bearing one:
+  `Children` is a map, so one node can point every output at the same
+  next node, and `nodeMaxDepth` — which the receive path runs on this
+  tree to validate the claimed ancestry depth — has no memoization, so a
+  shared child multiplies its work by the number of paths reaching it.
+  These mirror `roundpb.TreeFromProto`; keep the two decoders in step.
 - Child iteration during flattening is sorted by output index for
   deterministic serialization.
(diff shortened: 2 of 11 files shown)


ellemouton:

ellemouton:

litbot-9000:
### 📝 Doc gardening advisory

This PR touches three documented packages — `arkrpc`, `arkrpc/treeconv`, and
`rpc/roundpb` — and the docs for `arkrpc/treeconv` and `rpc/roundpb` are
already current, but `arkrpc/CLAUDE.md` never picked up the new exported
option surface (`TreePathToTreeOption`, `WithMaxTreePathNodes`,
`DefaultMaxTreePathNodes`) that `tree_path_convert.go` added.

```diff
diff --git a/arkrpc/AGENTS.md b/arkrpc/AGENTS.md
index f14c8319..f46bd9d5 100644
--- a/arkrpc/AGENTS.md
+++ b/arkrpc/AGENTS.md
@@ -15,6 +15,12 @@ Proto source: `arkrpc/ark.proto`, `arkrpc/indexer.proto`.
   with sorted child indices. Re-exported under the narrower
   `arkrpc/treeconv` sub-package for callers that do not need the full gRPC
   surface.
+- `TreePathToTreeOption` / `WithMaxTreePathNodes` /
+  `DefaultMaxTreePathNodes` — Functional-option surface bounding how many
+  nodes `TreePathToTree` will decode (default 50,000, deliberately equal to
+  `roundpb.DefaultMaxTreeNodes` so the two decode boundaries agree; a value
+  of 0 disables the limit). `AncestryPathToTree` passes no options, so the
+  indexer receive path runs with the default bound.
 - `IncomingOOREvent` — Lightweight notification (wake-up hint). Carries only
   session_id, pk_script, event_id. Triggers the three-phase receive flow.
 - `OORRecipientEvent` — Phase 1 query response from
diff --git a/arkrpc/CLAUDE.md b/arkrpc/CLAUDE.md
index f14c8319..f46bd9d5 100644
--- a/arkrpc/CLAUDE.md
+++ b/arkrpc/CLAUDE.md
@@ -15,6 +15,12 @@ Proto source: `arkrpc/ark.proto`, `arkrpc/indexer.proto`.
   with sorted child indices. Re-exported under the narrower
   `arkrpc/treeconv` sub-package for callers that do not need the full gRPC
   surface.
+- `TreePathToTreeOption` / `WithMaxTreePathNodes` /
+  `DefaultMaxTreePathNodes` — Functional-option surface bounding how many
+  nodes `TreePathToTree` will decode (default 50,000, deliberately equal to
+  `roundpb.DefaultMaxTreeNodes` so the two decode boundaries agree; a value
+  of 0 disables the limit). `AncestryPathToTree` passes no options, so the
+  indexer receive path runs with the default bound.
 - `IncomingOOREvent` — Lightweight notification (wake-up hint). Carries only
   session_id, pk_script, event_id. Triggers the three-phase receive flow.
 - `OORRecipientEvent` — Phase 1 query response from
```

**How to apply** — save the block above to a file and `git apply` it, or run
the doc-gardening skill locally against the touched package:

```
/doc-gardening arkrpc
```

Either way, keep the `CLAUDE.md`/`AGENTS.md` pair byte-identical — that is
what `make doc-check` enforces.

---

Advisory only; this does not block merge. Run:
https://github.com/lightninglabs/wavelength/pull/1111/checks

ellemouton:

ellemouton:
Thanks — both observations are right, and I verified the first rather than taking it on trust.

**Check ordering.** Confirmed: the duplicate check is at `convert.go:409`, `parentOf[childIdx] = i` at `:415`, and the range check at `:419` — so the map entry is recorded before `childIdx` is bounds-verified. Your trace holds: any out-of-range child errors on its own iteration and the function returns, so no invalid index ever survives into a wired edge, and the map is discarded with the error. Leaving it as-is, since reordering is non-functional and would churn a reviewed PR.

**Forest, not a rooted tree.** Good catch, and I agree it is out of scope here — an unreachable node is never walked, so it does not affect the amplification this closes, and it is pre-existing rather than introduced by the single-parent check. I have recorded it as a follow-up rather than let it die with the review, framed as a decision rather than a fix: either assert connectivity at decode (every node but index 0 has exactly one parent) so "tree" is true by construction, or state in the doc that the decoder guarantees a *forest* and no caller may assume reachability. The risk of leaving it implicit is a future walk that starts at the root and assumes it has seen every node.

**Doc mirroring** — yes, `ca135de` fixed that after the advisory bot flagged it. It was a genuine miss: I had updated `CLAUDE.md` and not its `AGENTS.md` mirror, so an AGENTS.md reader would still have been told the pre-order check covers shared children.

On the tests you could not execute: `make unit pkg=rpc/roundpb` passes, as do the `lib/tree` and `round` consumer packages, and CI is green on this branch. Both inverted tests were confirmed to fail against the pre-fix decoder, so they pin the guard rather than restate it.

litbot-9000 (commented):
FINAL — The fix is correct and does what it claims. I verified it rather than taking the description on faith.

**What I checked**

- Restored the worktree to a clean PR head first: an earlier run of this bot had left a PROBE edit in `rpc/roundpb/convert.go` and an untracked probe test behind. Both are saved under `.scratch/`; nothing here was reviewed against a dirty tree.
- `go test ./rpc/roundpb/ ./lib/tree/ ./arkrpc/ ./round/ ./vtxo/` — all pass at head.
- **The premise is right.** The old comment claimed the pre-order check (`childIdx > i`) subsumed diamond detection. It does not: two parents at indices 1 and 2 can both name child 3 with every edge forward. The PR correctly retires that claim in code, tests, and AGENTS.md/CLAUDE.md.
- **Attacker reachability confirmed.** `waved/server.go:3389` → `registerRoundEventRoutes` → EventRouter → `round/from_proto.go:218` → `TreeFromProto`. Server-push round events are decoded here, so a malicious/compromised Ark server controls this input directly.
- **The amplification claim is real.** `Node.NumTx`, `countLeaves`, `Depth`, `ForEach`/`ForEachLeaf` (`lib/tree/node.go`) and the stack-based `tree_materialize.go` walk all re-visit a shared subtree once per path reaching it, with no memoization. A linear message really does buy super-linear downstream work.
- **The check catches more than the tested diamond.** I probed a single parent naming the same child from two different output slots (`Children{0:1, 1:1}`) — correctly rejected. Self-refs and back-edges were already handled by the pre-order check. Cycles are impossible once every edge is forward, so acyclicity holds.
- **Allocation ordering is fine.** `maxNodes` (default 50k) is enforced before `goNodes` and before `parentOf` are allocated, so neither is sized by unbounded attacker input.

**Headline finding (follow-up, not a blocker for this PR)**

The identical gap remains in the sibling decoder `arkrpc.TreePathToTree` (`arkrpc/tree_path_convert.go:61-82`), which has the same wiring loop, the same pre-order check, no single-parent check, and no node cap at all. That decoder sits at a boundary this codebase's own comments explicitly designate untrusted (`ValidateAncestryPathDepth`: "The receive path treats indexer responses as untrusted, so this is the trust boundary"). I confirmed with a probe that it accepts the shared-child DAG, and that the very validator meant to bound the walk becomes the amplifier — details in the inline comment. This is pre-existing, so it should not hold up a change that strictly reduces attack surface, but it is the more exploitable half of the same bug and deserves a same-day follow-up.

No objections to the code as written; the two other inline notes are minor/nit.

rpc/roundpb/convert.go
```diff
@@ -392,6 +406,14 @@ func TreeFromProto(pt *VTXOTree,
 					"back-reference)", i, childIdx)
 			}
 
+			if prev, dup := parentOf[childIdx]; dup {
```
litbot-9000, line 419: 🟠 **major** — This check is right, but its twin is missing where it matters more.

`arkrpc.TreePathToTree` (`arkrpc/tree_path_convert.go:61-82`) has a byte-for-byte equivalent wiring loop — same `childIdx <= uint32(i)` pre-order check, same bounds checks, **no** single-parent check, and no `maxNodes` cap either. So the exact shape this PR now rejects still decodes there.

That matters because it is an explicitly untrusted boundary. `ValidateAncestryPathDepth` in `arkrpc/ancestry_path_convert.go` says so itself: *"The receive path treats indexer responses as untrusted, so this is the trust boundary."* The reachable path is:

`vtxo/incoming_ancestry.go:48` `AncestryFromRPC` → `AncestryPathToTree` → `TreePathToTree` (decodes the DAG) → `ValidateAncestryPathDepth` → `treeMaxDepth` → `nodeMaxDepth`.

And `nodeMaxDepth` is where it bites. It recurses over *every path*, bounded only by `MaxAncestryTreeWalkDepth = 32`, with no memoization and no visit budget. Because `Children` is a `map[uint32]*Node`, one node with k outputs can point **all k** entries at the same next node, giving k^(depth-1) paths from only `depth` nodes. I measured it (k=2, probe test, since removed):

```
nodes=10 depth=10 -> 2^9  paths  walk=66µs
nodes=14 depth=14 -> 2^13 paths  walk=886µs
nodes=18 depth=18 -> 2^17 paths  walk=58ms
nodes=22 depth=22 -> 2^21 paths  walk=990ms
```

Clean doubling per level. Extrapolating to the depth-32 cap: ~2^31 paths ≈ 17 minutes of CPU from a **32-node** message. With k=3 outputs per node it is 3^31 — an effective hang. And `MaxAncestryPaths = 64` (`vtxo/incoming_ancestry.go:15`) means one indexer response can carry 64 of these. I also confirmed `TreePathToTree` has no node cap at all (it accepted a 200k-node path).

The depth cap was added to stop a deep *linear* chain from overflowing the stack; it does not bound a *wide* DAG, and the walker added to enforce it is what does the exponential work.

Not asking you to block this PR on it — this change strictly improves things and should land. But please either fold the same six lines into `TreePathToTree` here, or open the follow-up before this merges so the fix does not stop half-done. The AGENTS.md text this PR adds ("prevents a malicious server from encoding cycles, shared children ... and DoS-ing tree traversal") reads as a codebase-wide guarantee that is not yet true on the indexer path.

rpc/roundpb/convert.go
```diff
@@ -372,18 +372,32 @@ func TreeFromProto(pt *VTXOTree,
...
+	//    shared child hands the rest of the system exponential work
+	//    from a linear-sized message.
+	//
+	// 3. Output index bounds: outIdx must be within the parent
 	//    node's output count. Without this, downstream code that
 	//    accesses Outputs[outIdx] would panic.
+	//
+	// parentOf records the parent that claimed each child, so a
+	// second claim can name both of them.
+	parentOf := make(map[uint32]int, len(pt.Nodes))
```
litbot-9000, line 404: 🟡 **minor** — Now that you are tracking `parentOf`, the reachability check is nearly free — and without it a forest still decodes.

I confirmed both of these still pass validation:

- two nodes, neither claiming the other: node 1 is simply unreachable from `Root`;
- three nodes where node 1 is a detached second root carrying its own subtree (`1 -> 2`).

Single-parent plus forward-edges gives you "every node has at most one parent", which is a *forest*, not a tree. The result is that a server can pad a message with up to `DefaultMaxTreeNodes` (50k) nodes that no walk from `Root` will ever reach — and every one of them still gets fully deserialized and, more expensively, run through `tree.ComputeFinalKey` (a MuSig2 key aggregation, with a sort and a copy per node) in the loop below. A tree with one useful node can carry 49,999 nodes of pure crypto busywork.

It is bounded by the node cap, so this is linear rather than the exponential case above, and it is pre-existing. But you are one line from closing it:

```go
if len(parentOf) != len(pt.Nodes)-1 {
        return nil, fmt.Errorf("tree has %d nodes but only %d are "+
                "claimed as children; unreachable nodes or "+
                "multiple roots", len(pt.Nodes), len(parentOf))
}
```

Placed after the wiring loop, that pins the decoded shape to exactly one connected tree rooted at index 0, and lets the `ComputeFinalKey` loop skip work no one can reach. Worth doing while the bookkeeping is fresh in the function.

rpc/roundpb/convert.go
```diff
@@ -392,6 +406,14 @@ func TreeFromProto(pt *VTXOTree,
 					"back-reference)", i, childIdx)
 			}
 
+			if prev, dup := parentOf[childIdx]; dup {
+				return nil, fmt.Errorf("node[%d] child index "+
+					"%d is already a child of node[%d]; "+
+					"tree must not share children", i,
+					childIdx, prev)
+			}
+			parentOf[childIdx] = i
```
litbot-9000, line 425: ⚪ **nit** — The duplicate check runs before the range check, so an out-of-range index that repeats gets reported as a sharing violation rather than an out-of-range one. `Children{0: 999, 1: 999}` on a 2-node tree yields *"child index 999 is already a child of node[0]"* — technically true, but it sends whoever is debugging the message at 3am looking for a tree-shape bug when the real problem is a wild index.

Moving the `childIdx >= len(goNodes)` check above the `parentOf` lookup fixes it and also keeps `parentOf` from recording indices that were never valid. Purely a diagnostics point — no behavioural difference, both inputs are rejected either way.

Issue #844, opened 2026-07-01
round: round failure after forfeit sigs are sent strands VTXO in Forfeiting (no runtime release)


Roasbeef:
## Summary

While driving the real client+server round actors under the new deterministic simulation testing (DST) harness in darepo (`systest/dst`, built on Go 1.25 `testing/synctest`), we hit a runtime forfeit strand that needs no crash and no restart to trigger: if a refresh/leave round fails _after_ the client has already submitted its per-VTXO forfeit signatures, the client FSM lands in `ClientFailedState` without ever releasing the forfeit reservation. The VTXO stays stuck in `Forfeiting` until the daemon restarts (where the #823 startup sweep can rescue it) or the batch expires. A single lost `SubmitVTXOForfeitSigs` message on the wire is enough to trigger it.

The notable part: the server behaves correctly the whole way through. It fires its input-signature collection timeout, fails the round, and delivers the failure notification to the client. The client receives it, transitions to `ClientFailed`, and still strands the coin. So this isn't a lost-notification problem, it's a missing release on the post-signing failure path.

## Root cause

`releaseForfeitsOnFailure` (`round/transitions.go`) is the #653 fix: it prepends a `ReleaseForfeitReservation` to any transition that lands in `ClientFailedState`. But it's only wired into the _procedural_ pre-signing handlers (admission timeout, quote reject, signing failures, etc). The post-signing failure path is a _declarative_ table entry in `round/transition_table.go` that never routes through it:

```go
// InputSigSentState: Waiting for commitment tx confirmation.
{
        FromState: &InputSigSentState{},
        Transitions: []ClientTransitionEntry{
                ...
                {
                        Event:       &BoardingFailed{},
                        ToState:     &ClientFailedState{},
                        Description: "Commitment tx failed to confirm",
                        IsTerminal:  true,
                },
```

No `EmitsOutbox`, so the vtxo manager is never told to release. The DST trace shows it plainly: the transition fires with zero outbox events.

```
[INF] temp(4ccdda1f) State transition from_state=ForfeitSignaturesCollecting to_state=InputSigSent
      ... SubmitVTXOForfeitSigs dropped in flight ...
[ERR] Round failed err="round failed: input signature collection timeout" round_id=...abf0381a
[DBG] Routing server message to pending round event_type=*round.BoardingFailed
[INF] temp(4ccdda1f) State transition from_state=InputSigSent to_state="ClientFailed: input signature collection timeout"
[...]

PR #1112, opened 2026-08-06
vtxo: do not exit a forfeit whose signature already left

### LINKED ISSUE #845: vtxo: stale Forfeiting VTXO triggers spurious standard_vtxo_timeout unilateral exit that wedges on CPFP funds
## Summary

A VTXO that is stale-`Forfeiting` (already cooperatively forfeited on-chain, but not reconciled to terminal locally — see #844) is still treated as a live, expiring coin by the critical-expiry watcher. As its `batch_expiry` approaches, the watcher fires `standard_vtxo_timeout` → `ForceUnroll`, moving the coin to `UnilateralExit` and starting a **unilateral exit of an already-forfeited VTXO**. The exit then wedges indefinitely because the client's on-chain wallet has no confirmed UTXOs to fund the CPFP fee.

This is a downstream amplification of the reconciliation gap in #844: the strand isn't just inert, it actively (and pointlessly) tries to go on-chain.

## Why it's wrong / dangerous

- **Pointless:** the coin was consumed by a cooperative refresh whose commitment already confirmed. There is nothing to recover.
- **Losing if it ever completes:** the operator holds a fully-signed forfeit for the coin (spending it into a connector of the confirmed refresh commitment). If this exit ever funded its CPFP and confirmed, it would race — and lose to — that forfeit. It only stays harmless because the wallet is empty.
- **Noisy / alarming:** it emits `operator intervention may be required` every ~15-20 min forever.

## Observed (signet, 2026-07-01)

VTXO `522270b9…2483:1` (194,744 sat), already refreshed into a live coin and long superseded. Its DB `vtxos.status` walked `Forfeiting → UnilateralExit`; a `unilateral_exit_jobs` row was created with `exit_policy_kind=standard_vtxo_timeout`, `trigger=standard_vtxo_timeout`.

```
TXCF: Tx broadcast repeatedly failing; operator intervention may be required
  err="cpfp fee input unavailable: confirmed cpfp fee input shortfall:
       no confirmed wallet UTXOs available (need >= 882 sats)"
  txid=97f342…  label=unroll-node-97f342…  consecutive_failures=15+
```

(`97f342…` is a tree node in `5222`'s exit ancestry, hence `unroll-node-…`.)

Full lineage context and the round/commitment details are in #844.

## Root cause

The critical-expiry watcher runs on every non-terminal VTXO. A coin stranded in `Forfeiting` by #844 qualifies, so it gets a `standard_vtxo_timeout` exit even though it's cooperatively forfeited. Two independent gaps compound:

1. The coin should have been reconciled to terminal `Forfeited` (that's #844).
2. Even absent (1), critical-expiry should not fire a `standard_vtxo_timeout` unilateral exit on a coin that is pending forfeit reconciliation / already has forfeit signatures out — exiting such a coin is never the right move (it races the operator's forfeit).

## Proposed fix

- Primary: reconcile cooperatively-forfeited coins to terminal `Forfeited` (via operator/indexer status lookup) so they leave the expiry watcher's set entirely — the #844 track.
- Defense-in-depth: suppress `standard_vtxo_timeout`/`ForceUnroll(critical-expiry)` for VTXOs in `Forfeiting`/`UnilateralExit`-from-forfeit whose forfeit signatures have already left the box. Critical-expiry exit is only sound for coins the client can still unilaterally claim.

## Related

- #844 — the reconciliation gap this escalates from (round success-after-restart or post-signing failure strands `Forfeiting`).
- #823 — startup sweep for pre-signing `PendingForfeit` orphans (does not touch `Forfeiting`).
- #687 — programmatic monitoring/alerting for stuck unilateral-exit broadcasts (this is a case that would trip it spuriously).
- lightninglabs/darepo#618 — server-side durable-round-actor / status-reconcile track.

ellemouton:
Fixes #845.

This is the **defence-in-depth** half of that issue. The primary fix — reconciling a cooperatively-forfeited coin to terminal `Forfeited` so it leaves the expiry watcher's set entirely — is #844's track and is **not** attempted here. What this removes is the escalation that turns that strand from inert into an on-chain attempt.

## The defect

`ForfeitingState.ProcessEvent` escalated to `UnilateralExitState` on `ExpiryStatusCritical` with no guard (`vtxo/transitions.go`, the `BlockEpochEvent` branch). A coin stranded in `Forfeiting` by #844 is still a non-terminal, expiring coin as far as the watcher is concerned, so as `batch_expiry` approached it started a unilateral exit of an already-forfeited VTXO — which then wedged forever on CPFP funding, as in the signet trace on the issue.

## Why it is wrong once the signature has left

This is the line the fix draws. Before the forfeit signature goes out, the client is the only party able to spend the coin, so a unilateral exit is a real recovery and still escalates. After it goes out, the operator holds a fully-signed transaction spending this coin into a connector of the new commitment and can broadcast it at will.

An exit started past that point must confirm the whole ancestry and then wait out the exit CSV, while the forfeit is spendable immediately — so it races a transaction that has already won. Pointless if it never funds its CPFP; **losing** if it ever does.

There is precedent for exactly this reasoning in the same file: `ExpiredState` already refuses `ForceUnrollEvent` because an exit past expiry "must confirm the whole ancestry and then wait out the exit CSV while racing an already-spendable operator sweep".

## How the line is drawn

`forfeitSignatureIssued` reads state the FSM already keeps — no new field:

- `ForfeitTxID` is stamped by `ForfeitSignedEvent` ("forfeit tx has been signed and submitted")
- `ForfeitTx` is the retained signed transaction, persisted for crash recovery

Either alone is sufficient evidence the signature exists, so both are checked.

## Both doors, not one

The chain-resolver bridge admits expiry-driven exits through `ForceUnrollEvent` as well as through the block-epoch path, so guarding only the latter would leave the identical doomed exit reachable by another door. `ForceUnrollEvent` is therefore guarded too — but **only** for `UnrollTriggerCriticalExpiry`:

- **manual** — an explicit decision, should not be silently dropped
- **fraud_spend** — the operator has already moved against us, and exiting is the correct response

Only the automatic trigger, which fires on a coin nobody chose to exit, is suppressed.

## Testing

`TestForfeitingStateCriticalExpiry` **previously asserted the behaviour this issue was filed about**: it seeded both `ForfeitTxID` and `ForfeitTx` — a signature that had demonstrably left — and required escalation. It now covers both sides of the line, so the case that used to be the bug is the case that pins the fix.

`TestForfeitingStateForceUnrollTriggers` covers the three triggers with the signature already sent, so the pass-through cases are asserted rather than assumed.

Verified that only the two suppression subtests fail without the guard while the escalate/pass-through subtests stay green — the tests pin the change rather than restating it.

`make unit pkg=vtxo` plus `unroll` and `round` (the packages that consume the exit path) pass, `make lint-changed-local` 0 issues, `make commitmsg-lint` OK.

## Note on CI

wavelength CI has not reported on recent branches — its last run was 18:15 UTC today on an unrelated branch. Both use self-hosted runners. Local package tests and lint are the only validation so far.

### DIFF
--- a/vtxo/transitions.go
+++ b/vtxo/transitions.go
@@ -8,9 +8,11 @@ import (
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil/v2"
+	"github.com/btcsuite/btcd/chainhash/v2"
 	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/lightninglabs/wavelength/build"
+	"github.com/lightninglabs/wavelength/lib/actormsg"
 	"github.com/lightninglabs/wavelength/lib/arkscript"
 	"github.com/lightninglabs/wavelength/lib/tx"
 	"github.com/lightninglabs/wavelength/lib/types"
@@ -981,6 +983,27 @@ func (s *PendingForfeitState) ProcessEvent(ctx context.Context, event VTXOEvent,
 	}
 }
 
+// forfeitSignatureIssued reports whether this VTXO's forfeit signature has
+// already left the client.
+//
+// This is the line between an exit that can still work and one that cannot.
+// Before the signature goes out, the client is the only party able to spend
+// the coin, so a unilateral exit is a real recovery. After it goes out, the
+// operator holds a fully-signed transaction spending this coin into a
+// connector of the new commitment and can broadcast it whenever it likes. An
+// exit started past that point has to confirm the whole ancestry and then wait
+// out the exit CSV, while the forfeit is spendable immediately -- so the exit
+// races it and loses.
+//
+// Both fields are checked because either one alone is enough to mean the
+// signature exists: ForfeitTxID is stamped by ForfeitSignedEvent, and ForfeitTx
+// is the retained signed transaction restored on crash recovery.
+func (s *ForfeitingState) forfeitSignatureIssued() bool {
+	var zeroTxID chainhash.Hash
+
+	return s.ForfeitTxID != zeroTxID || s.ForfeitTx != nil
+}
+
 // ProcessEvent handles events in ForfeitingState. The VTXO is being forfeited
 // and waiting for the new commitment transaction to confirm.
 func (s *ForfeitingState) ProcessEvent(ctx context.Context, event VTXOEvent,
@@ -1042,7 +1065,18 @@ func (s *ForfeitingState) ProcessEvent(ctx context.Context, event VTXOEvent,
 		// while racing an already-spendable operator sweep — and
 		// escalating would abort the in-flight cooperative spend that
 		// IS the recovery. Staying put lets it finish.
-		if expiryStatus == ExpiryStatusCritical {
+		// The same reasoning rules out escalating once the forfeit
+		// signature has left, whatever the deadline says: the operator
+		// can spend this coin into a connector immediately, so the
+		// exit races a transaction that has already won
+		// (wavelength#845). This is defence in depth against a coin
+		// stranded in Forfeiting by the reconciliation gap in
+		// wavelength#844 -- such a coin is not expiring at all, it is
+		// already spent, so exiting it is pointless, and losing if the
+		// exit ever funded its CPFP and confirmed.
+		if expiryStatus == ExpiryStatusCritical &&
+			!s.forfeitSignatureIssued() {
+
 			blocksRemaining := BlocksUntilExpiry(s.VTXO, evt.Height)
 
 			// Non-terminal exit: no VTXOTerminatedNotification, so
@@ -1085,6 +1119,28 @@ func (s *ForfeitingState) ProcessEvent(ctx context.Context, event VTXOEvent,
 		}, nil
 
 	case *ForceUnrollEvent:
+		// A critical-expiry admission is refused once the forfeit
+		// signature has left, for the same reason the block-epoch path
+		// above declines to escalate: the operator can spend this coin
+		// into a connector immediately, so the exit races a
+		// transaction that has already won (wavelength#845). The
+		// chain-resolver bridge admits expiry-driven exits through
+		// this event too, so guarding only the block-epoch branch
+		// would leave the same doomed exit reachable by another door.
+		//
+		// Manual and fraud-spend triggers are deliberately still
+		// honoured. A fraud spend means the operator has already moved
+		// against us and exiting is the correct response, and a manual
+		// unroll is an explicit operator decision that should not be
+		// silently dropped. Only the automatic path, which fires on a
+		// coin nobody chose to exit, is suppressed.
+		if evt.Trigger == actormsg.UnrollTriggerCriticalExpiry &&
+			s.forfeitSignatureIssued() {
+			return &VTXOStateTransition{
+				NextState: s,
+			}, nil
+		}
+
 		// Client requested unilateral exit while a forfeit is
 		// mid-flight. The on-chain recovery path doesn't depend
 		// on the forfeit signature landing, so we escalate to
--- a/vtxo/transitions_test.go
+++ b/vtxo/transitions_test.go
@@ -1111,47 +1111,89 @@ func signForfeitParticipantRequest(t *testing.T,
 func TestForfeitingStateCriticalExpiry(t *testing.T) {
 	t.Parallel()
 
-	h := newVTXOTestHarness(t)
-	vtxo := h.newTestDescriptor()
-	vtxo.BatchExpiry = 1000
+	// Critical expiry escalates a forfeit that is still mid-flight, and
+	// must NOT escalate one whose signature has already left. Past that
+	// point the operator holds a fully-signed spend of this coin into a
+	// connector and can broadcast whenever it likes, so the exit races a
+	// transaction that has already won: pointless if it never funds its
+	// CPFP, and losing if it does (wavelength#845).
+	//
+	// This test previously asserted the opposite. It seeded both
+	// ForfeitTxID and ForfeitTx -- a signature that had demonstrably left
+	// -- and required escalation, which is the behaviour the issue was
+	// filed about.
+	tests := []struct {
+		name           string
+		signatureSent  bool
+		wantEscalation bool
+	}{{
+		name:           "signature not yet sent escalates",
+		signatureSent:  false,
+		wantEscalation: true,
+	}, {
+		name:           "signature already sent stays put",
+		signatureSent:  true,
+		wantEscalation: false,
+	}}
 
-	h.withExpiryConfig(&ExpiryConfig{
-		RefreshThresholdBlocks:  200,
-		CriticalThresholdBlocks: 50,
-		TreeDepthMultiplier:     1,
-	})
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
 
-	forfeitTx := wire.NewMsgTx(2)
-	h.withState(&ForfeitingState{
-		VTXO:              vtxo,
-		NewRoundID:        "round-stalled",
-		ConnectorOutpoint: h.newTestOutpoint(),
-		ForfeitTxID:       forfeitTx.TxHash(),
-		ForfeitTx:         forfeitTx,
-	})
+			h := newVTXOTestHarness(t)
+			vtxo := h.newTestDescriptor()
+			vtxo.BatchExpiry = 1000
 
-	// Block height within critical threshold while waiting for forfeit
-	// confirmation. Should escalate to chain resolver.
-	evt := h.newBlockEpochEvent(970)
+			h.withExpiryConfig(&ExpiryConfig{
+				RefreshThresholdBlocks:  200,
+				CriticalThresholdBlocks: 50,
+				TreeDepthMultiplier:     1,
+			})
 
-	// Setup mock for status update.
-	h.store.On(
-		"UpdateVTXOStatus",
-		h.ctx,
-		vtxo.Outpoint,
-		VTXOStatusUnilateralExit,
-	).Return(nil)
+			state := &ForfeitingState{
+				VTXO:              vtxo,
+				NewRoundID:        "round-stalled",
+				ConnectorOutpoint: h.newTestOutpoint(),
+			}
+			if tc.signatureSent {
+				forfeitTx := wire.NewMsgTx(2)
+				state.ForfeitTxID = forfeitTx.TxHash()
+				state.ForfeitTx = forfeitTx
+			}
+			h.withState(state)
 
-	_, err := h.sendEvent(evt)
-	require.NoError(t, err)
+			if tc.wantEscalation {
+				h.store.On(
+					"UpdateVTXOStatus",
+					h.ctx,
+					vtxo.Outpoint,
+					VTXOStatusUnilateralExit,
+				).Return(nil)
+			}
 
-	assertState[*UnilateralExitState](h)
+			// Block height inside the critical threshold.
+			_, err := h.sendEvent(h.newBlockEpochEvent(970))
+			require.NoError(t, err)
 
-	// Should emit ExpiringNotification but NOT a
-	// VTXOTerminatedNotification: the exit is observed, not fire-and-forget
-	// (wavelength#602).
-	assertOutboxContains[*ExpiringNotification](h)
-	assertOutboxLacks[*VTXOTerminatedNotification](h)
+			if !tc.wantEscalation {
+				// Staying in Forfeiting is the whole point: the
+				// coin is not expiring, it is spent, and the
+				// reconciliation that retires it is #844's job.
+				assertState[*ForfeitingState](h)
+				assertOutboxLacks[*ExpiringNotification](h)
+
+				return
+			}
+
+			assertState[*UnilateralExitState](h)
+
+			// Emits ExpiringNotification but NOT a
+			// VTXOTerminatedNotification: the exit is observed,
+			// not fire-and-forget (wavelength#602).
+			assertOutboxContains[*ExpiringNotification](h)
+			assertOutboxLacks[*VTXOTerminatedNotification](h)
+		})
+	}
 }
 
 // TestForfeitSignedEventPreservesForfeitTx verifies that handling
@@ -1860,4 +1902,83 @@ func TestForfeitSignatureValidity(t *testing.T) {
 
 	err = engine.Execute()
 	require.NoError(t, err, "VTXO input signature verification failed")
+}
+
+// TestForfeitingStateForceUnrollTriggers pins which unilateral-exit triggers
+// survive a forfeit whose signature has already left.
+//
+// The chain-resolver bridge admits expiry-driven exits through ForceUnrollEvent
+// as well as through the block-epoch path, so guarding only the latter would
+// leave the same doomed exit reachable by another door (wavelength#845).
+//
+// Manual and fraud-spend stay honoured on purpose. A fraud spend means the
+// operator has already moved against us and exiting is the correct response; a
+// manual unroll is an explicit decision that should not be silently dropped.
+// Only the automatic trigger, which fires on a coin nobody chose to exit, is
+// suppressed.
+func TestForfeitingStateForceUnrollTriggers(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name           string
+		trigger        actormsg.UnrollTrigger
+		wantEscalation bool
+	}{{
+		name:           "critical expiry suppressed",
+		trigger:        actormsg.UnrollTriggerCriticalExpiry,
+		wantEscalation: false,
+	}, {
+		name:           "manual still honoured",
+		trigger:        actormsg.UnrollTriggerManual,
+		wantEscalation: true,
+	}, {
+		name:           "fraud spend still honoured",
+		trigger:        actormsg.UnrollTriggerFraudSpend,
+		wantEscalation: true,
+	}}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			h := newVTXOTestHarness(t)
+			vtxo := h.newTestDescriptor()
+
+			// The signature has left in every case here; the
+			// trigger is the only variable.
+			forfeitTx := wire.NewMsgTx(2)
+			h.withState(&ForfeitingState{
+				VTXO:              vtxo,
+				NewRoundID:        "round-stalled",
+				ConnectorOutpoint: h.newTestOutpoint(),
+				ForfeitTxID:       forfeitTx.TxHash(),
+				ForfeitTx:         forfeitTx,
+			})
+
+			if tc.wantEscalation {
+				h.store.On(
+					"UpdateVTXOStatus",
+					h.ctx,
+					vtxo.Outpoint,
+					VTXOStatusUnilateralExit,
+				).Return(nil)
+			}
+
+			_, err := h.sendEvent(&ForceUnrollEvent{
+				Reason:  "test",
+				Trigger: tc.trigger,
+			})
+			require.NoError(t, err)
+
+			if !tc.wantEscalation {
+				assertState[*ForfeitingState](h)
+				assertOutboxLacks[*ExpiringNotification](h)
+
+				return
+			}
+
+			assertState[*UnilateralExitState](h)
+			assertOutboxContains[*ExpiringNotification](h)
+		})
+	}
 }


ellemouton:

litbot-9000:
This PR changes Go code in `vtxo`, and that package's `CLAUDE.md`/`AGENTS.md` pair is now stale on the new `ForfeitingState` exit guard — proposed doc update below.

```diff
diff --git a/vtxo/AGENTS.md b/vtxo/AGENTS.md
index 46bb45fa..810f1ae9 100644
--- a/vtxo/AGENTS.md
+++ b/vtxo/AGENTS.md
@@ -163,7 +163,28 @@ when the local wallet owns the receive script.
 - **Startup sweep of orphaned forfeits.** `Start` calls `releaseOrphanedForfeits` after actor recovery. Any VTXO still in `VTXOStatusPendingForfeit` is provably orphaned and returns to `LiveState`. A `Forfeiting` VTXO can also be stranded before signature handoff because actors persist their signed transaction while the round is still collecting peers. The manager releases it only when `forfeit_round_id` is present and the round store explicitly reports no durable checkpoint; a missing binding, lookup error, or existing round row fails closed and retains `Forfeiting`.
 - **Confirmed-forfeit startup reconcile.** The serialized `ReconcileExpiry` manager turn runs after the round actor starts. It lists recovered `Forfeiting` descriptors and re-drives `ForfeitConfirmedEvent` only when the joined `Settlement` has `Height > 0`, closing a crash between round finalization and the original best-effort confirmation Tell. A missing settlement, height zero, lookup error, missing actor, or actor error fails closed in `Forfeiting`.
 - **Atomic reservation cleanup.** `VTXOStore.UpdateVTXOStatusReleasingReservation` deletes the spending-reservation row in the same transaction as the VTXO status change when a VTXO leaves `SpendingState` (via `SpendReleasedEvent`, `SpendCompletedEvent`, or escalation to `UnilateralExitState`). This prevents the durable index from retaining stale rows that would mask a future orphan on the same outpoint.
-- `ForceUnrollEvent` unifies every unilateral-exit trigger (manual `Unroll` RPC, fraud spend, vHTLC recovery) behind the manager's admission gate. It carries a `Trigger actormsg.UnrollTrigger` (zero value admits as critical expiry) and an `ExitPolicy fn.Option[actormsg.ExitPolicy]` (None selects the standard VTXO timeout policy); both ride through to the emitted `ExpiringNotification` so the chain-resolver bridge admits the registry job under the right `StartTrigger` and persists the correct exit-spend policy. It is accepted in `LiveState`, `PendingForfeitState`, `SpendingState`, and `ForfeitingState`: each transitions to `UnilateralExitState` and emits `ExpiringNotification` (trigger + exit policy threaded through) + `VTXOStatusUpdate{UnilateralExit}`. It does **not** emit `VTXOTerminatedNotification` on intent — `UnilateralExitState` is **non-terminal**, so the actor stays alive to observe the exit. Truly terminal states (`Spent`, `Forfeited`, `Failed`) self-loop; the manager maps that self-loop back to `ForceUnrollResponse{Accepted: false, Reason: "already terminal"}`. A `ForceUnrollEvent` on a VTXO already in `UnilateralExitState` is an idempotent re-admission, not a no-op: the actor stays in `UnilateralExitState`, does not re-persist the status, and **re-emits** the `ExpiringNotification` under the same trigger/policy so the chain-resolver bridge re-admits the job (the first admission's best-effort Tell can be lost to a crash before the registry writes its record; the registry dedups against a live record, so a redundant re-admit is a benign no-op).
+- `ForceUnrollEvent` unifies every unilateral-exit trigger (manual `Unroll` RPC, fraud spend, vHTLC recovery) behind the manager's admission gate. It carries a `Trigger actormsg.UnrollTrigger` (zero value admits as critical expiry) and an `ExitPolicy fn.Option[actormsg.ExitPolicy]` (None selects the standard VTXO timeout policy); both ride through to the emitted `ExpiringNotification` so the chain-resolver bridge admits the registry job under the right `StartTrigger` and persists the correct exit-spend policy. It is accepted in `LiveState`, `PendingForfeitState`, `SpendingState`, and `ForfeitingState` (with the signature-issued exception below): each transitions to `UnilateralExitState` and emits `ExpiringNotification` (trigger + exit policy threaded through) + `VTXOStatusUpdate{UnilateralExit}`. It does **not** emit `VTXOTerminatedNotification` on intent — `UnilateralExitState` is **non-terminal**, so the actor stays alive to observe the exit. Truly terminal states (`Spent`, `Forfeited`, `Failed`) self-loop; the manager maps that self-loop back to `ForceUnrollResponse{Accepted: false, Reason: "already terminal"}`. A `ForceUnrollEvent` on a VTXO already in `UnilateralExitState` is an idempotent re-admission, not a no-op: the actor stays in `UnilateralExitState`, does not re-persist the status, and **re-emits** the `ExpiringNotification` under the same trigger/policy so the chain-resolver bridge re-admits the job (the first admission's best-effort Tell can be lost to a crash before the registry writes its record; the registry dedups against a live record, so a redundant re-admit is a benign no-op).
+- **A forfeit whose signature already left is never auto-exited.**
+  `ForfeitingState.forfeitSignatureIssued()` reports whether this VTXO's forfeit
+  signature has reached the operator — true when either `ForfeitTxID` is set
+  (stamped by `ForfeitSignedEvent`) or `ForfeitTx` is retained (restored on
+  crash recovery). Once it is true the operator holds a fully-signed
+  transaction spending the coin into a connector and can broadcast it
+  immediately, so an exit would have to confirm the whole ancestry and wait out
+  the exit CSV against a transaction that has already won. Both automatic exit
+  doors in `ForfeitingState` are therefore closed on that condition: the
+  `BlockEpochEvent` path escalates on `ExpiryStatusCritical` only while the
+  signature has not been issued, and a `ForceUnrollEvent` carrying
+  `actormsg.UnrollTriggerCriticalExpiry` (the empty-string zero value, so an
+  unset trigger counts) self-loops instead of escalating. The `ForceUnrollEvent`
+  guard matters independently because the chain-resolver bridge admits
+  expiry-driven exits through that event rather than the block epoch. Manual and
+  fraud-spend triggers are still honoured: a fraud spend means the operator has
+  already moved against us, and a manual unroll is an explicit decision that
+  must not be silently dropped. The self-loop leaves a non-terminal, non-expired
+  state, so `handleForceUnroll` still answers `ForceUnrollResponse{Accepted:
+  true}` — only the terminal and `ExpiredState` self-loops carry an explicit
+  refusal Reason.
 - `UnilateralExitState` is **non-terminal** and observed, not fire-and-forget. The actor survives until the unroll job reports a terminal outcome via the manager's `ExitOutcomeNotification`: `ExitOutcomeRecoverable` (the unroll failed with no on-chain footprint) drives `ExitFailedEvent` → `LiveState` + `VTXOStatusUpdate{Live}`, while `ExitOutcomeConfirmed` (the exit confirmed on-chain) drives `ExitConfirmedEvent` → terminal `SpentState` + `VTXOTerminatedNotification` (the actor is reaped here, gated on a terminal on-chain event rather than the user's intent). A recoverable failure of a **recovery-only** target (`ExitOutcomeNotification.ExitPolicyKind.Valid()`, e.g. a vHTLC refund) is the exception: the manager holds the coin in `UnilateralExitState` rather than reliving it, since it is a swap-contract output, not spendable wallet liquidity, and reliving it would inflate balance and feed coin selection and sweep-all. This guard short-circuits before any store access; the owning recovery subsystem is responsible for retrying or terminal-failing the refund. When the actor is absent (e.g. a daemon restart, since exiting VTXOs are excluded from `ListLiveVTXOs` recovery) the manager re-materializes a live actor from the persisted descriptor (recover) or persists `VTXOStatusSpent` directly (confirm).
 - `Manager.handleForceUnroll` uses `Ask` (not `Tell`) so FSM errors and self-loop no-ops surface as structured `ForceUnrollResponse{Accepted, Reason}` instead of a uniform `Accepted:true` that masks work that was never scheduled.
 - When `handleForceUnroll` targets an outpoint with no live actor, `spawnForceUnrollActor` re-materializes an actor from the persisted descriptor so the manager still owns the exit rather than letting the caller admit the unroll behind its back. This is the common shape for the vHTLC recovery target (materialized directly in the store, never admitted through the manager) and any exiting VTXO a restart left out of the live-recovery set. It guards both ends: a missing descriptor returns `ForceUnrollResponse{Accepted: false, Reason: "no such vtxo"}`, and an already-terminal descriptor (`statusToState(...).IsTerminal()`) returns `Reason: "already terminal"` — neither spawns an actor that would immediately reap itself.
diff --git a/vtxo/CLAUDE.md b/vtxo/CLAUDE.md
index 46bb45fa..810f1ae9 100644
--- a/vtxo/CLAUDE.md
+++ b/vtxo/CLAUDE.md
@@ -163,7 +163,28 @@ when the local wallet owns the receive script.
 - **Startup sweep of orphaned forfeits.** `Start` calls `releaseOrphanedForfeits` after actor recovery. Any VTXO still in `VTXOStatusPendingForfeit` is provably orphaned and returns to `LiveState`. A `Forfeiting` VTXO can also be stranded before signature handoff because actors persist their signed transaction while the round is still collecting peers. The manager releases it only when `forfeit_round_id` is present and the round store explicitly reports no durable checkpoint; a missing binding, lookup error, or existing round row fails closed and retains `Forfeiting`.
 - **Confirmed-forfeit startup reconcile.** The serialized `ReconcileExpiry` manager turn runs after the round actor starts. It lists recovered `Forfeiting` descriptors and re-drives `ForfeitConfirmedEvent` only when the joined `Settlement` has `Height > 0`, closing a crash between round finalization and the original best-effort confirmation Tell. A missing settlement, height zero, lookup error, missing actor, or actor error fails closed in `Forfeiting`.
 - **Atomic reservation cleanup.** `VTXOStore.UpdateVTXOStatusReleasingReservation` deletes the spending-reservation row in the same transaction as the VTXO status change when a VTXO leaves `SpendingState` (via `SpendReleasedEvent`, `SpendCompletedEvent`, or escalation to `UnilateralExitState`). This prevents the durable index from retaining stale rows that would mask a future orphan on the same outpoint.
-- `ForceUnrollEvent` unifies every unilateral-exit trigger (manual `Unroll` RPC, fraud spend, vHTLC recovery) behind the manager's admission gate. It carries a `Trigger actormsg.UnrollTrigger` (zero value admits as critical expiry) and an `ExitPolicy fn.Option[actormsg.ExitPolicy]` (None selects the standard VTXO timeout policy); both ride through to the emitted `ExpiringNotification` so the chain-resolver bridge admits the registry job under the right `StartTrigger` and persists the correct exit-spend policy. It is accepted in `LiveState`, `PendingForfeitState`, `SpendingState`, and `ForfeitingState`: each transitions to `UnilateralExitState` and emits `ExpiringNotification` (trigger + exit policy threaded through) + `VTXOStatusUpdate{UnilateralExit}`. It does **not** emit `VTXOTerminatedNotification` on intent — `UnilateralExitState` is **non-terminal**, so the actor stays alive to observe the exit. Truly terminal states (`Spent`, `Forfeited`, `Failed`) self-loop; the manager maps that self-loop back to `ForceUnrollResponse{Accepted: false, Reason: "already terminal"}`. A `ForceUnrollEvent` on a VTXO already in `UnilateralExitState` is an idempotent re-admission, not a no-op: the actor stays in `UnilateralExitState`, does not re-persist the status, and **re-emits** the `ExpiringNotification` under the same trigger/policy so the chain-resolver bridge re-admits the job (the first admission's best-effort Tell can be lost to a crash before the registry writes its record; the registry dedups against a live record, so a redundant re-admit is a benign no-op).
+- `ForceUnrollEvent` unifies every unilateral-exit trigger (manual `Unroll` RPC, fraud spend, vHTLC recovery) behind the manager's admission gate. It carries a `Trigger actormsg.UnrollTrigger` (zero value admits as critical expiry) and an `ExitPolicy fn.Option[actormsg.ExitPolicy]` (None selects the standard VTXO timeout policy); both ride through to the emitted `ExpiringNotification` so the chain-resolver bridge admits the registry job under the right `StartTrigger` and persists the correct exit-spend policy. It is accepted in `LiveState`, `PendingForfeitState`, `SpendingState`, and `ForfeitingState` (with the signature-issued exception below): each transitions to `UnilateralExitState` and emits `ExpiringNotification` (trigger + exit policy threaded through) + `VTXOStatusUpdate{UnilateralExit}`. It does **not** emit `VTXOTerminatedNotification` on intent — `UnilateralExitState` is **non-terminal**, so the actor stays alive to observe the exit. Truly terminal states (`Spent`, `Forfeited`, `Failed`) self-loop; the manager maps that self-loop back to `ForceUnrollResponse{Accepted: false, Reason: "already terminal"}`. A `ForceUnrollEvent` on a VTXO already in `UnilateralExitState` is an idempotent re-admission, not a no-op: the actor stays in `UnilateralExitState`, does not re-persist the status, and **re-emits** the `ExpiringNotification` under the same trigger/policy so the chain-resolver bridge re-admits the job (the first admission's best-effort Tell can be lost to a crash before the registry writes its record; the registry dedups against a live record, so a redundant re-admit is a benign no-op).
+- **A forfeit whose signature already left is never auto-exited.**
+  `ForfeitingState.forfeitSignatureIssued()` reports whether this VTXO's forfeit
+  signature has reached the operator — true when either `ForfeitTxID` is set
+  (stamped by `ForfeitSignedEvent`) or `ForfeitTx` is retained (restored on
+  crash recovery). Once it is true the operator holds a fully-signed
+  transaction spending the coin into a connector and can broadcast it
+  immediately, so an exit would have to confirm the whole ancestry and wait out
+  the exit CSV against a transaction that has already won. Both automatic exit
+  doors in `ForfeitingState` are therefore closed on that condition: the
+  `BlockEpochEvent` path escalates on `ExpiryStatusCritical` only while the
+  signature has not been issued, and a `ForceUnrollEvent` carrying
+  `actormsg.UnrollTriggerCriticalExpiry` (the empty-string zero value, so an
+  unset trigger counts) self-loops instead of escalating. The `ForceUnrollEvent`
+  guard matters independently because the chain-resolver bridge admits
+  expiry-driven exits through that event rather than the block epoch. Manual and
+  fraud-spend triggers are still honoured: a fraud spend means the operator has
+  already moved against us, and a manual unroll is an explicit decision that
+  must not be silently dropped. The self-loop leaves a non-terminal, non-expired
+  state, so `handleForceUnroll` still answers `ForceUnrollResponse{Accepted:
+  true}` — only the terminal and `ExpiredState` self-loops carry an explicit
+  refusal Reason.
 - `UnilateralExitState` is **non-terminal** and observed, not fire-and-forget. The actor survives until the unroll job reports a terminal outcome via the manager's `ExitOutcomeNotification`: `ExitOutcomeRecoverable` (the unroll failed with no on-chain footprint) drives `ExitFailedEvent` → `LiveState` + `VTXOStatusUpdate{Live}`, while `ExitOutcomeConfirmed` (the exit confirmed on-chain) drives `ExitConfirmedEvent` → terminal `SpentState` + `VTXOTerminatedNotification` (the actor is reaped here, gated on a terminal on-chain event rather than the user's intent). A recoverable failure of a **recovery-only** target (`ExitOutcomeNotification.ExitPolicyKind.Valid()`, e.g. a vHTLC refund) is the exception: the manager holds the coin in `UnilateralExitState` rather than reliving it, since it is a swap-contract output, not spendable wallet liquidity, and reliving it would inflate balance and feed coin selection and sweep-all. This guard short-circuits before any store access; the owning recovery subsystem is responsible for retrying or terminal-failing the refund. When the actor is absent (e.g. a daemon restart, since exiting VTXOs are excluded from `ListLiveVTXOs` recovery) the manager re-materializes a live actor from the persisted descriptor (recover) or persists `VTXOStatusSpent` directly (confirm).
 - `Manager.handleForceUnroll` uses `Ask` (not `Tell`) so FSM errors and self-loop no-ops surface as structured `ForceUnrollResponse{Accepted, Reason}` instead of a uniform `Accepted:true` that masks work that was never scheduled.
 - When `handleForceUnroll` targets an outpoint with no live actor, `spawnForceUnrollActor` re-materializes an actor from the persisted descriptor so the manager still owns the exit rather than letting the caller admit the unroll behind its back. This is the common shape for the vHTLC recovery target (materialized directly in the store, never admitted through the manager) and any exiting VTXO a restart left out of the live-recovery set. It guards both ends: a missing descriptor returns `ForceUnrollResponse{Accepted: false, Reason: "no such vtxo"}`, and an already-terminal descriptor (`statusToState(...).IsTerminal()`) returns `Reason: "already terminal"` — neither spawns an actor that would immediately reap itself.
```

**How to apply:** save the diff above and `git apply` it, or run the
`doc-gardening` skill locally scoped to this package (`/doc-gardening vtxo`)
and commit the result.

[doc-gardening advisory run](https://github.com/lightninglabs/wavelength/actions/runs/31430680320)

ellemouton:

vtxo/transitions.go
```diff
@@ -1085,6 +1119,28 @@ func (s *ForfeitingState) ProcessEvent(ctx context.Context, event VTXOEvent,
...
+		// this event too, so guarding only the block-epoch branch
+		// would leave the same doomed exit reachable by another door.
+		//
+		// Manual and fraud-spend triggers are deliberately still
+		// honoured. A fraud spend means the operator has already moved
+		// against us and exiting is the correct response, and a manual
+		// unroll is an explicit operator decision that should not be
+		// silently dropped. Only the automatic path, which fires on a
+		// coin nobody chose to exit, is suppressed.
+		if evt.Trigger == actormsg.UnrollTriggerCriticalExpiry &&
```
litbot-9000, line 1137: 🔴 **blocker** — This guard suppresses the manual RPC unroll, which is exactly the path the comment three lines above says is "deliberately still honoured".

`UnrollTriggerCriticalExpiry` is the **zero value** (`lib/actormsg/vtxo_admission.go:357`: `UnrollTriggerCriticalExpiry UnrollTrigger = ""`), and the manual RPC handler never sets `Trigger`:

```go
// waved/rpc_server.go:4923, in func (r *RPCServer) Unroll
resp, askErr := vtxoMgrRef.Ask(
        admissionCtx, &actormsg.ForceUnrollRequest{
                Outpoint: outpoint,
                Reason:   "manual RPC request",
        },
).Await(admissionCtx).Unpack()
```

`Manager.handleForceUnroll` forwards `req.Trigger` verbatim into the `ForceUnrollEvent` (`vtxo/manager.go:1038-1042`), so a hand-typed `wavecli unroll` on a forfeit-signature-issued VTXO arrives here with `Trigger == ""` and takes this branch. The only two callers that set a non-zero trigger are `vhtlcrecovery/coordinator/service.go:476` and `fraud/actor.go:497` — the subsystem paths — while the actual operator-facing path is the one that gets dropped.

It also fails **silently and dishonestly**. The self-loop returns `NextState: s`; back in `handleForceUnroll`, `ForfeitingState` is neither terminal nor `*ExpiredState`, so neither refusal branch fires and the manager falls through to `return ... &ForceUnrollResponse{Accepted: true}` (`vtxo/manager.go:1104-1106`). `RPCServer.Unroll` maps that straight onto `UnrollResponse{Created: true}`. The operator is told the exit was created; no job exists, and `GetUnrollStatus` will never show progress. For a change whose whole purpose is to stop a stranded coin from misbehaving, that is the worst possible failure shape.

The new test does not catch it because `TestForfeitingStateForceUnrollTriggers` passes `actormsg.UnrollTriggerManual` explicitly — a value no `waved` manual caller actually produces. It pins the intent, not the wiring.

This is a merge-ordering problem more than a design problem: #1113 carries a `waved: record manual unrolls under the manual trigger` commit that sets `Trigger: actormsg.UnrollTriggerManual` at `waved/rpc_server.go:4936`, which makes this guard behave as documented. **#1112 is not safe to land on main by itself.** Either land it after (or squashed with) that commit, or cherry-pick the one-line `Trigger:` fix into this PR.

vtxo/transitions.go
```diff
@@ -1042,7 +1065,18 @@ func (s *ForfeitingState) ProcessEvent(ctx context.Context, event VTXOEvent,
...
+		// The same reasoning rules out escalating once the forfeit
+		// signature has left, whatever the deadline says: the operator
+		// can spend this coin into a connector immediately, so the
+		// exit races a transaction that has already won
+		// (wavelength#845). This is defence in depth against a coin
+		// stranded in Forfeiting by the reconciliation gap in
+		// wavelength#844 -- such a coin is not expiring at all, it is
+		// already spent, so exiting it is pointless, and losing if the
+		// exit ever funded its CPFP and confirmed.
+		if expiryStatus == ExpiryStatusCritical &&
```
litbot-9000, line 1077: 🟠 **major** — The predicate is one step to the left of the fact that actually matters, and the gap is a fund-loss window rather than a wasted-fee window.

The justification in the commit message is "the operator holds a fully-signed transaction spending this coin into a connector of the new commitment and can broadcast it whenever it likes." That last clause is not true in general, and #844 says so in its own words:

> The saving grace is that a forfeit tx spends a connector output of the round's commitment tx. A round that failed before broadcast has no commitment, hence the forfeit can never confirm.

`ConnectorOutpoint` is documented the same way here — `vtxo/states.go:137`: "the connector from the new commitment tx". So `forfeitSignatureIssued()` collapses two states that need opposite handling:

- **commitment confirmed, client missed the event** (#845's signet trace, and #844's restart-during-a-successful-round variant) — the coin really is gone, the exit is pointless, suppression is right;
- **commitment never broadcast** (#844's primary DST repro: round fails *after* `SubmitVTXOForfeitSigs`) — the forfeit is permanently unbroadcastable, the coin is still wholly the client's, and the unilateral exit is a real recovery.

In the second case this change removes the last automatic escape. To be fair to the PR, it is not the only escape in general: #844's reconcile landed and is the designed recovery — `InputSigSentState` parks the failure and probes `QueryRoundStatus`, releasing the reservation on an authoritative dead answer (`round/transition_table.go:295-303`, `round/transitions.go:4676-4742`). But that recovery is explicitly *not* guaranteed to terminate, by the code's own admission:

- `round/transitions.go:4660-4665` — any non-dead answer self-loops and holds the reservation, "the re-armed reconcile timeout keeps probing if nothing resolves";
- `round/fsm_timeouts.go:44-48` — "An operator that predates the `QueryRoundStatus` RPC never answers", so the client "keeps converging on a bounded, low-rate probe cadence for as long as the reservation is parked";
- `Manager.releasePreCheckpointForfeits` (`vtxo/manager.go:2491+`) releases only when storage *positively proves* the round never reached the `InputSigSent` checkpoint — precisely excluding this case, since the signature did leave.

So for an operator that is gone, unreachable, or too old to answer, the coin now parks in `Forfeiting` with no automatic exit at all and rides to `batch_expiry`, where the operator sweeps the shared output. The trade-off is asymmetric and points the other way: leaving the escalation in costs wasted CPFP fees when the coin really was forfeited, while removing it costs the entire VTXO when it was not.

Combined with the manual-trigger defect above, #1112 landing alone leaves that coin with *no* escape hatch at all, automatic or human — which is why I'd hold this until the trigger fix is in front of it.

The durable fact you want is "the new commitment tx exists on chain", not "a signature left". `NewRoundID` and `ConnectorOutpoint` are both in hand here; gating on a confirmed/known commitment (or on the same round-status reconcile #844 already added) would suppress #845's case without opening this one. If that is out of scope for a defence-in-depth patch, please say so explicitly in the commit message and in the code comment — right now both assert the operator can always broadcast, which is the specific claim #844 refutes.

vtxo/transitions.go
```diff
@@ -1085,6 +1119,28 @@ func (s *ForfeitingState) ProcessEvent(ctx context.Context, event VTXOEvent,
...
+		// Manual and fraud-spend triggers are deliberately still
+		// honoured. A fraud spend means the operator has already moved
+		// against us and exiting is the correct response, and a manual
+		// unroll is an explicit operator decision that should not be
+		// silently dropped. Only the automatic path, which fires on a
+		// coin nobody chose to exit, is suppressed.
+		if evt.Trigger == actormsg.UnrollTriggerCriticalExpiry &&
+			s.forfeitSignatureIssued() {
+			return &VTXOStateTransition{
+				NextState: s,
```
litbot-9000, line 1140: 🟡 **minor** — Both suppression branches are completely silent — no log, no notification, no status change. A coin that is genuinely stranded now sits in `Forfeiting` until `batch_expiry` with nothing emitted at any point.

That is a regression in observability, even granting the change is right. The pre-PR behaviour was noisy in a bad way (#845: `operator intervention may be required` every 15-20 minutes forever) but it *did* surface the strand; that log line is how the issue was found in the first place. Silence is the wrong replacement for a spurious alarm — the right one is a quiet, specific signal.

The file already has the pattern and the precedent for exactly this shape. `LiveState`'s `ExpiryStatusUnknown` branch (`vtxo/transitions.go:294-300`) self-loops and warns because it is "a persistent data fault worth a warning until an operator notices" — a VTXO suppressed at critical expiry because its forfeit signature left is the same category:

```go
build.LoggerFromContext(ctx).WithPrefix(Subsystem).WarnS(
        ctx, "Critical-expiry exit suppressed: forfeit signature "+
                "already issued", nil,
        slog.String("outpoint", s.VTXO.Outpoint.String()),
        slog.String("round_id", string(s.NewRoundID)),
        slog.String("forfeit_txid", s.ForfeitTxID.String()),
        slog.Int("batch_expiry", int(s.VTXO.BatchExpiry)),
)
```

Log the round id: it is the handle an operator needs to ask the server whether that round is actually dead, which is the one question that resolves the case. Worth adding at both suppression sites (here and the `BlockEpochEvent` branch), rate-limited or on the transition edge if per-block noise is a concern.

litbot-9000 (changes requested):
FINAL — The core judgement in this PR is right: a coin whose forfeit really was consumed should not be started on a unilateral exit, and the block-epoch branch genuinely was unguarded. The `forfeitSignatureIssued()` predicate reads only state the FSM already keeps, and the rewritten `TestForfeitingStateCriticalExpiry` does invert a test that had been asserting the bug. I verified the author's testing claim rather than taking it: with both guards removed, exactly the two suppression subtests fail (`TestForfeitingStateCriticalExpiry/signature_already_sent_stays_put`, `TestForfeitingStateForceUnrollTriggers/critical_expiry_suppressed`) and the escalate/pass-through subtests stay green. `go test ./vtxo/` passes clean on the head commit, and CI is now fully green across all 20 checks — the PR body's note about CI being down is stale.

Two problems, one of which decides merge order.

**The blocker is a wiring defect, not a design one.** `UnrollTriggerCriticalExpiry` is the zero value (`lib/actormsg/vtxo_admission.go:357`), and `RPCServer.Unroll` builds its `ForceUnrollRequest` without setting `Trigger` (`waved/rpc_server.go:4922-4926`). So the manual RPC unroll — the one path the new comment singles out as "deliberately still honoured" — is precisely what the new `ForceUnrollEvent` guard swallows. Worse, it swallows it while reporting success: `ForfeitingState` is neither terminal nor `ExpiredState`, so `handleForceUnroll` falls through to `Accepted: true` (`vtxo/manager.go:1104-1106`) and the RPC returns `Created: true` for a job that was never scheduled. The new test misses this because it passes `UnrollTriggerManual` explicitly, a value no `waved` caller produces. #1113's `waved: record manual unrolls under the manual trigger` commit fixes it at `waved/rpc_server.go:4936`. **#1112 is therefore not safe to land on main by itself** — sequence it behind that commit or cherry-pick the one-liner in.

**The design concern is that the predicate is one step to the left of the fact that matters.** A forfeit tx spends a connector output of the *new commitment*, so "the operator can broadcast it whenever it likes" only holds once that commitment exists — #844 states this explicitly ("A round that failed before broadcast has no commitment, hence the forfeit can never confirm"), and `vtxo/states.go:137` documents `ConnectorOutpoint` the same way. `forfeitSignatureIssued()` merges the confirmed-commitment case (#845, suppression correct) with the failed-round case (#844's primary DST repro, where the coin is still wholly the client's and the exit is real recovery). I checked whether #844's landed reconcile closes that hole and it mostly does — `InputSigSentState` parks and probes `QueryRoundStatus`, releasing on an authoritative dead answer — but the code itself documents that it need not terminate: a non-dead answer self-loops indefinitely (`round/transitions.go:4660-4665`) and "an operator that predates the QueryRoundStatus RPC never answers" (`round/fsm_timeouts.go:44-48`), while `releasePreCheckpointForfeits` deliberately excludes post-checkpoint coins. Against a gone or too-old operator the coin now parks in `Forfeiting` with no automatic escape and rides to `batch_expiry`. The trade is asymmetric the other way: leaving the escalation in costs wasted CPFP fees; removing it can cost the whole VTXO. Combined with the blocker above, #1112 alone leaves such a coin no escape at all, automatic or human.

On the earlier interim state I inherited: H1 is confirmed and I re-derived it from source. H2 is confirmed but was overstated as written — the guard is not removing "the only recovery" in general, since #844's reconcile shipped and handles the honest-operator case; the residual hole is specifically the never-answers operator. I have scoped the comment accordingly.

Two smaller notes that don't need their own thread. First, the suppression is entirely silent at both sites — worth a `WarnS` carrying the round id, following the precedent at `vtxo/transitions.go:294`; details inline. Second, a scope observation: coins that already walked `Forfeiting → UnilateralExit` under the old behaviour (#845's signet coin among them) are not rescued here, and `RPCServer.Unroll` short-circuits on `VTXOStatusUnilateralExit` before reaching the manager, so an operator cannot re-drive them either. That is consistent with the PR's stated defence-in-depth scope, but it means the reported symptom on existing daemons still needs the #844 track or hand intervention — worth a sentence in the PR description so nobody reads this as fixing the observed incident.

The direction is sound and the test work is careful. Fix the trigger wiring (or land behind #1113), and say plainly in the comment which of the two strand causes this guard is and isn't safe for.

ellemouton:
This review is right, and it goes further than I had. **My recommendation is to close this PR in favour of #1113.**

**Your main point is the decisive one, and I had not spotted it.** `forfeitSignatureIssued()` is `true` from state *entry* in every live flow — `LiveState.handleForfeitRequest` (`transitions.go:450-451`) and `PendingForfeitState` → `ForfeitRequestEvent` (`:921-922`) both populate `ForfeitTxID` and `ForfeitTx` before `ForfeitSignedEvent` ever arrives. Two consequences follow, both bad for this PR:

1. The `BlockEpochEvent` critical-expiry branch is **effectively dead** in the live flow, reachable only for a crash-recovered state whose forfeit tx failed to load. So my "signature not yet sent escalates" subtest pins a shape the live flow never produces — the test passes and proves less than it appears to.
2. My docstring claims the fields witness the signature having "reached the operator". They do not. They witness *local signing and handoff to the round actor*; the operator cannot broadcast until `SubmitVTXOForfeitSigsToServer` on the `ForfeitSignaturesCollecting` success edge. The framing is more precise than the code can detect, exactly as you say.

**And the stalled-round case you ask about is the reason to close rather than soften the docstring.** A round that stalls mid-collection and neither completes, fails, nor emits `ForfeitReleasedEvent` reaches critical expiry with the operator holding nothing — so the forfeit is unspendable and a unilateral exit is both safe and the only recovery. Pre-PR that coin escalated; post-PR it stays put. That is a permanent lockout past batch expiry, which is strictly worse than the pointless-exit bug #845 was filed about.

#1113 reaches the opposite arrangement — refuse `UnrollTriggerManual`, keep critical expiry escalating — and states the reason explicitly: critical expiry is the safety net that stops a stalled round holding a coin past its batch expiry. That is correct where this PR is wrong, and it is not a difference that can be patched here, because *both* of this PR's halves rest on the same too-coarse discriminator. The real one is whether the commitment confirmed, which is what wavelength#844's reconciliation establishes and neither PR has.

Your minor point stands too: the suppressed self-loop returns `ForceUnrollResponse{Accepted: true}` for work that was never scheduled, where `ExpiredState` deliberately supplies a refusal `Reason`. If any of this is revived, that asymmetry should come with it.

Leaving the PR open pending @ellemouton's call rather than closing it unilaterally, but I would not merge it.

ellemouton:
Closing — superseded by #1113, which merged.

This PR drew the line at "has the forfeit signature left the client", and review established that discriminator is too coarse in both directions:

- `forfeitSignatureIssued()` is true from state *entry* in every live flow, since `handleForfeitRequest` populates both fields before `ForfeitSignedEvent` arrives. So the block-epoch guard was effectively dead code in production, and its test pinned a shape the live flow never produces.
- The suppression swallowed the **manual** RPC unroll — the one path this PR claimed was deliberately honoured — because `Trigger` is the empty-string zero value and the RPC never set it. It failed dishonestly too, returning `Created: true` for a job never scheduled.
- On a stalled round the operator holds nothing, so the forfeit is unspendable and a unilateral exit is the only recovery. Suppressing it there converts a wasted-fee window into a permanent lockout past batch expiry — worse than the bug the issue was filed about.

#1113 arrives at the opposite and correct arrangement: refuse the manual trigger, keep critical expiry as the safety net, and set the trigger properly at the RPC boundary. It also closes the `PendingForfeit` door this PR left open and fixes the `Accepted: true`-on-a-no-op reporting that both PRs shared.

Nothing here is worth salvaging separately — both halves rested on the same discriminator. #845 is addressed by #1113.

ellemouton:
Superseded by #1113.

Issue #577, opened 2026-05-28
Unable to send out second onchain payment while first is still unconfirmed


Liongrass:
I was successfully send out an onchain payment.

While this first payment was still unconfirmed I tried to send a second payment. This payment failed with:

```
{
  "error": {
    "code": "EXECUTION_FAILED",
    "message": "send: rpc error: code = Internal desc = leave vtxos: rpc error: code = Internal desc = leave request failed: reserve leave inputs: reserve forfeit 70c81a058a0c920c47bdbfff35fe3349cc5afc3c268c13d2c07bd0b129b78237:0: process event: forfeiting: bad event: *round.PendingForfeitEvent"
  }
}
```

Only when the first payment confirmed I was able to send out the second payment.

PR #1113, opened 2026-08-07
multi: gate forfeit-committed VTXOs on the leave and unilateral-exit paths

ellemouton:
Fixes the user report behind #577, and folds in the #845 guard from
`agent/vtxo-no-exit-while-forfeiting`.

## What happened

A user ran `wavecli exit --outpoint … --onchain-address …` on signet and
got:

```
leave request failed: reserve leave inputs: reserve forfeit
90199df4…a48a1:0: process event: forfeiting: bad event:
*round.PendingForfeitEvent
```

The daemon log has the whole sequence. Their **first** exit at 16:01:04
succeeded (`VTXOs queued for leave queued_count=1`, round `019fd94b`,
quoted at 286 sat, commitment validated with `leave_output_count=1`).
Four seconds later `exit plan` answered `can_start=true
total_shortfall_sat=0` on a coin that was by then in `PendingForfeit`.
They then tried `--force-unroll-ack` with `--onchain-address`, which the
CLI refused on a flag-combination technicality, and finally re-ran the
plain exit at 16:06:30 — which is the error above. The commitment
confirmed at 16:07:33 and the VTXO went terminal `Forfeited`.

So the exit worked, and the "vanished" VTXO was correct behaviour. What
failed was everything that should have told them so.

## Changes

Each commit stands alone.

- **`vtxo`** — `CheckForfeitAdmission`, a pure predicate over the
  persisted lifecycle status. The admissible set is Live **and Expired**:
  an expired VTXO's value is recoverable only by forfeiting it into a
  round, so a `== Live` check would strand it permanently. Terminal
  states are refused explicitly, since the FSM self-loops on every event
  there and a caller would read that as an accepted reservation.

- **`waved`** — `LeaveVTXOs` filters both selection modes through it.
  Explicit outpoints get `FailedPrecondition` naming the round;
  `selection=all` drops them, because it enumerates `ListLiveVTXOs`,
  which returns every non-terminal VTXO — so one in-flight coin used to
  fail the whole batch. The filter runs before the `dry_run` echo so the
  preview stays a truthful validity probe.

- **`vtxo`** — `ForfeitingState` handles `PendingForfeitEvent`
  explicitly and reports outpoint, state and round instead of falling
  through to the `%T` catch-all.

- **`waved`** — the `Unroll` RPC built its request without `Trigger`, so
  every hand-typed unroll recorded as `UnrollTriggerCriticalExpiry`. The
  #845 guard below needs that distinction to exist.

- **`waved`** — `exitPlanEntry` priced the CPFP funding without ever
  reading `desc.Status`. It now reports a round-committed VTXO as a
  per-outpoint failure, which keeps the existing batch semantics and
  excludes it from the aggregate `can_start`.

- **`multi`** — `exit` prints a stderr notice naming which of the two
  exits ran (stdout keeps its JSON), and the flag-conflict message
  explains why the two cannot combine and points at the **cooperative**
  way out rather than the dangerous one.

- **`vtxo`** — the #845 guard: once `forfeitSignatureIssued()`, an
  expiry-driven escalation out of `ForfeitingState` is suppressed on
  both doors (block-epoch and `ForceUnrollEvent`). Manual and
  fraud-spend stay honoured.

## Why manual stays honoured

An earlier revision of this branch refused the manual trigger too. That
was wrong, and the asymmetry is worth recording.

A doomed exit costs CPFP fees and a wedged job, not principal: the
forfeit spends a connector of the new commitment, so if that commitment
confirms the user already receives what the round owed them. But if the
operator is unreachable and the commitment never confirms, the forfeit
can never confirm either — and the unilateral exit is then the only
thing that recovers the coin, since the reconcile is probing an
operator that never answers and critical expiry is (correctly)
suppressed. Refusing manual there loses principal to save fees.

So manual stays open. Making it *informed* rather than *refused* is a
consent problem, and belongs at the CLI alongside the existing `--yes`
gates on `leave --all`, `recovery escalate` and `vtxos refresh`.

## Deliberately not here

- An on-demand "is this round dead" probe. Both guards in this PR are
  backstops for the same absence: the real discriminator is whether the
  round is alive, and no FSM state answers it. #844's reconcile resolves
  it eventually — it probed twice during this incident and correctly
  held the reservations — but nothing lets a user ask on demand. That is
  the actual long-term fix these patches stand in for.
- The `--yes` consent gate for a manual force-unroll on a
  signature-issued coin.
- A dedicated `EXIT_INFEASIBILITY_REASON_*` value for the exit preview;
  it needs a proto change, and the error string carries state and round
  today.
- **Cross-repo**: `lumos/itest/leave_double_send_test.go:152` asserts the
  error *contains* `"forfeiting: bad event: *round.PendingForfeitEvent"`.
  That assertion encodes the old behaviour and will fail once the client
  pin bumps, so it needs a matching lumos PR.
- Unrelated, found in the same log: an automatic refresh cohort has
  failed 65 times over 9 days with `VTXO amount is below minimum: got
  0.00000744 BTC, want 0.00001000 BTC`. Two 744-sat VTXOs can never be
  admitted and take a healthy 10,737-sat coin down with them on every
  retry, since the cohort ships as one round. Needs its own issue.

## Testing

New coverage: the admission predicate across every lifecycle status;
strict vs. non-strict leave filtering; expired VTXOs staying leavable;
the dry-run refusal; the FSM rejection naming its round; the
force-unroll trigger matrix; the exit-preview refusal and its scoping;
and both CLI messages.

`go test` green on `vtxo`, `waved`, `wallet`, `round`, `unroll`,
`fraud`, `swapwallet` (tagged) and `cmd/wavecli/...`. `make lint-local`
(full tree) reports 0 issues; `make commitmsg-lint` passes.


Roasbeef:
One structural suggestion, fine as a follow-up: wire `CheckForfeitAdmission` into `handleReserveForfeit` itself, so the gate lives at the manager chokepoint rather than only at the RPC layer.

As written, the pre-check covers `LeaveVTXOs` and the exit preview, but `ReserveForfeitRequest` has more callers than that: the refresh path, the sweep-all onchain send, and the send-intent replayer all reserve through the same manager handler. A refresh that races a coin into `Forfeiting` still reaches the child FSM and only gets the improved backstop error, and any future caller has to remember to run the predicate on its own. Evaluating it once inside `handleReserveForfeit` (before dispatching `PendingForfeitEvent`) covers all of them uniformly, and the FSM stays the authority for the residual race exactly as the comments here already frame it.

Two ways to get the state to check: read the descriptor from the store and reuse the predicate as-is, or ask the child actor directly with a small non-mutating `GetStateRequest` message (I prototyped the latter on `roasbeef/vtxo-forfeit-reserve-admission`, which reads the live FSM state through the mailbox instead of a persisted snapshot). Either works; the descriptor route is cheaper since it avoids N extra child asks per reservation.

One thing I checked while convincing myself a hard rejection at the manager is safe: the send-intent replayer re-reserves its persisted anchors at startup, which looks like it would trip a non-Live rejection. It doesn't, because `releaseOrphanedForfeits` sweeps every `PendingForfeit` coin back to Live during `Manager.Start` before any replay runs. That ordering is load-bearing though, so worth a comment (or a test pinning it) if the gate moves down.

waved/rpc_leave_admission.go
```diff
@@ -0,0 +1,79 @@
...
+// strict selects. A caller naming outpoints explicitly must be told that the
+// specific coin it asked for is unavailable; silently dropping it would
+// report a queued leave that never happens. A caller asking for "all" has
+// asked for whatever is leavable, so an in-flight VTXO is simply not part of
+// the answer — and must be dropped, because ListLiveVTXOs returns every
+// non-terminal VTXO, PendingForfeit and Forfeiting included.
+//
+// This is an advisory pre-check. The reservation is only settled inside the
+// VTXO actor's FSM, so a concurrent claim can still take a coin between this
+// filter and the dispatch below; the FSM stays the authority for that race.
```
litbot-9000: 🟠 **major** — **"the FSM stays the authority for that race" is not true at the `PendingForfeit` door — and that is the door where the damage is.**

This PR closes the `Forfeiting` door (the new `*PendingForfeitEvent` case in `vtxo/transitions.go`), but the reservation path has no status check of its own: `Manager.handleReserveForfeit` (`vtxo/manager.go:2755-2768`) validates only that an actor exists and that the outpoint is not spend-reserved, then sends `PendingForfeitEvent` and trusts the FSM. `PendingForfeitState` self-loops on that event (`vtxo/transitions.go:936-942`, *"Duplicate commit while already pending is harmless"*).

Probing both doors through the real FSM via `ReserveForfeitRequest`:

```
PROBE forfeiting door:  err=reserve forfeit …:42: vtxo is already committed to a round
                        (outpoint …:42, state forfeiting, round round-first-claim)
PROBE pending door:     err=<nil> resp=*actormsg.ReserveForfeitResponse
                        state=*vtxo.PendingForfeitState
```

So in the race this comment describes — the pre-check passes while the coin is Live, a concurrent claim (an automatic refresh cohort, which the PR body itself reports firing repeatedly in the user's log) moves it to `PendingForfeit`, then the wallet reserves — the reservation **succeeds** and the leave is reported queued.

That is exactly the hazard `vtxo/forfeit_admission.go:32-37` documents as the reason to refuse `PendingForfeit` in the first place:

> a second request carrying a different destination would otherwise report success while the first destination is what actually gets paid, and which of the two behaviours a caller meets is decided purely by how far the round happens to have progressed in the preceding seconds.

The predicate closes that in the advisory pre-check — which the same doc-comment correctly says is not the authority. So the hazard survives precisely in the window the pre-check cannot cover, and the user is told their leave to address X is queued while the coin is consumed by the earlier claim.

This is pre-existing FSM behaviour and the PR does not make it worse. But the PR asserts a guarantee here (and at `vtxo/forfeit_admission.go:9-15`) that the code does not provide, and it closed the symmetric door while leaving this one open. Either give `PendingForfeitState` the same explicit refusal the new `ForfeitingState` case got — a genuine round replay can be told apart by round ID, which is what makes the duplicate "harmless" in the first place — or narrow both comments to say the FSM is authoritative only for `Forfeiting`.

vtxo/transitions.go
```diff
@@ -1085,6 +1119,28 @@ func (s *ForfeitingState) ProcessEvent(ctx context.Context, event VTXOEvent,
...
+		// Manual and fraud-spend triggers are deliberately still
+		// honoured. A fraud spend means the operator has already moved
+		// against us and exiting is the correct response, and a manual
+		// unroll is an explicit operator decision that should not be
+		// silently dropped. Only the automatic path, which fires on a
+		// coin nobody chose to exit, is suppressed.
+		if evt.Trigger == actormsg.UnrollTriggerCriticalExpiry &&
+			s.forfeitSignatureIssued() {
+			return &VTXOStateTransition{
+				NextState: s,
```
litbot-9000, line 1162: 🟠 **major** — **This suppression reports `Accepted: true` to the caller while scheduling nothing.**

`Manager.handleForceUnroll` (`vtxo/manager.go:1057-1107`) detects a self-loop no-op in exactly two cases: prior+new both terminal (`"already terminal"`), and `NewState` is `*ExpiredState` (`"batch expired; ..."`). This new self-loop is `ForfeitingState` → `ForfeitingState`, which is neither, so it falls straight through to the success path.

Probe against the real FSM (the manager test double's `Ask` runs `ProcessEvent` for real):

```
PROBE: accepted=true reason="" state=*vtxo.ForfeitingState
```

The manager also logs `"Force-unroll accepted by VTXO actor"` on the way out, so the daemon log records the same claim.

This breaks an invariant the package documents about itself. `vtxo/CLAUDE.md` (unchanged by this PR) states:

> `Manager.handleForceUnroll` uses `Ask` (not `Tell`) so FSM errors and self-loop no-ops surface as structured `ForceUnrollResponse{Accepted, Reason}` instead of a uniform `Accepted:true` that masks work that was never scheduled.

And the `ExpiredState` branch immediately above the fall-through exists for exactly this reason: *"the caller would be told the exit was accepted when nothing was scheduled."* This PR adds a third self-loop and does not extend the check to cover it.

On reachability, to be fair: this is latent, not live. `waved: record manual unrolls under the manual trigger` lands **before** the guard commit on this branch, and the other production callers (`fraud/actor.go:497`, `vhtlcrecovery/coordinator/service.go:476`) set explicit triggers — so nothing reaches this branch today. But `UnrollTriggerCriticalExpiry` is the **empty-string zero value** (`lib/actormsg/vtxo_admission.go:357`), so any caller that omits `Trigger` lands here — which is precisely the bug this PR fixes one commit earlier. The guard is defence-in-depth against a caller that forgets the field, and when it fires against that caller it reports a successful exit admission for a job that does not exist.

The new tests don't catch it because `TestForfeitingStateForceUnrollTriggers` asserts on the FSM transition (state stays put) and never on the `ForceUnrollResponse` the manager builds from it.

Suggested fix: make the refusal explicit rather than a bare self-loop and give `handleForceUnroll` a matching case, so the response carries `Accepted: false` with an actionable reason alongside the `ExpiredState` treatment — e.g. `"forfeit signature already issued; recover via the round holding this VTXO"`.

waved/rpc_exit_plan.go
```diff
@@ -228,6 +229,20 @@ func (r *RPCServer) exitPlanEntry(ctx context.Context, raw string,
...
 	}
 
+	// A VTXO committed to a cooperative round cannot be exited
+	// unilaterally, and no amount of wallet funding changes that: the
+	// operator holds a forfeit for it, so the exit path refuses the coin
+	// outright. Pricing it anyway answers a question that is not the
+	// blocking one and reports a ready exit for a coin that will be
+	// refused — which is exactly what sends a user reaching for
+	// --force-unroll-ack (wavelength#577, #845).
+	if commitErr := exitPlanRoundCommitment(desc); commitErr != nil {
```
litbot-9000: 🟠 **major** — **The preview now refuses exits that `Unroll` will actually perform — including the one this PR argues is the last lever that saves principal.**

The justifying comment above says pricing a committed coin *"reports a ready exit for a coin that will be refused"*. That is not true for the manual path, and this PR is the reason it isn't. `RPCServer.Unroll` short-circuits only on `VTXOStatusUnilateralExit`; it has no `PendingForfeit`/`Forfeiting` check, runs `preflightUnrollFeasibility` (dust / economics / funding only), then sends `ForceUnrollRequest{Trigger: UnrollTriggerManual}`. On the receiving side:

- `ForfeitingState` — the new guard fires only for `UnrollTriggerCriticalExpiry`, so a manual trigger falls through and escalates to `UnilateralExitState`;
- `PendingForfeitState` — accepts `ForceUnrollEvent` unconditionally (*"Client requested unilateral exit while forfeit is still pending"*).

So `exit plan` reports `cannot exit unilaterally: vtxo is already committed to a round (…)` for a coin that `exit --force-unroll-ack` will exit without complaint. The preview contradicts the command it previews.

The cost is concentrated on exactly the scenario the PR body singles out as principal-saving:

> if the operator is unreachable and the commitment never confirms, the forfeit can never confirm either — and the unilateral exit is then the only thing that recovers the coin … Refusing manual there loses principal to save fees.

A user in that state does the sensible thing and runs `exit plan` first — which is what the incident user did — and is now told the exit is impossible. Worse, this returns via `entry.Err`, so the entry carries **no** `FundingShortfallSat`, no `RecommendedTotalFundingSat` and no funding address. The user cannot even find out how much CPFP funding to deposit for the one action that would recover the coin.

The underlying goal is good: stop a user reaching for `--force-unroll-ack` when a cooperative leave already succeeded. But an error that hides the numbers is the wrong instrument, because the same states are also where the manual exit is legitimate. Suggest keeping the pricing and surfacing the commitment as an advisory alongside it — set `CanStart=false` with a reason naming the round (the `InfeasibilityReason` enum is the natural home, and the PR already notes a dedicated value needs a proto change), rather than failing the entry outright. That preserves the warning, keeps the preview consistent with `Unroll`, and still gives the funding figures the recovery path needs.

waved/rpc_leave_admission.go
```diff
@@ -0,0 +1,79 @@
...
+			return nil, status.Errorf(codes.Internal, "get VTXO "+
+				"%s: %v", op, err)
+		}
+
+		if admitErr := vtxo.CheckForfeitAdmission(
+			desc,
+		); admitErr != nil {
+
+			if !strict {
+				continue
```
litbot-9000, line 101: 🟡 **minor** — The `selection=all` drop is silent — nothing is logged, and the response cannot be distinguished from an empty wallet.

Dropping inadmissible coins under `--all` is the right default (it matches what `--all` means elsewhere, and it fixes the real bug where one in-flight coin failed the whole batch). The problem is that nothing anywhere records that it happened. `admitLeaveTargets` has no log statement at all, and the caller-side summary at `waved/rpc_server.go:2491-2494` reports only what survived:

```go
r.server.log.InfoS(ctx, "VTXOs queued for leave",
        slog.Int("queued_count", len(queued)),
        slog.Int("error_count", len(resp.Errors)),
)
```

So a `leave --all` over a wallet whose coins are all committed to a round logs `queued_count=0 error_count=0` and returns `{"queued_outpoints": [], "status": "queued"}` — byte-identical to the same call against a wallet with no VTXOs. The operator cannot tell "nothing to leave" from "everything was skipped, retry once the round confirms", and the daemon log holds no record of which coins were skipped or why.

That is the same class of failure this PR exists to fix — the PR body's own diagnosis is *"the exit worked … What failed was everything that should have told them so."* And `LeaveVTXOs` already carries a scar from an empty preview being misread: the H-5 comment at `waved/rpc_server.go:2384-2393` describes a user reading an empty list as a no-op.

Cheap fix: log each dropped outpoint with its admission error in the `!strict` branch, and/or return a `skipped_count` so the caller can see the batch was filtered. The error from `CheckForfeitAdmission` already names the round, so the log line is actionable as-is.

litbot-9000 (changes requested):
FINAL — **CI first, since it drives the verdict: the red is infrastructure, not a data race.** Both failed jobs carry the same GitHub check-run annotation — *"The self-hosted runner lost communication with the server"* — on two distinct runners in one deployment (`lightninglabs-runnerdeploy-44c8q-4lbfb` and `-msh6k`). They died 20 seconds apart (01:22:06Z and 01:21:46Z) despite starting 100 seconds apart, and every step from `Run unit-race` onward has a null conclusion, i.e. never completed. No test ever reported a failure. I confirmed it locally too: `go test -race -count=1 ./vtxo/... ./waved/... ./wallet/...` passes clean (exit 0). Re-run the two jobs; there is nothing to fix here. (Confirmed with the team: CI runs on ARC and the runner instances can be preempted, so runner-loss annotations of this shape are a known flake mode, not a signal about this PR.) (I inherited an earlier interim note asserting the same conclusion with no evidence attached, and re-derived it from scratch rather than trusting it.)

The core of this PR is right and the commit hygiene is genuinely good — in particular `waved: record manual unrolls under the manual trigger` lands *before* the guard commit, so no intermediate commit strands the manual unroll. I verified the headline claim rather than taking it: `vtxo.PendingForfeitEvent` is an alias for `round.PendingForfeitEvent` (`vtxo/events.go:47`), so the new `ForfeitingState` case really does replace the `forfeiting: bad event: *round.PendingForfeitEvent` string from the incident, and a probe through the real FSM returns the new round-naming error.

**Merge order vs #1112: land #1113, and close #1112.** #1112 is not safe standalone — its guard keys on `UnrollTriggerCriticalExpiry`, which is the empty-string zero value, while `RPCServer.Unroll` on main builds its request without `Trigger`. So on main alone the guard swallows the manual RPC unroll (the one path its own comment calls "deliberately still honoured") and reports it as created. #1113 contains both #1112's guard and the trigger fix that makes it behave as documented, correctly sequenced. There is nothing in #1112 that #1113 does not already carry.

**`CheckForfeitAdmission` against every lifecycle status** — I enumerated all nine (`vtxo/interfaces.go:259-316`). Live/Expired admit; PendingForfeit/Forfeiting → `ErrForfeitInFlight`; Spending → `ErrVTXOLiquidityLocked`; UnilateralExit → `ErrExitInFlight`; and the three genuinely terminal ones (Forfeited, Spent, Failed) fall to `default` → `ErrVTXOTerminal`. Complete, and `default` means a future status fails closed, which is the right direction. Expired is genuinely safe: `ExpiredState` accepts `PendingForfeitEvent` by reusing `LiveState.handleForfeitRequest`, and a `== Live` check really would strand the coin, since forfeiting into a round is the only way to recover an expired VTXO's value.

**The cross-repo break is called out adequately** — `lumos/itest/leave_double_send_test.go:152` is named with file and line under "Deliberately not here" and the reason is stated correctly. Worth linking the matching lumos PR before the client pin bumps so it isn't discovered by red CI.

**The race between the pre-check and the FSM is not benign in both directions**, which is the main finding. The PR closed the `Forfeiting` door and asserts twice that the FSM is authoritative for the rest; it isn't at the `PendingForfeit` door, where the reservation still returns success. Details inline, with probe output. Alongside that: the new suppression self-loop reports `Accepted: true` for an exit that was never scheduled, and the new exit preview refuses coins that `Unroll` will happily exit.

On the `selection=all` silent drop — dropping is the right default and I don't think it should change, but it is currently invisible in both the response and the daemon log. Filed as minor.

ellemouton:
@claude one last re-review pls

Issue #1116, opened 2026-08-07
Bug: Show all vtxos in `wavecli ark vtxos list`


Liongrass:
Currently, only `VTXO_STATUS_LIVE` vtxos are shown by default in the command `wavecli ark vtxos list`.

This may make it hard to track vtxos that are in other statuses.

I believe that ideally, the command should show all vtxos, with the exception of `forfeited` vtxos, and accept an additional flag `--include-forfeited`.

Additionally, the `--status` flag could accept multiple arguments. Currently it only accepts one.

PR #1118, opened 2026-08-07
Show all vtxos except forfeited and spent, add option to show all

Liongrass:
This pull request addresses #1116 

As of now, `wavecli ark vtxos list` only shows live VTXOs. This means that any application building on that RPC call has to make repeated RPC calls to fetch all VTXOs, regarding of their status. By default, any VTXO that isn't `live` appears to be simply missing.

Available statuses include: `live`, `pending_forfeit`, `forfeiting`, `forfeited`, `spent`, `unilateral_exit`, `failed`, `spending`

In the first commit, we expand this to all VTXOs, except forfeited and spent VTXOs. I believe this best represents the VTXOs a user would care about.

In the second commit, we add a new `--show-all` flag that gives the user or application the option to query all VTXOs, regardless of their status, and then filter themselves.

Future work to consider:
- Allow multiple ` --status` flags to be passed
- Distinguish between successful/completed unilateral exit, and unilateral exit in progress

### DIFF
--- a/cmd/wavecli/waveclicommands/cmd_vtxos.go
+++ b/cmd/wavecli/waveclicommands/cmd_vtxos.go
@@ -48,7 +48,9 @@ func newVTXOsListCmd() *cobra.Command {
 		Short: "List VTXOs",
 		Long: "Returns VTXOs known to the wallet, " +
 			"optionally filtered by status and " +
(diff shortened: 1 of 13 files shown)


darioAnongba:
superseded by #1299

PR #1121, opened 2026-08-07
actor: durable mailbox backpressure watermarks + depth metrics

Roasbeef:
In this PR, we give the durable mailbox the backpressure signal it never had. A channel mailbox bounds its queue with `ErrMailboxFull`, but a durable mailbox lands every `Tell` as a database row: a consumer that stops draining grows its backlog for as long as the outage lasts, with nothing pushing back on producers and nothing on a dashboard showing the hole getting deeper. This is the second PR in the durable-actor hardening series (after #1119), covering the backpressure gap called out in the DLQ/backpressure review.

## Watermarks

Two thresholds on `DurableMailboxConfig` (flowing through `DurableActorConfig`) bound the persistent backlog. Past the soft watermark the mailbox logs one warning per breach episode. At or past the hard watermark, `Send` refuses the message with the new `ErrMailboxSaturated` sentinel, the durable analogue of `ErrMailboxFull`: the message was not enqueued, and the caller sheds, stashes, or retries after the consumer drains. The check runs before encoding or promise registration, so a refusal needs no cleanup and `TrySend`/`TryTell` inherit it for free. Messages with priority >= `RestartPriority` are always exempt, since the `RestartMessage` that would un-wedge a stuck actor must not be refused by the very backlog it exists to drain.

Depth is read through a new narrow `MailboxDepthStore` surface (discovered by type assertion on the `DeliveryStore`, same pattern as keeping test doubles small elsewhere), backed by a `COUNT(*)` that the prefix of `idx_mailbox_messages_available` covers. The read is TTL-cached at one second with a local sent-since-probe delta on top, so the common send path pays no extra query. The estimate is deliberately one-sided (local sends push it up immediately, acks only surface at the next probe): overshooting is the safe direction for an admission check. A failed probe fails open, because a broken monitoring read must not become message loss.

Both thresholds default to zero (disabled). The `multi:` commit is the complete inventory of which mailboxes opt in: the serverconn egress sender, the OOR registry and per-session actors, and the ledger, credit, and unroll actors, all at the shared defaults (soft 1000, hard 10000).

## serverconn classification

`deliverToActor` now classifies `ErrMailboxSaturated` exactly like `ErrMailboxFull`: the envelope defers, the cursor stalls, and the ingress loop re-pulls after a backoff. A backed-up durable actor therefore exerts backpressure on the operator stream instead of deepening its own backlog, riding the deferral machinery #1093 put in place (episode logging, deferred counter, redrive) with no changes downstream of the sentinel check.

[...]

PR #1119, opened 2026-08-07
actor: wire up dead-letter handling with requeue, monitor, and operator surface

Roasbeef:
In this PR, we close the dead-letter half of #705: we had a full dead-letter *store* but no dead-letter *handling*. When a durable actor exhausted its `TellRetryPolicy`, the framework wrote the message to `dead_letters` and then nothing ever looked at it again. No log, no metric, no operator surface, no recovery path. For value-bearing subsystems (OOR, rounds, unroll) a silently parked message can mean stuck or invisible funds.

The contract this PR establishes: a dead letter is never silent and never final by default. The daemon surfaces it, an operator can recover it, and nothing ages out unless explicitly opted into.

## Faithful requeue needs the full routing identity

The original `dead_letters` projection dropped every routing field of the source mailbox row (priority, retry budget, ask plumbing, and the per-key FIFO tag), which made reconstruction impossible. A new actor-delivery migration widens the table and `MoveMailboxToDeadLetter` now projects the full identity. Rows written before the migration stay readable and requeue with fresh-enqueue defaults.

`RequeueDeadLetter` then re-enqueues the payload as a *new* mailbox message in one transaction: fresh UUIDv7, attempts reset, available immediately, routing preserved, dead letter deleted. The fresh ID is a correctness requirement rather than a convenience: the retry-exhaustion path marks the original ID processed before dead-lettering, so a same-ID requeue would be silently skipped by dedup (and masked by the enqueue's `ON CONFLICT DO NOTHING`). A successful requeue also fires the targeted mailbox wake, so a resident consumer picks the message up without waiting out its poll interval.

The operator ops live on a new `actor.DeadLetterStore` interface, deliberately kept out of `DeliveryStore`: the runtime only ever writes dead letters, and widening the runtime interface would force every test double in the tree to grow methods the runtime never calls.

## A monitor that owns observation and retention

A daemon-owned `DeadLetterMonitor` (OutboxPublisher lifecycle shape) scans for newly parked entries and surfaces each exactly once: error-level detail lines (bounded per scan, since a poison storm parks many at once and one summary line carries the same alerting weight), a startup summary of the backlog left by previous runs, and an `OnDeadLetter` hook feeding metrics. The since-scan boundary is inclusive at second granularity, so the monitor keeps a boundary set of reported IDs and prunes it as the watermark advances.

[...]

PR #1123, opened 2026-08-08
actor: Add a supervision kernel for the durable actor runtime

Roasbeef:
In this PR, we add a supervision kernel to the durable actor runtime, closing
the BEAM gap in its failure management. It is the fourth and last part of the
durable-actor hardening series, after #1119 and #1121.

### The limping actor

Today a panic inside a durable behavior's `Receive` is recovered by the worker
loop and converted into a nack (see the existing recovery in
`baselib/actor/durable_actor.go` and `TestDurableActorPanicRecovery`). The
message is redelivered, and it is redelivered into the very same behavior
instance whose in-memory state the panic may have left half-mutated. Every
message after the panic then runs against state nobody can vouch for. The actor
does not die, which sounds like a virtue, but what it actually does is limp:
the runtime has an excellent story for a process that crashes (checkpoint,
restart message, dedup, lease expiry) and no story at all for a behavior that
corrupts itself in place.

### Panic means restart

The worker loops now belong to a supervision goroutine that owns them a
generation at a time. Each generation runs under its own context derived from
the actor's lifetime context, so cancelling it drains every worker without
touching the mailbox. A recovered panic becomes a `behaviorPanic`, which is
what lets the runtime tell "the behavior returned an error" (an ordinary,
retryable message failure) apart from "the behavior panicked" (its state is
suspect). The panicking worker hands that error to supervision, which cancels
the generation, runs the behavior's `OnStop` hook if it has one (bounded by
`CleanupTimeout`), reloads the persisted FSM checkpoint, prepends a
`RestartMessage` at `RestartPriority`, and starts a fresh generation of the
configured worker count.

Two orderings inside that sequence are load-bearing.

The first is that the delivery's ack, nack, and dead-letter bookkeeping runs to
completion *before* the panic reaches supervision, so the message that
triggered the panic burns its attempt exactly as it does today. A deterministic
poison message therefore climbs to `max_attempts` and dead-letters rather than
restarting the actor forever, which is precisely the crash loop a naive
restart-on-panic would introduce.

The second is that on the classic path, where the whole `Receive` runs inside
one framework transaction, that bookkeeping is deliberately redone *outside*
the transaction. Supervision returns the panic from the transaction to force a
rollback of the behavior's own partial writes, because committing them
alongside the nack would persist exactly the torn state the restart exists to
escape, and the checkpoint reload would then hand it straight back.

The restart deliberately bypasses the `Once`-guarded `Start` and `Stop`. Those
guard the actor's public lifecycle, which a restart does not touch: the ID, the
`DurableMailbox`, and the cached `Ref` are the same objects afterwards, so
callers holding an `ActorRef` observe nothing beyond a pause in processing, and
senders keep enqueueing across the restart gap. In-flight Ask promises are
completed rather than dropped: the panicking turn's promise takes the panic
error, a sibling worker's turn sees its generation context cancelled and takes
that context error, and a message that had not yet reached the behavior is
redelivered afterwards with its promise still registered on the surviving
mailbox. `DurableAsk` responses travel through the outbox, so a restart only
delays them.

### Ordering the hand-off under a worker pool

`RestartPriority` makes the claim query hand the restart message out first, and
that is where the guarantee used to stop: it orders the *claims*, not the
*turns*. With `NumWorkers > 1` the supervisor launched every worker at once, so
one worker took the restart while a sibling immediately took the row behind it,
and a normal turn ran against a behavior that was still rebuilding itself from
the checkpoint. Latent today (the only pool adopter is the stateless
`serverconn` egress) and pre-existing on the boot path, but supervision makes
restart-under-pool routine, so a priority alone is no longer enough.

A generation now warms up one worker at a time. The supervisor launches a
single worker and the rest of the pool waits until that worker has resolved the
generation's restart hand-off. For a row supervision enqueued itself (every
supervised restart) the barrier waits for it unconditionally, so the guarantee
is exact rather than timing-dependent on the path this PR creates. The boot
hand-off is different in kind, since an owner prepends it before `Start` and
the actor never sees it; there the barrier orders the first claim and releases
on an idle tick, which orders the common case without claiming more than it can
know.

The release rule is shaped so the barrier cannot wedge a pool. A first claim of
anything other than a restart releases it *before* that message is processed,
so a generation with nothing to order pays nothing. A restore that fails is
dead-lettered and one that panics tears the generation down; both resolve the
hand-off. Whatever ends the warm-up worker releases it, including a closed
mailbox. And a `Stop` mid-barrier releases it through the generation context,
so shutdown never parks behind a restore. A single-worker actor is already
strictly sequential and gets no barrier at all.

### What "restart" does not mean, and the adopter wiring

This is the part worth reading closely, because the name oversells it. The
framework does **not** rebuild the behavior. It stops the workers, optionally
calls `OnStop`, and redelivers a `RestartMessage`; the same Go value keeps
serving afterwards with whatever fields the panic left behind. The clean slate
exists exactly where the `RestartMessage` handler rebuilds in-memory state from
the durable row, and nowhere else.

That made the first cut of this PR vacuous in practice. All six production
construction sites are Read/Commit, and three of them (credit op, OOR session,
OOR registry) explicitly no-op'd the restart message on the reasoning that
"restore already ran at construction", which was true of the boot-time message
and false of a supervised one. The other two never registered `RestartTLVType`
at all. Worse, the panic path bypasses the `commitFailed` reload guards those
behaviors already carry, since those are armed only on commit-error paths, so a
panic after an in-memory FSM advance left the behavior ahead of the durable row
and the restart delivered a no-op.

So this PR wires the adopters:

- **credit op** and **OOR session** arm their existing `commitFailed` guard, so
  the next turn reloads from the durable row before dispatching. Arming rather
  than reloading inline keeps the restart message free of IO, which matters
  because it is delivered with `max_attempts` 1.
- **OOR registry** drops the staged handoff and re-runs the same non-terminal
  restore the boot path uses; that restore already skips resident sessions, so
  it is idempotent by construction.
- **unroll** gains a sealed-surface adapter for `RestartTLVType` (its `Msg`
  interface is sealed, so the bare framework type would fail the mailbox cast
  and dead-letter) plus explicit handling that re-runs `restoreCheckpoint`. Its
  dispatch previously rejected unknown messages with a typed error, so a
  restart would have nack-looped.
- **serverconn egress** gains the same adapter and a documented no-op. It keeps
  no per-turn state a checkpoint could restore: every turn reads its whole
  input from the durable message and consumes it in one Commit, and the
  in-memory state the connector holds belongs to the connection rather than to
  a turn. Arriving at "no restore needed" by accident is not the same as
  choosing it, so it says so.

Two obligations follow for handler authors, both documented. A restart message
is delivered once and **dead-lettered rather than nacked** when its turn fails
(a nacked row at `attempts == max_attempts` is neither leasable nor reapable,
so it would strand forever), which makes restore handlers responsible for their
own idempotency. And `OnStop` now runs mid-life, once per restart, so it must
be idempotent and must leave the behavior able to serve a new generation. A
panic escaping `OnStop` is recovered rather than allowed to take the process
down, since a restart calls it precisely when the behavior's invariants are
broken.

An actor whose codec still cannot carry a restart message degrades to cycling
its worker generation with **no** `OnStop` and no restore. That check happens
before the teardown, not after: a teardown the behavior cannot be rebuilt from
is strictly worse than leaving it running.

### Restart intensity: default off

Restarts can be bounded by `MaxRestarts` inside `RestartWindow`, tracked in a
sliding window. It defaults **off** (`DefaultMaxRestarts` is
`UnlimitedRestarts`, and a zero value normalizes to it).

The first cut of this PR defaulted it on at five per sixty seconds, and the
arithmetic against real config is unkind: the default Tell retry policy gives a
message five attempts, each of which panics and restarts, so *one* poison Tell
burns the whole budget. Two poison messages in a minute would permanently kill
an actor that `main` would have dead-lettered twice and kept serving; the
serverconn egress sender would die that way while its heartbeat kept the
connection looking healthy.

The asymmetry decides it. Restarting forever is strictly no worse than the
nack-and-continue loop supervision replaces (both feed the same message to the
same behavior, both rate-limited by the nack backoff), while silent terminal
death is genuinely new and genuinely invisible: a terminated actor still holds
its ID and its mailbox rows. Nothing outside this package's tests calls `Watch`
yet, so a finite budget today is unobserved by construction. It is therefore
opt-in, documented as such on the config field, with `RecommendedMaxRestarts`
(the BEAM intensity) available for owners that do the wiring.

Supervision also keeps the mailbox tidy: each restart deletes the restart row
the previous one enqueued before writing its own, so at most one is pending at
a time rather than one per restart.

### The Watch contract

`(*DurableActor).Watch(ctx)` gives another component a way to observe the
terminal event. It returns a channel that receives exactly one
`TerminationInfo` and is then closed, carrying the reason (stopped, context
cancelled, restart intensity exceeded, restart failed), the failure behind it,
the lifetime restart count, and whether the restart budget was exhausted.
Delivery is non-blocking by construction: each watcher gets a buffer-of-one
channel written exactly once, so a watcher that is slow or gone can never park
the actor's shutdown path, which is this repo's #1093 invariant. Registering
after termination returns a channel already loaded with the notification, and
cancelling the watching context releases the registration. `Stop` publishes for
an actor that was never started, so a watcher on one does not wait forever.
Nothing outside `baselib/actor` consumes it yet; adopters come later.

### Testing

Tests live in `baselib/actor/supervision_test.go`, driven through the existing
`mockDeliveryStore` / `mockTxAwareStore` harness and the clock injection
pattern. The one that carries the most weight drives a Read/Commit behavior
shaped like the real adopters (in-memory mirror plus reload guard), has it
advance the mirror past the durable row and panic, and asserts the next turn
sees durable truth; it fails with the stale value if the restart handler
no-ops, which is what every Read/Commit adopter did before this PR.

The rest: panic restores from checkpoint, identity stays stable, a returned
error does not restart, a poison message dead-letters across restarts, a
restart racing `Stop` reports a graceful stop, a failed checkpoint load
terminates with `TerminationRestartFailed`, a panicking `OnStop` is recovered
and not run twice, two workers panicking in the *same* generation cost one unit
of budget, the intensity limit terminates with the right `Watch` reason,
`Watch` reports a graceful stop, eight unread watchers do not park shutdown, a
cancelled watcher deregisters, a watcher on a never-started actor still fires, a
four-worker pool has every worker drained by a restart, a four-worker pool runs zero normal turns while a restore is gated and drains its backlog once the gate opens, that barrier releases on a failed restore and on `Stop`, a pool with no hand-off reaches full width, a panicking turn's error reaches
`ExecTx` (which is what makes a real store roll back), restart rows neither
pile up nor strand, a `Stoppable` behavior with an unsupported codec is left
running, the sliding window and unlimited opt-out, the budget defaults, and the
termination reason strings.

`go test -count=1` and `-race` pass for `baselib/actor` and for every adopter
touched (`credit`, `oor`, `unroll`, `serverconn`, `ledger`,
`internal/actortest`, also under `-tags test_sqlite`). Every commit on the
branch compiles independently. `make fmt-changed-check`,
`make lint-changed-local` (0 issues), `sg scan` on the changed non-test files
(no new warnings), and `make commitmsg-lint` are all green.


litbot-9000:
This PR's supervision-kernel changes landed new restart-handling behavior in `credit`, `oor`, `serverconn`, and `unroll` whose per-package `CLAUDE.md`/`AGENTS.md` invariants were not updated alongside the code (`baselib/actor`'s docs were already updated in this PR and need no further change).

```diff
diff --git a/credit/AGENTS.md b/credit/AGENTS.md
index 13ce4f8a..449455fd 100644
--- a/credit/AGENTS.md
+++ b/credit/AGENTS.md
@@ -52,6 +52,19 @@ credit redemptions against the swap-server credit ledger, as a crash-safe
 - Every external call the behavior makes (`CreateCredit`, `SendOOR`,
   `StartPay`, `RedeemCredit`) must stay idempotent by op key or payment hash,
   since a redelivered message or a reload-after-`commitFailed` re-runs it.
+- A supervised restart is not a no-op for `opBehavior`. The framework reuses
+  the same behavior instance across a restart, so a panic that left `rec`
+  advanced past the last durable checkpoint would otherwise survive it: the
+  `actor.RestartMessage` case arms the same `commitFailed` reload guard the
+  failed-commit path uses, and the next driving turn rebuilds `rec` from the
+  durable row before it dispatches. The reload is deferred to that turn rather
+  than run on the restart turn itself because the framework delivers the
+  restart message with `MaxAttempts` 1 and dead-letters (rather than retries) a
+  restart turn that fails.
+- `CreditDurableMsg` is unsealed (`actor.TLVMessage` alone), so the codec
+  registers the framework's `actor.RestartMessage` directly under
+  `actor.RestartTLVType`; no package-local adapter type is needed for the
+  decoded message to satisfy the mailbox's cast.
 - Auto-redeem is receive-triggered, not a periodic sweep (except a single
   boot-time reconcile); `triggerRedeem` fires only after the settled receive's
   terminal snapshot commits, so a crash before that leaves no half-applied
diff --git a/credit/CLAUDE.md b/credit/CLAUDE.md
index 13ce4f8a..449455fd 100644
--- a/credit/CLAUDE.md
+++ b/credit/CLAUDE.md
@@ -52,6 +52,19 @@ credit redemptions against the swap-server credit ledger, as a crash-safe
 - Every external call the behavior makes (`CreateCredit`, `SendOOR`,
   `StartPay`, `RedeemCredit`) must stay idempotent by op key or payment hash,
   since a redelivered message or a reload-after-`commitFailed` re-runs it.
+- A supervised restart is not a no-op for `opBehavior`. The framework reuses
+  the same behavior instance across a restart, so a panic that left `rec`
+  advanced past the last durable checkpoint would otherwise survive it: the
+  `actor.RestartMessage` case arms the same `commitFailed` reload guard the
+  failed-commit path uses, and the next driving turn rebuilds `rec` from the
+  durable row before it dispatches. The reload is deferred to that turn rather
+  than run on the restart turn itself because the framework delivers the
+  restart message with `MaxAttempts` 1 and dead-letters (rather than retries) a
+  restart turn that fails.
+- `CreditDurableMsg` is unsealed (`actor.TLVMessage` alone), so the codec
+  registers the framework's `actor.RestartMessage` directly under
+  `actor.RestartTLVType`; no package-local adapter type is needed for the
+  decoded message to satisfy the mailbox's cast.
 - Auto-redeem is receive-triggered, not a periodic sweep (except a single
   boot-time reconcile); `triggerRedeem` fires only after the settled receive's
   terminal snapshot commits, so a crash before that leaves no half-applied
diff --git a/oor/AGENTS.md b/oor/AGENTS.md
index bb983c14..158ce181 100644
--- a/oor/AGENTS.md
+++ b/oor/AGENTS.md
@@ -122,6 +122,23 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/oor.<Sym
 - Server-side lineage-cap rejection surfaces as a typed `*ErrLineageTooLarge`
   via `ClassifySubmitError`, so wallet callers can switch on the cause
   without depending on the `oorpb` proto type.
+- A supervised restart is not a no-op for either behavior. The framework reuses
+  the same behavior instance across a restart, so whatever the panic left in
+  goroutine-owned memory outlives it unless the `actor.RestartMessage` handler
+  rebuilds from durable truth. `sessionBehavior` arms the same `commitFailed`
+  reload guard its failed-commit path uses, so the next turn stops the stale
+  FSM and rebuilds it from the registry row before dispatching; the rebuild is
+  deferred to that turn because `restore()` starts an FSM goroutine that must
+  outlive the turn, and because the framework delivers the restart message with
+  `MaxAttempts` 1 and dead-letters (rather than retries) a restart turn that
+  fails. `oorRegistryBehavior` instead restores inline: it drops the staged
+  `pendingHandoff` (it belonged to the turn that died) and re-runs
+  `handleRestoreNonTerminal`, which is cheap and idempotent because it skips
+  sessions that are already resident.
+- `OORDurableMsg` is unsealed (`actor.TLVMessage` alone), so the shared codec
+  registers the framework's `actor.RestartMessage` directly under
+  `actor.RestartTLVType`; no package-local adapter type is needed for the
+  decoded message to satisfy the mailbox's cast.
 
 ## Deep Docs
 
diff --git a/oor/CLAUDE.md b/oor/CLAUDE.md
index bb983c14..158ce181 100644
--- a/oor/CLAUDE.md
+++ b/oor/CLAUDE.md
@@ -122,6 +122,23 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/oor.<Sym
 - Server-side lineage-cap rejection surfaces as a typed `*ErrLineageTooLarge`
   via `ClassifySubmitError`, so wallet callers can switch on the cause
   without depending on the `oorpb` proto type.
+- A supervised restart is not a no-op for either behavior. The framework reuses
+  the same behavior instance across a restart, so whatever the panic left in
+  goroutine-owned memory outlives it unless the `actor.RestartMessage` handler
+  rebuilds from durable truth. `sessionBehavior` arms the same `commitFailed`
+  reload guard its failed-commit path uses, so the next turn stops the stale
+  FSM and rebuilds it from the registry row before dispatching; the rebuild is
+  deferred to that turn because `restore()` starts an FSM goroutine that must
+  outlive the turn, and because the framework delivers the restart message with
+  `MaxAttempts` 1 and dead-letters (rather than retries) a restart turn that
+  fails. `oorRegistryBehavior` instead restores inline: it drops the staged
+  `pendingHandoff` (it belonged to the turn that died) and re-runs
+  `handleRestoreNonTerminal`, which is cheap and idempotent because it skips
+  sessions that are already resident.
+- `OORDurableMsg` is unsealed (`actor.TLVMessage` alone), so the shared codec
+  registers the framework's `actor.RestartMessage` directly under
+  `actor.RestartTLVType`; no package-local adapter type is needed for the
+  decoded message to satisfy the mailbox's cast.
 
 ## Deep Docs
 
diff --git a/serverconn/AGENTS.md b/serverconn/AGENTS.md
index a88da6d0..22e40275 100644
--- a/serverconn/AGENTS.md
+++ b/serverconn/AGENTS.md
@@ -100,6 +100,21 @@ background ingress polling with event routing.
   (`validateInboundEnvelope`); a mismatch is always a permanent
   `*mailboxconn.StatusError` — there is no legacy zero-version fallback,
   since client and operator are always deployed with a negotiated version.
+- A supervised restart of the egress actor has nothing to rebuild, and
+  `ServerConnectionActor.Receive` consumes the restart message as an explicit
+  no-op to say so. Every egress turn reads its whole input from the durable
+  message, converts it, sends it over the edge, and consumes the message in one
+  Commit; nothing carries over between turns. The in-memory state the connector
+  does hold belongs to the CONNECTION, not to a turn (the unary response
+  registry, the last-send timestamp, the cached incompatibility, the ingress
+  cancel), and it is owned by the ingress loop and its callers, so a panicking
+  egress turn cannot leave any of it half-written.
+- `ServerConnMsg` is sealed, so `NewServerConnCodec` registers a package-local
+  `restartMsg` adapter (embedding `actor.RestartMessage`) under
+  `actor.RestartTLVType` rather than the bare framework type. The durable
+  mailbox casts every decoded message to `ServerConnMsg`; an unadapted restart
+  would fail that cast and dead-letter instead of being handled. Any future
+  sealed message surface that adopts supervision needs the same adapter.
 
 ## Deep Docs
 
diff --git a/serverconn/CLAUDE.md b/serverconn/CLAUDE.md
index a88da6d0..22e40275 100644
--- a/serverconn/CLAUDE.md
+++ b/serverconn/CLAUDE.md
@@ -100,6 +100,21 @@ background ingress polling with event routing.
   (`validateInboundEnvelope`); a mismatch is always a permanent
   `*mailboxconn.StatusError` — there is no legacy zero-version fallback,
   since client and operator are always deployed with a negotiated version.
+- A supervised restart of the egress actor has nothing to rebuild, and
+  `ServerConnectionActor.Receive` consumes the restart message as an explicit
+  no-op to say so. Every egress turn reads its whole input from the durable
+  message, converts it, sends it over the edge, and consumes the message in one
+  Commit; nothing carries over between turns. The in-memory state the connector
+  does hold belongs to the CONNECTION, not to a turn (the unary response
+  registry, the last-send timestamp, the cached incompatibility, the ingress
+  cancel), and it is owned by the ingress loop and its callers, so a panicking
+  egress turn cannot leave any of it half-written.
+- `ServerConnMsg` is sealed, so `NewServerConnCodec` registers a package-local
+  `restartMsg` adapter (embedding `actor.RestartMessage`) under
+  `actor.RestartTLVType` rather than the bare framework type. The durable
+  mailbox casts every decoded message to `ServerConnMsg`; an unadapted restart
+  would fail that cast and dead-letter instead of being handled. Any future
+  sealed message surface that adopts supervision needs the same adapter.
 
 ## Deep Docs
 
diff --git a/unroll/AGENTS.md b/unroll/AGENTS.md
index f7ed9727..e192e2ac 100644
--- a/unroll/AGENTS.md
+++ b/unroll/AGENTS.md
@@ -305,6 +305,22 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/unroll.<
   operator-sourced OOR artifacts flow into proof assembly, so a
   zero- or short-output node maps to a retryable error rather than
   a goroutine panic.
+- **A supervised restart restores the checkpoint inline.** The
+  framework reuses the same `behavior` instance across a restart, so
+  a panic that advanced `b.pending` or `b.sweepTx` past the last
+  `Staged` checkpoint would otherwise survive it. The `restartMsg`
+  case re-runs the constructor's `restoreCheckpoint`, overwriting
+  both from the durable row. It drives no FSM transition and writes
+  nothing, so it returns before the Commit like the status probe.
+  The restore gets exactly one shot: the framework delivers the
+  restart message with `MaxAttempts` 1 and dead-letters (rather than
+  retries) a restart turn that fails.
+- **The restart message needs a sealed adapter.** `Msg` is sealed, so
+  `newCodec` registers a package-local `restartMsg` (embedding
+  `actor.RestartMessage`) under `actor.RestartTLVType` rather than
+  the bare framework type. The durable mailbox casts every decoded
+  message to `Msg`; an unadapted restart would fail that cast and
+  dead-letter instead of restoring.
 
 ## Deep Docs
 
diff --git a/unroll/CLAUDE.md b/unroll/CLAUDE.md
index f7ed9727..e192e2ac 100644
--- a/unroll/CLAUDE.md
+++ b/unroll/CLAUDE.md
@@ -305,6 +305,22 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/unroll.<
   operator-sourced OOR artifacts flow into proof assembly, so a
   zero- or short-output node maps to a retryable error rather than
   a goroutine panic.
+- **A supervised restart restores the checkpoint inline.** The
+  framework reuses the same `behavior` instance across a restart, so
+  a panic that advanced `b.pending` or `b.sweepTx` past the last
+  `Staged` checkpoint would otherwise survive it. The `restartMsg`
+  case re-runs the constructor's `restoreCheckpoint`, overwriting
+  both from the durable row. It drives no FSM transition and writes
+  nothing, so it returns before the Commit like the status probe.
+  The restore gets exactly one shot: the framework delivers the
+  restart message with `MaxAttempts` 1 and dead-letters (rather than
+  retries) a restart turn that fails.
+- **The restart message needs a sealed adapter.** `Msg` is sealed, so
+  `newCodec` registers a package-local `restartMsg` (embedding
+  `actor.RestartMessage`) under `actor.RestartTLVType` rather than
+  the bare framework type. The durable mailbox casts every decoded
+  message to `Msg`; an unadapted restart would fail that cast and
+  dead-letter instead of restoring.
 
 ## Deep Docs
```

**How to apply:** save the diff above and `git apply` it, or run the doc-gardening skill locally (`/doc-gardening` scoped to `credit oor serverconn unroll`) to regenerate the same updates. Each `CLAUDE.md` must stay byte-identical to its sibling `AGENTS.md` (`make doc-check` enforces the pair).

---
<sub>Posted by the per-PR doc-gardening advisory. Advisory only — this does not block merge. Run log: https://github.com/lightninglabs/wavelength/pull/1123/checks</sub>

Roasbeef:
An adversarial audit pass ran against the first 5 commits; the 13 commits on top address what it found. The audit's headline was blunt and correct: the kernel mechanics held up, but the two central product claims did not. Dispositions:

**Fixed:**

1. **Restart-from-checkpoint was vacuous for every production adopter (major).** The behavior instance is reused across generations, and all six adopters either no-op'd the RestartMessage or could not decode it, so the panic path bypassed their existing commitFailed reload guards and "restart" restored nothing. Now every adopter is wired for real: credit and the OOR session arm their commitFailed reload seam (deferred, because restore starts goroutines that must outlive the turn), the OOR registry clears the dead turn's pending hand-off and re-runs its idempotent restore, unroll registers a sealed RestartMessage adapter and re-runs restoreCheckpoint (its old default arm would have nack-looped the restart), and serverconn's egress gets the adapter plus a documented no-op with the statelessness argument spelled out. A regression test drives a Right-path behavior into in-memory divergence, panics it, and proves the reload (verified non-vacuous: reverting the guard fails it). The docs now state the honest contract: the instance is reused, and the clean slate exists exactly where the handler rebuilds from the checkpoint.

2. **Default-on intensity was a silent kill switch (major).** One poison Tell burned the whole 5/60s budget through the retry schedule; two in a minute permanently killed an actor nothing observes (no production Watch callers). Reversed: the default is now UnlimitedRestarts (zero normalizes to unlimited), with RecommendedMaxRestarts = 5 for owners that opt in AND wire Watch; the config GoDoc states the trade explicitly.

3. **Bare OnStop could crash the process (major).** runStopHook now recovers panics into a termination (never a process crash), is idempotent per generation (deduping the double call on the restart-failed path), and the Stoppable contract is rewritten for mid-life, more-than-once invocation.

4. **The classic tx path committed a panicking turn's torn writes together with the nack (major),** so the checkpoint the restart restored could contain the corruption. The panic now forces rollback of the behavior's writes inside the tx closure, with the nack redone detached and bounded so the poison message still burns its attempt.

5. Ordering and hygiene: the codec Supports probe runs before teardown (an unsupported codec degrades to generation-cycling without a mid-life OnStop), restart rows are deduped to at most one pending per mailbox and can no longer strand past max_attempts (a failed restart turn dead-letters), Watch on a never-started actor is well-defined and leak-free, and the two dead-code/race notes carry comments.

Deliberately unchanged from the first cut: no Watch adopters outside baselib/actor, no supervision for the non-durable path, one-for-one only, no inter-restart backoff.

Roasbeef:
@claude review this

Roasbeef:
P1 resolved in 26b6248 with the barrier rather than the doc-narrowing minimum, plus 60e2221 (tests) and c0fc477 (docs).

A generation now launches one warm-up worker and holds the remaining NumWorkers-1 behind a barrier until the restart hand-off resolves. The barrier is in-process and edge-triggered rather than store-polled, for a load-bearing reason: the restart row deliberately does NOT use a deterministic per-actor ID (a fixed ID would collide with processed_messages and dedup-skip the next restart's restore), so there is no stable key to poll. Instead the warm-up worker classifies its first claim: a RestartMessage holds the barrier across the whole restore turn; anything else opens it immediately (RestartPriority would have won the claim, so no hand-off was pending). On the supervised-restart path the guarantee is exact (a required flag makes the barrier ignore the idle tick until the worker's first claim, closing a fan-out race found during implementation); the boot hand-off remains first-claim-ordered plus idle-tick best-effort, and the docs state that distinction rather than paper over it. Deferred open on every worker exit covers panic, closed mailbox, and cancelled generation; a dead-lettered restore, the codec-unsupported degrade, and Stop mid-barrier all release cleanly. Single-worker actors get a nil barrier and an unchanged path.

The barrier test was verified non-vacuous (disabling the barrier flips the central never-assertion), which required first making the mock store's claim order priority-aware like the real query. The invariant wording in the package guides now names the barrier instead of overclaiming RestartPriority alone.

CI re-running on the new head.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #1136, opened 2026-08-12
waved: authenticate outbound mailbox RPCs with the auth header

Roasbeef:
## What this does

The operator authorizes a mailbox RPC one of two ways: from the TLS client
certificate bound to the caller's mailbox ID at registration, or from the
`x-mailbox-auth-sig` metadata header (`verifyMailboxMetadataAuth`). An operator
that terminates TLS at a load balancer never sees a client certificate, so the
header is the only path left, and lumos is adding the knob that requires it
(lightninglabs/lumos#780).

We send that header on exactly one transport today, and it is the wrong one.
`NewAuthenticatedMailboxClient` wraps the swap server edge in
`swapclientserver/service.go`, while the operator edge this daemon actually
talks Ark over is built raw in `connectOperatorClients`: the gRPC arm
(`outbound_clients.go:41`) and the REST arm (`:62`) both hand back an unwrapped
client, and `dialServer` installs only the metrics interceptors.

The signature we *do* compute (`server.go:3474`) goes into **envelope headers**,
which is the first-contact `Send` proof the server verifies in
`HandleUnknownClient` during registration. That is a different mechanism from
the per-RPC gRPC metadata the interceptor reads. `Pull` and `AckUpTo` carried
nothing.

This PR wraps both operator arms so every `Send`, `Pull` and `AckUpTo` carries
the header.

## Why sign unconditionally

Signing is not keyed off the operator's TLS posture. The server checks the
header first and still falls through to the certificate binding when no header
is present, so a client that always signs works against either deployment shape
without having to learn which one it is talking to. Making it conditional would
put the operator's TLS choice into client config for no gain.

## Why memoize

The digest is `TaggedHash("mailbox-auth", identityPubKey || recipientMailboxID)`
and covers nothing request-specific, so one signature stays valid for the life
of the key. `signMailboxAuth` costs a round trip to whichever wallet backend
holds that key, so signing per call would put a wallet round trip in front of
every `Pull` on a long-poll loop.

The runtime setup path now takes the same memo, so the two signatures it and the
edge would otherwise compute for the same recipient collapse to one. The daemon
addresses one recipient mailbox in practice, so the map holds a single entry.

## Sequencing

**This has to land and roll out before an operator sets
`mailbox.requiredirectauth`.** That knob rejects any mailbox RPC that no TLS
binding and no valid header authorizes, and `isMailboxRPC` covers all three
methods, so an operator who set it against a client built before this PR would
reject every mailbox call its own clients make. `mailbox.allowdirectauth` is the
staging step: it authenticates the clients that sign and changes nothing for the
ones that do not.

## Verification

- `go build ./waved/... ./serverconn/...`, `go vet ./waved/` — clean. (`go build
  ./...` reports `function main is undeclared in the main package` for the repo
  root both with and without this change; confirmed pre-existing by stashing.)
- `go test ./waved/ ./serverconn/ -count=1` — pass, and `-race` on the
  `TestConnectOperatorClients` cases.

`TestConnectOperatorClientsREST` needed updating: it called `Pull` with an empty
`PullRequest`, which the wrapper now rejects with "mailbox recipient is
required" before dialing. It now addresses a real compound mailbox, primes the
memo to stand in for the wallet backend, and additionally asserts the header
actually reached the server — so it covers the REST arm rather than only
proving the transport is wired.

## Context

- lightninglabs/lumos#780 — the daemon-side knob this makes usable
- lightninglabs/lightning-infra#3265 — the deployment topology behind it

### DIFF
--- a/serverconn/mailbox_auth.go
+++ b/serverconn/mailbox_auth.go
@@ -43,8 +43,24 @@ const MailboxTLSBindTagStr = "mailbox-tls-bind"
 //
 //	senderCompressedPubKey || recipientMailboxID
 //
-// Including the recipient mailbox ID (the server's pubkey-derived ID)
-// prevents cross-server replay of the signature.
+// Including the recipient mailbox ID binds the signature to what it
+// addresses, but how much that buys depends on the caller, and only
+// Send gets the strong version:
+//
+//   - Send passes the compound operator:client recipient, which
+//     embeds the operator's pubkey-derived ID, so a Send signature is
+//     useless at any other operator.
+//   - Pull and AckUpTo pass the client's own plain mailbox ID
+//     (ingress.go, via ConnectorConfig.LocalMailboxID), which carries
+//     no operator component. That digest is identical at every
+//     operator, so one Pull/Ack signature authorizes that client's
+//     mailbox at every operator its identity key is known to. Identity
+//     keys are a deterministic derivation, so a wallet driving two
+//     operators presents the same credential to both.
+//
+// Do not read this function as preventing cross-server replay on its
+// own. Closing the Pull/Ack case means folding a server identity into
+// the digest, which is a wire change on both sides.
 func MailboxAuthMessage(senderPubKey *btcec.PublicKey,
 	recipientMailboxID string) []byte {
 
--- a/waved/outbound_clients.go
+++ b/waved/outbound_clients.go
@@ -110,8 +110,12 @@ func (s *Server) mailboxAuthSigner() serverconn.MailboxAuthSigner {
 // Memoizing is what makes signing per RPC affordable: the digest is
 // TaggedHash("mailbox-auth", identityPubKey || recipientMailboxID), so it does
 // not vary with the request, while signMailboxAuth costs a round-trip to
-// whichever wallet backend holds the key. The daemon addresses one recipient
-// mailbox in practice, so the map holds a single entry.
+// whichever wallet backend holds the key.
+//
+// The map holds two entries, not one. Send addresses the compound
+// operator:client mailbox, while Pull and AckUpTo address this client's own
+// plain mailbox ID, so the two arms sign different recipients and cache
+// separately. Keying on the recipient is what keeps them from colliding.
 func (s *Server) mailboxAuthSig(ctx context.Context,
 	recipientMailboxID string) (*schnorr.Signature, error) {
 
--- a/waved/outbound_clients.go
+++ b/waved/outbound_clients.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/lightninglabs/wavelength/arkrpc"
@@ -18,6 +19,12 @@ import (
 	"github.com/lightninglabs/wavelength/serverconn"
 	"google.golang.org/grpc"
 )
+
+// mailboxAuthSignTimeout bounds a single mailbox auth signing round trip to
+// the wallet backend. A mailbox RPC that cannot get its signature in this long
+// should fail on its own rather than hold the caller open, since the ingress
+// puller hands down a context with no deadline of its own.
+const mailboxAuthSignTimeout = 30 * time.Second
 
 // operatorClients holds the daemon-owned outbound clients used to talk to the
 // Ark operator directly and through the mailbox edge.
@@ -89,8 +96,13 @@ func (s *Server) connectOperatorClients() (*operatorClients, error) {
 }
 
 // mailboxAuthSigner returns the signer that stamps x-mailbox-auth-sig on
-// outbound mailbox RPCs. It reads the identity key lazily, per RPC, because
-// the mailbox edge is built before the wallet has derived one.
+// outbound mailbox RPCs.
+//
+// The identity key is read per RPC rather than captured at construction.
+// deriveIdentityKeyEarly runs before connectOperatorClients and fails the
+// startup path if it cannot derive, so the key is in fact already there by
+// the time the edge is built; reading it late costs nothing and keeps this
+// closure from pinning a descriptor that startup has not finished with.
 func (s *Server) mailboxAuthSigner() serverconn.MailboxAuthSigner {
 	return func(ctx context.Context, recipientMailboxID string) (string,
 		error) {
@@ -116,6 +128,16 @@ func (s *Server) mailboxAuthSigner() serverconn.MailboxAuthSigner {
 // operator:client mailbox, while Pull and AckUpTo address this client's own
 // plain mailbox ID, so the two arms sign different recipients and cache
 // separately. Keying on the recipient is what keeps them from colliding.
+//
+// The wallet call deliberately happens with the mutex released. Holding it
+// across the round trip would serialize the whole mailbox edge behind one
+// signature: sync.Mutex.Lock is not context-aware, so the egress workers and
+// the ingress puller would block on it with their own deadlines silently
+// ignored, and a wedged wallet would stall egress, ingress, heartbeat and ack
+// together with nothing in the logs to say why. Two callers racing the same
+// cold recipient may both sign, which is harmless — the digest is
+// deterministic, so they produce the same signature and the second store is a
+// no-op.
 func (s *Server) mailboxAuthSig(ctx context.Context,
 	recipientMailboxID string) (*schnorr.Signature, error) {
 
@@ -125,19 +147,30 @@ func (s *Server) mailboxAuthSig(ctx context.Context,
 	}
 
 	s.mailboxAuthSigsMu.Lock()
-	defer s.mailboxAuthSigsMu.Unlock()
+	sig, ok := s.mailboxAuthSigs[recipientMailboxID]
+	s.mailboxAuthSigsMu.Unlock()
 
-	if sig, ok := s.mailboxAuthSigs[recipientMailboxID]; ok {
+	if ok {
 		return sig, nil
 	}
 
-	sig, err := s.signMailboxAuth(ctx, recipientMailboxID)
+	// Bound the signing round trip rather than inheriting the caller's
+	// context. The ingress puller builds its context with no deadline at
+	// all, so a hung wallet would otherwise park this call forever, and
+	// the memo miss is on the path every Pull takes.
+	signCtx, cancel := context.WithTimeout(ctx, mailboxAuthSignTimeout)
+	defer cancel()
+
+	sig, err := s.signMailboxAuth(signCtx, recipientMailboxID)
 	if err != nil {
 		return nil, err
 	}
 
+	s.mailboxAuthSigsMu.Lock()
+	defer s.mailboxAuthSigsMu.Unlock()
+
 	if s.mailboxAuthSigs == nil {
-		s.mailboxAuthSigs = make(map[string]*schnorr.Signature, 1)
+		s.mailboxAuthSigs = make(map[string]*schnorr.Signature, 2)
 	}
 	s.mailboxAuthSigs[recipientMailboxID] = sig
 
--- a/waved/outbound_clients.go
+++ b/waved/outbound_clients.go
@@ -1,13 +1,16 @@
 package waved
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"fmt"
 	"net/http"
 	"os"
 	"strings"
 
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/lightninglabs/wavelength/arkrpc"
 	mailboxpb "github.com/lightninglabs/wavelength/mailbox/pb"
 	"github.com/lightninglabs/wavelength/rpc/restclient"
@@ -27,7 +30,16 @@ type operatorClients struct {
 
 // connectOperatorClients builds the outbound clients for the configured
 // operator transport.
+//
+// Both mailbox edges are wrapped so every Send, Pull and AckUpTo carries the
+// x-mailbox-auth-sig metadata header. The operator authorizes a mailbox RPC
+// either from the TLS client certificate bound to the caller's mailbox ID or
+// from that header, and an operator that terminates TLS at a proxy never sees
+// a client certificate. Signing unconditionally means the daemon works against
+// either posture without the operator's TLS choice leaking into client config.
 func (s *Server) connectOperatorClients() (*operatorClients, error) {
+	sign := s.mailboxAuthSigner()
+
 	switch s.cfg.Server.Transport {
 	case "", RPCTransportGRPC:
 		conn, err := s.dialServer()
@@ -36,9 +48,11 @@ func (s *Server) connectOperatorClients() (*operatorClients, error) {
 		}
 
 		return &operatorClients{
-			conn:    conn,
-			ark:     arkrpc.NewArkServiceClient(conn),
-			mailbox: mailboxpb.NewMailboxServiceClient(conn),
+			conn: conn,
+			ark:  arkrpc.NewArkServiceClient(conn),
+			mailbox: serverconn.NewAuthenticatedMailboxClient(
+				mailboxpb.NewMailboxServiceClient(conn), sign,
+			),
 			cleanup: conn.Close,
 		}, nil
 
@@ -59,8 +73,11 @@ func (s *Server) connectOperatorClients() (*operatorClients, error) {
 			ark: restclient.NewArkServiceClientFromClient(
 				transport,
 			),
-			mailbox: restclient.NewMailboxServiceClientFromClient(
-				transport,
+			mailbox: serverconn.NewAuthenticatedMailboxClient(
+				restclient.NewMailboxServiceClientFromClient(
+					transport,
+				),
+				sign,
 			),
 			cleanup: func() error { return nil },
 		}, nil
@@ -69,6 +86,58 @@ func (s *Server) connectOperatorClients() (*operatorClients, error) {
 		return nil, fmt.Errorf("unknown server transport %q",
 			s.cfg.Server.Transport)
 	}
+}
+
+// mailboxAuthSigner returns the signer that stamps x-mailbox-auth-sig on
+// outbound mailbox RPCs. It reads the identity key lazily, per RPC, because
+// the mailbox edge is built before the wallet has derived one.
+func (s *Server) mailboxAuthSigner() serverconn.MailboxAuthSigner {
+	return func(ctx context.Context, recipientMailboxID string) (string,
+		error) {
+
+		sig, err := s.mailboxAuthSig(ctx, recipientMailboxID)
+		if err != nil {
+			return "", err
+		}
+
+		return hex.EncodeToString(sig.Serialize()), nil
+	}
+}
+
+// mailboxAuthSig returns the mailbox auth signature for recipientMailboxID,
+// signing once per recipient and serving every later call from memory.
+//
+// Memoizing is what makes signing per RPC affordable: the digest is
+// TaggedHash("mailbox-auth", identityPubKey || recipientMailboxID), so it does
+// not vary with the request, while signMailboxAuth costs a round-trip to
+// whichever wallet backend holds the key. The daemon addresses one recipient
+// mailbox in practice, so the map holds a single entry.
+func (s *Server) mailboxAuthSig(ctx context.Context,
+	recipientMailboxID string) (*schnorr.Signature, error) {
+
+	if s.clientKeyDesc.PubKey == nil {
+		return nil, fmt.Errorf("identity key not yet derived; wallet " +
+			"not ready")
+	}
+
+	s.mailboxAuthSigsMu.Lock()
+	defer s.mailboxAuthSigsMu.Unlock()
+
+	if sig, ok := s.mailboxAuthSigs[recipientMailboxID]; ok {
+		return sig, nil
+	}
+
+	sig, err := s.signMailboxAuth(ctx, recipientMailboxID)
+	if err != nil {
+		return nil, err
+	}
+
+	if s.mailboxAuthSigs == nil {
+		s.mailboxAuthSigs = make(map[string]*schnorr.Signature, 1)
+	}
+	s.mailboxAuthSigs[recipientMailboxID] = sig
+
+	return sig, nil
 }
 
 // serverClientTLSCerts returns the optional client certificate used by the
--- a/waved/outbound_clients.go
+++ b/waved/outbound_clients.go
@@ -124,10 +124,18 @@ func (s *Server) mailboxAuthSigner() serverconn.MailboxAuthSigner {
 // not vary with the request, while signMailboxAuth costs a round-trip to
 // whichever wallet backend holds the key.
 //
-// The map holds two entries, not one. Send addresses the compound
-// operator:client mailbox, while Pull and AckUpTo address this client's own
-// plain mailbox ID, so the two arms sign different recipients and cache
-// separately. Keying on the recipient is what keeps them from colliding.
+// There is one entry per distinct recipient, which is more than the operator
+// edge alone accounts for. That edge contributes two: Send addresses the
+// compound operator:client mailbox while Pull and AckUpTo address this
+// client's own plain mailbox ID, so the two arms sign different recipients and
+// must not collide. RPCServer.SignMailboxAuth also routes here, and the swap
+// edge behind it addresses a per-swap mailbox (client:payment_hash), so a
+// long-lived daemon accumulates one further entry per out-swap. Each is a
+// mailbox ID and a 64-byte signature and none is ever evicted, so the map
+// grows with swaps performed rather than staying at a fixed size. Bounding it
+// would need an eviction policy, which is not worth the machinery at the sizes
+// involved — but it is growth, not a constant, and a reader should know that
+// before assuming otherwise.
 //
 // The wallet call deliberately happens with the mutex released. Holding it
 // across the round trip would serialize the whole mailbox edge behind one
--- a/waved/server_outbound_clients_test.go
+++ b/waved/server_outbound_clients_test.go
@@ -353,6 +353,12 @@ func TestMailboxAuthSigMemo(t *testing.T) {
 	_, err = s.mailboxAuthSig(t.Context(), compound)
 	require.ErrorContains(t, err, "no wallet backend available")
 
+	// A failed sign caches nothing. Caching it would pin the failure for
+	// the life of the process, so every later RPC to that mailbox would
+	// fail without ever retrying the wallet.
+	require.Len(t, s.mailboxAuthSigs, 1)
+	require.NotContains(t, s.mailboxAuthSigs, compound)
+
 	// The miss must not have disturbed the entry that was already there.
 	got, err = s.mailboxAuthSig(t.Context(), plain)
 	require.NoError(t, err)
--- a/waved/server_outbound_clients_test.go
+++ b/waved/server_outbound_clients_test.go
@@ -1,15 +1,20 @@
 package waved
 
 import (
+	"encoding/hex"
 	"net/http"
 	"net/http/httptest"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/lightninglabs/wavelength/arkrpc"
 	mailboxpb "github.com/lightninglabs/wavelength/mailbox/pb"
 	"github.com/lightninglabs/wavelength/rpcauth"
+	"github.com/lightninglabs/wavelength/serverconn"
+	"github.com/lightningnetwork/lnd/keychain"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/encoding/protojson"
 	"gopkg.in/macaroon-bakery.v2/bakery"
@@ -38,6 +43,19 @@ func TestConnectOperatorClientsREST(t *testing.T) {
 	macHex, err := rpcauth.HexFromFile(macaroonPath)
 	require.NoError(t, err)
 
+	// Record the mailbox auth header the REST edge sent, guarded because
+	// the handler runs on the test server's goroutine.
+	var (
+		authSigMu  sync.Mutex
+		gotAuthSig string
+	)
+	pulledAuthSig := func() string {
+		authSigMu.Lock()
+		defer authSigMu.Unlock()
+
+		return gotAuthSig
+	}
+
 	server := httptest.NewServer(
 		http.HandlerFunc(
 			func(w http.ResponseWriter, r *http.Request) {
@@ -63,6 +81,12 @@ func TestConnectOperatorClientsREST(t *testing.T) {
 					)
 
 				case "/v1/mailbox/pull":
+					authSigMu.Lock()
+					gotAuthSig = r.Header.Get(
+						serverconn.AuthHeaderKey,
+					)
+					authSigMu.Unlock()
+
 					msg, marshalErr = protojson.Marshal(
 						&mailboxpb.PullResponse{},
 					)
@@ -81,6 +105,20 @@ func TestConnectOperatorClientsREST(t *testing.T) {
 	)
 	defer server.Close()
 
+	// The mailbox edge signs every call with the daemon identity key, so
+	// the fixture needs one. Priming the memo stands in for the wallet
+	// backend that would produce the signature in a running daemon.
+	clientKey, err := btcec.NewPrivateKey()
+	require.NoError(t, err)
+
+	clientMailbox := serverconn.CompoundMailboxID(
+		serverconn.PubKeyMailboxID(operatorKey.PubKey()),
+		serverconn.PubKeyMailboxID(clientKey.PubKey()),
+	)
+
+	authSig, err := serverconn.SignMailboxAuth(clientKey, clientMailbox)
+	require.NoError(t, err)
+
 	s := &Server{
 		cfg: &Config{
 			Server: &ServerConfig{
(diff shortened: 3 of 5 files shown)


litbot-9000:
This PR's Go changes leave the per-package docs for **`serverconn`** and **`waved`** stale — the mailbox auth edge now signs `Pull` and `AckUpTo` as well as `Send`, the signature is memoized per recipient, and the digest's cross-server binding is weaker than the existing docs imply.

```diff
diff --git a/serverconn/AGENTS.md b/serverconn/AGENTS.md
index a88da6d0..9f2066da 100644
--- a/serverconn/AGENTS.md
+++ b/serverconn/AGENTS.md
@@ -14,7 +14,7 @@ background ingress polling with event routing.
 - `UnaryFacade` — Implements `mailboxrpc.RPCClient` for generated RPC stubs (low-latency path). Bounded waits come from the caller's context plus the response registry TTL; there is no separate timeout entry point. `SendRPC` gates on `ConnectorConfig.MaxInFlightUnary` and fails a send that would exceed it with `codes.ResourceExhausted`.
 - `ConnectorConfig` — Wiring configuration (edge address, mailbox IDs, dispatchers, store, durable unary builder, `EgressWorkers`, `MaxInFlightUnary`). `EgressWorkers` sizes the egress worker pool (default `DefaultEgressWorkers` = 4); `<= 1` keeps the legacy single sender. The `DurableUnaryBuilder` field must be set to handle `DurableUnaryQuery` message types; otherwise those messages are rejected. The `AuthSignature` field holds the Schnorr auth sig injected into every outbound envelope via `mergeAuthHeaders` (auth header always wins over caller-provided headers).
 - `PubKeyMailboxID` — Derives canonical mailbox ID from a public key (hex-encoded compressed SEC). Panics on nil.
-- `MailboxAuthDigest` / `MailboxAuthMessage` — BIP-340 tagged hash digest construction for mailbox auth signatures. Uses `chainhash.TaggedHash` with the `MailboxAuthTagStr` domain separator.
+- `MailboxAuthDigest` / `MailboxAuthMessage` — BIP-340 tagged hash digest construction for mailbox auth signatures. Uses `chainhash.TaggedHash` with the `MailboxAuthTagStr` domain separator. The digest is `senderCompressedPubKey || recipientMailboxID` and nothing else, so it does not vary per request and how much cross-server binding it buys depends entirely on which recipient the caller passes (see Invariants).
 - `SignMailboxAuth` / `VerifyMailboxAuth` / `ParseMailboxPubKey` — Schnorr sign/verify helpers for pubkey-derived mailbox identity.
 - `AuthHeaderKey` — Envelope header key (`x-mailbox-auth-sig`) for the Schnorr auth signature.
 - `GenerateClientTLSCert` — Creates an ephemeral P-256 mTLS client cert with the secp256k1 identity pubkey hex as Subject CN. Returns error on nil key.
@@ -28,9 +28,16 @@ background ingress polling with event routing.
   secp256k1 identity via a BIP-340 Schnorr signature, complementing
   `GenerateClientTLSCert` (the cert alone proves nothing; this signature
   proves the TLS key and the identity key are held by the same party).
+- `MailboxAuthSigner` — `func(ctx, recipientMailboxID) (hexSig, error)`; the
+  seam `NewAuthenticatedMailboxClient` calls to mint the header value. The
+  signer is invoked lazily, per RPC, so the mailbox edge can be built before
+  the wallet has derived an identity key.
 - `NewAuthenticatedMailboxClient` — `mailboxpb.MailboxServiceClient` decorator
-  that signs and attaches the `x-mailbox-auth-sig` header to every `Send`
-  before forwarding to the wrapped edge transport.
+  that signs and attaches the `x-mailbox-auth-sig` header to every `Send`,
+  `Pull` and `AckUpTo` before forwarding to the wrapped edge transport. The
+  recipient signed over is the envelope recipient for `Send` and `MailboxId`
+  for `Pull`/`AckUpTo`. A nil signer returns `next` unchanged, which is how
+  unauthenticated test transports opt out.
 - `AckState` — Four-cursor watermark state machine (PullCursor, DispatchCommittedTo, AckTarget, AckCommittedTo).
 - `SendUnaryRequest` — Durable typed unary request that becomes a real unary RPC after commit. The response arrives via KIND_RESPONSE and, if no in-memory waiter exists, falls back to durable route dispatch via the EventRouter.
 - `DurableUnaryRequestBuilder` — Interface for proof-gated request-body construction. Implementations build the actual proto request (e.g., with signed proofs) at send time, not at persist time. The interface is provided via `ConnectorConfig.DurableUnaryBuilder`.
@@ -93,6 +100,18 @@ background ingress polling with event routing.
   per-key FIFO lane key regardless of whether the message was constructed
   fresh or decoded from TLV. The `cachedCorrelationKey` field is populated
   during `Decode` via `tlv.TlvType8` so restarts do not lose FIFO routing.
+- `MailboxAuthMessage` does **not** prevent cross-server replay on its own,
+  and the two callers differ. `Send` passes the compound
+  `operator:client` recipient, which embeds the operator's pubkey-derived
+  ID, so a `Send` signature is useless at any other operator. `Pull` and
+  `AckUpTo` pass the client's own plain mailbox ID (`ingress.go`, via
+  `ConnectorConfig.LocalMailboxID`), which carries no operator component —
+  that digest is identical at every operator, so one `Pull`/`Ack` signature
+  authorizes that client's mailbox at every operator its identity key is
+  known to, and identity keys are a deterministic derivation, so a wallet
+  driving two operators presents the same credential to both. Closing the
+  `Pull`/`Ack` case means folding a server identity into the digest, which
+  is a wire change on both sides.
 - Every outbound envelope is stamped with the runtime-bound mailbox
   transport and Ark protocol version pair (`stampEnvelopeVersions`/
   `versionStampingMailboxClient`), overwriting any caller-provided value.
diff --git a/serverconn/CLAUDE.md b/serverconn/CLAUDE.md
index a88da6d0..9f2066da 100644
--- a/serverconn/CLAUDE.md
+++ b/serverconn/CLAUDE.md
@@ -14,7 +14,7 @@ background ingress polling with event routing.
 - `UnaryFacade` — Implements `mailboxrpc.RPCClient` for generated RPC stubs (low-latency path). Bounded waits come from the caller's context plus the response registry TTL; there is no separate timeout entry point. `SendRPC` gates on `ConnectorConfig.MaxInFlightUnary` and fails a send that would exceed it with `codes.ResourceExhausted`.
 - `ConnectorConfig` — Wiring configuration (edge address, mailbox IDs, dispatchers, store, durable unary builder, `EgressWorkers`, `MaxInFlightUnary`). `EgressWorkers` sizes the egress worker pool (default `DefaultEgressWorkers` = 4); `<= 1` keeps the legacy single sender. The `DurableUnaryBuilder` field must be set to handle `DurableUnaryQuery` message types; otherwise those messages are rejected. The `AuthSignature` field holds the Schnorr auth sig injected into every outbound envelope via `mergeAuthHeaders` (auth header always wins over caller-provided headers).
 - `PubKeyMailboxID` — Derives canonical mailbox ID from a public key (hex-encoded compressed SEC). Panics on nil.
-- `MailboxAuthDigest` / `MailboxAuthMessage` — BIP-340 tagged hash digest construction for mailbox auth signatures. Uses `chainhash.TaggedHash` with the `MailboxAuthTagStr` domain separator.
+- `MailboxAuthDigest` / `MailboxAuthMessage` — BIP-340 tagged hash digest construction for mailbox auth signatures. Uses `chainhash.TaggedHash` with the `MailboxAuthTagStr` domain separator. The digest is `senderCompressedPubKey || recipientMailboxID` and nothing else, so it does not vary per request and how much cross-server binding it buys depends entirely on which recipient the caller passes (see Invariants).
 - `SignMailboxAuth` / `VerifyMailboxAuth` / `ParseMailboxPubKey` — Schnorr sign/verify helpers for pubkey-derived mailbox identity.
 - `AuthHeaderKey` — Envelope header key (`x-mailbox-auth-sig`) for the Schnorr auth signature.
 - `GenerateClientTLSCert` — Creates an ephemeral P-256 mTLS client cert with the secp256k1 identity pubkey hex as Subject CN. Returns error on nil key.
@@ -28,9 +28,16 @@ background ingress polling with event routing.
   secp256k1 identity via a BIP-340 Schnorr signature, complementing
   `GenerateClientTLSCert` (the cert alone proves nothing; this signature
   proves the TLS key and the identity key are held by the same party).
+- `MailboxAuthSigner` — `func(ctx, recipientMailboxID) (hexSig, error)`; the
+  seam `NewAuthenticatedMailboxClient` calls to mint the header value. The
+  signer is invoked lazily, per RPC, so the mailbox edge can be built before
+  the wallet has derived an identity key.
 - `NewAuthenticatedMailboxClient` — `mailboxpb.MailboxServiceClient` decorator
-  that signs and attaches the `x-mailbox-auth-sig` header to every `Send`
-  before forwarding to the wrapped edge transport.
+  that signs and attaches the `x-mailbox-auth-sig` header to every `Send`,
+  `Pull` and `AckUpTo` before forwarding to the wrapped edge transport. The
+  recipient signed over is the envelope recipient for `Send` and `MailboxId`
+  for `Pull`/`AckUpTo`. A nil signer returns `next` unchanged, which is how
+  unauthenticated test transports opt out.
 - `AckState` — Four-cursor watermark state machine (PullCursor, DispatchCommittedTo, AckTarget, AckCommittedTo).
 - `SendUnaryRequest` — Durable typed unary request that becomes a real unary RPC after commit. The response arrives via KIND_RESPONSE and, if no in-memory waiter exists, falls back to durable route dispatch via the EventRouter.
 - `DurableUnaryRequestBuilder` — Interface for proof-gated request-body construction. Implementations build the actual proto request (e.g., with signed proofs) at send time, not at persist time. The interface is provided via `ConnectorConfig.DurableUnaryBuilder`.
@@ -93,6 +100,18 @@ background ingress polling with event routing.
   per-key FIFO lane key regardless of whether the message was constructed
   fresh or decoded from TLV. The `cachedCorrelationKey` field is populated
   during `Decode` via `tlv.TlvType8` so restarts do not lose FIFO routing.
+- `MailboxAuthMessage` does **not** prevent cross-server replay on its own,
+  and the two callers differ. `Send` passes the compound
+  `operator:client` recipient, which embeds the operator's pubkey-derived
+  ID, so a `Send` signature is useless at any other operator. `Pull` and
+  `AckUpTo` pass the client's own plain mailbox ID (`ingress.go`, via
+  `ConnectorConfig.LocalMailboxID`), which carries no operator component —
+  that digest is identical at every operator, so one `Pull`/`Ack` signature
+  authorizes that client's mailbox at every operator its identity key is
+  known to, and identity keys are a deterministic derivation, so a wallet
+  driving two operators presents the same credential to both. Closing the
+  `Pull`/`Ack` case means folding a server identity into the digest, which
+  is a wire change on both sides.
 - Every outbound envelope is stamped with the runtime-bound mailbox
   transport and Ark protocol version pair (`stampEnvelopeVersions`/
   `versionStampingMailboxClient`), overwriting any caller-provided value.
diff --git a/waved/AGENTS.md b/waved/AGENTS.md
index 5a340df5..acd5d35d 100644
--- a/waved/AGENTS.md
+++ b/waved/AGENTS.md
@@ -12,8 +12,9 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
 
 - `Server` — main daemon. Owns the wallet, DB, chainsource actor, gRPC
   server, and `ActorSystem`. Caches `localMailboxID` (pubkey-derived),
-  `authSigHex` (Schnorr auth), and a single `clk` (`clock.Clock`) shared by
-  all sub-stores for deterministic time injection.
+  `authSigHex` (Schnorr auth), `mailboxAuthSigs` (per-recipient mailbox auth
+  signature memo, guarded by `mailboxAuthSigsMu`), and a single `clk`
+  (`clock.Clock`) shared by all sub-stores for deterministic time injection.
 - `RPCServer` — implements the gRPC `DaemonService`. Most write RPCs
   (`Board`, `SendVTXO`, `SendOOR`, `SweepBoardingUTXOs`, `SendOnChain`)
   validate input locally then `Ask` the relevant actor; `GetRound` and
@@ -50,6 +51,26 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
   `serverconn.PubKeyMailboxID`, not config strings. The operator's remote
   mailbox ID and pubkey are fetched via direct gRPC (`fetchCurrentOperatorPubKey`)
   before the mailbox runtime starts.
+- Every outbound mailbox edge is wrapped in
+  `serverconn.NewAuthenticatedMailboxClient`, so `Send`, `Pull` and `AckUpTo`
+  all carry the `x-mailbox-auth-sig` header. Both `connectOperatorClients`
+  arms (gRPC and REST) wrap, and so does `newMailboxEdge`; the operator
+  authorizes a mailbox RPC either from the TLS client certificate bound to
+  the caller's mailbox ID or from that header, and an operator that
+  terminates TLS at a proxy never sees a client certificate. Signing
+  unconditionally keeps the operator's TLS posture out of client config. A
+  new mailbox client built anywhere else must be wrapped too, or it silently
+  loses the header against an operator that requires it.
+- `mailboxAuthSig` memoizes the signature per recipient mailbox ID and is the
+  only path that should reach `signMailboxAuth` (`mailboxAuthSigner`,
+  `connectAndBootstrapMailbox` and `RPCServer.SignMailboxAuth` all go through
+  it). The digest covers only the identity key and the recipient, so one
+  signature stays valid for the life of the key; signing per RPC would put a
+  wallet round trip in front of every `Pull` on a long-poll loop. The map
+  normally holds two entries, not one — `Send` addresses the compound
+  `operator:client` mailbox while `Pull`/`AckUpTo` address this client's own
+  plain mailbox ID — so keying on the recipient is what keeps them from
+  colliding. It returns an error until `clientKeyDesc.PubKey` is derived.
 - All sub-stores share the single `s.clk` clock assigned in `NewServer`; new
   code must not call `clock.NewDefaultClock()` directly, use `s.clk`.
 - Actor startup order in `startWalletDependentActors`: VTXO manager, then
diff --git a/waved/CLAUDE.md b/waved/CLAUDE.md
index 5a340df5..acd5d35d 100644
--- a/waved/CLAUDE.md
+++ b/waved/CLAUDE.md
@@ -12,8 +12,9 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
 
 - `Server` — main daemon. Owns the wallet, DB, chainsource actor, gRPC
   server, and `ActorSystem`. Caches `localMailboxID` (pubkey-derived),
-  `authSigHex` (Schnorr auth), and a single `clk` (`clock.Clock`) shared by
-  all sub-stores for deterministic time injection.
+  `authSigHex` (Schnorr auth), `mailboxAuthSigs` (per-recipient mailbox auth
+  signature memo, guarded by `mailboxAuthSigsMu`), and a single `clk`
+  (`clock.Clock`) shared by all sub-stores for deterministic time injection.
 - `RPCServer` — implements the gRPC `DaemonService`. Most write RPCs
   (`Board`, `SendVTXO`, `SendOOR`, `SweepBoardingUTXOs`, `SendOnChain`)
   validate input locally then `Ask` the relevant actor; `GetRound` and
@@ -50,6 +51,26 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
   `serverconn.PubKeyMailboxID`, not config strings. The operator's remote
   mailbox ID and pubkey are fetched via direct gRPC (`fetchCurrentOperatorPubKey`)
   before the mailbox runtime starts.
+- Every outbound mailbox edge is wrapped in
+  `serverconn.NewAuthenticatedMailboxClient`, so `Send`, `Pull` and `AckUpTo`
+  all carry the `x-mailbox-auth-sig` header. Both `connectOperatorClients`
+  arms (gRPC and REST) wrap, and so does `newMailboxEdge`; the operator
+  authorizes a mailbox RPC either from the TLS client certificate bound to
+  the caller's mailbox ID or from that header, and an operator that
+  terminates TLS at a proxy never sees a client certificate. Signing
+  unconditionally keeps the operator's TLS posture out of client config. A
+  new mailbox client built anywhere else must be wrapped too, or it silently
+  loses the header against an operator that requires it.
+- `mailboxAuthSig` memoizes the signature per recipient mailbox ID and is the
+  only path that should reach `signMailboxAuth` (`mailboxAuthSigner`,
+  `connectAndBootstrapMailbox` and `RPCServer.SignMailboxAuth` all go through
+  it). The digest covers only the identity key and the recipient, so one
+  signature stays valid for the life of the key; signing per RPC would put a
+  wallet round trip in front of every `Pull` on a long-poll loop. The map
+  normally holds two entries, not one — `Send` addresses the compound
+  `operator:client` mailbox while `Pull`/`AckUpTo` address this client's own
+  plain mailbox ID — so keying on the recipient is what keeps them from
+  colliding. It returns an error until `clientKeyDesc.PubKey` is derived.
 - All sub-stores share the single `s.clk` clock assigned in `NewServer`; new
   code must not call `clock.NewDefaultClock()` directly, use `s.clk`.
 - Actor startup order in `startWalletDependentActors`: VTXO manager, then
```

**How to apply:** save the diff above and `git apply` it, or run the doc-gardening skill locally (`/doc-gardening serverconn` then `/doc-gardening waved`) to regenerate the same updates. Each `CLAUDE.md` must stay byte-identical to its sibling `AGENTS.md`.

**Note on `make doc-check`:** it reports two failures on this branch — `./.claude-pr/AGENTS.md` missing, and `./vtxo/CLAUDE.md` diverged from `./vtxo/AGENTS.md`. Both are pre-existing on the base commit and outside this PR's changed packages, so this advisory leaves them alone; the `serverconn` and `waved` pairs it touches pass.

---
Posted by the per-PR documentation advisor — https://github.com/lightninglabs/wavelength/pull/1136/checks

waved/outbound_clients.go
```diff
@@ -71,6 +88,62 @@ func (s *Server) connectOperatorClients() (*operatorClients, error) {
...
+	}
+
+	s.mailboxAuthSigsMu.Lock()
+	defer s.mailboxAuthSigsMu.Unlock()
+
+	if sig, ok := s.mailboxAuthSigs[recipientMailboxID]; ok {
+		return sig, nil
+	}
+
+	sig, err := s.signMailboxAuth(ctx, recipientMailboxID)
```
litbot-9000: 🟠 **major** — The wallet round trip happens with `mailboxAuthSigsMu` held, and on the path that matters the context has no deadline.

`connectAndBootstrapMailbox` primes the memo for `remoteMailboxID` (the compound `operator:client` ID) at `waved/server.go:3974`, so `Send` is a memo hit. But ingress addresses the other key the follow-up commit's comment now correctly names: `serverconn/ingress.go:501` and `:712` pull and ack with `a.cfg.LocalMailboxID`, the bare client pubkey (`waved/server.go:3962`). That is a cold miss, so the first `Pull` (or `AckUpTo`) enters this critical section and calls out to the wallet. `StartIngress` builds its context with `context.WithCancel(ctx)` and no timeout (`serverconn/actor.go:1220`), and `pullBatch` adds none, so `signTaggedSchnorr` reaches `lndSvc.Signer.SignMessage` (`waved/server.go:5043`) on a context that never expires.

Blast radius: `StartEgress()` is already running by then (`waved/server.go:4029`, while ingress does not start until `:1556` -> `startMailboxIngress` -> `:2609`), so its worker pool is concurrently calling `Send` -> `authContext` -> `mailboxAuthSig`. Those callers each wrap their own send deadline (`serverconn/actor.go:876`, `:1043`), but they block on `sync.Mutex.Lock()`, which is not context-aware — their deadlines are silently ignored. A wedged or unreachable lnd therefore converts "one RPC times out" into "the entire operator mailbox edge stalls indefinitely": egress, heartbeat, ack and pull all parked on one mutex, with no log line and no metric to say why.

`a2ab2309` widens this rather than narrowing it. Routing `RPCServer.SignMailboxAuth` through the memo is right for cost, but it puts every subserver-stamped `Send`, `Pull` and `AckUpTo` behind the same mutex, so a subserver now queues behind the ingress cold miss too.

The error path is worse than a one-shot window. A failed sign caches nothing, so every subsequent mailbox RPC re-enters the critical section and re-attempts the round trip. A wallet that fails slowly serializes the whole edge for as long as it keeps failing, not just once.

Two changes fix it: (1) do not hold the mutex across the wallet call — lock, check the memo, unlock, sign, re-lock and store; a duplicate concurrent sign is harmless because the digest is deterministic; (2) bound the signing context with an explicit timeout rather than inheriting the caller's, so a hung wallet fails the RPC instead of the daemon.

waved/outbound_clients.go
```diff
@@ -71,6 +88,62 @@ func (s *Server) connectOperatorClients() (*operatorClients, error) {
...
+// whichever wallet backend holds the key.
+//
+// The map holds two entries, not one. Send addresses the compound
+// operator:client mailbox, while Pull and AckUpTo address this client's own
+// plain mailbox ID, so the two arms sign different recipients and cache
+// separately. Keying on the recipient is what keeps them from colliding.
+func (s *Server) mailboxAuthSig(ctx context.Context,
+	recipientMailboxID string) (*schnorr.Signature, error) {
+
+	if s.clientKeyDesc.PubKey == nil {
```
litbot-9000, line 152: 🟠 **major** — This read introduces a data race on `s.clientKeyDesc`, because `initRPCClients` writes that field after the egress workers that reach this code are already running.

Ordering inside `connectAndBootstrapMailbox` (`waved/server.go`, line numbers at `a2ab2309`):

- `:4029` `s.runtime.StartEgress()` — `DurableActor.Start()` spawns `numWorkers` goroutines (`baselib/actor/durable_actor.go:504-527`) that drain the durable outbox and call `Edge.Send`, i.e. the wrapped client, i.e. this function.
- `:4031` `startActorOutboxPublisher`
- `:4041` `s.initRPCClients(ctx)` — which at `waved/server.go:3795` does `s.clientKeyDesc = *identityDesc`.

So a plain struct assignment to `clientKeyDesc` (a `keychain.KeyDescriptor`: a `KeyLocator` plus a `*btcec.PublicKey`) races unsynchronized reads on the egress goroutines — here at line 122, and again inside `signMailboxAuth` (`waved/server.go:4996`) and `signTaggedSchnorr` (`:5043`, `:5095`, which also read `KeyLocator`). On a restart with a non-empty durable outbox there is real outbound traffic in that window. `-race` would flag it; the unit suite does not drive this ordering, which is why `unit-race` does not catch it — it will surface in an itest, or not at all until it matters.

To be fair about the consequence: both writes resolve the same key. `deriveIdentityKeyEarly` derives `Family: identityKeyFamily, Index: 0`, and `IndexerProofKey` -> `proofKeyBackend.DeriveKey` uses the same locator against the same backend (`waved/server.go:1636`), so today the second write stores an identical descriptor and no wrong signature results. What you have is a genuine memory-model race with a benign value, plus a landmine: the moment those two derivations diverge, or a reader observes a new `PubKey` alongside a stale `KeyLocator`, this signs under a mismatched key and the operator rejects the RPC — which, once lumos#780 lands and the knob is on, means the fleet cannot reach its mailbox.

Cheapest fix is ordering, not locking: move `s.initRPCClients(ctx)` above `s.runtime.StartEgress()`, or drop the redundant `clientKeyDesc` write in `initRPCClients` since `deriveIdentityKeyEarly` has already stored the same value. Note that `mailboxAuthSigsMu` cannot help — no writer of `clientKeyDesc` takes it, so moving this nil check inside the lock would change nothing.

serverconn/mailbox_auth.go
```diff
@@ -43,8 +43,24 @@ const MailboxTLSBindTagStr = "mailbox-tls-bind"
...
-// Including the recipient mailbox ID (the server's pubkey-derived ID)
-// prevents cross-server replay of the signature.
+// Including the recipient mailbox ID binds the signature to what it
+// addresses, but how much that buys depends on the caller, and only
+// Send gets the strong version:
+//
+//   - Send passes the compound operator:client recipient, which
+//     embeds the operator's pubkey-derived ID, so a Send signature is
+//     useless at any other operator.
+//   - Pull and AckUpTo pass the client's own plain mailbox ID
```
litbot-9000, line 53: 🟠 **major** — Good catch, and the comment is accurate — but writing the gap down is not the same as deciding about it, and this PR is what makes the gap load-bearing.

Before this change the operator edge sent no per-RPC credential at all. After it, every `Pull` and `AckUpTo` on the always-on operator edge carries a 64-byte signature with no operator component, no nonce and no expiry: a static bearer token for that client's inbox, valid for the life of the identity key at every operator that accepts direct auth. And the motivating deployment is precisely the one that exposes it — an operator terminating TLS at a load balancer means an intermediary sees the header in cleartext, and proxies and gateways routinely log request headers.

The consequence is not only disclosure. On the lumos side `AckUpTo` deletes: `db/mailbox_store.go:392-425` upserts the cursor and then runs `DeleteAckedMailboxEnvelopes`, and the cursor is monotonic so it cannot be walked back. A replayed Pull/Ack credential lets the holder permanently destroy that client's undelivered mail at a *different* operator, silently — the victim re-pulls from its own cursor and simply finds nothing. On the round and OOR paths that is lost protocol traffic, not just lost privacy.

I am not asking for the wire change here; the comment is right that folding a server identity into the Pull/Ack digest is a both-sides change. What I am asking is that this not land as a code comment only:

- Open a tracking issue and reference it from this comment, so the next reader finds the decision rather than re-deriving the risk.
- Say explicitly in the lumos#780 rollout notes that enabling `requiredirectauth` on a proxy-terminated deployment puts a cross-operator-replayable credential in front of the proxy, and that the proxy must not log `x-mailbox-auth-sig`.

If the answer is "accepted risk until the digest changes", that is a defensible call — it just needs to be a recorded one, made before an operator flips the knob, rather than a paragraph discovered later by whoever is debugging the incident.

waved/server_outbound_clients_test.go
```diff
@@ -81,6 +109,20 @@ func TestConnectOperatorClientsREST(t *testing.T) {
 	)
 	defer server.Close()
 
+	// The mailbox edge signs every call with the daemon identity key, so
+	// the fixture needs one. Priming the memo stands in for the wallet
+	// backend that would produce the signature in a running daemon.
+	clientKey, err := btcec.NewPrivateKey()
```
litbot-9000, line 115: 🟡 **minor** — `Static Checks` is red on this file and it is yours, not infrastructure. The failing step is step 5, `Check changed code format` (`make fmt-changed-check`); every later step in the job is `skipped` after it, and the annotation's generic "Process completed with exit code 2" is just the make failure. This is not ARC preemption — the job ran to a real step failure, all 15 steps carry real conclusions with no nulls, and there is no "self-hosted runner lost communication" annotation.

Reproduced at `a2ab2309` with `GOWORK=off make fmt-changed-check base=$(git merge-base HEAD origin/main)`. There are now four spots, not two, because `a2ab2309`'s gRPC test repeats both patterns: the nested `PubKeyMailboxID` calls inside `CompoundMailboxID` want breaking onto their own lines at lines 116-123 and again at 232-239, and the `require.Equal` wants the same treatment at 163-169 and 271-277. `make fmt-changed` fixes all four.

waved/server_outbound_clients_test.go
```diff
@@ -90,6 +132,12 @@ func TestConnectOperatorClientsREST(t *testing.T) {
 				MacaroonPath: macaroonPath,
 			},
 		},
+		clientKeyDesc: keychain.KeyDescriptor{
+			PubKey: clientKey.PubKey(),
+		},
+		mailboxAuthSigs: map[string]*schnorr.Signature{
+			clientMailbox: authSig,
+		},
 	}
```
litbot-9000, line 146: 🟡 **minor** — Priming `mailboxAuthSigs` directly means both tests short-circuit the code this PR adds beyond the wiring. `mailboxAuthSig` returns on the memo hit, so the signing path, the lazy map allocation, the `clientKeyDesc == nil` guard and the mutex are all uncovered.

What the pair does prove is worth having, and `a2ab2309` improved it: the wrapper is installed on both arms now, not just REST, and reverting either one fails a test. That is the regression these guard.

The gap left is the memo semantics the design leans on. A third case with a stub `MailboxAuthSigner` — asserting that two `Pull` calls to the same mailbox invoke the signer once, and that a `Pull` for a different mailbox ID invokes it again — would pin the two-entry behaviour the compound-vs-plain recipient split depends on, and would give you somewhere to pin behaviour when the lock scope changes.

litbot-9000 (changes requested):
FINAL — re-verified at head `a2ab2309`.

**Note the tree moved twice during review.** The branch was rebased onto `b2acfd0b` (so `587be266` -> `2d03ab81`, `a7aec0e6` -> `1c36dc84`) and gained `a2ab2309 waved: close two paths that would send mailbox RPCs unsigned`. Everything below is re-resolved against that head; the `waved/server.go` line numbers moved by ~10 in the rebase.

The direction is right and the payload agrees across the two repos — the client digest was traced into the lumos verifier and they match on all three RPCs, so `mailbox.requiredirectauth` is safe to enable against this client. Two implementation defects and one design gap still hold, plus a red CI that is the author's.

**`a2ab2309` closes the coverage gap and is a good change.** `newMailboxEdge`'s fallback is now wrapped (`waved/server.go:2863`), which was the one remaining place a refactor could silently produce an unauthenticated edge, and `TestConnectOperatorClientsGRPC` now proves the header on the gRPC arm rather than only the REST one. Both were raised in the previous pass and are done.

**One caveat on that commit:** routing `RPCServer.SignMailboxAuth` through `mailboxAuthSig` is correct for cost, but it widens the lock problem below. A subserver that stamps every `Send`, `Pull` and `AckUpTo` now contends on `mailboxAuthSigsMu` — the same mutex the ingress cold miss holds across a wallet round trip. The memo saves a round trip per RPC and buys a queue behind whoever is signing.

**What was verified, and how.** Built the tree and ran `go test ./waved/ ./serverconn/ -race -count=1` at this head. Reproduced the CI failure locally with `GOWORK=off make fmt-changed-check base=$(git merge-base HEAD origin/main)`. Read the lumos side at `d4673986` for the verifier, the resolver, the REST gateway and the mailbox store.

**CI triage.** At `a2ab2309` (run 31552276183) `Static Checks` fails on step 5, `Check changed code format` — the author's `gofmt`/`golines` drift in `waved/server_outbound_clients_test.go`, reproduced locally at this head, now four spots rather than two because the new gRPC test repeats both patterns. Not ARC preemption: the job ran to a real step failure, all 15 steps carry real conclusions with no nulls, and there is no "self-hosted runner lost communication" annotation. The rest of the run is green or still pending.

**Unrelated flake you will probably hit, flagged so it is not mistaken for this branch.** `go test ./waved/ -race` fails here on `TestSignCreditAccountAuthorizationValidatesEnvelope/too_far`, expecting `InvalidArgument` and getting `Internal`. It is not this PR's: the table builds `tooFarExpiry = now + CreditAccountMaxAuthTTL + 1m` at `waved/rpc_wallet_test.go:131-133`, while `validateCreditAccountAuthExpiry` compares against a fresh `time.Now()` at `waved/rpc_wallet.go:192-194`. The subtests are `t.Parallel()`, and the package takes ~107s under `-race`, so once more than a minute of wall clock passes the expiry is no longer "too far", validation falls through to signing on a zero-valued `RPCServer`, and the code becomes `Internal`. It passes in isolation and passes at the merge base (where the package runs in 11s), so it is a latent timing bug in the credit-account tests already on `main` that any slow run of `waved` can trip — including `unit-race`. Worth a separate fix: freeze the comparison clock or widen the margin.

**Cross-repo payload agreement: confirmed, both arms.** For `Pull`/`AckUpTo`, lumos resolves the caller from the *claimed* mailbox ID (`server_mtls.go:610` `ResolveMailboxClientID`, via `mailboxMetadataAuthTarget`) and verifies `TaggedHash("mailbox-auth", senderPubKey || claimedID)`. Ingress claims `LocalMailboxID` — the bare client pubkey hex, no colon (`serverconn/ingress.go:501`, `:712`) — so the resolver returns it unchanged and the sender key is the client identity key. The signer here signs `TaggedHash("mailbox-auth", identityPubKey || recipientMailboxID)` with the same key over the same string: digests identical. For `Send`, lumos takes the sender from `Envelope.Sender` and the recipient from `Envelope.Recipient` (`server_mtls.go:518-524`); `serverconn` sets those to `LocalMailboxID` and `RemoteMailboxID` (`actor.go:860-861`, `heartbeat.go:102-103`, `unary_facade.go:129-130`), and the compound `operator:client` recipient is signed verbatim on both sides. No mismatch anywhere.

**REST arm: refuted, but on an invisible cross-repo dependency.** The header does reach the verifier. `rpc/restclient/client.go:199-207` copies outgoing gRPC metadata onto raw HTTP headers, and lumos's `gatewayHeaderMatcher` has an explicit case for `serverconn.AuthHeaderKey` (`gatewayserver.go:169-171`) that hands it back as gRPC metadata, where `hasGatewayAuth` routes it into `verifyMailboxMetadataAuth` (`server_mtls.go:415-425`). Note what that rests on: `runtime.DefaultHeaderMatcher` alone would have dropped a custom `x-mailbox-auth-sig` header — grpc-gateway only forwards custom headers under a `Grpc-Metadata-` prefix. The REST arm works only because lumos special-cased the key. Worth a line in the commit message so nobody "cleans up" that matcher case later and silently breaks every REST client.

**Cold start: refuted.** Traced rather than reasoned about. `connectAndBootstrapMailbox` (`waved/server.go:3885`) calls `deriveIdentityKeyEarly` as its first statement at `:3887` and only reaches `connectOperatorClients` at `:3896`; the whole thing runs either synchronously when the wallet is ready or from the goroutine that waits on `walletReady` (`:1469-1500`, call at `:1546`). The mailbox edge therefore cannot exist before the identity key does, and the `"identity key not yet derived"` branch is unreachable on this path. No cold-start regression — nothing that previously succeeded unauthenticated now fails, and startup cannot wedge on it. One consequence: the comment on `mailboxAuthSigner` (`waved/outbound_clients.go:93`, "the mailbox edge is built before the wallet has derived one") states the opposite of the actual ordering and should be corrected or dropped.

**Key rotation: refuted.** `clientKeyDesc` is written twice per startup — `deriveIdentityKeyEarly` (`waved/server.go:4908`/`:4932`/`:4956`) and `initRPCClients` (`:3795`) — but both resolve `Family: identityKeyFamily, Index: 0` against the same backend (`refreshProofKeyBackend`, `:1636`), so the descriptor is identical and the memo cannot serve a signature from a superseded key. The other two `initRPCClients` call sites (`:1855`, `:2140`) are guarded on `s.runtime != nil` and cannot fire before the runtime exists. No re-derivation, rotation or re-unlock path changes the key mid-process, so the memo needs no invalidation path. That second write does create a data race, though — see the inline comment.

**Bottom line:** the lock scope and the `clientKeyDesc` race are both small fixes and should land with this PR. The replayable Pull/Ack credential is not this PR's fault, but this PR is what puts it on the wire on the always-on operator edge, so it needs an explicit decision and a tracking issue before an operator flips `requiredirectauth` — not just a comment.

PR #1140, opened 2026-08-12
multi: confine the daemon's lnd spending to one account

ellemouton:
### Motivation

When `waved` shares an lnd node with another daemon, every spend it makes comes
out of the same pool of coins the other one relies on: CPFP fee inputs for
unilateral exit, boarding sweeps, change, and the address operators deposit
into. A busy neighbour can leave this daemon unable to pay for an exit, and vice
versa.

This is the client-side half of giving each daemon that shares an lnd node its
own pocket of funds.

### Changes

- `lndbackend.BoardingBackend` gains an account (`WithAccount` option,
  defaulting to lnd's `default`), exposed via `Account()` / `IsDefaultAccount()`.
- `ListUnspentDefaultAccount` → `ListUnspentWalletAccount`, filtered on that
  account.
- `lndUnrollWallet` uses it for CPFP fee-input selection, fresh change scripts,
  `FundPsbt` and `FinalizePsbt` (which previously passed `""`, i.e. lnd's
  default account, while funding from a different one).
- The deposit address and the exit-funding UTXO accounting move onto it too.
- New `lnd.account` config option and `waved` flag, plus sample config and CLI
  guide entries.

Default behaviour is unchanged: an empty `lnd.account` keeps lnd's `default`
account, so existing deployments are unaffected.

### Notes for reviewers

- **Observation stays unscoped on purpose.** Imported boarding and exit taproot
  scripts live in lnd's separate watch-only account and must remain visible to
  boarding detection. Only spending, signing and change are narrowed.

- **Fee-input accounting *is* scoped**, unlike observation. It feeds the
  preflight that decides whether an exit can pay for itself, so counting coins
  the funding path cannot actually spend would green-light an exit that then
  cannot be funded.

- **The account must be taproot-scoped** — but not for the reason you might
  expect. It is not the CPFP witness type (the fee-bump code handles p2wkh,
  np2wkh and p2pkh fine). It is that every lnd address this daemon derives asks
  for a taproot address, and lnd resolves a custom account name within the key
  scope the requested address type implies, so an account under any other scope
  resolves as "account not found".

- `FundPsbt` only sets an explicit `ChangeType` on the default account. A custom
  account lives in exactly one key scope, which already fixes its change type,
  and lnd rejects a request that supplies both.

### Testing

Unit tests cover the account option, the empty-config fallback, and the scoped
versus unscoped enumerations.

**Known gap, flagged deliberately:** the custom-account path is not exercised at
runtime. Both harnesses give each client daemon its *own* lnd node, where the
default account is already correct, so no integration test sets `lnd.account`.
Closing this needs a harness change to put `waved` on a shared node, and I'd
suggest doing that before the option is enabled on any shared deployment.

### DIFF
--- a/lndbackend/AGENTS.md
+++ b/lndbackend/AGENTS.md
@@ -12,10 +12,13 @@ MuSig2) for the round actor, and proof-key derivation/signing.
   Implements `wallet.BoardingBackend` and `wallet.OutputLeaser`.
   `GetTransaction` returns `*wallet.TxInfo`; `GetBlock` fetches raw blocks via
   `chainKit` for TxProof merkle inclusion. `ListUnspent` spans every wallet
-  account including imported watch-only scripts; `ListUnspentDefaultAccount`
-  restricts to the default account for CPFP fee-input selection (watch-only
-  outputs are unsignable). `LeaseOutput`/`ReleaseOutput` forward to
-  walletKit, casting `wallet.LockID` <-> `wtxmgr.LockID`.
+  account including imported watch-only scripts; `ListUnspentWalletAccount`
+  restricts to the backend's own account for CPFP fee-input selection
+  (watch-only outputs are unsignable). The account defaults to
+  `DefaultWalletAccount` (lnd's "default") and is overridden with the
+  `WithAccount` option; `Account()`/`IsDefaultAccount()` expose it to PSBT
+  builders. `LeaseOutput`/`ReleaseOutput` forward to walletKit, casting
+  `wallet.LockID` <-> `wtxmgr.LockID`.
 - `ClientWallet` — Adapts lndclient's remote signer to `input.Signer` +
   MuSig2 (`round.ClientWallet`), so the round actor can sign VTXO tree
   branches and forfeit transactions via LND's remote signer without a local
@@ -39,9 +42,23 @@ MuSig2) for the round actor, and proof-key derivation/signing.
 - `ClientWallet.signOutputRawWithLocator` always forwards the key locator
   when set (including family != 0, index == 0), working around an lndclient
   gap that otherwise breaks the family-6/index-0 identity signing path.
-- CPFP fee-input selection must use `ListUnspentDefaultAccount`, not
+- CPFP fee-input selection must use `ListUnspentWalletAccount`, not
   `ListUnspent`: offering a watch-only (imported script) output as a fee
   input makes the child PSBT unsignable.
+- Every lnd call that spends, derives change, or signs must use the same
+  account (`Account()`). A custom account lives in one key scope, so lnd
+  rejects an explicit `ChangeType` for it on the `Psbt`/`Raw` templates —
+  guard that with `IsDefaultAccount()`.
+- A custom account must be **taproot-scoped**, and not because of CPFP
+  witness types: lnd resolves a custom account name *within the key scope
+  the requested address type implies* (`keyScopeForAccountAddr`), and every
+  address this daemon derives asks for taproot, so an account under any
+  other scope fails address derivation outright with "account not found".
+- Observation must stay unscoped. Imported boarding scripts live in lnd's
+  watch-only account, which belongs to no wallet account, so an account
+  filter hides them — and lnd treats an empty account name as *every*
+  account but `"default"` as a real filter, so scoping an observation path
+  breaks it even with no account configured.
 
 ## Deep Docs
 
--- a/lndbackend/boarding_backend.go
+++ b/lndbackend/boarding_backend.go
@@ -22,6 +22,17 @@ import (
 	"github.com/lightningnetwork/lnd/lnwallet"
 )
 
+// DefaultWalletAccount is the LND wallet account this backend spends from
+// when no other account is configured. LND's "default" account is the one
+// every wallet has and the one a freshly funded node puts its coins in, so
+// using it preserves the behaviour of deployments that never think about
+// accounts at all.
+//
+// Deployments that share one LND node between several daemons should give
+// each daemon its own account instead, so that one daemon's spending cannot
+// consume another's funds.
+const DefaultWalletAccount = lnwallet.DefaultAccountName
+
 // BoardingBackend implements the wallet.BoardingBackend interface by
 // wrapping an lndclient.WalletKitClient and ChainKitClient. This
 // provides the concrete integration with an LND node for boarding
@@ -35,19 +46,65 @@ type BoardingBackend struct {
 	// This enables TxProof construction when boarding UTXOs confirm.
 	chainKit lndclient.ChainKitClient
 
+	// account is the LND wallet account whose coins this backend may
+	// select, derive addresses from, and sign for. It never restricts
+	// which UTXOs the backend can *observe*: imported boarding scripts
+	// live in LND's separate watch-only account and are still enumerated
+	// by ListUnspent.
+	account string
+
 	// Log is an optional logger for this backend. If None, the backend
 	// falls back to extracting a logger from context.
 	Log fn.Option[btclog.Logger]
 }
 
+// Option customises a BoardingBackend at construction time.
+type Option func(*BoardingBackend)
+
+// WithAccount pins the backend to a named LND wallet account, so that coin
+// selection, change, address derivation and signing only ever touch that
+// account's funds. An empty name leaves the backend on DefaultWalletAccount.
+func WithAccount(account string) Option {
+	return func(b *BoardingBackend) {
+		if account == "" {
+			return
+		}
+
+		b.account = account
+	}
+}
+
 // NewBoardingBackend creates a new LND boarding backend.
 func NewBoardingBackend(walletKit lndclient.WalletKitClient,
-	chainKit lndclient.ChainKitClient) *BoardingBackend {
+	chainKit lndclient.ChainKitClient, opts ...Option) *BoardingBackend {
 
-	return &BoardingBackend{
+	backend := &BoardingBackend{
 		walletKit: walletKit,
 		chainKit:  chainKit,
+		account:   DefaultWalletAccount,
 	}
+
+	for _, opt := range opts {
+		opt(backend)
+	}
+
+	return backend
+}
+
+// Account returns the LND wallet account this backend spends from. Callers
+// that build PSBTs against the same LND node must fund, finalize and derive
+// change with this account so that every step stays inside the same pocket of
+// funds.
+func (l *BoardingBackend) Account() string {
+	return l.account
+}
+
+// IsDefaultAccount reports whether the backend spends from LND's built-in
+// default account. Custom accounts live in exactly one key scope, which fixes
+// their change address type, and LND rejects an explicit change type for them;
+// callers building PSBTs use this to decide whether they may ask for one.
+func (l *BoardingBackend) IsDefaultAccount() bool {
+	return l.account == DefaultWalletAccount
 }
 
 // WalletKit returns the underlying LND WalletKit client for
@@ -109,25 +166,28 @@ func (l *BoardingBackend) ImportTaprootScript(ctx context.Context,
 // The result spans every wallet account, including imported watch-only
 // script outputs (e.g. boarding scripts tracked via ImportTaprootScript).
 // Callers that need outputs the wallet can unilaterally sign, such as CPFP
-// fee-input selection, must use ListUnspentDefaultAccount instead.
+// fee-input selection, must use ListUnspentWalletAccount instead.
 func (l *BoardingBackend) ListUnspent(ctx context.Context, minConfs,
 	maxConfs int32) ([]*wallet.Utxo, error) {
 
 	return l.listUnspent(ctx, minConfs, maxConfs)
 }
 
-// ListUnspentDefaultAccount returns UTXOs from LND's default wallet account
-// only. This excludes imported watch-only script outputs (boarding and exit
-// scripts imported via ImportTaprootScript), which the wallet merely tracks
-// and cannot unilaterally sign. CPFP fee-input selection must use this
-// variant: offering a watch-only output as a fee input makes the child PSBT
-// unsignable and the fee bump fails with "PSBT is not finalizable".
-func (l *BoardingBackend) ListUnspentDefaultAccount(ctx context.Context,
+// ListUnspentWalletAccount returns UTXOs from the backend's configured LND
+// wallet account only. This excludes imported watch-only script outputs
+// (boarding and exit scripts imported via ImportTaprootScript), which the
+// wallet merely tracks and cannot unilaterally sign. CPFP fee-input selection
+// must use this variant: offering a watch-only output as a fee input makes the
+// child PSBT unsignable and the fee bump fails with "PSBT is not finalizable".
+//
+// When the backend is pinned to a non-default account, this is also what keeps
+// the daemon from spending another daemon's coins on the same LND node.
+func (l *BoardingBackend) ListUnspentWalletAccount(ctx context.Context,
 	minConfs, maxConfs int32) ([]*wallet.Utxo, error) {
 
 	return l.listUnspent(
 		ctx, minConfs, maxConfs,
-		lndclient.WithUnspentAccount(lnwallet.DefaultAccountName),
+		lndclient.WithUnspentAccount(l.account),
 	)
 }
 
--- a/waved/config.go
+++ b/waved/config.go
@@ -913,6 +913,22 @@ type LndConfig struct {
 	// RPCTimeout is the maximum duration for individual RPC calls to
 	// lnd. If zero, DefaultRPCTimeout is used.
 	RPCTimeout time.Duration `mapstructure:"rpctimeout"`
+
+	// Account is the lnd wallet account this daemon may spend from. It
+	// bounds coin selection, change, address derivation and signing, but
+	// not what the daemon can observe: imported boarding scripts live in
+	// lnd's watch-only account and stay visible either way.
+	//
+	// Empty selects lndbackend.DefaultWalletAccount, lnd's built-in
+	// "default" account, which is where a freshly funded node keeps its
+	// coins. Set this when several daemons share one lnd node, so that one
+	// daemon's spending cannot drain another's funds. The named account
+	// must already exist on the node (lnd's `wallet accounts create`) and
+	// must be taproot-scoped: every lnd address this daemon derives for
+	// itself asks for a taproot address, and lnd resolves a custom account
+	// name within the key scope implied by the requested address type, so
+	// an account created under any other scope resolves as "not found".
+	Account string `mapstructure:"account"`
 }
 
 // ServerConfig holds connection parameters for the ark operator's mailbox
--- a/waved/server.go
+++ b/waved/server.go
@@ -1649,8 +1649,65 @@ func (s *Server) initLndBackend(ctx context.Context) error {
 		"pubkey", lndServices.NodePubkey,
 	)
 
+	if err := s.validateLndAccount(ctx, lndServices); err != nil {
+		return err
+	}
+
 	// In lnd mode the wallet is immediately ready.
 	s.markWalletReady()
+
+	return nil
+}
+
+// validateLndAccount refuses to start when the configured lnd account cannot
+// serve this daemon, rather than letting the daemon boot and fail later.
+//
+// Every misconfiguration here surfaces late and badly otherwise, and one of
+// them is silent. A name that does not exist makes lnd's ListUnspent filter
+// every output away rather than error, so the exit preflight simply concludes
+// there is nothing to spend. An account under the wrong key scope is worse
+// than that: coin selection and signing resolve it by name across all scopes
+// and work fine, while address derivation pins the taproot scope and fails —
+// so the daemon runs until the moment a unilateral exit or a boarding sweep
+// needs a fresh script. A watch-only account fails later still, at signing,
+// once inputs are already leased.
+//
+// One lookup filtered by both name and address type distinguishes all three.
+//
+// NOTE: this confirms the taproot-scoped account exists, which is what address
+// derivation needs. It does not rule out the same name also existing under
+// another key scope, in which case funding's lookupFirstCustomAccount could
+// resolve to the other one. lnd only allowed that on older versions, so it is
+// not guarded against here.
+func (s *Server) validateLndAccount(ctx context.Context,
+	lnd *lndclient.GrpcLndServices) error {
+
+	account := s.lndWalletAccount()
+	if account == lndbackend.DefaultWalletAccount {
+		return nil
+	}
+
+	accounts, err := lnd.WalletKit.ListAccounts(
+		ctx, account, walletrpc.AddressType_TAPROOT_PUBKEY,
+	)
+	if err != nil || len(accounts) == 0 {
+		return fmt.Errorf("lnd wallet account %q not found under the "+
+			"taproot key scope: this daemon derives taproot "+
+			"addresses, and lnd resolves an account name within "+
+			"the key scope the requested address type implies, so "+
+			"the account must be created taproot-scoped (lookup "+
+			"error: %v)", account, err)
+	}
+
+	if accounts[0].GetWatchOnly() {
+		return fmt.Errorf("lnd wallet account %q is watch-only, so "+
+			"lnd cannot sign for it: coin selection would succeed "+
+			"and signing would fail after the inputs were "+
+			"already leased", account)
+	}
+
+	s.log.InfoS(ctx, "Using lnd wallet account",
+		"account", account)
 
 	return nil
 }
@@ -4099,6 +4156,9 @@ func (s *Server) initWalletActor(ctx context.Context,
 		lndSvc := s.lnd.UnsafeFromSome()
 		backend := lndbackend.NewBoardingBackend(
 			lndSvc.WalletKit, lndSvc.ChainKit,
+			lndbackend.WithAccount(
+				s.lndWalletAccount(),
+			),
 		)
 		backend.Log = fn.Some(s.subLogger(lndbackend.Subsystem))
 		boardingBackend = backend
@@ -5168,6 +5228,16 @@ func networkToLndclient(network string) (lndclient.Network, error) {
 	}
 }
 
+// lndWalletAccount returns the lnd wallet account this daemon spends from,
+// falling back to lnd's built-in account when none is configured.
+func (s *Server) lndWalletAccount() string {
+	if s.cfg.Lnd == nil || s.cfg.Lnd.Account == "" {
+		return lndbackend.DefaultWalletAccount
+	}
+
+	return s.cfg.Lnd.Account
+}
+
 // lndUnrollWallet composes the LND-backed signing/key-derivation wallet
 // with the boarding backend's ListUnspent to satisfy the
 // UnilateralExitWallet interface needed by the package executor.
@@ -5176,19 +5246,20 @@ type lndUnrollWallet struct {
 	boardingBackend *lndbackend.BoardingBackend
 }
 
-// ListUnspent returns UTXOs from LND's default wallet account only.
+// ListUnspent returns UTXOs from the boarding backend's configured LND wallet
+// account only.
 //
 // The boarding backend's unfiltered enumeration also surfaces imported
 // watch-only script outputs (boarding and exit scripts tracked via
 // ImportTaprootScript). Those are not signable by LND's FinalizePsbt, so
 // offering them as CPFP fee inputs makes the child PSBT unsignable and the
 // fee bump fails with "PSBT is not finalizable". Restricting fee selection
-// to the default account keeps selection aligned with what the finalize
+// to the wallet's own account keeps selection aligned with what the finalize
 // path can actually sign, mirroring the lwwallet and btcwallet adapters.
 func (w *lndUnrollWallet) ListUnspent(ctx context.Context, minConfs,
 	maxConfs int32) ([]*wallet.Utxo, error) {
 
-	return w.boardingBackend.ListUnspentDefaultAccount(
+	return w.boardingBackend.ListUnspentWalletAccount(
 		ctx, minConfs, maxConfs,
 	)
 }
@@ -5199,7 +5270,7 @@ func (w *lndUnrollWallet) NewWalletPkScript(ctx context.Context) ([]byte,
 	error) {
 
 	addr, err := w.boardingBackend.WalletKit().NextAddr(
-		ctx, lnwallet.DefaultAccountName,
+		ctx, w.boardingBackend.Account(),
 		walletrpc.AddressType_TAPROOT_PUBKEY, true,
 	)
 	if err != nil {
@@ -5226,7 +5297,7 @@ func (w *lndUnrollWallet) FinalizePsbt(ctx context.Context,
 	}
 
 	_, finalTx, err := w.boardingBackend.WalletKit().FinalizePsbt(
-		ctx, packet, "",
+		ctx, packet, w.boardingBackend.Account(),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("LND FinalizePsbt: %w", err)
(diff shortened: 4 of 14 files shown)


litbot-9000:
**Doc drift advisory:** this PR's Go changes in `.`, `cmd/waved`, `lndbackend` and `waved` leave `waved/CLAUDE.md` (and its `AGENTS.md` twin) stale — the new `lnd.account` config, the `ListUnspentDefaultAccount` → `ListUnspentWalletAccount` rename, and the account-scoping invariant it introduces are not reflected there. (`lndbackend`'s docs were already updated in this PR; the root and `cmd/waved` docs describe `lnd.*` generically and need no change.)

```diff
diff --git a/waved/AGENTS.md b/waved/AGENTS.md
index 91762c69..1e7d0041 100644
--- a/waved/AGENTS.md
+++ b/waved/AGENTS.md
@@ -23,9 +23,10 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
   (`rpc_fees.go`).
 - `Config` — daemon configuration: wallet backend selection, mailbox/chain
   backend wiring, `OORConfig`/`OORLimitsConfig` (receive safety caps),
-  `UnrollConfig` (unilateral-exit fee-bump cadence and cap), and
-  `MaxOperatorFeeSat` (the #270 seal-time fee-cap validated in
-  `Config.Validate()`).
+  `UnrollConfig` (unilateral-exit fee-bump cadence and cap),
+  `LndConfig` (LND connection parameters plus `Account`, the wallet
+  account this daemon may spend from), and `MaxOperatorFeeSat` (the #270
+  seal-time fee-cap validated in `Config.Validate()`).
 - `WalletState` — `None` / `Locked` / `Ready` wallet lifecycle.
 - `UnrollConfig` / `OORConfig` — subsystem tunables; see `Config.Validate()`
   for the invariants each enforces.
@@ -98,6 +99,24 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
   (`newSweepWallet`, one of `lndUnrollWallet` / `lwUnrollWallet` /
   `btcwUnrollWallet`), which is structurally compatible with both
   `unroll.SweepWallet` and the wallet actor's `SweepSigner`.
+- On the LND backend, every path that spends, derives change, or signs must
+  name the same wallet account. `Config.Lnd.Account` (`--lnd.account`, empty
+  = `lndbackend.DefaultWalletAccount`) is threaded into each
+  `lndbackend.NewBoardingBackend` call with `lndbackend.WithAccount`, and
+  `Server.lndWalletAccount()` covers the paths that reach `lndSvc.WalletKit`
+  directly (`wallet_testhooks.go`). Mixing accounts across fund/finalize
+  makes LND produce a PSBT it cannot sign. `lndUnrollWallet.FundPsbt` sets an
+  explicit taproot `ChangeType` only when `IsDefaultAccount()` holds: a custom
+  account lives in exactly one key scope, so LND rejects a change type
+  alongside it — and that scope must be taproot, since the v3 CPFP child needs
+  a Schnorr-signable fee input.
+- Account scoping bounds spending, not observation. Funding and
+  exit-affordability paths use `BoardingBackend.ListUnspentWalletAccount`
+  (renamed from `ListUnspentDefaultAccount`), while the unfiltered
+  `ListUnspent` still spans every account so imported boarding scripts stay
+  visible. `listBackingWalletUnspent` deliberately uses the scoped variant:
+  counting the whole node's coins would green-light an exit the funding path
+  cannot pay for.
 - `LeaveVTXOs` filters its targets through `admitLeaveTargets` before
   dispatch, which runs `vtxo.CheckForfeitAdmission` over each descriptor.
   The two selection modes fail in opposite directions on purpose: an
diff --git a/waved/CLAUDE.md b/waved/CLAUDE.md
index 91762c69..1e7d0041 100644
--- a/waved/CLAUDE.md
+++ b/waved/CLAUDE.md
@@ -23,9 +23,10 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
   (`rpc_fees.go`).
 - `Config` — daemon configuration: wallet backend selection, mailbox/chain
   backend wiring, `OORConfig`/`OORLimitsConfig` (receive safety caps),
-  `UnrollConfig` (unilateral-exit fee-bump cadence and cap), and
-  `MaxOperatorFeeSat` (the #270 seal-time fee-cap validated in
-  `Config.Validate()`).
+  `UnrollConfig` (unilateral-exit fee-bump cadence and cap),
+  `LndConfig` (LND connection parameters plus `Account`, the wallet
+  account this daemon may spend from), and `MaxOperatorFeeSat` (the #270
+  seal-time fee-cap validated in `Config.Validate()`).
 - `WalletState` — `None` / `Locked` / `Ready` wallet lifecycle.
 - `UnrollConfig` / `OORConfig` — subsystem tunables; see `Config.Validate()`
   for the invariants each enforces.
@@ -98,6 +99,24 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/waved.<S
   (`newSweepWallet`, one of `lndUnrollWallet` / `lwUnrollWallet` /
   `btcwUnrollWallet`), which is structurally compatible with both
   `unroll.SweepWallet` and the wallet actor's `SweepSigner`.
+- On the LND backend, every path that spends, derives change, or signs must
+  name the same wallet account. `Config.Lnd.Account` (`--lnd.account`, empty
+  = `lndbackend.DefaultWalletAccount`) is threaded into each
+  `lndbackend.NewBoardingBackend` call with `lndbackend.WithAccount`, and
+  `Server.lndWalletAccount()` covers the paths that reach `lndSvc.WalletKit`
+  directly (`wallet_testhooks.go`). Mixing accounts across fund/finalize
+  makes LND produce a PSBT it cannot sign. `lndUnrollWallet.FundPsbt` sets an
+  explicit taproot `ChangeType` only when `IsDefaultAccount()` holds: a custom
+  account lives in exactly one key scope, so LND rejects a change type
+  alongside it — and that scope must be taproot, since the v3 CPFP child needs
+  a Schnorr-signable fee input.
+- Account scoping bounds spending, not observation. Funding and
+  exit-affordability paths use `BoardingBackend.ListUnspentWalletAccount`
+  (renamed from `ListUnspentDefaultAccount`), while the unfiltered
+  `ListUnspent` still spans every account so imported boarding scripts stay
+  visible. `listBackingWalletUnspent` deliberately uses the scoped variant:
+  counting the whole node's coins would green-light an exit the funding path
+  cannot pay for.
 - `LeaveVTXOs` filters its targets through `admitLeaveTargets` before
   dispatch, which runs `vtxo.CheckForfeitAdmission` over each descriptor.
   The two selection modes fail in opposite directions on purpose: an
```

**How to apply:** save the diff above to a file and `git apply` it, or run
the doc-gardening skill locally (`/doc-gardening waved`) and let it
regenerate the pair. Keep `waved/CLAUDE.md` and `waved/AGENTS.md`
byte-identical (`cp waved/CLAUDE.md waved/AGENTS.md`).

Note: `make doc-check` currently reports two failures that predate this PR and
are outside its scope — `vtxo/CLAUDE.md` and `vtxo/AGENTS.md` have diverged in
the committed tree, and an untracked `.claude-pr/CLAUDE.md` on the runner has
no `AGENTS.md` sibling. This advisory leaves both alone.

---
[doc-gardening PR advisory run](https://github.com/lightninglabs/wavelength/actions/workflows/doc-gardening-pr.yml?query=branch%3Aagent%2Flnd-account-isolation)

waved/wallet_testhooks.go
```diff
@@ -99,9 +99,16 @@ func (s *Server) listBackingWalletUnspent(ctx context.Context, minConfs,
...
 			lndSvc.WalletKit, lndSvc.ChainKit,
+			lndbackend.WithAccount(s.cfg.Lnd.Account),
 		)
 
-		utxos, err := backend.ListUnspent(ctx, minConfs, maxConfs)
+		// Scoped, not unfiltered: callers use this to decide whether a
+		// unilateral exit can pay for itself, and the funding path can
+		// only spend this account. Counting the whole node's coins
+		// would green-light an exit it cannot fund.
+		utxos, err := backend.ListUnspentWalletAccount(
```
litbot-9000: **Blocking — this scopes an observation path, and it changes behaviour with `lnd.account` empty.**

`listBackingWalletUnspent` has three callers, not one:

1. `ListWalletUnspent` (right above) → the exit preflight. Scoping this is correct and is the point of the PR.
2. `RPCServer.fetchUnconfirmedBoardingBalance` (`waved/rpc_server.go`) — sums zero-conf UTXOs *paying to known boarding scripts*, feeding `GetBalanceResponse.BoardingUnconfirmedSat` via `fetchBoardingBalance`.
3. `RPCServer.ListUnconfirmedBoardingUTXOs` (`waved/rpc_wallet_helpers.go`) — consumed by `swapwallet/history.go` for activity correlation. Its own doc comment says it reaches for this helper because `ListWalletUnspent` "intentionally removes these outputs for fee-input callers, so activity correlation needs this narrowly scoped inverse view".

(2) and (3) exist *only* to see imported boarding outputs. `ListUnspentWalletAccount` excludes exactly those — as your own doc comment on it says. So both go permanently empty/zero on the lnd path.

And this isn't gated on a custom account. Chain of custody for `account = "default"`:

- `walletrpc.WalletKit.ListUnspent` passes `req.Account` straight to `Wallet.ListUnspentWitness(minConfs, maxConfs, req.Account)` (`lnrpc/walletrpc/walletkit_server.go:448`).
- which passes it to `btcwallet.Wallet.ListUnspent(minconf, maxconf, accountName)` (`lnwallet/btcwallet/btcwallet.go:1053`).
- which does `filter := accountName != ""` and then `if filter && outputAcctName != accountName { continue }` (btcwallet `wallet/wallet.go:2912`, `:2969`).
- `ImportTaprootScript` outputs resolve to `waddrmgr.ImportedAddrAccount` / `"imported"`, never `"default"`.

So pre-PR `ListUnspent` (account `""`, unfiltered) saw boarding outputs; post-PR `ListUnspentWalletAccount("default")` does not. Operators on current mainnet lose the pending-deposit balance and the pending-deposit activity view, with no error and no log line. Given real coins are being deposited right now, that's the failure mode most worth avoiding.

Suggested fix: leave `listBackingWalletUnspent` unfiltered (it is documented as "the active wallet backend's raw unspent-output query without any filtering") and move the scoping up into `ListWalletUnspent`, which is the fee-input/preflight entry point and already post-filters boarding scripts anyway. That puts the split exactly where your review note says it should be: spending scoped, observation not.

Worth a regression test too — `fakeWalletKit.ListUnspent` currently ignores `req.Account`, which is why the full suite passes with this in place.

waved/config.go
```diff
@@ -913,6 +913,22 @@ type LndConfig struct {
...
+	// Empty selects lndbackend.DefaultWalletAccount, lnd's built-in
+	// "default" account, which is where a freshly funded node keeps its
+	// coins. Set this when several daemons share one lnd node, so that one
+	// daemon's spending cannot drain another's funds. The named account
+	// must already exist on the node (lnd's `wallet accounts create`) and
+	// must be taproot-scoped: every lnd address this daemon derives for
+	// itself asks for a taproot address, and lnd resolves a custom account
+	// name within the key scope implied by the requested address type, so
+	// an account created under any other scope resolves as "not found".
+	Account string `mapstructure:"account"`
```
litbot-9000, line 931: No startup validation for this, and the misconfiguration failure modes are asymmetric — some loud, one silent, none at startup:

- **Nonexistent / typo'd name**: `ListUnspent` does *not* error. btcwallet just filters every output out (`filter && outputAcctName != accountName`), so the exit preflight silently sees zero UTXOs and concludes the exit cannot pay for itself. `FundPsbt`/`FinalizePsbt` do error (`newAccountNotFoundError` via `lookupFirstCustomAccount`), and so does `NextAddr`.
- **Right name, wrong key scope** (e.g. created under BIP84): this one is nastier, because the two lnd lookup paths disagree. `FundPsbt`/`FinalizePsbt` go through `lookupFirstCustomAccount`, which scans *all* of `waddrmgr.DefaultKeyScopes` and takes the first hit — so funding and signing work fine. But `NextAddr(account, TAPROOT_PUBKEY)` goes through `keyScopeForAccountAddr`, which pins BIP0086 and returns `account not found`. So the daemon boots, lists coins, funds PSBTs — and then fails at `NewWalletPkScript` during a unilateral exit or a boarding sweep. Exactly when funds are at stake.

A cheap startup check collapses all of that into a refuse-to-start: when `Account != ""`, look the account up filtered by name *and* `KeyScopeBIP0086` during lnd wallet init, and fail with a message naming both the account and the required scope. The doc comment and sample config are accurate and well written, but a comment cannot stop `lnd.account=swapd` from being pasted into a config where the account was created under the wrong scope.

waved/server.go
```diff
@@ -5240,24 +5252,33 @@ func (w *lndUnrollWallet) FundPsbt(ctx context.Context, packetBytes []byte,
...
+		LockExpirationSeconds: uint64(lockExpiry.Seconds()),
+	}
+
+	// The default account spans all key scopes, so LND needs to be told
+	// which one to take change from; we ask for taproot because the v3
+	// CPFP child this funds needs a Schnorr-signable fee input. A custom
+	// account lives in exactly one key scope, which already fixes its
+	// change type, and LND rejects the request outright if a change type
+	// is supplied alongside one.
+	if w.boardingBackend.IsDefaultAccount() {
```
litbot-9000, line 5337: Guard is correct — verified against the pinned lnd. `keyScopeFromChangeAddressType(CHANGE_ADDRESS_TYPE_P2TR)` returns `&waddrmgr.KeyScopeBIP0086` (`walletkit_server.go:2378`), which reaches `BtcWallet.FundPsbt`'s `default:` branch and trips `couldn't select a custom change type for custom accounts` (`lnwallet/btcwallet/psbt.go:118-123`). And it is the `GetPsbt()` template branch in play here, which does forward the scope (`walletkit_server.go:1730`). So a custom account genuinely cannot supply both.

One nit on the comment's rationale: "we ask for taproot because the v3 CPFP child this funds needs a Schnorr-signable fee input" conflates change with fee inputs. `ChangeType` governs the *change* output this call creates, not the fee input, which comes from `ListUnspentWalletAccount`. The real reason to keep P2TR here is that without it lnd defaults the default-account change scope to BIP0084 (`psbt.go:98-102`), so you would silently start emitting p2wkh change — a behaviour change on the default path. That is a better justification than the one written, and it is the one that would stop someone deleting the line.

lndbackend/AGENTS.md
```diff
@@ -39,9 +42,14 @@ MuSig2) for the round actor, and proof-key derivation/signing.
...
   when set (including family != 0, index == 0), working around an lndclient
   gap that otherwise breaks the family-6/index-0 identity signing path.
-- CPFP fee-input selection must use `ListUnspentDefaultAccount`, not
+- CPFP fee-input selection must use `ListUnspentWalletAccount`, not
   `ListUnspent`: offering a watch-only (imported script) output as a fee
   input makes the child PSBT unsignable.
+- Every lnd call that spends, derives change, or signs must use the same
+  account (`Account()`). A custom account lives in one key scope, so lnd
+  rejects an explicit `ChangeType` for it — guard that with
+  `IsDefaultAccount()`. A custom account must be taproot-scoped, since
```
litbot-9000: This gives the taproot-scope reason as "unilateral exit's v3 CPFP child needs a Schnorr-signable fee input" — which is the reason the PR description explicitly says it is *not*. The correct reason (the one in the description and in `waved/config.go`) is `keyScopeForAccountAddr`: the requested address type fixes the key scope the account name is resolved in, so a non-BIP0086 account is simply not found.

Worth fixing here and in `CLAUDE.md:51`, because this is the file a future agent reads instead of the PR description, and the wrong reason points somewhere wrong — it implies a p2wkh-scoped account would merely produce a worse fee input, when in fact address derivation fails outright.

waved/server.go
```diff
@@ -5168,6 +5169,16 @@ func networkToLndclient(network string) (lndclient.Network, error) {
 	}
 }
 
+// lndWalletAccount returns the lnd wallet account this daemon spends from,
+// falling back to lnd's built-in account when none is configured.
+func (s *Server) lndWalletAccount() string {
```
litbot-9000, line 5233: Non-blocking, pick one: this guards `s.cfg.Lnd == nil`, but the four sites that pass the account into `WithAccount` dereference `s.cfg.Lnd.Account` unguarded (`initWalletActor`, `initUnrollSubsystem`, `newSweepWallet`, `listBackingWalletUnspent`). Either `s.cfg.Lnd` can be nil while `s.lnd.IsSome()` — in which case those four panic — or it cannot, and this guard is dead code. Reading the init path I believe it is the latter, but the inconsistency reads as uncertainty. Routing all five through this helper would settle it and drop the duplication.

Separately, and outside this diff: `lndSvc.Client.WalletBalance` (`walletBalanceFetchers` / `metricsWalletBalance`) has no account filter — lndclient hardcodes an empty `WalletBalanceRequest` — so with a custom account `OnchainWalletConfirmedSat` and the metrics gauge report the *whole node*, including the neighbour daemon's coins, while the exit preflight counts only this daemon's pocket. That is the same "the balance says I have money, the exit says I do not" confusion this PR exists to remove, just relocated. Follow-up rather than a blocker, but it should land before `lnd.account` is set anywhere real.

litbot-9000 (changes requested):
The mechanism is right and the reasoning in the description is right — I checked the three load-bearing claims against the actual lnd this repo pins (`v0.21.0-beta.rc2.0.20260630214209`) rather than taking them on trust, and all three hold:

- **Taproot scope requirement** — confirmed, and for exactly the reason you give, not the CPFP one. `btcwallet.keyScopeForAccountAddr` (`lnwallet/btcwallet/btcwallet.go:443`) maps the *requested address type* to a key scope and then does `AccountNumber(addrKeyScope, accountName)`. A `TAPROOT_PUBKEY` request against an account created under any other scope is a plain `ErrAccountNotFound`.
- **`ChangeType` conditional** — confirmed. `keyScopeFromChangeAddressType(P2TR)` returns a non-nil `KeyScopeBIP0086`, and `BtcWallet.FundPsbt` (`lnwallet/btcwallet/psbt.go:118-123`) errors with `couldn't select a custom change type for custom accounts` for anything that isn't `default`/`imported`. The `IsDefaultAccount()` guard is necessary and correctly placed.
- **Preflight vs funding use the same account** — confirmed. `walletExitFundingSnapshot` → `ListWalletUnspent` → `listBackingWalletUnspent`, and the CPFP path `txconfirm/broadcaster.go` → `lndUnrollWallet.ListUnspent` → `ListUnspentWalletAccount`, and `FundPsbt`/`FinalizePsbt` all land on `boardingBackend.Account()`. No mismatch.

But the observation/spending line is drawn one level too high in one place, and it changes behaviour **today, with `lnd.account` empty**, on mainnet. Details inline on `waved/wallet_testhooks.go`. Short version: `listBackingWalletUnspent` is not only the preflight's dispatcher — two other callers reach for it precisely *because* it was unfiltered, and scoping it to `default` drops imported boarding scripts (they live in `waddrmgr.ImportedAddrAccount`), so `GetBalance.boarding_unconfirmed_sat` becomes permanently 0 and `ListUnconfirmedBoardingUTXOs` permanently empty. That's the silent default-behaviour change, so: request changes. Everything else below is non-blocking.

One correction to the description: `FinalizePsbt(ctx, packet, "")` was not a live bug before this PR. lnd defaults an empty account to `DefaultAccountName` (`walletkit_server.go:2477`) and the old `FundPsbt` already passed `lnwallet.DefaultAccountName`, so the two agreed. It was a latent inconsistency that this PR makes load-bearing — worth saying it that way, since "we were already funding from a different account" implies a bug on current mainnet that isn't there.

**On the harness gap.** I think it's acceptable-with-follow-up for *merging*, and not acceptable for *enabling*. The option is opt-in and off by default, so merging it to unblock lightninglabs/lumos#786's submodule bump is fine once the regression above is fixed. But the first time anyone sets `lnd.account` on the shared prod node, the code path has never run. Before that switch is flipped I'd want either the shared-lnd harness change you describe, or a documented manual regtest run on a shared node covering: derive deposit address → receive → boarding detect → exit preflight → unilateral exit with CPFP → boarding sweep. And I'd want the startup validation below in this PR, since it converts most of the misconfiguration space from "fails when funds are at stake" into "fails at startup".

**What I actually ran** (worktree off `48dabc01`): `go build ./...` — clean, with the same pre-existing `function main is undeclared in the main package` note that `main` produces, so not introduced here. `go test ./lndbackend/... ./waved/... ./unroll/... ./txconfirm/... ./wallet/...` — all pass. Note they pass *with* the regression above: `fakeWalletKit.ListUnspent` ignores `req.Account`, so no unit test can currently observe the boarding-output exclusion.

**What I did not verify**: nothing against a real lnd node — no regtest, no itest, no shared-node run. The lnd behaviour claims are read from the pinned module source, not observed. I also did not check the operator-facing side (whether the accounts are actually created taproot-scoped in the deployment tooling), or lumos#786.

— claudell (⚡)

litbot-9000:
## Cross-PR review: wavelength#1140 ↔ lightninglabs/lumos#786

This is a **seam review of the pair**, not a review of this PR on its own — a separate per-PR review covers that, and I've deliberately not repeated it. I only looked at what neither single-PR reviewer can see: the API contract across the submodule boundary, the merge ordering, whether the two halves agree about lnd's account/key-scope rules, and the combined rollout and test-coverage picture.

(Note: this PR was force-pushed to `bb4d0391` while I was reading. It's a rebase onto `e8848d4e`; the account-relevant hunks are unchanged apart from one reflow of the `lnd.account` flag string. Findings below are against the new head.)

### Merge order verdict

**This PR must merge first, then `make submodule-update` in lumos, then lumos#786. The dependency lumos#786 states is real — I verified it by building.**

lumos currently pins `ce7e9832` (*Merge pull request #1136*), which predates this PR. Building lumos#786 against that pin gives nine compile errors — `undefined: clientlnd.WithAccount`, `too many arguments in call to clientlnd.NewBoardingBackend`, `ListUnspentWalletAccount undefined`, `Account undefined` ×4. Swap this PR into `client/` and lumos builds clean (`go build ./...` exit 0, `go test ./lndbackend/...` ok). `ce7e9832` is an ancestor of this PR's head, so the bump afterwards is a clean fast-forward.

**The exported API is exactly what lumos consumes** — `WithAccount(string) Option`, the variadic `NewBoardingBackend`, `Account()`, `ListUnspentWalletAccount(ctx, minConfs, maxConfs)`, signature for signature. Nothing to change here.

### `IsDefaultAccount()` is exported but the other half doesn't call it

The one substantive seam defect is on lumos's side, but it's worth flagging here because it's about *this* API being bypassed.

Your `lndUnrollWallet.FundPsbt` gates the change type correctly:

```go
if w.boardingBackend.IsDefaultAccount() {
        req.ChangeType = walletrpc.ChangeAddressType_CHANGE_ADDRESS_TYPE_P2TR
}
```

lumos#786's `serverTxConfirmWallet.FundPsbt` in `server_fraud.go` is the structural twin — same `Psbt` template, same job (CPFP fee wallet for `txconfirm`) — and it **deletes the `ChangeType` unconditionally** instead. Correct for a custom account, wrong for the default one: per the lnd path below, the fraud responder's change silently moves **P2TR → P2WKH** for every operator on merge, since `lnd.account` defaults to empty. `IsDefaultAccount()` therefore has no consumer in lumos today. I've suggested the fix over there; noting it here so the pair is reviewed as one thing.

### Both descriptions of lnd's rules are correct, and both repos pin the same lnd

Worth stating plainly since the two PRs word the rule differently: they are consistent, and I found no version skew. Both `go.mod`s pin `github.com/lightningnetwork/lnd v0.21.0-beta.rc2.0.20260630214209-40c64f9db30d`, no `replace` on either side, one copy in the module cache. Traced in that source:

- **This PR's claim** (Psbt/Raw): `keyScopeFromChangeAddressType` (`walletkit_server.go:2378`) → `BtcWallet.FundPsbt` (`lnwallet/btcwallet/psbt.go:93-133`). Default account + nil change scope → falls back to **BIP-0084**; custom account + non-nil change scope → `"couldn't select a custom change type for custom accounts"`. ✅ Exactly as described.
- **lumos's claim** (PsbtCoinSelect): `req.ChangeType` only maps to `P2TRChangeAddress`/`P2WKHChangeAddress` (`:1792`) → `handleChange` → `NewAddress(addrType, true, account)` → `keyScopeForAccountAddr` → `AccountNumber(BIP0084, name)` → not-found for a taproot account. ✅ Also exactly as described, and not in conflict — the two templates genuinely have opposite change rules.
- The **"must be taproot-scoped"** rationale in your notes checks out: `NextAddr(..., TAPROOT_PUBKEY, ...)` resolves the account name through `keyScopeForAccountAddr` under BIP-0086, so an account created under any other scope is not found.

One caveat neither PR documents: `lookupFirstCustomAccount` — used by the Psbt-template `FundPsbt` and by `FinalizePsbt` — walks `waddrmgr.DefaultKeyScopes` in the order **BIP0049Plus, BIP0084, BIP0086, BIP0044** and takes the first hit, whereas `NextAddr`/`handleChange` pin the scope from the requested address type. lnd's own comment notes older versions allowed the same account name under several scopes. So "an account under any other scope resolves as not found" holds only when the name is unique to one scope; if it exists under two, funding/finalizing and address derivation resolve to *different* scopes, silently. Worth a sentence in `waved/config.go`'s doc comment and in `lndbackend/CLAUDE.md`.

### Combined rollout risk

Neither half validates the account at startup. `lndWalletAccount()` here and `walletAccount()` there are plain getters; no config validation on either side. `lndclient.WalletKit.ListAccounts` is already used in `harness/harness.go:2580`, so a probe is cheap. Without one:

- **Typo or wrong scope:** the daemon starts clean, serves traffic, and fails the *first time it needs to move money*. For waved that's the deposit address (`NextAddr` → "account not found") and unilateral-exit CPFP funding. For lumosd it's any round producing change, plus the fraud responder's CPFP child — which you'd discover during a fraud attempt.
- **Empty new account:** creating an account moves no coins. The node's balance still looks healthy while the configured pocket has nothing selectable. Your scoped fee-input preflight will correctly refuse the exit — good — but only at exit time.
- **Rollback is not symmetric.** On the lumos side the batch sweeper now sweeps into the configured account; reverting `lnd.account` to empty afterwards strands those funds where default-account selection can never see them.

Concrete order for a shared node, given there is no startup check:

1. `lncli wallet accounts create --name <daemon> --address_type p2tr` per daemon (waved, lumosd, swapd); confirm with `wallet accounts list` that each name appears under exactly one scope (`m/86'/...`).
2. **Fund each account before pointing any daemon at it** — `lncli newaddr --account <name> --address_type p2tr`. waved needs enough on hand for exit CPFP specifically, since that's the path that fails silently until it's needed.
3. Restart one daemon at a time, and manually verify it can derive its deposit address in its own account before letting traffic at it. That manual step is currently the only check.
4. Order between daemons matters less than funding first. What matters is never restarting a daemon onto an empty account while it has live obligations — pending exits here, in-flight rounds or expired batches on the lumos side.

A startup `ListAccounts` probe that fails fast when the configured name is missing, under a non-BIP-0086 scope, or present under more than one scope, is ~10 lines and converts every failure above from *at money time* to *at boot*. Given this is mainnet with real deposited coins, that's the highest-value thing missing from the pair.

### Coverage, taken together

**No — the partitioning behaviour is not tested end to end anywhere.** Each PR flags its own gap; combined, there's no coverage at all:

- Here: three unit tests (`TestListUnspentAccountScoping`, `TestListUnspentCustomAccount`, `TestWithAccountEmptyKeepsDefault`) assert an account string reaches the `lndclient` call. Nothing sets `lnd.account` in any harness or itest, and nothing covers the `FundPsbt` `ChangeType` gating — the exact behaviour lumos got wrong.
- lumos: zero. No test sets `lnd.account`, and none covers `walletAccount()`, `NewSweepPkScript`, or the fraud-responder wallet's account.

So "a custom account actually resolves in lnd, change lands inside it, `FinalizePsbt` can sign what `FundPsbt` selected, and a co-tenant can't touch the coins" is verified by reading only. The swapd itests lumos#786 points at live in neither repo and exercise neither of these daemons' paths. Your recommendation to land a shared-node harness before enabling the option anywhere is, on the aggregate picture, the right call.

Minimum viable test — one itest, either side, against a single lnd node: create two taproot accounts and fund both; bring the daemon up on account A; assert a derived address belongs to A, that a funded+finalized tx spends only A's UTXOs, **and that the change output belongs to A**; then lock A's coins from outside and assert funding fails cleanly instead of reaching into B or `default`. The change-output assertion is what would have caught the `ChangeType` divergence above.

---

**Verified:** built lumos#786 against its pinned submodule (fails, 9 errors) and against this PR (`go build ./...` clean, `lndbackend` tests pass); built `./waved/... ./lndbackend/... ./cmd/...` here (clean — the root-package `main is undeclared` error also reproduces on `main`, so it's pre-existing); compared both `go.mod` lnd pins (identical, no `replace`); traced `FundPsbt`/`FinalizePsbt`/`NextAddr`/`handleChange`/`lookupFirstCustomAccount`/`keyScopeForAccountAddr` in the pinned lnd source; grepped both repos for startup validation and account-related tests. **Not verified:** no lnd node was run, so every claim about runtime behaviour is read from lnd's source rather than observed; and I did not review the swapd half.

Cross-posted on lightninglabs/lumos#786.

— claudell ⚡

ellemouton:
@claude please review. Pushed fixes for the previous round:

- The account-scoping regression is fixed: `listBackingWalletUnspent` is unfiltered again and the scoping moved to a new `listSpendableWalletUnspent` used only by the fee-input/preflight path, with a regression test asserting observation stays unscoped for every account value (wavelength).
- Restored the gated `ChangeType` on the fraud path so the default account keeps P2TR change, and extracted the two opposite template rules into named helpers with tests (lumos).
- Added startup validation refusing to boot on an account that is missing, non-taproot-scoped, or watch-only, in both daemons.
- Corrected the taproot rationale in the docs (scope resolution, not CPFP witness types) and filed lightninglabs/lumos#787 for the node-wide treasury reporting.

The lnd side (lightningnetwork/lnd#11065) now also carries an itest and a clearer wrong-scope error.

lndbackend/boarding_backend.go
```diff
@@ -109,25 +166,28 @@ func (l *BoardingBackend) ImportTaprootScript(ctx context.Context,
 // The result spans every wallet account, including imported watch-only
 // script outputs (e.g. boarding scripts tracked via ImportTaprootScript).
 // Callers that need outputs the wallet can unilaterally sign, such as CPFP
-// fee-input selection, must use ListUnspentDefaultAccount instead.
+// fee-input selection, must use ListUnspentWalletAccount instead.
```
Roasbeef, line 169: Perhaps we should just _remove_ this all together/

Roasbeef (approved):
LGTM 🪡

lndbackend/boarding_backend.go
ellemouton: I don't think we can drop it, unfortunately — the unfiltered variant has four callers that specifically need it, all of which go blind if it becomes account-scoped:

- `wallet.Ark.unconfirmedBoardingBalance` (`wallet/wallet.go:1094`)
- boarding detection on each block epoch (`wallet/wallet.go:1276`)
- seed-restore boarding recovery (`waved/wallet_recovery.go:199`)
- `listBackingWalletUnspent`, feeding `GetBalance.boarding_unconfirmed_sat` and `ListUnconfirmedBoardingUTXOs`

They all look for imported boarding scripts, which `ImportTaprootScript` puts in lnd's watch-only account (`waddrmgr.ImportedAddrAccount`) — that account belongs to no wallet account, so any account filter excludes them. Scoping this was in fact the blocking regression on the first round of this PR: it silently zeroed the pending-deposit balance and activity views, and did so even with `lnd.account` unset, since lnd reads an empty account name as *every* account but `"default"` as a real filter.

Where I think you're right is that "unfiltered" is doing too much work implicitly. Two options, happy either way:

1. Rename it to say so at every call site — `ListUnspentAllAccounts` or similar. It's on the `wallet.BoardingBackend` interface so it touches the lwwallet/btcwbackend implementations too, which is why I didn't do it unprompted here.
2. Replace "everything" with "explicitly the imported account", since boarding scripts are the only thing these callers want. More precise, but it changes behaviour on the lwwallet/btcwbackend paths where accounts don't exist, so it'd want its own testing.

I'd lean towards (1) as a follow-up unless you'd rather it land here. Let me know which you meant and I'll do it.

PR #1144, opened 2026-08-13
arkscript: Tighten custom policy validation

sputn1ck:
## What changed

- require custom predicates to parse as complete standalone script fragments
- reject unsupported tapscript opcodes in custom predicates
- restrict typed CSV values to canonical non-zero block delays
- apply the checks consistently across direct policy admission and the standard
  VTXO, vHTLC, and checkpoint builders
- add regression coverage for malformed predicates and noncanonical CSV
  encodings

## Why

Policy templates should have one interpretation across semantic validation,
script generation, persistence, and recovery. These changes align compiler and
validator behavior so malformed or ambiguous custom templates fail closed.

## Compatibility and rollout

Built-in policy templates and canonical production delays are unchanged.
Persisted custom templates that use unsupported predicate fragments or
noncanonical CSV values will no longer compile and should be audited before
rollout.

## Checks

- `go test ./lib/arkscript ./lib/tx/checkpoint ./lib/tx/oor`
- `make lint-changed-local`
- `make build`
- `make fmt-changed`
- `make tidy-module-check`
- `git diff --check`

### DIFF
--- a/docs/arkscript_spec.md
+++ b/docs/arkscript_spec.md
@@ -111,7 +111,7 @@ type Condition struct {
 }
 ```
 
-Script encoding (from `Condition.Script`, `lib/arkscript/node.go:120-141`):
+Script encoding (from `Condition.Script` in `lib/arkscript/node.go`):
 
 ```
 <predicate> <inner>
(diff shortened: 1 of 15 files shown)


sputn1ck:
@claude please review

bhandras (approved):
Reviewed at `dfb8ad8b`. Approving the client/compiler hardening.

The policy compiler now fails closed on the consensus-sensitive constructions in scope: incomplete predicate fragments, tapscript OP_SUCCESS opcodes, and noncanonical CSV encodings including the disable/type/reserved-bit cases. Direct policy validation compiles each leaf before reasoning about its typed structure, and the standard VTXO/vHTLC constructors apply the same canonical CSV guard.

The focused policy tests, full unit/race jobs, SQLite/PostgreSQL system tests, lint, static checks, and cross-compilation are green. The remaining scoped-doc advisory failure is an external review-action limit rather than a code failure.

This approval is for the client-side compiler and structural validator scope. Operator-specific policy terms such as the negotiated minimum exit window remain an admission-layer responsibility and should still be enforced by the service consuming these policies.

PR #1157, opened 2026-08-19
docs: nightly doc-gardening sweep 2026-08-19

litbot-9000:
Nightly documentation-graph sweep via `.claude/skills/doc-gardening` (scope: `all`), anchored to the last merged sweep (`7a88cb4e`, 2026-08-10).

Workflow run: https://github.com/lightninglabs/wavelength/actions

## What changed

The per-package docs for this window's feature work (retry-safe receive scripts, lnd account isolation) were already updated by their own PRs, so this sweep records the one invariant that fell through the gaps plus two map corrections:

- **`db/CLAUDE.md`** — new invariant: an `ExecTx` closure may run more than once, so variables it captures from the enclosing function are per-attempt, not per-call, and captured out-parameters must be reset inside the closure. This is the shape `cd9596eb` fixed in `AdmitIdempotentOwnedReceiveScript`, where a serialization retry returned the durable winner alongside the failed attempt's `created` flag.
- **`docs/postgres_isolation.md`** — same rule written up against the write-path audit, since a relaxed isolation level induces more retries and so makes the shape more reachable. `docs/index.md` entry updated to match.
- **`CLAUDE.md`** (root) — Quick Commands gained the three verification targets that exist in the `Makefile` but were unlisted (`doc-check`, `schema-check`, `sample-conf-check`); the duplicate `make systest` row is collapsed. The skills section now also points at `doc-gardening` and `bug-report`, which closes the loop from a failing `make doc-check` to the skill that fixes it.

`ARCHITECTURE.md` needed no change: no new packages and no layer moves this window, and every changed package is already in the layer tables.

## Review notes

Please review the updated **`CLAUDE.md`/`AGENTS.md`** and **`docs/`** diffs. The `db` and root `CLAUDE.md`/`AGENTS.md` pairs are byte-identical mirrors, so read one of each.

Verified before commit:
- `make doc-check` passes.
- Every symbol named in the new prose exists in the tree (`AdmitIdempotentOwnedReceiveScript`, `created` reset at `db/oor_artifact_store.go`; all three `make` targets present in the `Makefile`).
- No leaked tool/formatting artifacts in any `.md`.
- Staged with `git add -- '*.md'`, so the commit is docs-only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,14 +16,16 @@ This file is a **map**, not a manual. Follow links for details.
 | `make fmt-changed-check` | Verify changed Go source files are formatted |
 | `make unit pkg=<pkg> case=<test>` | Run unit tests |
 | `make unit log="stdlog trace" pkg=<pkg> case=<test>` | Unit tests with debug logs |
-| `make systest` | System integration tests (use `db=postgres` for PostgreSQL) |
+| `make systest` | Run system integration tests (sqlite) |
+| `make systest db=postgres` | Run system integration tests (postgres) |
 | `make tidy-module-check` | Verify module files are tidy |
 | `make rpc` | Regenerate protobuf stubs |
 | `make sqlc` | Regenerate type-safe DB queries |
 | `make ast-lint` | Check ast-grep style rules |
+| `make doc-check` | Verify documentation cross-links (CI canonical) |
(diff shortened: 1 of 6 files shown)


litbot-9000:
Superseded by tonight's doc-gardening sweep.

PR #1166, opened 2026-08-20
multi: Use Lightning Labs mainnet mempool endpoint

sputn1ck:
## Summary

- route the lwwallet mainnet Esplora default through `mempool.staging.lightningcluster.com`
- use the same mainnet instance for recommended-fee estimates
- align configuration examples and operator documentation with the new defaults

## Testing

- `make lint-changed-local`
- `make unit pkg=chainfees case=TestDefaultMempoolSpaceURL`
- `make unit pkg=lwwallet case=TestDefaultEsploraURL`
- `make unit pkg=waved case=TestMempoolSpaceFeeAccessorsAreNilSafe`
- `make unit pkg=waved case=TestConfigValidateWalletDefaults`
- live probes of `/api/blocks/tip/height` and `/api/v1/fees/recommended`

### DIFF
--- a/chainfees/mempoolspace.go
+++ b/chainfees/mempoolspace.go
@@ -32,10 +32,10 @@ const (
 	// wall-clock time but not the number of bytes read.
 	maxMempoolSpaceResponseBytes = 64 << 10
 
-	// The test networks default to Lightning Labs-operated mempool
-	// instances, matching the lwwallet Esplora defaults; only mainnet uses
-	// the public mempool.space endpoint.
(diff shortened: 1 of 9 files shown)


GeorgeTsagk:
With this I'm not hitting ~that many~ any rate limits. Two open questions:
a) Will we be able to sustain the load, when lots of users use this and/or griefers find out?
b) Why not add a federation of explorers, which enables the fallback option in case you get rate limited or the service goes down?

whiteyhat:
Great work 💯

PR #1176, opened 2026-08-21
tree: add asset-aware materialization context

darioAnongba:
## Summary

- add an outpoint-keyed tree context for asset amounts, commitment roots, sealed transition packages, and per-node signing tweaks
- derive untweaked MuSig2 internal keys for asset outputs while requiring a recorded tweak for every signed asset-tree node
- preserve the existing public signer API and copy MuSig2 participant slices before operations that may sort them
- defensively copy stored byte slices so callers cannot mutate published tree data

## Runtime isolation

This branch adds no asset runtime entry point. Bitcoin-only leaves keep `AssetAmount` at zero, `BuildStructure` returns a nil `AssetContext`, and public signer calls continue to use the tree-wide sweep tweak.

### DIFF
--- a/lib/tree/asset_tree_context.go
+++ b/lib/tree/asset_tree_context.go
@@ -1,0 +1,162 @@
+package tree
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/btcsuite/btcd/wire/v2"
+)
+
+// AssetTreeContext contains the data needed to materialize and sign an asset
+// tree. Builders populate the context before sharing the tree.
+type AssetTreeContext struct {
+	// Nodes do not have inputs until materialization, so the structure pass
+	// records amounts by node first.
+	amountsByNode map[*Node]uint64
+
+	// Input outpoints remain stable when paths are extracted or serialized.
+	amountsByInput map[wire.OutPoint]uint64
+
+	leafRootsByInput map[wire.OutPoint][]byte
+	tweaksByInput    map[wire.OutPoint][]byte
+	packagesByInput  map[wire.OutPoint][]byte
+	assetRef         string
+}
+
+// NewAssetTreeContext returns an empty asset tree context.
+func NewAssetTreeContext() *AssetTreeContext {
+	return &AssetTreeContext{
+		amountsByNode:    make(map[*Node]uint64),
+		amountsByInput:   make(map[wire.OutPoint]uint64),
+		leafRootsByInput: make(map[wire.OutPoint][]byte),
+		tweaksByInput:    make(map[wire.OutPoint][]byte),
+		packagesByInput:  make(map[wire.OutPoint][]byte),
+	}
+}
+
+// SetNodeAssetAmount records a node's subtree asset amount.
+func (c *AssetTreeContext) SetNodeAssetAmount(node *Node, amount uint64) {
+	c.amountsByNode[node] = amount
+	if node.Input != (wire.OutPoint{}) {
+		// Extracted paths contain cloned nodes, but preserve their
+		// inputs.
+		c.amountsByInput[node.Input] = amount
+	}
+}
+
+// NodeAssetAmount returns a node's subtree asset amount.
+func (c *AssetTreeContext) NodeAssetAmount(node *Node) uint64 {
+	if amount, ok := c.amountsByNode[node]; ok {
+		return amount
+	}
+
+	// Path extraction clones nodes but preserves their inputs.
+	return c.amountsByInput[node.Input]
+}
+
+// SetSigningTweak records the taproot tweak for a tree transaction.
+func (c *AssetTreeContext) SetSigningTweak(input wire.OutPoint, tweak []byte) {
+	c.tweaksByInput[input] = append([]byte(nil), tweak...)
+}
+
+// SigningTweak returns the taproot tweak for a tree transaction.
+func (c *AssetTreeContext) SigningTweak(input wire.OutPoint) []byte {
+	return append([]byte(nil), c.tweaksByInput[input]...)
+}
+
+// SetSealedPackage records a tree transaction's sealed transfer package.
+func (c *AssetTreeContext) SetSealedPackage(input wire.OutPoint, pkg []byte) {
+	c.packagesByInput[input] = append([]byte(nil), pkg...)
+}
+
+// SealedPackage returns a tree transaction's sealed transfer package.
+func (c *AssetTreeContext) SealedPackage(input wire.OutPoint) []byte {
+	return append([]byte(nil), c.packagesByInput[input]...)
+}
+
+// SetAssetRef records the tree's asset reference.
+func (c *AssetTreeContext) SetAssetRef(ref string) {
+	c.assetRef = ref
+}
+
+// AssetRef returns the tree's asset reference.
+func (c *AssetTreeContext) AssetRef() string {
+	return c.assetRef
+}
+
+// IsEmpty reports whether the context carries no asset data at all.
+func (c *AssetTreeContext) IsEmpty() bool {
+	if c == nil {
+		return true
+	}
+
+	return len(c.amountsByNode) == 0 && len(c.amountsByInput) == 0 &&
+		len(c.leafRootsByInput) == 0 && len(c.tweaksByInput) == 0 &&
+		len(c.packagesByInput) == 0 && c.assetRef == ""
+}
+
+// tweakLookup returns the signing tweak for a node's input.
+func (c *AssetTreeContext) tweakLookup() func(*Node) []byte {
+	return func(node *Node) []byte {
+		return c.SigningTweak(node.Input)
+	}
+}
+
+// aggregateAssetAmounts records the asset amount of every subtree.
+func aggregateAssetAmounts(node *Node, leafAssets map[*Node]uint64,
+	ctx *AssetTreeContext) (uint64, error) {
+
+	if len(node.Children) == 0 {
+		amount := leafAssets[node]
+		ctx.SetNodeAssetAmount(node, amount)
+
+		return amount, nil
+	}
+
+	var total uint64
+	for _, child := range node.Children {
+		childAmount, err := aggregateAssetAmounts(
+			child, leafAssets, ctx,
+		)
+		if err != nil {
+			return 0, err
+		}
+
+		if childAmount > math.MaxUint64-total {
+			return 0, fmt.Errorf("asset amount overflow " +
+				"aggregating subtree totals")
+		}
+		total += childAmount
+	}
+
+	ctx.SetNodeAssetAmount(node, total)
+
+	return total, nil
+}
+
+// SetLeafAssetRoot records an asset leaf's commitment root.
+func (c *AssetTreeContext) SetLeafAssetRoot(input wire.OutPoint, root []byte) {
+	if c == nil || len(root) == 0 {
+		return
+	}
+
+	c.leafRootsByInput[input] = append([]byte(nil), root...)
+}
+
+// LeafAssetRoot returns an asset leaf's commitment root.
+func (c *AssetTreeContext) LeafAssetRoot(input wire.OutPoint) []byte {
+	if c == nil {
+		return nil
+	}
+
+	return append([]byte(nil), c.leafRootsByInput[input]...)
+}
+
+// LeafAssetRoot returns an asset leaf's commitment root.
+func (t *Tree) LeafAssetRoot(input wire.OutPoint) []byte {
+	if t == nil {
+		return nil
+	}
+
+	return t.AssetContext.LeafAssetRoot(input)
+}
--- a/lib/tree/asset_tree_context_test.go
+++ b/lib/tree/asset_tree_context_test.go
@@ -1,0 +1,220 @@
+package tree
+
+import (
+	"bytes"
+	"math"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/wire/v2"
+	"github.com/stretchr/testify/require"
+)
+
+// assetTestLeaves returns leaves with the given asset amounts.
+func assetTestLeaves(t *testing.T, amounts ...uint64) []LeafDescriptor {
+	t.Helper()
+
+	leaves := make([]LeafDescriptor, len(amounts))
+	for i, amount := range amounts {
+		_, owner := createTestKey(t)
+		leaves[i] = LeafDescriptor{
+			PkScript: []byte{
+				0x51,
+				byte(i),
+			},
+			Amount:      10_000,
+			CoSignerKey: owner,
+			AssetAmount: amount,
+		}
+	}
+
+	return leaves
+}
+
+// TestBuildStructureAssetAggregation tests subtree amount calculation.
+func TestBuildStructureAssetAggregation(t *testing.T) {
+	t.Parallel()
+
+	_, operatorKey := createTestKey(t)
+	cfg := StructureConfig{OperatorKey: operatorKey, Radix: 2}
+
+	t.Run("btc-only trees have no asset context", func(t *testing.T) {
+		t.Parallel()
+
+		structure, err := BuildStructure(
+			assetTestLeaves(t, 0, 0, 0), cfg,
+		)
+		require.NoError(t, err)
+		require.Nil(t, structure.AssetContext)
+	})
+
+	t.Run("subtree totals cover every node", func(t *testing.T) {
+		t.Parallel()
+
+		structure, err := BuildStructure(
+			assetTestLeaves(t, 800, 200, 500), cfg,
+		)
+		require.NoError(t, err)
+		require.NotNil(t, structure.AssetContext)
+
+		assetCtx := structure.AssetContext
+		require.EqualValues(
+			t, 1_500, assetCtx.NodeAssetAmount(structure.Root),
+		)
+
+		// Every node's total must equal the sum of its children,
+		// and each leaf must carry one of the input amounts.
+		var leafTotal uint64
+		for node := range structure.Root.NodesIter() {
+			if len(node.Children) == 0 {
+				leafTotal += assetCtx.NodeAssetAmount(node)
+				continue
+			}
+
+			var childSum uint64
+			for _, child := range node.Children {
+				childSum += assetCtx.NodeAssetAmount(child)
+			}
+			require.Equal(
+				t, childSum, assetCtx.NodeAssetAmount(node),
+			)
+		}
+		require.EqualValues(t, 1_500, leafTotal)
+	})
+
+	t.Run("mixed asset and btc leaves", func(t *testing.T) {
+		t.Parallel()
+
+		structure, err := BuildStructure(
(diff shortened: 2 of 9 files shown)


lib/tree/node.go
```diff
@@ -216,6 +223,37 @@ func NewBranchNode(input wire.OutPoint, groups [][]LeafDescriptor,
...
+		return nil, fmt.Errorf("failed to aggregate keys: %w", err)
+	}
+
+	return aggKey.PreTweakedKey, nil
+}
+
+// sortableCopy returns a copy of the cosigner set for MuSig2 aggregation.
+// AggregateKeys sorts the slice it is handed in place, which would silently
+// reorder a caller's set — and callers that identify a participant by
+// position (the operator first, say) would then read the wrong key.
```
darioAnongba: The comment mentions the behaviour of AggregateKeys, this is not how comments on code are written. The comments of a function describe it, and not the behaviour of other functions.

lib/tree/node.go
```diff
@@ -828,14 +884,17 @@ func (n *Node) SigHash(prevOutFetcher txscript.PrevOutputFetcher) ([]byte,
 	)
 }
 
-// NewSignerSession creates a new MuSig2 signing session for this node.
+// NewSignerSession creates a new MuSig2 signing session for this node. The
+// cosigner set is copied first: session creation sorts the slice it is
+// handed in place, and concurrent sessions over nodes sharing one backing
+// array would race on that sort.
```
darioAnongba: Describe what the function does. If the set is copied, you can add a comment when doing it to explain why, not in the description of the function.

lib/tree/node.go
```diff
@@ -65,6 +65,13 @@ type Node struct {
 	// that must sign this node's input. This is cached to avoid repeated
 	// MuSig2 aggregations during signature verification.
 	FinalKey *btcec.PublicKey
+
+	// Sequence is the input sequence of this node's transaction. Zero
+	// selects the flow-V1 default (MaxTxInSequenceNum); V2 trees use an
+	// RBF-signalling sequence. Sequence is consensus-visible: it changes
+	// the node's txid and sighash, so every node of a tree must carry
+	// the same value.
```
darioAnongba: we know what a sequence is, no need to explain that it is consensus visible here or what it does to the tx.

lib/tree/asset_tree_context.go
```diff
@@ -0,0 +1,210 @@
+package tree
```
darioAnongba, line 1: comments can be greatly improved here.

lib/tree/asset_tree_context.go
```diff
@@ -0,0 +1,162 @@
...
+	"math"
+
+	"github.com/btcsuite/btcd/wire/v2"
+)
+
+// AssetTreeContext contains the data needed to materialize and sign an asset
+// tree. Builders populate the context before sharing the tree.
+type AssetTreeContext struct {
+	// Nodes do not have inputs until materialization, so the structure pass
+	// records amounts by node first.
```
bhandras: nit: note that these are non-standard godoc's (below too for `amountsByInput`).

GeorgeTsagk:
Hey! Great work, currently reviewing this (side by side with #1177 )

lib/tree/signing.go
```diff
@@ -207,8 +218,17 @@ func NewSignerSession(signer input.MuSig2Signer,
 			return fmt.Errorf("failed to create tx: %w", err)
 		}
 
+		taprootTweak := sweepTapscriptRoot
+		if tweakLookup != nil {
+			taprootTweak = tweakLookup(node)
```
GeorgeTsagk, line 223: **Signers trust operator-supplied tweaks; presence is checked, correctness is not.**

The tweak lookup fails closed on a missing tweak, but a wrong tweak is signed without complaint. Since the tweak encodes the asset commitment, any signer using this path is trusting whoever populated the context for the exact value that determines the asset distribution. Please document on `newSignerSession`/`NewTreeSignerSession` that this hook is operator-trusted until client-side verification lands, and link the follow-up where signers recompute the expected tweak from the sealed package before signing.

lib/tree/asset_tree_context.go
```diff
@@ -0,0 +1,162 @@
...
+	c.amountsByNode[node] = amount
+	if node.Input != (wire.OutPoint{}) {
+		// Extracted paths contain cloned nodes, but preserve their
+		// inputs.
+		c.amountsByInput[node.Input] = amount
+	}
+}
+
+// NodeAssetAmount returns a node's subtree asset amount.
+func (c *AssetTreeContext) NodeAssetAmount(node *Node) uint64 {
```
GeorgeTsagk, line 48: **Amount lookups silently return zero if the re-indexing choreography is missed.**

The dual keying (node pointer during structure building, outpoint after materialization) only works because the materializer in #1177 re-indexes every node once inputs are assigned. Any other clone/rehydrate/serialize path that misses this silently gets zero from `NodeAssetAmount`. Consider keying everything by outpoint and returning an error (not 0) before materialization, or moving the amount onto `Node`. At minimum, document the re-indexing requirement on `SetNodeAssetAmount`.

lib/tree/tree.go
```diff
@@ -132,6 +136,7 @@ func (t *Tree) ExtractPathForCoSigners(targetKeys ...*btcec.PublicKey) (*Tree,
 		BatchOutpoint:      t.BatchOutpoint,
 		BatchOutput:        t.BatchOutput,
 		SweepTapscriptRoot: t.SweepTapscriptRoot,
+		AssetContext:       t.AssetContext,
```
GeorgeTsagk, line 139: **Extracted paths carry the whole batch's asset data.**

Both extractors prune the node graph but return the same `AssetContext` pointer, which holds amounts and sealed packages for every input in the batch. Latent today since the context has no wire format, but once it is serialized for client delivery, every client can enumerate the whole batch's asset distribution. Please project the context down to the extracted path's outpoints at extraction time, before a transport format cements this shape.

lib/tree/asset_tree_context_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	assetCtx := NewAssetTreeContext()
+	require.True(t, assetCtx.IsEmpty())
+
+	assetCtx.SetLeafAssetRoot(wire.OutPoint{Index: 1}, []byte{0x01})
+	require.False(t, assetCtx.IsEmpty())
+}
+
+// TestAssetTreeContextAmountsByIdentity tests node identity keying.
+func TestAssetTreeContextAmountsByIdentity(t *testing.T) {
```
GeorgeTsagk, line 161: **The pointer-to-outpoint fallback that makes extracted paths work is untested.**

No test exercises `ExtractPath*` followed by `NodeAssetAmount` on a cloned node, which is exactly the fallback invariant from the comment on `NodeAssetAmount`. Please add a regression test.

litbot-9000:
This PR's Go changes in `lib/tree` have left that package's `CLAUDE.md`/`AGENTS.md` stale — the new `AssetTreeContext` type, the `LeafDescriptor.AssetAmount` field, the `Tree`/`Structure` asset-context fields, the exported `ComputeInternalKey`, and the cosigner-copy and asset-tweak signing invariants are all undocumented.

```diff
diff --git a/lib/tree/AGENTS.md b/lib/tree/AGENTS.md
index 55fabdea..e8fc9aa1 100644
--- a/lib/tree/AGENTS.md
+++ b/lib/tree/AGENTS.md
@@ -4,7 +4,9 @@
 
 VTXO tree construction, materialization, and MuSig2 signing session management.
 Builds the Merkle-like transaction tree structure used in Ark rounds, from leaf
-descriptors through branch nodes to the batch output.
+descriptors through branch nodes to the batch output. Trees may optionally carry
+an asset context alongside the Bitcoin structure for asset-aware materialization
+and signing.
 
 ## Key Types
 
@@ -13,12 +15,14 @@ descriptors through branch nodes to the batch output.
 - `LeafDescriptor` — Describes a single VTXO leaf: amount, owner pubkeys, cosigner keys, CSV delay.
 - `VTXODescriptor` — Interface for VTXO metadata needed by tree construction (amount, cosigners, owner key).
 - `ConnectorDescriptor` — Describes a connector output for forfeit transaction construction.
-- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization.
+- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization. Carries an `AssetContext` when any leaf has a non-zero `AssetAmount`.
 - `StructureConfig` — Configuration for tree building (radix, partition weight function).
+- `AssetTreeContext` — Asset-side state hanging off a `Tree`/`Structure`: per-subtree asset amounts, per-input leaf asset roots, taproot signing tweaks, sealed transfer packages, and the tree's asset ref. Created via `NewAssetTreeContext`; nil for Bitcoin-only trees.
 - `SignerSession` — MuSig2 signing session for tree transactions, wrapping `input.MuSig2Signer`.
 - `Materializer` / `BTCMaterializer` — Interface and implementation for materializing tree nodes into actual Bitcoin transactions.
 - `TreeAssembler` — Two-pass builder (`BuildStructure` then `Materialize`) driven by `TreeConfig`.
 - `Queue[T]` — Generic queue used internally for BFS tree traversal.
+- `ComputeInternalKey` / `ComputeFinalKey` — Untweaked and taproot-tweaked MuSig2 aggregate keys for a cosigner set.
 
 ## Relationships
 
@@ -30,6 +34,9 @@ descriptors through branch nodes to the batch output.
 - `DefaultRadix` is 2 (binary tree). Each internal node has at most 2 children.
 - `NumLeafOutputs` is 2 per leaf transaction (VTXO output + sweep output).
 - Cosigner keys must be deduplicated (`UniqueCosigners`) before computing the final MuSig2 key.
+- Cosigner slices must be copied before being handed to `musig2.AggregateKeys` or
+  `MuSig2CreateSession` — both sort their input in place, and a `Node`'s
+  `CoSigners` order is load-bearing for downstream readers.
 - Tree materialization is deterministic given the same leaf descriptors and operator key.
 - `ValidateVTXODescriptors` / `ValidateConnectorDescriptor` must pass before tree construction.
 - **Cache-aliasing invariant**: a `*Tree` is effectively immutable once published from
@@ -37,6 +44,17 @@ descriptors through branch nodes to the batch output.
   pointer through caches and ancestry-fragment slices. Silently mutating a shared
   tree's nodes or root would corrupt every aliasing reader. Callers that need to
   transform a tree must clone it first.
+- Asset trees are additive to the Bitcoin tree: `LeafDescriptor.AssetAmount`
+  drives a separate aggregation pass (`Structure.AssetContext`) that never
+  changes the Bitcoin node hierarchy. Subtree totals must not overflow
+  `uint64`; `BuildStructure` fails rather than wrapping.
+- `AssetTreeContext` keys amounts by both `*Node` and input outpoint.
+  `ExtractPathForCoSigners` / `ExtractPathForIndices` clone nodes but preserve
+  their inputs and share the parent's `AssetContext` pointer, so lookups on an
+  extracted path resolve through the outpoint map.
+- When a tree has an `AssetContext`, `NewTreeSignerSession` signs each node with
+  the per-input taproot tweak recorded in the context instead of the tree's
+  `SweepTapscriptRoot`. A missing tweak is a hard error, never a fallback.
 
 ## Deep Docs
 
diff --git a/lib/tree/CLAUDE.md b/lib/tree/CLAUDE.md
index 55fabdea..e8fc9aa1 100644
--- a/lib/tree/CLAUDE.md
+++ b/lib/tree/CLAUDE.md
@@ -4,7 +4,9 @@
 
 VTXO tree construction, materialization, and MuSig2 signing session management.
 Builds the Merkle-like transaction tree structure used in Ark rounds, from leaf
-descriptors through branch nodes to the batch output.
+descriptors through branch nodes to the batch output. Trees may optionally carry
+an asset context alongside the Bitcoin structure for asset-aware materialization
+and signing.
 
 ## Key Types
 
@@ -13,12 +15,14 @@ descriptors through branch nodes to the batch output.
 - `LeafDescriptor` — Describes a single VTXO leaf: amount, owner pubkeys, cosigner keys, CSV delay.
 - `VTXODescriptor` — Interface for VTXO metadata needed by tree construction (amount, cosigners, owner key).
 - `ConnectorDescriptor` — Describes a connector output for forfeit transaction construction.
-- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization.
+- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization. Carries an `AssetContext` when any leaf has a non-zero `AssetAmount`.
 - `StructureConfig` — Configuration for tree building (radix, partition weight function).
+- `AssetTreeContext` — Asset-side state hanging off a `Tree`/`Structure`: per-subtree asset amounts, per-input leaf asset roots, taproot signing tweaks, sealed transfer packages, and the tree's asset ref. Created via `NewAssetTreeContext`; nil for Bitcoin-only trees.
 - `SignerSession` — MuSig2 signing session for tree transactions, wrapping `input.MuSig2Signer`.
 - `Materializer` / `BTCMaterializer` — Interface and implementation for materializing tree nodes into actual Bitcoin transactions.
 - `TreeAssembler` — Two-pass builder (`BuildStructure` then `Materialize`) driven by `TreeConfig`.
 - `Queue[T]` — Generic queue used internally for BFS tree traversal.
+- `ComputeInternalKey` / `ComputeFinalKey` — Untweaked and taproot-tweaked MuSig2 aggregate keys for a cosigner set.
 
 ## Relationships
 
@@ -30,6 +34,9 @@ descriptors through branch nodes to the batch output.
 - `DefaultRadix` is 2 (binary tree). Each internal node has at most 2 children.
 - `NumLeafOutputs` is 2 per leaf transaction (VTXO output + sweep output).
 - Cosigner keys must be deduplicated (`UniqueCosigners`) before computing the final MuSig2 key.
+- Cosigner slices must be copied before being handed to `musig2.AggregateKeys` or
+  `MuSig2CreateSession` — both sort their input in place, and a `Node`'s
+  `CoSigners` order is load-bearing for downstream readers.
 - Tree materialization is deterministic given the same leaf descriptors and operator key.
 - `ValidateVTXODescriptors` / `ValidateConnectorDescriptor` must pass before tree construction.
 - **Cache-aliasing invariant**: a `*Tree` is effectively immutable once published from
@@ -37,6 +44,17 @@ descriptors through branch nodes to the batch output.
   pointer through caches and ancestry-fragment slices. Silently mutating a shared
   tree's nodes or root would corrupt every aliasing reader. Callers that need to
   transform a tree must clone it first.
+- Asset trees are additive to the Bitcoin tree: `LeafDescriptor.AssetAmount`
+  drives a separate aggregation pass (`Structure.AssetContext`) that never
+  changes the Bitcoin node hierarchy. Subtree totals must not overflow
+  `uint64`; `BuildStructure` fails rather than wrapping.
+- `AssetTreeContext` keys amounts by both `*Node` and input outpoint.
+  `ExtractPathForCoSigners` / `ExtractPathForIndices` clone nodes but preserve
+  their inputs and share the parent's `AssetContext` pointer, so lookups on an
+  extracted path resolve through the outpoint map.
+- When a tree has an `AssetContext`, `NewTreeSignerSession` signs each node with
+  the per-input taproot tweak recorded in the context instead of the tree's
+  `SweepTapscriptRoot`. A missing tweak is a hard error, never a fallback.
 
 ## Deep Docs
```

**How to apply:** save the diff above to a file and `git apply` it, or run the doc-gardening skill locally with `/doc-gardening lib/tree` and let it regenerate the pair.

Advisory only — this check does not fail CI. Posted by the per-PR doc-gardening workflow: https://github.com/lightninglabs/wavelength/actions

lib/tree/signing.go
darioAnongba: Good one. This will be handled in the client verification PR(s) by validating the sealed package and deriving the expected roots and tweaks before signing. Some pieces were indeed missing from the PoC.

lib/tree/asset_tree_context.go
darioAnongba: I don't think it should be moved onto `Node` because the materialization and decoding paths populate the outpoint index and missing final amounts are rejected. I added a test though covering the amount lookup after extraction.

lib/tree/tree.go
darioAnongba: No change needed IMO. The wire codex serializes only nodes in the extracted graph and does not serialize sealed packages, so won't be disclosed.

lib/tree/asset_tree_context_test.go
darioAnongba: Yeah, added a test that extracts a cloned path and verifies its amounts are resolved through the preserved outpoints.

lib/tree/leaf.go
```diff
@@ -19,4 +19,7 @@ type LeafDescriptor struct {
 	// CoSignerKey is the public key of the leaf owner who must participate
 	// in signing this leaf's transaction along with the operator.
 	CoSignerKey *btcec.PublicKey
+
+	// AssetAmount is the number of asset units in the leaf output.
+	AssetAmount uint64
```
GeorgeTsagk, line 24: **`LeafDescriptor.AssetAmount` is silently dropped by every public entry point.**

`TreeAssembler.BuildTree` (`btc_tree_assembler.go`, not in this diff) calls `BuildStructure` (which computes a full `AssetContext` when any leaf carries `AssetAmount > 0`) and then returns a `Tree` without it. A caller who sets the field gets a valid BTC-only tree with the asset value silently gone. For a value-bearing field, please reject `AssetAmount > 0` in the BTC path or thread the context through.

lib/tree/tree.go
```diff
@@ -321,8 +325,14 @@ func (t *Tree) NewTreeSignerSession(wallet input.MuSig2Signer,
 			"fetcher: %w", err)
 	}
 
-	return NewSignerSession(
+	var tweakLookup func(*Node) []byte
+	if t.AssetContext != nil {
```
GeorgeTsagk, line 329: **`NewTreeSignerSession` gates per-node tweaks on context presence, not tweak presence, and the wiring is untested.**

Any non-nil `AssetContext` makes per-node tweaks mandatory, but `BuildStructure` produces a context carrying only amounts, so signing such a tree fails on every node with "missing taproot tweak". Gate on the tweak map instead. Separately, deleting the `if t.AssetContext != nil` block keeps the whole `lib/tree` suite green; the one production integration point of the signing change needs a test that goes through `NewTreeSignerSession`.

GeorgeTsagk (changes requested):

lib/tree/leaf.go
darioAnongba: `TreeAssembler` is explicitly BTC-only. Asset trees use `BuildAssetTree`, so passing an asset amount here would be API misuse (aka a bug introduced by a dev + not tested). I'm generally against defending against developer misuse since we'll enter into infinite validation otherwise.

lib/tree/tree.go
darioAnongba: A non-nil asset context must contain every node tweak and falling back could sign with the wrong key. The asset-tree itest coming exercises `NewTreeSignerSession`. No change needed as using fake data here to test doesn't add value (itest does).

GeorgeTsagk (approved):
LGTM

PR #1177, opened 2026-08-21
tapassets: materialize asset VTXO trees

darioAnongba:
## Summary

- build asset VTXO trees through tap-sdk custom-anchor commits
- validate each node’s anchor transaction, output scripts, carrier values, asset identity, and per-issuance amount conservation
- cache completed sealed packages for restart replay and retry interrupted caller-funded preparations
- accept confirmed proofs or serialized proof paths as root inputs, including multiple issuances, and check proof depth and size before committing


go.mod
```diff
@@ -24,8 +24,8 @@ require (
 	github.com/lightninglabs/loop v0.33.0-beta
 	github.com/lightninglabs/neutrino v0.18.0
 	github.com/lightninglabs/neutrino/cache v1.1.4
-	github.com/lightninglabs/taproot-assets v0.7.1-0.20260706193822-2adfadc58e3c
-	github.com/lightninglabs/taproot-assets/taprpc v1.1.1-0.20260706193822-2adfadc58e3c
+	github.com/lightninglabs/taproot-assets v0.7.1-0.20260804093026-95bd8d412bef
+	github.com/lightninglabs/taproot-assets/taprpc v1.1.1-0.20260804093026-95bd8d412bef
```
darioAnongba: why do we need to depend directly on taproot-assets and taprpc? ideally we shoul depend only on tap-sdk.

go.mod
```diff
@@ -63,10 +63,12 @@ require (
 	github.com/btcsuite/btcd/txscript/v2 v2.0.0
 	github.com/btcsuite/btcd/wire/v2 v2.0.0
 	github.com/lightninglabs/go-wasmsqlite v0.0.0-20260811033710-d14cd0d80aa0
+	github.com/lightninglabs/tap-sdk v0.1.1-0.20260818172001-f657858b0fcc
```
darioAnongba, line 66: wait on the release of tap-sdk to merge this?

tapassets/driver.go
```diff
@@ -0,0 +1,292 @@
...
+	plan, err := builder.Build(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	result, err := plan.Commit(ctx, tapsdk.CustomAnchorCommitOptions{
+		Publish: tapsdk.CustomAnchorPublishMetadata{
+			SkipAnchorTxBroadcast: true,
+			ExternalBroadcast:     true,
+			Label:                 "wavelength-asset-transition",
```
darioAnongba: why is this label needed? why hardcoded?

tapassets/journal.go
```diff
@@ -0,0 +1,23 @@
...
+	Load(context.Context, string) ([]byte, error)
+
+	Store(context.Context, string, []byte) error
+}
+
+// ErrReconciliationRequired reports a tapd commit that may have succeeded.
+// Retrying without reconciling the recorded attempt could spend the same
+// asset state twice.
+var ErrReconciliationRequired = errors.New("asset transition reconciliation " +
+	"required")
```
darioAnongba: why not defined under ErrStoreNotFound?

go.mod
darioAnongba: Wavelength already imports taproot-assets/proof in db and taprpc in harness, so tap-sdk cannot be the module's only direct dependency. I changed both pins to the versions required by this tap-sdk revision.

tapassets/driver.go
darioAnongba: It was not needed here. I removed the hard-coded label.

tapassets/journal.go
darioAnongba: Agreed. Both package errors are now in the same var block.

litbot-9000:
This PR's Go changes leave `lib/tree`'s package doc stale — the new `AssetTreeContext` type, the `LeafDescriptor.AssetAmount` field, the per-node asset signing tweak, and the cosigner-slice copying are not yet reflected in `lib/tree/CLAUDE.md` / `AGENTS.md`. (`tapassets/` has no `CLAUDE.md` yet, so it is left to the nightly full sweep.)

```diff
diff --git a/lib/tree/AGENTS.md b/lib/tree/AGENTS.md
index 55fabdea..f07338c2 100644
--- a/lib/tree/AGENTS.md
+++ b/lib/tree/AGENTS.md
@@ -8,12 +8,13 @@ descriptors through branch nodes to the batch output.
 
 ## Key Types
 
-- `Tree` — Complete VTXO or connector tree: root outpoint, root output, node hierarchy, and traversal helpers. Built via `BuildVTXOTree` or `BuildConnectorTree`.
+- `Tree` — Complete VTXO or connector tree: root outpoint, root output, node hierarchy, traversal helpers, and an optional `AssetContext`. Built via `BuildVTXOTree` or `BuildConnectorTree`.
 - `Node` — Single tree node representing a transaction in the tree (branch or leaf).
-- `LeafDescriptor` — Describes a single VTXO leaf: amount, owner pubkeys, cosigner keys, CSV delay.
+- `LeafDescriptor` — Describes a single VTXO leaf: amount, owner pubkeys, cosigner keys, CSV delay, and an optional `AssetAmount` for asset trees.
 - `VTXODescriptor` — Interface for VTXO metadata needed by tree construction (amount, cosigners, owner key).
 - `ConnectorDescriptor` — Describes a connector output for forfeit transaction construction.
-- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization.
+- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization; carries the `AssetContext` when the leaves declare asset amounts.
+- `AssetTreeContext` — Side table holding the asset state of a tree: per-subtree asset amounts, per-input taproot signing tweaks, leaf commitment roots, sealed transfer packages, and the asset reference. Keeps asset data out of `Node` so the Bitcoin tree stays asset-agnostic.
 - `StructureConfig` — Configuration for tree building (radix, partition weight function).
 - `SignerSession` — MuSig2 signing session for tree transactions, wrapping `input.MuSig2Signer`.
 - `Materializer` / `BTCMaterializer` — Interface and implementation for materializing tree nodes into actual Bitcoin transactions.
@@ -23,7 +24,7 @@ descriptors through branch nodes to the batch output.
 ## Relationships
 
 - **Depends on**: `lib/arkscript` (taproot script construction, policy templates, `SpendInfo`).
-- **Depended on by**: `round` (tree construction/validation), `oor` (tree references), `db` (tree serialization).
+- **Depended on by**: `round` (tree construction/validation), `oor` (tree references), `db` (tree serialization), `tapassets` (asset tree assembly and materialization, which populates `AssetTreeContext`).
 
 ## Invariants
 
@@ -37,6 +38,23 @@ descriptors through branch nodes to the batch output.
   pointer through caches and ancestry-fragment slices. Silently mutating a shared
   tree's nodes or root would corrupt every aliasing reader. Callers that need to
   transform a tree must clone it first.
+- **Cosigner-slice copy invariant**: `musig2.AggregateKeys` and local
+  `input.MuSig2Signer` implementations sort their cosigner input in place, so
+  `ComputeInternalKey`, `ComputeFinalKey`, and `Node.NewSignerSession` copy the
+  slice before handing it over. A shared `Node.CoSigners` must never be passed
+  directly, or the node's cosigner order (and the keys derived from it) would
+  change under aliasing readers.
+- `Tree.AssetContext` and `Structure.AssetContext` are nil for Bitcoin-only
+  trees. `BuildStructure` allocates the context only when at least one leaf
+  declares a non-zero `AssetAmount`, and `aggregateAssetAmounts` then records
+  the asset total of every subtree (erroring on `uint64` overflow).
+- Asset trees sign with a per-node taproot tweak taken from
+  `AssetTreeContext.SigningTweak` instead of the tree-wide
+  `SweepTapscriptRoot`; a node with no recorded tweak is a hard error rather
+  than a silent fallback.
+- `ExtractPathForCoSigners` / `ExtractPathForIndices` clone nodes but share the
+  parent's `*AssetTreeContext` pointer. Asset lookups therefore key on the
+  node's input outpoint, which survives cloning, not on the node pointer.
 
 ## Deep Docs
 
diff --git a/lib/tree/CLAUDE.md b/lib/tree/CLAUDE.md
index 55fabdea..f07338c2 100644
--- a/lib/tree/CLAUDE.md
+++ b/lib/tree/CLAUDE.md
@@ -8,12 +8,13 @@ descriptors through branch nodes to the batch output.
 
 ## Key Types
 
-- `Tree` — Complete VTXO or connector tree: root outpoint, root output, node hierarchy, and traversal helpers. Built via `BuildVTXOTree` or `BuildConnectorTree`.
+- `Tree` — Complete VTXO or connector tree: root outpoint, root output, node hierarchy, traversal helpers, and an optional `AssetContext`. Built via `BuildVTXOTree` or `BuildConnectorTree`.
 - `Node` — Single tree node representing a transaction in the tree (branch or leaf).
-- `LeafDescriptor` — Describes a single VTXO leaf: amount, owner pubkeys, cosigner keys, CSV delay.
+- `LeafDescriptor` — Describes a single VTXO leaf: amount, owner pubkeys, cosigner keys, CSV delay, and an optional `AssetAmount` for asset trees.
 - `VTXODescriptor` — Interface for VTXO metadata needed by tree construction (amount, cosigners, owner key).
 - `ConnectorDescriptor` — Describes a connector output for forfeit transaction construction.
-- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization.
+- `Structure` — Intermediate tree layout built by `BuildStructure` before materialization; carries the `AssetContext` when the leaves declare asset amounts.
+- `AssetTreeContext` — Side table holding the asset state of a tree: per-subtree asset amounts, per-input taproot signing tweaks, leaf commitment roots, sealed transfer packages, and the asset reference. Keeps asset data out of `Node` so the Bitcoin tree stays asset-agnostic.
 - `StructureConfig` — Configuration for tree building (radix, partition weight function).
 - `SignerSession` — MuSig2 signing session for tree transactions, wrapping `input.MuSig2Signer`.
 - `Materializer` / `BTCMaterializer` — Interface and implementation for materializing tree nodes into actual Bitcoin transactions.
@@ -23,7 +24,7 @@ descriptors through branch nodes to the batch output.
 ## Relationships
 
 - **Depends on**: `lib/arkscript` (taproot script construction, policy templates, `SpendInfo`).
-- **Depended on by**: `round` (tree construction/validation), `oor` (tree references), `db` (tree serialization).
+- **Depended on by**: `round` (tree construction/validation), `oor` (tree references), `db` (tree serialization), `tapassets` (asset tree assembly and materialization, which populates `AssetTreeContext`).
 
 ## Invariants
 
@@ -37,6 +38,23 @@ descriptors through branch nodes to the batch output.
   pointer through caches and ancestry-fragment slices. Silently mutating a shared
   tree's nodes or root would corrupt every aliasing reader. Callers that need to
   transform a tree must clone it first.
+- **Cosigner-slice copy invariant**: `musig2.AggregateKeys` and local
+  `input.MuSig2Signer` implementations sort their cosigner input in place, so
+  `ComputeInternalKey`, `ComputeFinalKey`, and `Node.NewSignerSession` copy the
+  slice before handing it over. A shared `Node.CoSigners` must never be passed
+  directly, or the node's cosigner order (and the keys derived from it) would
+  change under aliasing readers.
+- `Tree.AssetContext` and `Structure.AssetContext` are nil for Bitcoin-only
+  trees. `BuildStructure` allocates the context only when at least one leaf
+  declares a non-zero `AssetAmount`, and `aggregateAssetAmounts` then records
+  the asset total of every subtree (erroring on `uint64` overflow).
+- Asset trees sign with a per-node taproot tweak taken from
+  `AssetTreeContext.SigningTweak` instead of the tree-wide
+  `SweepTapscriptRoot`; a node with no recorded tweak is a hard error rather
+  than a silent fallback.
+- `ExtractPathForCoSigners` / `ExtractPathForIndices` clone nodes but share the
+  parent's `*AssetTreeContext` pointer. Asset lookups therefore key on the
+  node's input outpoint, which survives cloning, not on the node pointer.
 
 ## Deep Docs
```

**How to apply:** save the diff above to a file and `git apply` it, or run the
doc-gardening skill locally against this package (`/doc-gardening lib/tree`).
`lib/tree/CLAUDE.md` and `lib/tree/AGENTS.md` must stay byte-identical.

---
Posted by the per-PR documentation advisory run:
https://github.com/lightninglabs/wavelength/actions

tapassets/transition_helpers.go
```diff
@@ -0,0 +1,292 @@
...
+		InternalKey: tapsdk.InternalKey{
+			PubKey: pubKey,
+		},
+		Tapscript: tapsdk.CustomAnchorTapscriptPlan{
+			TapLeaves: sdkLeaves,
+		},
+	}
+}
+
+func deterministicKey(digest tapsdk.Hash, domain string) tapsdk.PubKey {
```
GeorgeTsagk, line 252: **The OP_TRUE internal keys are publicly recomputable private keys, and the "own BTC = own assets" rule is undocumented.**

The OP_TRUE internal keys are derived from a public seed, so anyone can recompute the private key. Harmless while the script is literally OP_TRUE, but a footgun if any future path treats these as real keys. Prefer a NUMS point, or document why a known key is safe here.

Related: this PR commits to "BTC ownership = asset ownership" for all in-tree outputs. Reasonable rule, but please state it as a protocol decision, including that the operator CSV sweep takes the assets with the carrier sats, and that unrolled users present proofs whose final hops are OP_TRUE witnesses. Worth confirming the wider tapd recipient ecosystem accepts those proofs.

tapassets/tree_materializer.go
```diff
@@ -0,0 +1,971 @@
...
+// TreeMaterializerConfig configures asset tree materialization.
+type TreeMaterializerConfig struct {
+	// Wallet is the tap-sdk wallet used to commit node transitions.
+	Wallet *tapsdk.Wallet
+
+	// Store records commit attempts and their sealed packages.
+	Store Store
+
+	// AssetRef is the asset carried by every leaf of this tree.
+	AssetRef tapsdk.AssetRef
```
GeorgeTsagk, line 80: **One asset (or group) per tree is a hard structural limit, and it is undocumented.**

tap-sdk's `AssetRef` encodes either an asset ID or a group key, so one tree can carry multiple issuances of a grouped asset. The hard limit is no mixed groups per tree, meaning mixed-asset rounds need one tree per group. Fine as scoping, but please document it on `AssetTreeRequest`/`TreeMaterializerConfig`.

tapassets/journal.go
```diff
@@ -0,0 +1,23 @@
...
+	"context"
+	"errors"
+)
+
+var (
+	// ErrStoreNotFound reports that a journal key has no durable value.
+	ErrStoreNotFound = errors.New("asset transition state not found")
+
+	// ErrReconciliationRequired reports a commit whose outcome is unknown.
+	ErrReconciliationRequired = errors.New("asset transition " +
```
GeorgeTsagk: **The journal can halt the build, but nothing can un-halt it; package state is duplicated across two stores.**

The durable-commit flow is good: replay without re-commit, digest pinning, outcome classification, `WithoutCancel` post-commit writes. Two asks. `ErrReconciliationRequired` has no resolver anywhere in the stack; please document the intended operational path, since an unresolvable halt state needs a runbook. And sealed packages now live in both the `Store` and the `AssetTreeContext`; please state which is authoritative and when they may diverge.

tapassets/transition_helpers.go
darioAnongba: Yeah, OP_TRUE on the asset layer is really the core of the implementation and design and most things revolve around this. IMO comments should not describe the "why", protocols decisions are extensively documented in the design docs (litbucket and issues) and will be also added in the README, AGENTS.md and CLAUDE.md of the tapassets package in a subsequent PR.

> Worth confirming the wider tapd recipient ecosystem accepts those proofs.

the PoC proves this works with tapd, the e2e tests in lumos use a real tapd, also tested manually.

tapassets/journal.go
darioAnongba: Good one, changed. The store now caches completed packages only. The tree contesxt is derived from the validated cache result 👍

tapassets/tree_materializer.go
darioAnongba: This is deliberate for now at least, could always be improved but on my mental model it's clean to have one tree per AssetRef, an AssetRef being either an asset ID for single issuance assets or group key for grouped assets, meaning multiple issuances in the same tree.

tapassets/tree_materializer.go
```diff
@@ -0,0 +1,972 @@
...
+		if len(outputs) == 0 {
+			return nil, fmt.Errorf("committed result misses "+
+				"output %d", spec.index)
+		}
+		if err := m.validateCommitOutputs(
+			spec, committedTx, outputs,
+		); err != nil {
+			return nil, err
+		}
+		root := outputs[0].taprootMerkleRoot
```
GeorgeTsagk, line 649: **`validateCommit` never binds the committed merkle root to the expected tap leaves.**

Here and at `:774`, `taprootMerkleRoot` is taken from tapd as ground truth and only checked to compose with `spec.internalKey` into the committed script. The expected leaves (sweep leaf for branches, VTXO policy for leaves) are never hashed into the comparison, so a root that omits the sweep leaf or adds an anyone-can-spend leaf passes. The SDK's `ValidateAnchorOutputs` backstops this in the real driver path, but the seam should defend itself: recompute the expected root from `spec.tapLeaves` plus the asset leaf construction. The tamper tests cannot catch this class because the fake computes scripts from the request; add a case with a self-consistent root that drops the sweep leaf.

tapassets/transition_helpers.go
```diff
@@ -0,0 +1,292 @@
...
+	issuanceID     tapsdk.AssetID
+	issuanceKnown  bool
+	anchorOutpoint tapsdk.Outpoint
+}
+
+type assetSourceVerifier struct {
+	sources []*assetSpendSource
+}
+
+func (v *assetSourceVerifier) VerifyConfirmedProof(ctx context.Context,
```
GeorgeTsagk, line 44: **`assetSourceVerifier` is not bound to its input.**

Both verify methods iterate all sources and return on the first success, so input A's proof can be ratified by input B's caller-supplied verifier. The SDK revalidates underneath, but the per-input binding the `TreeRootAssetInput.Verifier` API suggests is not what the code does.

tapassets/tree_materializer.go
```diff
@@ -0,0 +1,972 @@
...
+
+	return committedTx, nil
+}
+
+func validateIssuanceConservation(sources []*assetSpendSource,
+	outputs []commitOutput) error {
+
+	expected := make(map[tapsdk.AssetID]uint64, len(sources))
+	for _, source := range sources {
+		expected[source.issuanceID] = source.amount
```
GeorgeTsagk, line 693: **The repeated-issuance overwrite from the earlier thread is still live at this head.**

`expected[source.issuanceID] = source.amount` (and `sourceByIssuance[source.issuanceID] = source` at `:851`) still overwrite instead of accumulating, so two root inputs carrying the same issuance make a legitimate configuration unbuildable and mislead the conservation check. The existing codex comment on this line covers it; confirming it still applies after the rebase.

GeorgeTsagk (changes requested):

tapassets/tree_materializer.go
darioAnongba: The SDK already validates sealed-package outputs against the committed virtual packets and requested tapscript sibling. Recomputing it here would duplicate SDK validation. The SDK is not an external deps we're defending against.

tapassets/transition_helpers.go
darioAnongba: The verifier validates proofs, not input indexes. The SDK subsequently binds each verified proof to the requested asset reference, issuance, amount, script key, and outpoint.

tapassets/tree_materializer.go
darioAnongba: The batch anchor merges inputs from the same issuance into one proof path before tree materialization, so the materializer receives one source per issuance. This is covered by the multi-tranche integration flow.

GeorgeTsagk (approved):
LGTM

PR #1189, opened 2026-08-24
arkscript/oor: Prepare channel-backed OOR funding

sputn1ck:
## Summary

- define the channel VTXO policy with an immediate 3-of-3 cooperative path, a CSV-delayed 2-of-2 channel-funding path, and a longer funder-refund path
- prepare and validate an OOR transfer before either party commits it
- persist the source lineage needed to materialize the unpublished channel point during recovery

This is layer 1 of 5 in the Ark channels stack. The next layer is #1190.

## Testing

- `go test ./lib/arkscript ./oor ./unroll ./vtxo ./db`
- `make build`
- `make lint-changed-local`

### DIFF
--- a/db/vtxo_store.go
+++ b/db/vtxo_store.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"math"
 	"slices"
 
 	"github.com/btcsuite/btcd/btcec/v2"
@@ -119,6 +120,125 @@ func (s *VTXOPersistenceStore) SaveVTXO(
 			int32(desc.Outpoint.Index), desc.Ancestry,
 		)
 	})
+}
+
+// SaveRecoveryOnlyVTXO atomically persists an application-owned descriptor in
+// the recovery-only state. Unlike SaveVTXO, this method never exposes the row
+// as live between insertion and the status transition, so concurrent wallet
+// selection cannot mistake a channel or contract output for wallet liquidity.
+// Replays are accepted only when the existing immutable descriptor matches.
+func (s *VTXOPersistenceStore) SaveRecoveryOnlyVTXO(ctx context.Context,
+	desc *vtxo.Descriptor) error {
+
+	if desc == nil {
+		return fmt.Errorf("descriptor must be provided")
+	}
+	if desc.Status != vtxo.VTXOStatusRecoveryOnly {
+		return fmt.Errorf("recovery-only descriptor has status %s",
+			desc.Status)
+	}
+
+	return s.db.ExecTx(ctx, WriteTxOption(), func(q RoundStore) error {
+		key := sqlc.GetVTXOParams{
+			OutpointHash:  desc.Outpoint.Hash[:],
+			OutpointIndex: int32(desc.Outpoint.Index),
+		}
+		existing, err := q.GetVTXO(ctx, key)
+		switch {
+		case err == nil:
+			if err := validateRecoveryOnlyVTXORow(
+				existing, desc,
+			); err != nil {
+				return err
+			}
+
+			return upsertAncestryPaths(
+				ctx, q, desc.Outpoint.Hash[:],
+				int32(desc.Outpoint.Index), desc.Ancestry,
+			)
+
+		case !errors.Is(err, sql.ErrNoRows):
+			return fmt.Errorf("load recovery-only VTXO: %w", err)
+		}
+
+		if err := s.ensureRoundExists(ctx, q, desc); err != nil {
+			return fmt.Errorf("ensure round: %w", err)
+		}
+		params, err := s.descriptorToInsertParams(ctx, q, desc)
+		if err != nil {
+			return fmt.Errorf("convert descriptor: %w", err)
+		}
+		if err := q.InsertVTXO(ctx, params); err != nil {
+			return fmt.Errorf("insert recovery-only VTXO: %w", err)
+		}
+		if err := q.UpdateVTXOStatus(ctx, sqlc.UpdateVTXOStatusParams{
+			OutpointHash:   desc.Outpoint.Hash[:],
+			OutpointIndex:  int32(desc.Outpoint.Index),
+			Status:         int32(desc.Status),
+			LastUpdateTime: s.clock.Now().Unix(),
+		}); err != nil {
+			return fmt.Errorf("mark VTXO recovery-only: %w", err)
+		}
+
+		return upsertAncestryPaths(
+			ctx, q, desc.Outpoint.Hash[:],
+			int32(desc.Outpoint.Index), desc.Ancestry,
+		)
+	})
+}
+
+// SetRecoveryOnlyVTXORelativeExpiry selects the delay used by the exact Ark
+// channel final-spend policy without making the descriptor wallet-owned.
+func (s *VTXOPersistenceStore) SetRecoveryOnlyVTXORelativeExpiry(
+	ctx context.Context, outpoint wire.OutPoint, delay uint32) error {
+
+	if delay > math.MaxInt32 {
+		return fmt.Errorf("recovery-only VTXO delay %d exceeds "+
+			"storage range", delay)
+	}
+
+	return s.db.ExecTx(ctx, WriteTxOption(), func(q RoundStore) error {
+		rows, err := q.SetRecoveryOnlyVTXORelativeExpiry(
+			ctx, sqlc.SetRecoveryOnlyVTXORelativeExpiryParams{
+				OutpointHash:   outpoint.Hash[:],
+				OutpointIndex:  int32(outpoint.Index),
+				Expiry:         int32(delay),
+				LastUpdateTime: s.clock.Now().Unix(),
+			},
+		)
+		if err != nil {
+			return err
+		}
+		if rows != 1 {
+			return fmt.Errorf("recovery-only VTXO %s not found",
+				outpoint)
+		}
+
+		return nil
+	})
+}
+
+// validateRecoveryOnlyVTXORow rejects an outpoint collision or a replay after
+// another subsystem has taken ownership of the same VTXO.
+func validateRecoveryOnlyVTXORow(row VTXORow, desc *vtxo.Descriptor) error {
+	status := vtxo.VTXOStatus(row.Status)
+	if status != vtxo.VTXOStatusRecoveryOnly &&
+		status != vtxo.VTXOStatusSpent {
+		return fmt.Errorf("recovery-only VTXO already has status %s",
+			status)
+	}
+	if row.RoundID != desc.RoundID || row.Amount != int64(desc.Amount) ||
+		!bytes.Equal(row.PkScript, desc.PkScript) ||
+		!bytes.Equal(row.PolicyTemplate, desc.PolicyTemplate) ||
+		row.Expiry != int32(desc.RelativeExpiry) ||
+		row.BatchExpiry != desc.BatchExpiry ||
+		row.ChainDepth != int32(desc.ChainDepth) ||
+		row.CreatedHeight != desc.CreatedHeight ||
+		!bytes.Equal(row.CommitmentTxid, desc.CommitmentTxID[:]) {
+		return fmt.Errorf("recovery-only VTXO descriptor mismatch")
+	}
+
+	return nil
 }
 
 // ensureRoundExists inserts a minimal confirmed round row when a VTXO refers
--- a/db/vtxo_store.go
+++ b/db/vtxo_store.go
@@ -219,7 +219,8 @@ func (s *VTXOPersistenceStore) SetRecoveryOnlyVTXORelativeExpiry(
 }
 
 // validateRecoveryOnlyVTXORow rejects an outpoint collision or a replay after
-// another subsystem has taken ownership of the same VTXO.
+// another subsystem has taken ownership of the same VTXO. Relative expiry is
+// excluded because channel recovery selects it after registration.
 func validateRecoveryOnlyVTXORow(row VTXORow, desc *vtxo.Descriptor) error {
 	status := vtxo.VTXOStatus(row.Status)
 	if status != vtxo.VTXOStatusRecoveryOnly &&
@@ -230,7 +231,6 @@ func validateRecoveryOnlyVTXORow(row VTXORow, desc *vtxo.Descriptor) error {
 	if row.RoundID != desc.RoundID || row.Amount != int64(desc.Amount) ||
 		!bytes.Equal(row.PkScript, desc.PkScript) ||
 		!bytes.Equal(row.PolicyTemplate, desc.PolicyTemplate) ||
-		row.Expiry != int32(desc.RelativeExpiry) ||
 		row.BatchExpiry != desc.BatchExpiry ||
 		row.ChainDepth != int32(desc.ChainDepth) ||
 		row.CreatedHeight != desc.CreatedHeight ||
--- a/db/vtxo_store_test.go
+++ b/db/vtxo_store_test.go
@@ -2484,3 +2484,62 @@ func TestVTXOPersistenceStoreListVTXOsByStatusesLight(t *testing.T) {
 		outpoints(live),
 	)
 }
+
+// TestVTXOStoreRecoveryOnlyExcludedFromWalletSets verifies that a descriptor
+// prepared for channel materialization remains available to the unroller but
+// cannot become wallet liquidity or spawn an ordinary VTXO actor.
+func TestVTXOStoreRecoveryOnlyExcludedFromWalletSets(t *testing.T) {
+	t.Parallel()
+
+	vtxoStore, _, _ := newVTXOStoreForTest(t)
+	ctx := t.Context()
+	desc := createTestVTXODescriptor(
+		t, testRoundIDDB("test-round-recovery-only"), 21,
+	)
+	desc.Status = vtxo.VTXOStatusRecoveryOnly
+
+	require.NoError(t, vtxoStore.SaveRecoveryOnlyVTXO(ctx, desc))
+
+	stored, err := vtxoStore.GetVTXO(ctx, desc.Outpoint)
+	require.NoError(t, err)
+	require.Equal(t, vtxo.VTXOStatusRecoveryOnly, stored.Status)
+	require.Len(t, stored.Ancestry, len(desc.Ancestry))
+	require.Equal(
+		t, desc.Ancestry[0].CommitmentTxID,
+		stored.Ancestry[0].CommitmentTxID,
+	)
+	require.Equal(
+		t, desc.Ancestry[0].TreeDepth, stored.Ancestry[0].TreeDepth,
+	)
+
+	live, err := vtxoStore.ListLiveVTXOs(ctx)
+	require.NoError(t, err)
+	require.Empty(t, live)
+
+	recoverable, err := vtxoStore.ListRecoverableVTXOs(ctx)
+	require.NoError(t, err)
+	require.Empty(t, recoverable)
+	require.ErrorContains(
+		t, vtxoStore.SetRecoveryOnlyVTXORelativeExpiry(
+			ctx, desc.Outpoint, ^uint32(0),
+		),
+		"exceeds storage range",
+	)
+
+	candidates, err := vtxoStore.ListSelectionCandidatesByStatus(
+		ctx, vtxo.VTXOStatusLive,
+	)
+	require.NoError(t, err)
+	require.Empty(t, candidates)
+
+	// Durable close actions may replay source preparation after the row has
+	// already been committed. Identical terms are idempotent, while an
+	// outpoint collision with different immutable terms must fail.
+	require.NoError(t, vtxoStore.SaveRecoveryOnlyVTXO(ctx, desc))
+	mismatch := *desc
+	mismatch.Amount++
+	require.ErrorContains(
+		t, vtxoStore.SaveRecoveryOnlyVTXO(ctx, &mismatch),
+		"descriptor mismatch",
+	)
+}
--- a/db/vtxo_store_test.go
+++ b/db/vtxo_store_test.go
@@ -2526,6 +2526,23 @@ func TestVTXOStoreRecoveryOnlyExcludedFromWalletSets(t *testing.T) {
 		"exceeds storage range",
 	)
 
+	// Channel registration sets the materialization delay after the row is
+	// first saved. Replaying source preparation must preserve that selected
+	// value instead of mistaking it for an outpoint collision.
+	selectedExpiry := desc.RelativeExpiry + 10
+	require.NoError(
+		t, vtxoStore.SetRecoveryOnlyVTXORelativeExpiry(
+			ctx, desc.Outpoint, selectedExpiry,
+		),
+	)
+	require.NoError(t, vtxoStore.SaveRecoveryOnlyVTXO(ctx, desc))
+	storedRow, err := vtxoStore.db.GetVTXO(ctx, sqlc.GetVTXOParams{
+		OutpointHash:  desc.Outpoint.Hash[:],
+		OutpointIndex: int32(desc.Outpoint.Index),
+	})
+	require.NoError(t, err)
+	require.Equal(t, int32(selectedExpiry), storedRow.Expiry)
+
 	candidates, err := vtxoStore.ListSelectionCandidatesByStatus(
 		ctx, vtxo.VTXOStatusLive,
 	)
@@ -2541,5 +2558,24 @@ func TestVTXOStoreRecoveryOnlyExcludedFromWalletSets(t *testing.T) {
 	require.ErrorContains(
 		t, vtxoStore.SaveRecoveryOnlyVTXO(ctx, &mismatch),
 		"descriptor mismatch",
+	)
+
+	// The setter must not repurpose an ordinary wallet VTXO as channel
+	// recovery state.
+	liveDescriptor := createTestVTXODescriptor(
+		t, testRoundIDDB("test-round-live-expiry-guard"), 22,
+	)
+	require.NoError(t, vtxoStore.SaveVTXO(ctx, liveDescriptor))
+	require.ErrorContains(
+		t, vtxoStore.SetRecoveryOnlyVTXORelativeExpiry(
+			ctx, liveDescriptor.Outpoint,
+			liveDescriptor.RelativeExpiry+10,
+		),
+		"not found",
+	)
+	storedLive, err := vtxoStore.GetVTXO(ctx, liveDescriptor.Outpoint)
+	require.NoError(t, err)
+	require.Equal(
+		t, liveDescriptor.RelativeExpiry, storedLive.RelativeExpiry,
 	)
 }
--- a/lib/arkscript/channel_vtxo.go
+++ b/lib/arkscript/channel_vtxo.go
@@ -1,0 +1,291 @@
+package arkscript
+
+import (
+	"bytes"
+	"fmt"
+	"math"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/wire/v2"
+)
+
+const (
+	// DefaultChannelReactionWindow gives either channel endpoint roughly
+	// three days to materialize before the original funder can refund.
+	DefaultChannelReactionWindow uint32 = 432
+)
+
+// ChannelVTXOParams defines the key roles and relative delays of a VTXO that
+// backs an unpublished Lightning channel point.
+type ChannelVTXOParams struct {
+	ClientArkKey   *btcec.PublicKey
+	HubArkKey      *btcec.PublicKey
+	ArkOperatorKey *btcec.PublicKey
+
+	ClientChannelKey *btcec.PublicKey
+	HubChannelKey    *btcec.PublicKey
+	FunderKey        *btcec.PublicKey
+
+	ChannelDelay   uint32
+	FunderDelay    uint32
+	MinExitDelay   uint32
+	ReactionWindow uint32
+}
+
+// ChannelVTXOPolicy is the compiled three-path policy for an Ark-backed
+// Lightning channel.
+type ChannelVTXOPolicy struct {
+	Template *PolicyTemplate
+	*CompiledPolicy
+
+	CooperativeClosure Node
+	ChannelClosure     Node
+	FunderClosure      Node
+
+	ChannelDelay uint32
+	FunderDelay  uint32
+}
+
+// NewChannelVTXOPolicy constructs the cooperative, materialization, and
+// funder-refund paths while enforcing their relative timing.
+func NewChannelVTXOPolicy(params ChannelVTXOParams) (*ChannelVTXOPolicy,
+	error) {
+
+	if err := params.validate(); err != nil {
+		return nil, err
+	}
+
+	cooperativeClosure := &Multisig{
+		Keys: []*btcec.PublicKey{
+			params.ClientArkKey,
+			params.HubArkKey,
+			params.ArkOperatorKey,
+		},
+	}
+	channelClosure := &CSV{
+		Lock: blockchain.LockTimeToSequence(
+			false, params.ChannelDelay,
+		),
+		Inner: &Multisig{
+			Keys: []*btcec.PublicKey{
+				params.ClientChannelKey,
+				params.HubChannelKey,
+			},
+		},
+	}
+	funderClosure := &CSV{
+		Lock: blockchain.LockTimeToSequence(
+			false, params.FunderDelay,
+		),
+		Inner: &Multisig{
+			Keys: []*btcec.PublicKey{
+				params.FunderKey,
+			},
+		},
+	}
+
+	template := &PolicyTemplate{
+		Leaves: []LeafTemplate{
+			{
+				Node: cooperativeClosure,
+			},
+			{
+				Node: channelClosure,
+			},
+			{
+				Node: funderClosure,
+			},
+		},
+	}
+	if err := template.ValidateArkPolicy(PolicyValidationOpts{
+		OperatorKey:  params.ArkOperatorKey,
+		MinExitDelay: params.MinExitDelay,
+	}); err != nil {
+		return nil, fmt.Errorf("validate channel VTXO policy: %w", err)
+	}
+
+	compiled, err := template.Compile()
+	if err != nil {
+		return nil, fmt.Errorf("compile channel VTXO policy: %w", err)
+	}
+
+	return &ChannelVTXOPolicy{
+		Template:           template,
+		CompiledPolicy:     compiled,
+		CooperativeClosure: cooperativeClosure,
+		ChannelClosure:     channelClosure,
+		FunderClosure:      funderClosure,
+		ChannelDelay:       params.ChannelDelay,
+		FunderDelay:        params.FunderDelay,
+	}, nil
+}
+
+// CooperativeSpendPath returns the immediate client-hub-Ark-operator path.
+func (p *ChannelVTXOPolicy) CooperativeSpendPath() (*SpendPath, error) {
+	return p.SpendPathForNode(p.CooperativeClosure, nil)
+}
+
+// ChannelSpendPath returns the delayed two-party materialization path.
+func (p *ChannelVTXOPolicy) ChannelSpendPath() (*SpendPath, error) {
+	return p.SpendPathForNode(p.ChannelClosure, nil)
+}
+
+// FunderSpendPath returns the later unilateral refund path.
+func (p *ChannelVTXOPolicy) FunderSpendPath() (*SpendPath, error) {
+	return p.SpendPathForNode(p.FunderClosure, nil)
+}
+
+// EncodeChannelVTXOArtifacts returns the semantic policy and its canonical
+// P2TR output script.
+func EncodeChannelVTXOArtifacts(params ChannelVTXOParams) ([]byte, []byte,
+	error) {
+
+	policy, err := NewChannelVTXOPolicy(params)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	template, err := policy.Template.Encode()
+	if err != nil {
+		return nil, nil, err
+	}
+	pkScript, err := policy.Template.PkScript()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return template, pkScript, nil
+}
+
+// ValidateChannelVTXOTemplate verifies that an untrusted template has the
+// exact key roles and delays expected by a channel intent.
+func ValidateChannelVTXOTemplate(template *PolicyTemplate,
+	params ChannelVTXOParams) error {
+
+	if template == nil {
+		return fmt.Errorf("channel VTXO template is required")
+	}
+
+	expected, err := NewChannelVTXOPolicy(params)
+	if err != nil {
+		return err
+	}
+	if len(template.Leaves) != len(expected.Template.Leaves) {
+		return fmt.Errorf("channel VTXO policy must contain exactly " +
+			"3 leaves")
+	}
+
+	actualScripts := make(map[string]struct{}, len(template.Leaves))
+	for i := range template.Leaves {
+		script, err := template.Leaves[i].Script()
+		if err != nil {
+			return fmt.Errorf("compile channel VTXO leaf %d: %w", i,
+				err)
+		}
+
+		actualScripts[string(script)] = struct{}{}
+	}
+
+	for i := range expected.Template.Leaves {
+		script, err := expected.Template.Leaves[i].Script()
+		if err != nil {
+			return err
+		}
+		if _, ok := actualScripts[string(script)]; !ok {
+			return fmt.Errorf("channel VTXO policy does not " +
+				"match expected key roles and delays")
+		}
+	}
+
+	return nil
+}
+
+// validate enforces key separation and the funder reaction window.
+func (p ChannelVTXOParams) validate() error {
+	keys := []struct {
+		name string
+		key  *btcec.PublicKey
+	}{
+		{
+			name: "client Ark",
+			key:  p.ClientArkKey,
+		},
+		{
+			name: "hub Ark",
+			key:  p.HubArkKey,
+		},
+		{
+			name: "Ark operator",
+			key:  p.ArkOperatorKey,
+		},
+		{
+			name: "client channel",
+			key:  p.ClientChannelKey,
+		},
+		{
+			name: "hub channel",
+			key:  p.HubChannelKey,
+		},
+		{
+			name: "funder",
+			key:  p.FunderKey,
+		},
+	}
+	for _, key := range keys {
+		if key.key == nil {
+			return fmt.Errorf("channel VTXO %s key is nil",
+				key.name)
+		}
+	}
+
+	if channelKeysEqual(p.ClientArkKey, p.HubArkKey) ||
+		channelKeysEqual(p.ClientArkKey, p.ArkOperatorKey) ||
+		channelKeysEqual(p.HubArkKey, p.ArkOperatorKey) {
+		return fmt.Errorf("channel VTXO cooperative keys must be " +
+			"controlled by three distinct parties")
+	}
+	if channelKeysEqual(p.ClientChannelKey, p.HubChannelKey) {
+		return fmt.Errorf("channel VTXO materialization keys must be " +
+			"distinct")
+	}
+
+	if p.MinExitDelay == 0 {
+		return fmt.Errorf("channel VTXO minimum exit delay must be " +
+			"non-zero")
+	}
+	if p.ChannelDelay < p.MinExitDelay {
+		return fmt.Errorf("channel delay %d is below Ark minimum %d",
+			p.ChannelDelay, p.MinExitDelay)
+	}
+	if p.ChannelDelay > wire.SequenceLockTimeMask ||
+		p.FunderDelay > wire.SequenceLockTimeMask {
+		return fmt.Errorf("channel VTXO delays must fit BIP-68 block " +
+			"mode")
+	}
+
+	reactionWindow := p.ReactionWindow
+	if reactionWindow == 0 {
+		reactionWindow = DefaultChannelReactionWindow
+	}
+	if reactionWindow > wire.SequenceLockTimeMask ||
+		p.ChannelDelay > math.MaxUint32-reactionWindow {
+		return fmt.Errorf("channel VTXO reaction window overflows")
+	}
+	minimumFunderDelay := p.ChannelDelay + reactionWindow
+	if p.FunderDelay < minimumFunderDelay {
+		return fmt.Errorf("funder delay %d must be at least %d to "+
+			"preserve the reaction window", p.FunderDelay,
+			minimumFunderDelay)
+	}
+
+	return nil
+}
+
+// channelKeysEqual compares public keys by their x-only tapscript encoding.
+func channelKeysEqual(a, b *btcec.PublicKey) bool {
+	return bytes.Equal(
+		schnorr.SerializePubKey(a), schnorr.SerializePubKey(b),
+	)
+}
--- a/lib/arkscript/channel_vtxo.go
+++ b/lib/arkscript/channel_vtxo.go
@@ -159,6 +159,110 @@ func EncodeChannelVTXOArtifacts(params ChannelVTXOParams) ([]byte, []byte,
 	return template, pkScript, nil
 }
 
+// ValidateChannelVTXOPolicy recognizes the canonical three-leaf channel VTXO
+// policy using the admitting operator's key and minimum exit delay. It does
+// not assign client and hub labels to the two symmetric endpoint key pairs.
+func ValidateChannelVTXOPolicy(template *PolicyTemplate,
+	operatorKey *btcec.PublicKey, minExitDelay uint32) error {
+
+	if template == nil {
+		return fmt.Errorf("channel VTXO template is required")
+	}
+	if operatorKey == nil {
+		return fmt.Errorf("channel VTXO operator key is required")
+	}
+	if minExitDelay == 0 {
+		return fmt.Errorf("channel VTXO minimum exit delay is required")
+	}
+	if len(template.Leaves) != 3 {
+		return fmt.Errorf("channel VTXO policy must contain exactly " +
+			"3 leaves")
+	}
+
+	var (
+		cooperativeKeys []*btcec.PublicKey
+		channelKeys     []*btcec.PublicKey
+		funderKey       *btcec.PublicKey
+		channelDelay    uint32
+		funderDelay     uint32
+	)
+	for i := range template.Leaves {
+		switch node := template.Leaves[i].Node.(type) {
+		case *Multisig:
+			if len(node.Keys) != 3 || cooperativeKeys != nil {
+				return fmt.Errorf("channel VTXO leaf %d is "+
+					"not the unique 3-of-3 "+
+					"cooperative path", i)
+			}
+
+			cooperativeKeys = node.Keys
+
+		case *CSV:
+			if err := validateCSVLock(node.Lock); err != nil {
+				return fmt.Errorf("channel VTXO leaf %d has "+
+					"invalid CSV: %w", i, err)
+			}
+
+			multisig, ok := node.Inner.(*Multisig)
+			if !ok {
+				return fmt.Errorf("channel VTXO leaf %d is "+
+					"not CSV-gated multisig", i)
+			}
+
+			delay := node.Lock & uint32(wire.SequenceLockTimeMask)
+			switch len(multisig.Keys) {
+			case 2:
+				if channelKeys != nil {
+					return fmt.Errorf("channel VTXO has " +
+						"multiple 2-of-2 " +
+						"materialization paths")
+				}
+
+				channelKeys = multisig.Keys
+				channelDelay = delay
+
+			case 1:
+				if funderKey != nil {
+					return fmt.Errorf("channel VTXO has " +
+						"multiple funder refund paths")
+				}
+
+				funderKey = multisig.Keys[0]
+				funderDelay = delay
+
+			default:
+				return fmt.Errorf("channel VTXO leaf %d has "+
+					"%d signing keys", i, len(
+					multisig.Keys,
+				))
+			}
+
+		default:
+			return fmt.Errorf("channel VTXO leaf %d has an "+
+				"unsupported shape", i)
+		}
+	}
+
+	if cooperativeKeys == nil || channelKeys == nil || funderKey == nil {
+		return fmt.Errorf("channel VTXO policy is missing a required " +
+			"path")
+	}
+
+	params := ChannelVTXOParams{
+		ClientArkKey:     cooperativeKeys[0],
+		HubArkKey:        cooperativeKeys[1],
+		ArkOperatorKey:   operatorKey,
+		ClientChannelKey: channelKeys[0],
+		HubChannelKey:    channelKeys[1],
+		FunderKey:        funderKey,
+		ChannelDelay:     channelDelay,
+		FunderDelay:      funderDelay,
+		MinExitDelay:     minExitDelay,
+	}
+
+	return ValidateChannelVTXOTemplate(template, params)
+}
+
 // ValidateChannelVTXOTemplate verifies that an untrusted template has the
 // exact key roles and delays expected by a channel intent.
 func ValidateChannelVTXOTemplate(template *PolicyTemplate,
--- a/lib/arkscript/channel_vtxo.go
+++ b/lib/arkscript/channel_vtxo.go
@@ -199,6 +199,20 @@ func ValidateChannelVTXOTemplate(template *PolicyTemplate,
 		}
 	}
 
+	actualPkScript, err := template.PkScript()
+	if err != nil {
+		return fmt.Errorf("compile channel VTXO output: %w", err)
+	}
+	expectedPkScript, err := expected.Template.PkScript()
+	if err != nil {
+		return fmt.Errorf("compile expected channel VTXO output: %w",
+			err)
+	}
+	if !bytes.Equal(actualPkScript, expectedPkScript) {
+		return fmt.Errorf("channel VTXO output does not match " +
+			"expected key roles and delays")
+	}
+
 	return nil
 }
 
--- a/oor/actor_durable_message.go
+++ b/oor/actor_durable_message.go
@@ -303,6 +303,10 @@ func decodeStartTransferPayloadWithLimits(raw []byte,
 	if _, err := stream.DecodeWithParsedTypes(reader); err != nil {
 		return startTransferPayload{}, err
 	}
+	if prepareOnly > 1 {
+		return startTransferPayload{}, fmt.Errorf("prepare-only flag " +
+			"must be 0 or 1")
+	}
 
 	inputs, err := decodeTransferInputSnapshotsWithLimits(
 		inputsRaw, limits,
--- a/oor/actor_durable_message.go
+++ b/oor/actor_durable_message.go
@@ -41,6 +41,10 @@ const (
 	// startPayloadDispatchRequestDataType stores the normalized caller
 	// recipient proof for a keyed dispatch.
 	startPayloadDispatchRequestDataType tlv.Type = 9
+
+	// startPayloadPrepareOnlyType stores whether admission must stop before
+	// signing. It is optional so pre-feature messages decode as false.
+	startPayloadPrepareOnlyType tlv.Type = 11
 )
 
 const (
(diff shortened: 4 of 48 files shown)


litbot-9000:
@bhandras: review reminder
@darioanongba: review reminder

vtxo/manager.go
```diff
@@ -1373,16 +1373,24 @@ func (m *Manager) confirmExitedVTXO(ctx context.Context,
...
 		return fn.Err[ManagerResp](
 			fmt.Errorf("load vtxo for confirm: %w", err),
 		)
 	}
-	if descriptor == nil || descriptor.Status != VTXOStatusUnilateralExit {
+	if descriptor == nil {
+		return fn.Ok[ManagerResp](&ExitOutcomeResp{})
+	}
+	isOrdinaryExit := descriptor.Status == VTXOStatusUnilateralExit
+	isRecoveryOnly := descriptor.Status == VTXOStatusRecoveryOnly &&
```
darioAnongba, line 1431: `Valid()` only knows the two vHTLC kinds and `unroll/registry.go` casts the unroll kind straight through, so `ark_channel_backing` never passes this gate. After a materialized force close the row stays recovery-only forever and `RestoreWatches` re-arms the watch every boot. No fund loss, but this branch never runs in prod and nothing tests it. Extend `Valid()` (or gate on "not the standard timeout kind") and add a test with a non-vHTLC kind.

db/vtxo_store.go
```diff
@@ -119,6 +120,125 @@ func (s *VTXOPersistenceStore) SaveVTXO(
...
+	status := vtxo.VTXOStatus(row.Status)
+	if status != vtxo.VTXOStatusRecoveryOnly &&
+		status != vtxo.VTXOStatusSpent {
+		return fmt.Errorf("recovery-only VTXO already has status %s",
+			status)
+	}
+	if row.RoundID != desc.RoundID || row.Amount != int64(desc.Amount) ||
+		!bytes.Equal(row.PkScript, desc.PkScript) ||
+		!bytes.Equal(row.PolicyTemplate, desc.PolicyTemplate) ||
+		row.Expiry != int32(desc.RelativeExpiry) ||
```
darioAnongba: +1. `SetRecoveryOnlyVTXORelativeExpiry` exists to mutate exactly this field. The tip works around it at one callsite (`ark_channel_source.go` copies `existing.RelativeExpiry` before the replay save) but not in the export path, so export replay after a close fails with "descriptor mismatch". Drop expiry from the immutable set here instead of patching callers.

oor/transitions.go
```diff
@@ -66,12 +66,16 @@ func unexpectedEvent(state State) *StateTransition {
 func failedState(reason string, current State) *Failed {
 	return &Failed{
 		Reason:         reason,
+		PrePONR:        len(prePONRInputOutpoints(current)) > 0,
```
darioAnongba: `PrePONR` comes from `len(prePONRInputOutpoints)`, which skips inputs with a nil `VTXO`. A pre-PONR failure with nil pointers would read as post-PONR, and `oorbridge/preparer.go` treats that as "don't re-prepare", wedging the intent. Unlikely today, but this is safety evidence; derive it from the state type.

lib/arkscript/channel_vtxo.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return template, pkScript, nil
+}
+
+// ValidateChannelVTXOTemplate verifies that an untrusted template has the
+// exact key roles and delays expected by a channel intent.
+func ValidateChannelVTXOTemplate(template *PolicyTemplate,
```
darioAnongba, line 268: +1 on comparing pkScripts. Also: nothing in the stack calls this except its own tests (the tip validates via terms + `VerifyBindsToPkScript` in `backing.go`, which is the right way). Fix it or delete it so nobody picks up the weak version later.

unroll/actor.go
```diff
@@ -1774,6 +1792,118 @@ func (b *behavior) handleSpendObserved(ctx context.Context,
 	return b.handleEvent(ctx, ax, &FailEvent{Reason: reason})
 }
 
+// ackPreSignedExitSpend completes an unroll when another authorized party
+// publishes the exact immutable final spend supplied by the exit policy.
+func (b *behavior) ackPreSignedExitSpend(ctx context.Context,
```
darioAnongba, line 1915: nit: the parent gets recorded as confirmed at `msg.SpendingHeight`, but it confirmed at least `ChannelDelay` earlier. Harmless for the FSM, skews anything that reads per-tx heights later. Real height, or an "unknown" sentinel?

oor/outgoing_snapshot.go
```diff
@@ -120,15 +177,30 @@ func NewOutgoingSnapshot(sessionID SessionID,
 	}
 
 	snap := &OutgoingSnapshot{
-		// Version 5 adds the FirstRejectUnixNanos record (bounded
-		// transient submit-reject retry window). Restore stays
-		// backward-compatible: a pre-v5 snapshot lacks the record and
-		// decodes FirstRejectUnixNanos to 0 (a fresh window).
-		Version:   5,
+		// Version 7 adds the pre-PONR terminal-failure marker. Older
```
darioAnongba, line 182: nit: `oor/AGENTS.md` still says 5, this is 7 (6 skipped?), and decode only rejects 0. Since 27/29 are odd TLVs, an older binary restores a v7 `Failed` with `PrePONR=false`, which the preparer reads as accepted. Worth deciding whether downgrade tolerance is ok here.

unroll/policy_resolvers.go
```diff
@@ -0,0 +1,48 @@
...
+
+// ResolveExitSpendPolicy delegates to the one resolver advertising the kind.
+func (r PolicyResolvers) ResolveExitSpendPolicy(ctx context.Context,
+	req ExitSpendPolicyRequest) (ExitSpendPolicy, error) {
+
+	for _, resolver := range r {
+		if resolver == nil {
+			continue
+		}
+		support, ok := resolver.(ResolverKindSupport)
```
darioAnongba, line 43: +1, a resolver without `ResolverKindSupport` is silently unreachable; `testExitPolicyResolver` has exactly that shape.

oor/actor_durable_message.go
```diff
@@ -307,6 +325,7 @@ func decodeStartTransferPayloadWithLimits(raw []byte,
 		IdempotencyKey:             string(idKey),
 		DispatchRequestData:        dispatchRequestData,
 		AdmissionDeadlineUnixNanos: int64(deadline),
+		PrepareOnly:                prepareOnly != 0,
```
darioAnongba, line 333: +1, any non-zero byte passes here while the snapshot codec rejects >1. Same check in both.

darioAnongba (commented):
Prepare/commit boundary is clean and abort-safe, and recovery-only never leaks into balance, selection, refresh or admission (checked all of them). Tests pass at this head. Two things I'd fix in this layer before the rest builds on it: the keyed `Prepared` lookup doesn't survive a restart, and `ExitPolicyKind.Valid()` never learned the channel kind, so the retirement path is dead at the tip. Inline.

Not anchorable (code this PR doesn't touch), `oor/registry.go` `validateOutgoingDispatchIdentity` / `handleLookupTransfer`, blocking: a keyed `Prepared` session has no dispatch-attempt row, `pendingOutgoingKeys` is same-boot only, and identity validation never compares `record.IdempotencyKey`. After a restart the lookup says "absent", we re-prepare, the wallet can't re-pick the still-reserved inputs, and the new commit collides with the unique key index. Open loops, inputs stranded until a manual abort. Compare the key in `validateOutgoingDispatchIdentity` and return the existing session, and fall back to a registry-row-by-key lookup in `handleLookupTransfer`. (Bots flagged this too; the restart consequence is the part I'd stress.)

Two design notes. `Prepared` has no TTL at the OOR layer, so if the coordinator record never gets created the reservations leak until someone aborts by session ID; `AdmissionDeadlineUnixNanos` is already plumbed, an auto-abort past it would make this self-healing. And `ExitPolicyKind` being two string types joined by a cast is what let the `Valid()` gap happen silently; a shared constant set with an exhaustiveness test would turn it into a test failure.

Long version: https://ark-channels.lightning.wiki/

oor/actor_durable_message.go
```diff
@@ -307,6 +325,7 @@ func decodeStartTransferPayloadWithLimits(raw []byte,
...
 	}, nil
```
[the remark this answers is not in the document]
sputn1ck: Addressed on the updated branch: the durable decoder now rejects values greater than 1 before setting `PrepareOnly`, so corrupted non-boolean values cannot enter the FSM. The OOR package tests pass on `beb51374`.

oor/actor_durable_message.go
sputn1ck: Addressed on the updated branch: the durable decoder now rejects values greater than 1 before setting `PrepareOnly`, matching the snapshot boolean validation. The OOR package tests pass on `beb51374`.

oor/transitions.go
```diff
@@ -179,10 +183,21 @@ func (s *Idle) ProcessEvent(ctx context.Context, event Event,
 			}
 		}
 
-		signReq := &RequestArkSignatures{
-			ArkPSBT:         ark,
-			CheckpointPSBTs: checkpoints,
-			TransferInputs:  evt.VTXOInputs,
+		if evt.PrepareOnly {
+			return &StateTransition{
+				NextState: &Prepared{
```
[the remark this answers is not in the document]
sputn1ck: Addressed in the cleaned history by `03fc3ee`: keyed prepared sessions are resolved from the durable session row before a new admission is attempted, and `TestOORRegistryPreparedTransferDedupAfterRestart` covers lookup and retry after restart.

oor/registry.go
```diff
@@ -646,7 +647,7 @@ func (r *oorRegistryBehavior) handleStartTransfer(ctx context.Context,
 	// daemon's lifetime (one leak per outgoing admission otherwise).
 	session, _, err := newSessionWithDispatchRequest(
 		ctx, req.Policy, req.Inputs, req.Recipients, req.IdempotencyKey,
-		req.DispatchRequestData, r.envConfig(),
+		req.DispatchRequestData, req.PrepareOnly, r.envConfig(),
```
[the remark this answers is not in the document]
sputn1ck: Addressed in the cleaned history by `03fc3ee`: the registry now consults the durable unfailed session keyed by idempotency key, rather than relying on the same-boot map. `TestOORRegistryPreparedTransferDedupAfterRestart` exercises the restart case.

db/vtxo_store.go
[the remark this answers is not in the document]
sputn1ck: Addressed in `d823571`: relative expiry is excluded from the immutable recovery-only descriptor comparison because the dedicated setter owns that field. The test now sets expiry between two `SaveRecoveryOnlyVTXO` calls and verifies replay preserves it.

sputn1ck: Addressed in `d823571` at the store boundary rather than at individual callers: relative expiry is mutable and no longer participates in immutable replay validation. The setter-then-save replay is covered directly.

db/vtxo_store_test.go
```diff
@@ -2011,3 +2011,62 @@ func TestVTXOStoreExpiredExcludedFromLiveSet(t *testing.T) {
...
+
+	live, err := vtxoStore.ListLiveVTXOs(ctx)
+	require.NoError(t, err)
+	require.Empty(t, live)
+
+	recoverable, err := vtxoStore.ListRecoverableVTXOs(ctx)
+	require.NoError(t, err)
+	require.Empty(t, recoverable)
+	require.ErrorContains(
+		t, vtxoStore.SetRecoveryOnlyVTXORelativeExpiry(
```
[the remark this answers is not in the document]
sputn1ck: Addressed in `d823571`: the test now covers a valid recovery-only expiry update, replay preserving the selected value, and rejection against an ordinary live VTXO without mutating it.

lib/arkscript/channel_vtxo.go
```diff
@@ -0,0 +1,291 @@
...
+		script, err := template.Leaves[i].Script()
+		if err != nil {
+			return fmt.Errorf("compile channel VTXO leaf %d: %w", i,
+				err)
+		}
+
+		actualScripts[string(script)] = struct{}{}
+	}
+
+	for i := range expected.Template.Leaves {
```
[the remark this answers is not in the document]
sputn1ck: Addressed in `7d6a403`: validation now derives and compares the complete output `PkScript` after the leaf checks, binding ordering and every template field that contributes to the Taproot output.

lib/arkscript/channel_vtxo.go
sputn1ck: Addressed in `7d6a403`: the validator now compares the derived output scripts directly, while retaining leaf checks for useful diagnostics.

unroll/policy_resolvers.go
```diff
@@ -0,0 +1,48 @@
...
+		if !ok || !support.SupportsKind(req.Kind) {
```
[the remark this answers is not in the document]
sputn1ck: Addressed in `798a51c`: advertised resolvers are routed by kind, while a resolver without `ResolverKindSupport` is now treated as a legacy fallback and is called instead of being silently skipped.

unroll/policy_resolvers.go
sputn1ck: Addressed in `798a51c`: non-advertising resolvers retain legacy fallback behavior and are no longer unreachable.

vtxo/manager.go
sputn1ck: Addressed in the rewritten `multi: Track channel recovery sources` commit: `ExitPolicyArkChannelBacking` is a valid non-standard completion policy, and `TestHandleExitOutcomeConfirmedNoActorPersistsChannelRecoverySpent` now proves a recovery-only source is retired to spent without a live actor.

oor/transitions.go
sputn1ck: Addressed in `6cbdd50`: pre-PONR is now derived from the protocol state type rather than nullable input pointers, with state coverage for the failure boundary.

unroll/actor.go
sputn1ck: Addressed in `d9b9183`: observing the CSV-delayed pre-signed spend now records the parent at `spending_height - csv_delay`, with underflow rejection, rather than assigning the child confirmation height.

oor/outgoing_snapshot.go
sputn1ck: Addressed in `b26c108`: the package docs now state version 7 and the pre-PONR marker semantics. Restore rejects future versions, so an older binary does not interpret a v7 snapshot while silently dropping the odd TLV.

litbot-9000:
The Go changes on this branch left the per-package docs for `db`, `lib/actormsg`, `lib/arkscript`, `oor`, `unroll`, and `vtxo` stale — recovery-only VTXOs, the Ark channel VTXO policy, the pre-signed exit-spend path, and the prepared-OOR boundary are all undocumented (`db/sqlc` was in scope but needed no change, and `oor/CLAUDE.md`'s snapshot-version bullet is amended rather than added).

<details>
<summary>Proposed <code>CLAUDE.md</code> / <code>AGENTS.md</code> updates</summary>

```diff
diff --git a/db/AGENTS.md b/db/AGENTS.md
index aee6255c..32606a75 100644
--- a/db/AGENTS.md
+++ b/db/AGENTS.md
@@ -38,6 +38,18 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/db.<Symb
   paginated listing (avoids deserializing full trees).
 - `VTXOPersistenceStore` — VTXO descriptor store
   (`InsertClientVTXO`, `FetchByOutpoint`). Persists `ChainDepth`.
+- `VTXOPersistenceStore.SaveRecoveryOnlyVTXO` — insert-and-transition in one
+  transaction for an application-owned descriptor (`vtxo.VTXOStatusRecoveryOnly`).
+  Unlike `SaveVTXO` it never leaves the row observable as live between the
+  insert and the status write, so concurrent wallet coin selection cannot
+  mistake a channel or contract output for spendable liquidity. A replay is
+  accepted only when `validateRecoveryOnlyVTXORow` finds the existing row still
+  `RecoveryOnly` or `Spent` **and** byte-identical on every immutable column;
+  relative expiry is excluded because channel recovery selects it afterwards.
+- `VTXOPersistenceStore.SetRecoveryOnlyVTXORelativeExpiry` — sets the exact
+  final-spend delay chosen by the Ark channel policy. The SQL is scoped to
+  `status = 9`, so an ordinary wallet VTXO can never be modified through it,
+  and a row count other than 1 is an error rather than a silent no-op.
 - `OORArtifactStore`, `OwnedReceiveScriptStore` — OOR session state
   and locally owned receive-script metadata.
 - `OwnedReceiveScriptRecord` — one owned receive-script row. Its
@@ -57,8 +69,14 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/db.<Symb
   different label fails with `ErrOwnedReceiveScriptReplayMismatch`.
 - `OORSessionRegistryStoreDB` / `OORSessionRegistryRecord` — control-plane
   registry of per-session durable OOR actors
-  (`UpsertSession`/`GetSession`/`ListSessions`/`ListNonTerminal`). One mutable
+  (`UpsertSession`/`GetSession`/`GetUnfailedSessionByIdempotencyKey`/
+  `ListSessions`/`ListNonTerminal`). One mutable
   row per session id is shared by both lifecycle directions.
+  `GetUnfailedSessionByIdempotencyKey` is the lookup for an outgoing identity
+  that is durable *before* its immutable `oor_dispatch_attempts` row exists
+  (an OOR session prepared but not yet committed). It filters out `status = 2`
+  (Failed), because a failed row deliberately releases its key, and returns
+  `ErrOORSessionNotFound` for an empty key.
 - `OORDispatchAttemptRecord` — immutable keyed outgoing identity, loaded by
   idempotency key or session id. It stores the canonical recipient record
   before the first transport enqueue and remains authoritative after the
@@ -112,6 +130,15 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/db.<Symb
   safety bounds enforced during `DeserializeTree`.
 - `resolveInputPackage` / `loadPackageBundleBySessionID` — two-stage
   OOR ancestry resolver (`oor_unroll_resolver.go`).
+- `OORArtifactPersistenceStore.ResolveUnrollPackages` /
+  `ResolveUnrollPackagesBySessionID` — the two entry points into that resolver,
+  sharing one `resolveUnrollPackages` traversal parameterized by how the target
+  package is loaded. The by-outpoint form is the ordinary path; the
+  by-session-id form resolves a chain *before* the local endpoint has installed
+  an outpoint binding, which is what an OOR sender needs while promoting a
+  recipient output into shared channel recovery state. The session variant fails
+  closed twice: the session id must equal `outpoint.Hash`, and the loaded
+  package must actually create the target outpoint.
 - `LatestMigrationVersion = 18` — current schema version.
 - `PendingIntentPersistenceStore` — implements `wallet.PendingIntentStore`,
   the persistence half of the generic restart-safe intent outbox (header
@@ -285,7 +312,11 @@ when adding one.
   `round_client_trees`, `client_tree_txids`.
 - `000004_vtxos` — `vtxos` plus the normalized `vtxo_ancestry_paths`
   side table. Routine queries skip the ancestry join; the unroller
-  loads ancestry only when resolving an exit.
+  loads ancestry only when resolving an exit. The `status` column's inline
+  comment enumerates the `vtxo.VTXOStatus` iota and is append-only alongside
+  it (`8 = Expired`, `9 = RecoveryOnly`); queries such as
+  `SetRecoveryOnlyVTXORelativeExpiry` hard-code those numbers, so a renumber
+  would silently retarget them.
 - `000005_oor` — OOR artifact store (packages, checkpoints, VTXO
   bindings), receiver-side polling state, and the
   `oor_session_registry` that owns per-session durable actors.
diff --git a/db/CLAUDE.md b/db/CLAUDE.md
index aee6255c..32606a75 100644
--- a/db/CLAUDE.md
+++ b/db/CLAUDE.md
@@ -38,6 +38,18 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/db.<Symb
   paginated listing (avoids deserializing full trees).
 - `VTXOPersistenceStore` — VTXO descriptor store
   (`InsertClientVTXO`, `FetchByOutpoint`). Persists `ChainDepth`.
+- `VTXOPersistenceStore.SaveRecoveryOnlyVTXO` — insert-and-transition in one
+  transaction for an application-owned descriptor (`vtxo.VTXOStatusRecoveryOnly`).
+  Unlike `SaveVTXO` it never leaves the row observable as live between the
+  insert and the status write, so concurrent wallet coin selection cannot
+  mistake a channel or contract output for spendable liquidity. A replay is
+  accepted only when `validateRecoveryOnlyVTXORow` finds the existing row still
+  `RecoveryOnly` or `Spent` **and** byte-identical on every immutable column;
+  relative expiry is excluded because channel recovery selects it afterwards.
+- `VTXOPersistenceStore.SetRecoveryOnlyVTXORelativeExpiry` — sets the exact
+  final-spend delay chosen by the Ark channel policy. The SQL is scoped to
+  `status = 9`, so an ordinary wallet VTXO can never be modified through it,
+  and a row count other than 1 is an error rather than a silent no-op.
 - `OORArtifactStore`, `OwnedReceiveScriptStore` — OOR session state
   and locally owned receive-script metadata.
 - `OwnedReceiveScriptRecord` — one owned receive-script row. Its
@@ -57,8 +69,14 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/db.<Symb
   different label fails with `ErrOwnedReceiveScriptReplayMismatch`.
 - `OORSessionRegistryStoreDB` / `OORSessionRegistryRecord` — control-plane
   registry of per-session durable OOR actors
-  (`UpsertSession`/`GetSession`/`ListSessions`/`ListNonTerminal`). One mutable
+  (`UpsertSession`/`GetSession`/`GetUnfailedSessionByIdempotencyKey`/
+  `ListSessions`/`ListNonTerminal`). One mutable
   row per session id is shared by both lifecycle directions.
+  `GetUnfailedSessionByIdempotencyKey` is the lookup for an outgoing identity
+  that is durable *before* its immutable `oor_dispatch_attempts` row exists
+  (an OOR session prepared but not yet committed). It filters out `status = 2`
+  (Failed), because a failed row deliberately releases its key, and returns
+  `ErrOORSessionNotFound` for an empty key.
 - `OORDispatchAttemptRecord` — immutable keyed outgoing identity, loaded by
   idempotency key or session id. It stores the canonical recipient record
   before the first transport enqueue and remains authoritative after the
@@ -112,6 +130,15 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/db.<Symb
   safety bounds enforced during `DeserializeTree`.
 - `resolveInputPackage` / `loadPackageBundleBySessionID` — two-stage
   OOR ancestry resolver (`oor_unroll_resolver.go`).
+- `OORArtifactPersistenceStore.ResolveUnrollPackages` /
+  `ResolveUnrollPackagesBySessionID` — the two entry points into that resolver,
+  sharing one `resolveUnrollPackages` traversal parameterized by how the target
+  package is loaded. The by-outpoint form is the ordinary path; the
+  by-session-id form resolves a chain *before* the local endpoint has installed
+  an outpoint binding, which is what an OOR sender needs while promoting a
+  recipient output into shared channel recovery state. The session variant fails
+  closed twice: the session id must equal `outpoint.Hash`, and the loaded
+  package must actually create the target outpoint.
 - `LatestMigrationVersion = 18` — current schema version.
 - `PendingIntentPersistenceStore` — implements `wallet.PendingIntentStore`,
   the persistence half of the generic restart-safe intent outbox (header
@@ -285,7 +312,11 @@ when adding one.
   `round_client_trees`, `client_tree_txids`.
 - `000004_vtxos` — `vtxos` plus the normalized `vtxo_ancestry_paths`
   side table. Routine queries skip the ancestry join; the unroller
-  loads ancestry only when resolving an exit.
+  loads ancestry only when resolving an exit. The `status` column's inline
+  comment enumerates the `vtxo.VTXOStatus` iota and is append-only alongside
+  it (`8 = Expired`, `9 = RecoveryOnly`); queries such as
+  `SetRecoveryOnlyVTXORelativeExpiry` hard-code those numbers, so a renumber
+  would silently retarget them.
 - `000005_oor` — OOR artifact store (packages, checkpoints, VTXO
   bindings), receiver-side polling state, and the
   `oor_session_registry` that owns per-session durable actors.
diff --git a/lib/actormsg/AGENTS.md b/lib/actormsg/AGENTS.md
index 7f163c7e..f95f41a1 100644
--- a/lib/actormsg/AGENTS.md
+++ b/lib/actormsg/AGENTS.md
@@ -18,7 +18,12 @@ package boundaries. Lives in `lib/` to break import cycles between `vtxo`,
 - `ReleaseSpendRequest` / `CompleteSpendRequest` — Spend lifecycle completion messages.
 - `ForceUnrollRequest` / `ForceUnrollResponse` — Ask-message that routes an operator or chain-resolver unroll trigger through the VTXO manager into the per-VTXO FSM. `ForceUnrollRequest.Trigger` (a `UnrollTrigger`) names *why* the coin is exiting, and `ForceUnrollRequest.ExitPolicy` (an `fn.Option[ExitPolicy]`) names *which* exit-spend policy the target unrolls under, so a single admission path carries manual, critical-expiry, fraud, and vHTLC-recovery intent through to the unroll registry. `ForceUnrollResponse.Accepted` is true when the request caused a state transition; when false, `Reason` distinguishes `"no such vtxo"` from `"already terminal"` so callers don't misread a silent self-loop as success.
 - `UnrollTrigger` — String-typed enum naming why a unilateral exit was started (`UnrollTriggerCriticalExpiry` is the empty-string zero value and preserves the historical critical-expiry admission, `UnrollTriggerManual`, `UnrollTriggerFraudSpend`). It mirrors the unroll package's `StartTrigger` so `vtxo` and `actormsg` can thread the trigger through `ForceUnroll` without importing `unroll` (which would form a cycle); the waved chain resolver bridge converts it back at the seam.
-- `ExitPolicyKind` / `ExitPolicyRef` / `ExitPolicy` — Durable exit-spend policy identity for a forced exit. `ExitPolicyKind` is a string-typed enum of the non-standard policies that ride `ForceUnroll` (`ExitPolicyVHTLCClaim`, `ExitPolicyVHTLCRefundWithoutReceiver`), with `Valid()` true only for those two vHTLC kinds; `ExitPolicyRef` is the policy-specific durable reference (e.g. a vHTLC recovery job id), kept a distinct type so `Kind` and `Ref` can't be transposed; `ExitPolicy` bundles the pair as one identity validated at the registry admission boundary. A `None` `ExitPolicy` selects the standard VTXO timeout policy.
+- `ExitPolicyKind` / `ExitPolicyRef` / `ExitPolicy` — Durable exit-spend policy identity for a forced exit. `ExitPolicyKind` is a string-typed enum of the non-standard policies that ride `ForceUnroll` (`ExitPolicyVHTLCClaim`, `ExitPolicyVHTLCRefundWithoutReceiver`, `ExitPolicyArkChannelBacking`), with `Valid()` true only for those kinds; `ExitPolicyRef` is the policy-specific durable reference (e.g. a vHTLC recovery job id), kept a distinct type so `Kind` and `Ref` can't be transposed; `ExitPolicy` bundles the pair as one identity validated at the registry admission boundary. A `None` `ExitPolicy` selects the standard VTXO timeout policy.
+- `ExitPolicyArkChannelBacking` — the pre-signed channel materialization spend
+  of a recovery-only channel-policy VTXO. Unlike the vHTLC kinds, the final
+  transaction is fixed before the unroll starts and either channel endpoint may
+  publish it, so `unroll` treats the exact pre-signed txid appearing on chain as
+  success rather than as an external conflict.
 - `RegisterIntentMsg` — Carries pre-composed cooperative intent package to
   round actor. The `TriggerRegistration bool` field controls whether the
   round FSM immediately fires `IntentRequested` after accepting the intent
diff --git a/lib/actormsg/CLAUDE.md b/lib/actormsg/CLAUDE.md
index 7f163c7e..f95f41a1 100644
--- a/lib/actormsg/CLAUDE.md
+++ b/lib/actormsg/CLAUDE.md
@@ -18,7 +18,12 @@ package boundaries. Lives in `lib/` to break import cycles between `vtxo`,
 - `ReleaseSpendRequest` / `CompleteSpendRequest` — Spend lifecycle completion messages.
 - `ForceUnrollRequest` / `ForceUnrollResponse` — Ask-message that routes an operator or chain-resolver unroll trigger through the VTXO manager into the per-VTXO FSM. `ForceUnrollRequest.Trigger` (a `UnrollTrigger`) names *why* the coin is exiting, and `ForceUnrollRequest.ExitPolicy` (an `fn.Option[ExitPolicy]`) names *which* exit-spend policy the target unrolls under, so a single admission path carries manual, critical-expiry, fraud, and vHTLC-recovery intent through to the unroll registry. `ForceUnrollResponse.Accepted` is true when the request caused a state transition; when false, `Reason` distinguishes `"no such vtxo"` from `"already terminal"` so callers don't misread a silent self-loop as success.
 - `UnrollTrigger` — String-typed enum naming why a unilateral exit was started (`UnrollTriggerCriticalExpiry` is the empty-string zero value and preserves the historical critical-expiry admission, `UnrollTriggerManual`, `UnrollTriggerFraudSpend`). It mirrors the unroll package's `StartTrigger` so `vtxo` and `actormsg` can thread the trigger through `ForceUnroll` without importing `unroll` (which would form a cycle); the waved chain resolver bridge converts it back at the seam.
-- `ExitPolicyKind` / `ExitPolicyRef` / `ExitPolicy` — Durable exit-spend policy identity for a forced exit. `ExitPolicyKind` is a string-typed enum of the non-standard policies that ride `ForceUnroll` (`ExitPolicyVHTLCClaim`, `ExitPolicyVHTLCRefundWithoutReceiver`), with `Valid()` true only for those two vHTLC kinds; `ExitPolicyRef` is the policy-specific durable reference (e.g. a vHTLC recovery job id), kept a distinct type so `Kind` and `Ref` can't be transposed; `ExitPolicy` bundles the pair as one identity validated at the registry admission boundary. A `None` `ExitPolicy` selects the standard VTXO timeout policy.
+- `ExitPolicyKind` / `ExitPolicyRef` / `ExitPolicy` — Durable exit-spend policy identity for a forced exit. `ExitPolicyKind` is a string-typed enum of the non-standard policies that ride `ForceUnroll` (`ExitPolicyVHTLCClaim`, `ExitPolicyVHTLCRefundWithoutReceiver`, `ExitPolicyArkChannelBacking`), with `Valid()` true only for those kinds; `ExitPolicyRef` is the policy-specific durable reference (e.g. a vHTLC recovery job id), kept a distinct type so `Kind` and `Ref` can't be transposed; `ExitPolicy` bundles the pair as one identity validated at the registry admission boundary. A `None` `ExitPolicy` selects the standard VTXO timeout policy.
+- `ExitPolicyArkChannelBacking` — the pre-signed channel materialization spend
+  of a recovery-only channel-policy VTXO. Unlike the vHTLC kinds, the final
+  transaction is fixed before the unroll starts and either channel endpoint may
+  publish it, so `unroll` treats the exact pre-signed txid appearing on chain as
+  success rather than as an external conflict.
 - `RegisterIntentMsg` — Carries pre-composed cooperative intent package to
   round actor. The `TriggerRegistration bool` field controls whether the
   round FSM immediately fires `IntentRequested` after accepting the intent
diff --git a/lib/arkscript/AGENTS.md b/lib/arkscript/AGENTS.md
index 0f1db272..c9884b47 100644
--- a/lib/arkscript/AGENTS.md
+++ b/lib/arkscript/AGENTS.md
@@ -44,6 +44,28 @@ validated invariants.
   one sender/receiver/operator key set, payment hash, and refund locktime;
   any other custom policy shape is rejected so vHTLC assumptions cannot be
   applied to it.
+- `ChannelVTXOPolicy` / `ChannelVTXOParams` / `NewChannelVTXOPolicy` — 3-leaf
+  policy for a VTXO backing an unpublished Lightning channel point
+  (`channel_vtxo.go`). The leaves are a 3-of-3 cooperative closure
+  (client Ark + hub Ark + Ark operator, no CSV), a `ChannelDelay`-gated 2-of-2
+  materialization closure (client channel + hub channel), and a longer
+  `FunderDelay`-gated single-key funder refund. Path accessors:
+  `CooperativeSpendPath()`, `ChannelSpendPath()`, `FunderSpendPath()`.
+- `DefaultChannelReactionWindow` (432 blocks, ≈3 days) — the minimum gap the
+  funder refund leaf must sit above the materialization leaf when
+  `ChannelVTXOParams.ReactionWindow` is left zero.
+- `EncodeChannelVTXOArtifacts(params)` — channel-VTXO counterpart to
+  `EncodeStandardVTXOArtifacts`; returns the encoded policy template bytes and
+  the canonical P2TR pkScript.
+- `ValidateChannelVTXOPolicy(template, operatorKey, minExitDelay)` — admission
+  check that recognizes the canonical three-leaf shape in an untrusted
+  template, recovers its key roles and delays, and re-derives the policy to
+  confirm the match. It deliberately does **not** label which of the two
+  symmetric endpoint keys is client and which is hub.
+- `ValidateChannelVTXOTemplate(template, params)` — stricter binding check for
+  a caller that already knows the exact intended roles and delays: rebuilds
+  the expected policy and requires both a leaf-script set match and an exact
+  pkScript match.
 - `CheckpointPolicy` — Parameters for OOR checkpoint taproot tree construction. `CheckpointTapScript` / `CheckpointPkScript` derive the checkpoint output.
 - `SpendInfo` — Witness script + control block needed to spend a specific leaf. Methods: `BuildSignDescriptor`, `CollabWitness`, `TimeoutWitness`.
 - `SpendPath` — Serializable spend path (leaf index + encoded leaf data) with `Witness` and `AttachTapLeafScript` helpers for PSBT integration.
@@ -116,6 +138,16 @@ validated invariants.
   (collaborative) and at least one non-operator leaf (exit/unilateral).
 - Exit delay must be >= `MinExitDelay` when `ValidateStandardVTXOPolicy` is
   used; `ValidatePolicy` with `MinExitDelay = 0` skips this check.
+- Channel VTXO construction enforces its own timing and key separation before
+  compiling: the three cooperative keys must be three distinct parties, the two
+  materialization keys must differ, `ChannelDelay >= MinExitDelay` (a zero
+  `MinExitDelay` is rejected, so the caller cannot opt out of the floor), both
+  delays must fit the BIP-68 block mode, and `FunderDelay >= ChannelDelay +
+  ReactionWindow`. The reaction window is what keeps the refund leaf from
+  opening while either endpoint can still materialize the channel, so it is a
+  construction-time invariant rather than an admission-time judgement. Key
+  comparison is by x-only tapscript encoding, so a parity flip cannot smuggle
+  the same party into two roles.
 - **Policy construction does not validate timing.** `NewVHTLCPolicy` compiles
   whatever timing tuple it is given; suitability depends on the lifecycle
   boundary and the current chain height, which the compiler does not know.
diff --git a/lib/arkscript/CLAUDE.md b/lib/arkscript/CLAUDE.md
index 0f1db272..c9884b47 100644
--- a/lib/arkscript/CLAUDE.md
+++ b/lib/arkscript/CLAUDE.md
@@ -44,6 +44,28 @@ validated invariants.
   one sender/receiver/operator key set, payment hash, and refund locktime;
   any other custom policy shape is rejected so vHTLC assumptions cannot be
   applied to it.
+- `ChannelVTXOPolicy` / `ChannelVTXOParams` / `NewChannelVTXOPolicy` — 3-leaf
+  policy for a VTXO backing an unpublished Lightning channel point
+  (`channel_vtxo.go`). The leaves are a 3-of-3 cooperative closure
+  (client Ark + hub Ark + Ark operator, no CSV), a `ChannelDelay`-gated 2-of-2
+  materialization closure (client channel + hub channel), and a longer
+  `FunderDelay`-gated single-key funder refund. Path accessors:
+  `CooperativeSpendPath()`, `ChannelSpendPath()`, `FunderSpendPath()`.
+- `DefaultChannelReactionWindow` (432 blocks, ≈3 days) — the minimum gap the
+  funder refund leaf must sit above the materialization leaf when
+  `ChannelVTXOParams.ReactionWindow` is left zero.
+- `EncodeChannelVTXOArtifacts(params)` — channel-VTXO counterpart to
+  `EncodeStandardVTXOArtifacts`; returns the encoded policy template bytes and
+  the canonical P2TR pkScript.
+- `ValidateChannelVTXOPolicy(template, operatorKey, minExitDelay)` — admission
+  check that recognizes the canonical three-leaf shape in an untrusted
+  template, recovers its key roles and delays, and re-derives the policy to
+  confirm the match. It deliberately does **not** label which of the two
+  symmetric endpoint keys is client and which is hub.
+- `ValidateChannelVTXOTemplate(template, params)` — stricter binding check for
+  a caller that already knows the exact intended roles and delays: rebuilds
+  the expected policy and requires both a leaf-script set match and an exact
+  pkScript match.
 - `CheckpointPolicy` — Parameters for OOR checkpoint taproot tree construction. `CheckpointTapScript` / `CheckpointPkScript` derive the checkpoint output.
 - `SpendInfo` — Witness script + control block needed to spend a specific leaf. Methods: `BuildSignDescriptor`, `CollabWitness`, `TimeoutWitness`.
 - `SpendPath` — Serializable spend path (leaf index + encoded leaf data) with `Witness` and `AttachTapLeafScript` helpers for PSBT integration.
@@ -116,6 +138,16 @@ validated invariants.
   (collaborative) and at least one non-operator leaf (exit/unilateral).
 - Exit delay must be >= `MinExitDelay` when `ValidateStandardVTXOPolicy` is
   used; `ValidatePolicy` with `MinExitDelay = 0` skips this check.
+- Channel VTXO construction enforces its own timing and key separation before
+  compiling: the three cooperative keys must be three distinct parties, the two
+  materialization keys must differ, `ChannelDelay >= MinExitDelay` (a zero
+  `MinExitDelay` is rejected, so the caller cannot opt out of the floor), both
+  delays must fit the BIP-68 block mode, and `FunderDelay >= ChannelDelay +
+  ReactionWindow`. The reaction window is what keeps the refund leaf from
+  opening while either endpoint can still materialize the channel, so it is a
+  construction-time invariant rather than an admission-time judgement. Key
+  comparison is by x-only tapscript encoding, so a parity flip cannot smuggle
+  the same party into two roles.
 - **Policy construction does not validate timing.** `NewVHTLCPolicy` compiles
   whatever timing tuple it is given; suitability depends on the lifecycle
   boundary and the current chain height, which the compiler does not know.
diff --git a/oor/AGENTS.md b/oor/AGENTS.md
index e34dde69..607ec31f 100644
--- a/oor/AGENTS.md
+++ b/oor/AGENTS.md
@@ -24,6 +24,29 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/oor.<Sym
 - `Session` / `ReceiveSession` — outgoing/incoming FSM state containers;
   `OutgoingSnapshot` / `IncomingSnapshot` are their durable, versioned
   serializations.
+- `Prepared` — outgoing state holding a deterministic OOR package (unsigned Ark
+  PSBT, checkpoint PSBTs, retained transfer inputs, recipient outputs, caller
+  key, dispatch proof) with **no** signature or transport released. Entered
+  from `Idle` when `StartTransferEvent.PrepareOnly` is set, and left only by an
+  explicit `CommitPreparedEvent` (→ `AwaitingArkSignatures`, emitting
+  `RequestArkSignatures`) or `AbortPreparedEvent` / `FailEvent` (→ `Failed`,
+  emitting `ReleaseInputsRequest`). The motivating caller is a channel
+  coordinator that needs the output outpoint to be stable while it negotiates
+  the lnd channel, before committing the transfer.
+- `NewPreparedSessionWithIdempotencyKey` — prepared-mode sibling of
+  `NewSessionWithIdempotencyKey`; both funnel into one unexported constructor
+  so the two admission modes cannot drift.
+- `StartTransferRequest.PrepareOnly` / `StartTransferEvent.PrepareOnly` — the
+  durable flag that selects prepared mode. Carried as an optional TLV record
+  (`startPayloadPrepareOnlyType`) so pre-feature mailbox messages decode as
+  `false`; a value other than 0 or 1 is rejected.
+- `OutgoingSnapshotRecipients(raw)` — extracts canonical Ark recipients from a
+  durable outgoing snapshot, re-attaching each output's `VTXOPolicyTemplate`
+  from the snapshot's `RecipientOutputs`. Lets an idempotent replay recover
+  response metadata without rebuilding the transfer or selecting new inputs.
+- `Failed.PrePONR` / `OutgoingSnapshot.PrePONRFailure` — records that a terminal
+  failure happened before the operator co-signed the checkpoints, so the source
+  VTXO reservation was provably safe to release.
 - `OutboxHandler` / `LocalPersistenceOutboxHandler` — handles the local
   persistence outbox events (mark-inputs-spent, incoming metadata query,
   VTXO materialization, ack); everything else is handled inline by the
@@ -77,9 +100,37 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/oor.<Sym
   messages and return as fresh events.
 - Snapshots are versioned per direction (`OutgoingSnapshot.Version = 7`,
   `IncomingSnapshot.Version = 1`); restore rejects zero and unknown future
-  versions. Outgoing v5 adds `FirstRejectUnixNanos`; v7 adds the pre-PONR
-  terminal-failure marker. Older snapshots decode absent optional records to
-  their conservative zero values.
+  versions. Outgoing v5 adds `FirstRejectUnixNanos`; v7 adds both the
+  `RecipientOutputs` record (semantic policy metadata a prepared or pre-submit
+  transfer needs after restart) and the pre-PONR terminal-failure marker — v6
+  was skipped, so no snapshot ever carries it. The current value lives in one
+  place, `outgoingSnapshotVersion`, and both the codec and
+  `OutgoingStateFromSnapshot` reject anything above it. Older snapshots decode
+  absent optional records to their conservative zero values.
+- **Preparation has no implied side effect.** `OutboxForState(*Prepared)`
+  returns no outbox events, so a restart that restores a prepared session
+  re-enters `Prepared` and waits; only an explicit `CommitPreparedEvent` may
+  release signatures. Aborting is the safe direction: `Prepared` has not
+  exposed a signature and has not crossed the operator co-sign boundary, so
+  `abortPreparedTransition` releases **every** input.
+- Pre-PONR-ness is derived from the protocol phase (`isPrePONRState`:
+  `Prepared`, `AwaitingArkSignatures`, `AwaitingSubmitAccepted`), not from
+  whether a corrupt or partially restored state happens to have retained its
+  input pointers. `failedState` stamps it on every `Failed` transition.
+- Outgoing idempotency-key admission consults
+  `GetUnfailedSessionByIdempotencyKey` before the in-memory
+  `pendingOutgoingKeys` map in both `handleStartTransfer` and
+  `handleLookupTransfer`. A prepared session is durable before its immutable
+  `oor_dispatch_attempts` row exists, so the mutable-row lookup is what makes
+  an exact replay find it; failed rows release the key, pending and completed
+  rows retain it. A key that resolves to a non-outgoing session is
+  `ErrIdempotencyKeyConflict`.
+- `GetStateRequest` is answered from the durable snapshot when the child is
+  gone. Terminal actors are deliberately short-lived and reaped, but a status
+  reader must not race that reap, so `routeAsk` falls back to
+  `persistedState`, which validates the record's flow version and decodes the
+  direction-appropriate snapshot. Every other message to an unknown session
+  still errors.
 - `StartTransferRequest.IdempotencyKey` dedup reads
   `oor_dispatch_attempts`, not the mutable session row. The table has one row
   per caller key and one caller key per deterministic session id. Once the
diff --git a/oor/CLAUDE.md b/oor/CLAUDE.md
index e34dde69..607ec31f 100644
--- a/oor/CLAUDE.md
+++ b/oor/CLAUDE.md
@@ -24,6 +24,29 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/oor.<Sym
 - `Session` / `ReceiveSession` — outgoing/incoming FSM state containers;
   `OutgoingSnapshot` / `IncomingSnapshot` are their durable, versioned
   serializations.
+- `Prepared` — outgoing state holding a deterministic OOR package (unsigned Ark
+  PSBT, checkpoint PSBTs, retained transfer inputs, recipient outputs, caller
+  key, dispatch proof) with **no** signature or transport released. Entered
+  from `Idle` when `StartTransferEvent.PrepareOnly` is set, and left only by an
+  explicit `CommitPreparedEvent` (→ `AwaitingArkSignatures`, emitting
+  `RequestArkSignatures`) or `AbortPreparedEvent` / `FailEvent` (→ `Failed`,
+  emitting `ReleaseInputsRequest`). The motivating caller is a channel
+  coordinator that needs the output outpoint to be stable while it negotiates
+  the lnd channel, before committing the transfer.
+- `NewPreparedSessionWithIdempotencyKey` — prepared-mode sibling of
+  `NewSessionWithIdempotencyKey`; both funnel into one unexported constructor
+  so the two admission modes cannot drift.
+- `StartTransferRequest.PrepareOnly` / `StartTransferEvent.PrepareOnly` — the
+  durable flag that selects prepared mode. Carried as an optional TLV record
+  (`startPayloadPrepareOnlyType`) so pre-feature mailbox messages decode as
+  `false`; a value other than 0 or 1 is rejected.
+- `OutgoingSnapshotRecipients(raw)` — extracts canonical Ark recipients from a
+  durable outgoing snapshot, re-attaching each output's `VTXOPolicyTemplate`
+  from the snapshot's `RecipientOutputs`. Lets an idempotent replay recover
+  response metadata without rebuilding the transfer or selecting new inputs.
+- `Failed.PrePONR` / `OutgoingSnapshot.PrePONRFailure` — records that a terminal
+  failure happened before the operator co-signed the checkpoints, so the source
+  VTXO reservation was provably safe to release.
 - `OutboxHandler` / `LocalPersistenceOutboxHandler` — handles the local
   persistence outbox events (mark-inputs-spent, incoming metadata query,
   VTXO materialization, ack); everything else is handled inline by the
@@ -77,9 +100,37 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/oor.<Sym
   messages and return as fresh events.
 - Snapshots are versioned per direction (`OutgoingSnapshot.Version = 7`,
   `IncomingSnapshot.Version = 1`); restore rejects zero and unknown future
-  versions. Outgoing v5 adds `FirstRejectUnixNanos`; v7 adds the pre-PONR
-  terminal-failure marker. Older snapshots decode absent optional records to
-  their conservative zero values.
+  versions. Outgoing v5 adds `FirstRejectUnixNanos`; v7 adds both the
+  `RecipientOutputs` record (semantic policy metadata a prepared or pre-submit
+  transfer needs after restart) and the pre-PONR terminal-failure marker — v6
+  was skipped, so no snapshot ever carries it. The current value lives in one
+  place, `outgoingSnapshotVersion`, and both the codec and
+  `OutgoingStateFromSnapshot` reject anything above it. Older snapshots decode
+  absent optional records to their conservative zero values.
+- **Preparation has no implied side effect.** `OutboxForState(*Prepared)`
+  returns no outbox events, so a restart that restores a prepared session
+  re-enters `Prepared` and waits; only an explicit `CommitPreparedEvent` may
+  release signatures. Aborting is the safe direction: `Prepared` has not
+  exposed a signature and has not crossed the operator co-sign boundary, so
+  `abortPreparedTransition` releases **every** input.
+- Pre-PONR-ness is derived from the protocol phase (`isPrePONRState`:
+  `Prepared`, `AwaitingArkSignatures`, `AwaitingSubmitAccepted`), not from
+  whether a corrupt or partially restored state happens to have retained its
+  input pointers. `failedState` stamps it on every `Failed` transition.
+- Outgoing idempotency-key admission consults
+  `GetUnfailedSessionByIdempotencyKey` before the in-memory
+  `pendingOutgoingKeys` map in both `handleStartTransfer` and
+  `handleLookupTransfer`. A prepared session is durable before its immutable
+  `oor_dispatch_attempts` row exists, so the mutable-row lookup is what makes
+  an exact replay find it; failed rows release the key, pending and completed
+  rows retain it. A key that resolves to a non-outgoing session is
+  `ErrIdempotencyKeyConflict`.
+- `GetStateRequest` is answered from the durable snapshot when the child is
+  gone. Terminal actors are deliberately short-lived and reaped, but a status
+  reader must not race that reap, so `routeAsk` falls back to
+  `persistedState`, which validates the record's flow version and decodes the
+  direction-appropriate snapshot. Every other message to an unknown session
+  still errors.
 - `StartTransferRequest.IdempotencyKey` dedup reads
   `oor_dispatch_attempts`, not the mutable session row. The table has one row
   per caller key and one caller key per deterministic session id. Once the
diff --git a/unroll/AGENTS.md b/unroll/AGENTS.md
index 125b61c1..9f11babb 100644
--- a/unroll/AGENTS.md
+++ b/unroll/AGENTS.md
@@ -117,6 +117,24 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/unroll.<
 - `ExitSpendPolicyResolver` — interface for looking up the final spend
   policy by `(ExitPolicyKind, ExitPolicyRef)`. Implemented by
   `vhtlcrecovery/unrollpolicy.ExitSpendPolicyResolver`.
+- `ResolverKindSupport` — optional interface a resolver implements to
+  advertise which `ExitPolicyKind`s it covers (`SupportsKind`). The registry
+  consults it at admission time so an unroutable kind fails closed before an
+  actor is spawned.
+- `PolicyResolvers` (`policy_resolvers.go`) — an `ExitSpendPolicyResolver`
+  that fans one `RegistryConfig.ExitSpendPolicyResolver` slot out over several
+  policy families, dispatching by kind. It is itself both an
+  `ExitSpendPolicyResolver` and a `ResolverKindSupport`, so it composes. A
+  child that does **not** implement `ResolverKindSupport` is treated as a
+  legacy resolver: `SupportsKind` admits any kind if such a child is present,
+  and resolution falls back to the first legacy child only after explicit
+  kind routing has found no match. That ordering is what preserves a
+  standalone pre-advertisement resolver's former behaviour when it is wrapped.
+- `PreSignedExitSpendPolicy` — `ExitSpendPolicy` extension for policies whose
+  final spend is fixed before the actor reaches CSV maturity.
+  `PreSignedSpendTx()` returns an isolated copy of the immutable, fully signed
+  transaction. Implemented by the Ark channel materialization policy
+  (`actormsg.ExitPolicyArkChannelBacking`).
 
 ### Feasibility & Funding
 
@@ -280,6 +298,27 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/unroll.<
   network fallback or a policy-owned emergency rate permits construction
   without a fresh estimate; vHTLC policies use the latter because another
   valid leaf can race either recovery action.
+- **A pre-signed final spend published by the peer is success, not a
+  conflict.** `handleSpendObserved` classifies a target-outpoint spend in five
+  cases; case 4 (`ackPreSignedExitSpend`) sits ahead of the catch-all
+  `FailEvent`. When the resolved exit policy implements
+  `PreSignedExitSpendPolicy` and its transaction's txid equals the observed
+  spending txid, the actor caches the full transaction as `b.sweepTx` (the txid
+  alone is not enough — checkpoint replay and exit-cost accounting both need the
+  exact bytes) and drives `TxConfirmedEvent` / `SweepBroadcastedEvent` /
+  `TxConfirmedEvent` idempotently, skipping any leg the planner state already
+  records. Either authorized party may publish the transaction after CSV
+  maturity, so the durable actor has to converge no matter who reached the chain
+  first. Two guards keep it from blessing the wrong spend: the transaction must
+  actually consume `cfg.TargetOutpoint` (otherwise a hard error, not a silent
+  fall-through), and a recorded sweep txid that differs from the observed one is
+  an error rather than an overwrite.
+- **A CSV-delayed spend dates its parent, not itself.** When the target's own
+  confirmation was never recorded, `ackPreSignedExitSpend` derives the parent
+  height as `SpendingHeight - policy.CSVDelay()` instead of using
+  `SpendingHeight` directly, which would make an already-confirmed spend look
+  immature to the planner. A negative result is a hard error: the observed spend
+  could not have satisfied its own CSV.
 - **Reissue fails hard on missing state.** The
   `ReissueInFlightTransactions` and `ReissueSweepConfirmation`
   outbox branches return errors on a missing proof node or nil
diff --git a/unroll/CLAUDE.md b/unroll/CLAUDE.md
index 125b61c1..9f11babb 100644
--- a/unroll/CLAUDE.md
+++ b/unroll/CLAUDE.md
@@ -117,6 +117,24 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/unroll.<
 - `ExitSpendPolicyResolver` — interface for looking up the final spend
   policy by `(ExitPolicyKind, ExitPolicyRef)`. Implemented by
   `vhtlcrecovery/unrollpolicy.ExitSpendPolicyResolver`.
+- `ResolverKindSupport` — optional interface a resolver implements to
+  advertise which `ExitPolicyKind`s it covers (`SupportsKind`). The registry
+  consults it at admission time so an unroutable kind fails closed before an
+  actor is spawned.
+- `PolicyResolvers` (`policy_resolvers.go`) — an `ExitSpendPolicyResolver`
+  that fans one `RegistryConfig.ExitSpendPolicyResolver` slot out over several
+  policy families, dispatching by kind. It is itself both an
+  `ExitSpendPolicyResolver` and a `ResolverKindSupport`, so it composes. A
+  child that does **not** implement `ResolverKindSupport` is treated as a
+  legacy resolver: `SupportsKind` admits any kind if such a child is present,
+  and resolution falls back to the first legacy child only after explicit
+  kind routing has found no match. That ordering is what preserves a
+  standalone pre-advertisement resolver's former behaviour when it is wrapped.
+- `PreSignedExitSpendPolicy` — `ExitSpendPolicy` extension for policies whose
+  final spend is fixed before the actor reaches CSV maturity.
+  `PreSignedSpendTx()` returns an isolated copy of the immutable, fully signed
+  transaction. Implemented by the Ark channel materialization policy
+  (`actormsg.ExitPolicyArkChannelBacking`).
 
 ### Feasibility & Funding
 
@@ -280,6 +298,27 @@ For field-level detail, use `go doc github.com/lightninglabs/wavelength/unroll.<
   network fallback or a policy-owned emergency rate permits construction
   without a fresh estimate; vHTLC policies use the latter because another
   valid leaf can race either recovery action.
+- **A pre-signed final spend published by the peer is success, not a
+  conflict.** `handleSpendObserved` classifies a target-outpoint spend in five
+  cases; case 4 (`ackPreSignedExitSpend`) sits ahead of the catch-all
+  `FailEvent`. When the resolved exit policy implements
+  `PreSignedExitSpendPolicy` and its transaction's txid equals the observed
+  spending txid, the actor caches the full transaction as `b.sweepTx` (the txid
+  alone is not enough — checkpoint replay and exit-cost accounting both need the
+  exact bytes) and drives `TxConfirmedEvent` / `SweepBroadcastedEvent` /
+  `TxConfirmedEvent` idempotently, skipping any leg the planner state already
+  records. Either authorized party may publish the transaction after CSV
+  maturity, so the durable actor has to converge no matter who reached the chain
+  first. Two guards keep it from blessing the wrong spend: the transaction must
+  actually consume `cfg.TargetOutpoint` (otherwise a hard error, not a silent
+  fall-through), and a recorded sweep txid that differs from the observed one is
+  an error rather than an overwrite.
+- **A CSV-delayed spend dates its parent, not itself.** When the target's own
+  confirmation was never recorded, `ackPreSignedExitSpend` derives the parent
+  height as `SpendingHeight - policy.CSVDelay()` instead of using
+  `SpendingHeight` directly, which would make an already-confirmed spend look
+  immature to the planner. A negative result is a hard error: the observed spend
+  could not have satisfied its own CSV.
 - **Reissue fails hard on missing state.** The
   `ReissueInFlightTransactions` and `ReissueSweepConfirmation`
   outbox branches return errors on a missing proof node or nil
diff --git a/vtxo/AGENTS.md b/vtxo/AGENTS.md
index 4dfb2cdb..10efb15d 100644
--- a/vtxo/AGENTS.md
+++ b/vtxo/AGENTS.md
@@ -14,6 +14,15 @@ when the local wallet owns the receive script.
 ## Key Types
 
 - `VTXOState` — Sealed interface for all states (Live, Spending, Spent, PendingForfeit, Forfeiting, Forfeited, UnilateralExit, Expired, Failed).
+- `VTXOStatus` — persisted status enum, **append-only** so the numeric values
+  already written by SQL queries never shift. `VTXOStatusExpired` (8) and
+  `VTXOStatusRecoveryOnly` (9) are the two most recent additions.
+- `VTXOStatusRecoveryOnly` — an application-owned output whose descriptor exists
+  only so the common unroller can recover its on-chain ancestry (an Ark channel
+  backing VTXO is the motivating case). It is deliberately excluded from wallet
+  balance, coin selection, refresh, and VTXO actor recovery, and it has **no
+  VTXO actor**: the owning application FSM controls its lifecycle and admits the
+  exact final-spend policy to unroll.
 - `Descriptor` — Complete VTXO metadata: `Outpoint`, `Amount`, `PolicyTemplate`
   (the authoritative semantic policy for ownership/spend semantics),
   `PkScript`, `ClientKey` (keychain.KeyDescriptor), `OperatorKey`,
@@ -163,6 +172,23 @@ when the local wallet owns the receive script.
   `ListRecoverableVTXOs` includes them and is what `Manager.Start` recovers
   actors from. Coin selection is already `VTXOStatusLive`-scoped via
   `ListSelectionCandidatesByStatus`, so it excludes them by construction.
+- **Recovery-only descriptors are actorless by design, so the manager handles
+  them directly.** Three sites carry the exception, and each one is narrowed so
+  an ordinary wallet coin can never take the same path:
+  - `SumPendingBalance` and `exactSpendUnavailableError` group
+    `VTXOStatusRecoveryOnly` with the terminal/separately-accounted statuses, so
+    it contributes nothing to pending balance and an exact-spend request naming
+    it fails with `ErrInsufficientSpendableFunds`.
+  - `confirmExitedVTXO` accepts a recovery-only row **only** when
+    `req.ExitPolicyKind.Valid()`. The policy check is what stops an ordinary
+    timeout job from retiring an application-owned row; a plain
+    `VTXOStatusUnilateralExit` row still takes the pre-existing path.
+  - `completePersistedSpend` (formerly `isPersistedSpent`) recognizes exactly
+    two actorless completions: a terminal `Spent` row means a previous attempt
+    already finished, and a `RecoveryOnly` row is retired to `Spent` via
+    `UpdateVTXOStatusReleasingReservation`, which drops the durable spend
+    reservation in the same transaction. Every other status still requires its
+    normal VTXO actor, so the rename widened the contract but not the surface.
 - VTXO actor state is the single source of truth for availability.
 - Forfeit transaction is not broadcast until the connector output's round confirms (atomic replacement).
 - Refresh is auto-triggered at configurable height before expiry.
diff --git a/vtxo/CLAUDE.md b/vtxo/CLAUDE.md
index 4dfb2cdb..10efb15d 100644
--- a/vtxo/CLAUDE.md
+++ b/vtxo/CLAUDE.md
@@ -14,6 +14,15 @@ when the local wallet owns the receive script.
 ## Key Types
 
 - `VTXOState` — Sealed interface for all states (Live, Spending, Spent, PendingForfeit, Forfeiting, Forfeited, UnilateralExit, Expired, Failed).
+- `VTXOStatus` — persisted status enum, **append-only** so the numeric values
+  already written by SQL queries never shift. `VTXOStatusExpired` (8) and
+  `VTXOStatusRecoveryOnly` (9) are the two most recent additions.
+- `VTXOStatusRecoveryOnly` — an application-owned output whose descriptor exists
+  only so the common unroller can recover its on-chain ancestry (an Ark channel
+  backing VTXO is the motivating case). It is deliberately excluded from wallet
+  balance, coin selection, refresh, and VTXO actor recovery, and it has **no
+  VTXO actor**: the owning application FSM controls its lifecycle and admits the
+  exact final-spend policy to unroll.
 - `Descriptor` — Complete VTXO metadata: `Outpoint`, `Amount`, `PolicyTemplate`
   (the authoritative semantic policy for ownership/spend semantics),
   `PkScript`, `ClientKey` (keychain.KeyDescriptor), `OperatorKey`,
@@ -163,6 +172,23 @@ when the local wallet owns the receive script.
   `ListRecoverableVTXOs` includes them and is what `Manager.Start` recovers
   actors from. Coin selection is already `VTXOStatusLive`-scoped via
   `ListSelectionCandidatesByStatus`, so it excludes them by construction.
+- **Recovery-only descriptors are actorless by design, so the manager handles
+  them directly.** Three sites carry the exception, and each one is narrowed so
+  an ordinary wallet coin can never take the same path:
+  - `SumPendingBalance` and `exactSpendUnavailableError` group
+    `VTXOStatusRecoveryOnly` with the terminal/separately-accounted statuses, so
+    it contributes nothing to pending balance and an exact-spend request naming
+    it fails with `ErrInsufficientSpendableFunds`.
+  - `confirmExitedVTXO` accepts a recovery-only row **only** when
+    `req.ExitPolicyKind.Valid()`. The policy check is what stops an ordinary
+    timeout job from retiring an application-owned row; a plain
+    `VTXOStatusUnilateralExit` row still takes the pre-existing path.
+  - `completePersistedSpend` (formerly `isPersistedSpent`) recognizes exactly
+    two actorless completions: a terminal `Spent` row means a previous attempt
+    already finished, and a `RecoveryOnly` row is retired to `Spent` via
+    `UpdateVTXOStatusReleasingReservation`, which drops the durable spend
+    reservation in the same transaction. Every other status still requires its
+    normal VTXO actor, so the rename widened the contract but not the surface.
 - VTXO actor state is the single source of truth for availability.
 - Forfeit transaction is not broadcast until the connector output's round confirms (atomic replacement).
 - Refresh is auto-triggered at configurable height before expiry.
```

</details>

**How to apply:** save the diff above and `git apply` it on this branch, or run the nightly gardener locally with `/doc-gardening` scoped to the packages named above. Re-run `make doc-check` afterwards.

This check is advisory only — it never fails the build.

[Workflow run](https://github.com/lightninglabs/wavelength/actions/runs/34279121839)

litbot-9000:

litbot-9000:
@bhandras: review reminder
@roasbeef: review reminder

PR #1190, opened 2026-08-24
arkchannel: Add durable channel lifecycle

sputn1ck:
## Summary

- add the durable channel FSM for VTXO promotion, hub-funded receive intents, activation, cooperative Ark settlement, materialization, and failure
- persist channel terms, backing transactions, close state, and recovery evidence
- define the authenticated mailbox and RPC contracts used by the two channel endpoints

This is layer 2 of 5 in the Ark channels stack. It depends on #1189; the next layer is #1191.

## Testing

- `go test ./arkchannel/... ./db ./serverconn ./rpc/arkchannelrpc`
- `make build`
- `make lint-changed-local`


litbot-9000:
### 📚 Per-package doc advisory

This PR's Go changes touch four documented packages — `db`, `db/sqlc`, `mailbox/rpc`, and `serverconn`. `db`, `db/sqlc` and `serverconn` were already updated in the PR; the remaining drift is the new correlation-ID/idempotency-key split in `serverconn.UnaryFacade.SendRPC` (undocumented in both `serverconn` and `mailbox/rpc`), the relaxed `RetryWithKey` contract in `mailbox/rpc`, and the migration list in `db/CLAUDE.md`, where `000022`–`000024` were inserted ahead of `000021` (and `000020` was never listed).

```diff
diff --git a/db/AGENTS.md b/db/AGENTS.md
index 8fede44f..63158871 100644
--- a/db/AGENTS.md
+++ b/db/AGENTS.md
@@ -407,6 +407,17 @@ when adding one.
   post-step detects the old refresh/exit collision and reconstructs the
   missing net exit-send row from the surviving refresh-send and exit-fee rows,
   repairing the overstated VTXO balance atomically with the key rewrite.
+- `000020_taproot_asset_vtxo_state` — adds the asset columns on `vtxos`
+  (`taproot_asset_root`, `taproot_asset_ref`, `taproot_asset_amount`, and the
+  optional `taproot_asset_sealed_package`). Units are an 8-byte BLOB so the
+  full uint64 range survives on both SQLite and Postgres, and a column-group
+  check makes the asset identity all-or-nothing.
+- `000021_round_admission_deadlines` — records accepted attempt expiry and
+  closure separately from signature checkpoints. Deadline constraints only
+  shorten the saved budget. Startup closes interrupted admissions without
+  changing checkpoints or releasing reservations; the existing VTXO recovery
+  boundary remains authoritative. `CommitState` closes admission atomically
+  with its signature-bearing checkpoint.
 - `000022_ark_channels` — creates `ark_channels`, one revisioned row per
   Ark-to-lnd channel. It stores only facts that cross the two databases;
   Lightning commitments, HTLCs, invoices, and payment attempts remain in lnd.
@@ -425,13 +436,6 @@ when adding one.
   Existing rows are backfilled from `updated_at` only when preparation had
   begun: phase 1 or 2 with an OOR session.
 
-- `000021_round_admission_deadlines` — records accepted attempt expiry and
-  closure separately from signature checkpoints. Deadline constraints only
-  shorten the saved budget. Startup closes interrupted admissions without
-  changing checkpoints or releasing reservations; the existing VTXO recovery
-  boundary remains authoritative. `CommitState` closes admission atomically
-  with its signature-bearing checkpoint.
-
 ## Deep Docs
 
 - [ARCHITECTURE.md](../ARCHITECTURE.md) — System-wide package map.
diff --git a/db/CLAUDE.md b/db/CLAUDE.md
index 8fede44f..63158871 100644
--- a/db/CLAUDE.md
+++ b/db/CLAUDE.md
@@ -407,6 +407,17 @@ when adding one.
   post-step detects the old refresh/exit collision and reconstructs the
   missing net exit-send row from the surviving refresh-send and exit-fee rows,
   repairing the overstated VTXO balance atomically with the key rewrite.
+- `000020_taproot_asset_vtxo_state` — adds the asset columns on `vtxos`
+  (`taproot_asset_root`, `taproot_asset_ref`, `taproot_asset_amount`, and the
+  optional `taproot_asset_sealed_package`). Units are an 8-byte BLOB so the
+  full uint64 range survives on both SQLite and Postgres, and a column-group
+  check makes the asset identity all-or-nothing.
+- `000021_round_admission_deadlines` — records accepted attempt expiry and
+  closure separately from signature checkpoints. Deadline constraints only
+  shorten the saved budget. Startup closes interrupted admissions without
+  changing checkpoints or releasing reservations; the existing VTXO recovery
+  boundary remains authoritative. `CommitState` closes admission atomically
+  with its signature-bearing checkpoint.
 - `000022_ark_channels` — creates `ark_channels`, one revisioned row per
   Ark-to-lnd channel. It stores only facts that cross the two databases;
   Lightning commitments, HTLCs, invoices, and payment attempts remain in lnd.
@@ -425,13 +436,6 @@ when adding one.
   Existing rows are backfilled from `updated_at` only when preparation had
   begun: phase 1 or 2 with an OOR session.
 
-- `000021_round_admission_deadlines` — records accepted attempt expiry and
-  closure separately from signature checkpoints. Deadline constraints only
-  shorten the saved budget. Startup closes interrupted admissions without
-  changing checkpoints or releasing reservations; the existing VTXO recovery
-  boundary remains authoritative. `CommitState` closes admission atomically
-  with its signature-bearing checkpoint.
-
 ## Deep Docs
 
 - [ARCHITECTURE.md](../ARCHITECTURE.md) — System-wide package map.
diff --git a/mailbox/rpc/AGENTS.md b/mailbox/rpc/AGENTS.md
index 26787ed6..a790cf23 100644
--- a/mailbox/rpc/AGENTS.md
+++ b/mailbox/rpc/AGENTS.md
@@ -65,6 +65,19 @@ and server-side routing need without including any transport implementation.
   `Retry`/`RetryWithKey` (or thread an explicit `RPCOptions.IdempotencyKey`);
   leaving the field empty makes the transport mint a fresh key per attempt,
   which defeats operator-side deduplication.
+- `RPCOptions.CorrelationID` and `RPCOptions.IdempotencyKey` are separate
+  identities and must not be aliased. Correlation matches one response to one
+  transport attempt; idempotency spans every attempt of the logical operation.
+  A transport that defaulted the correlation ID to the key would give the
+  server response a repeated mailbox message ID, so an acknowledged earlier
+  attempt could tombstone the retry's answer. `serverconn.UnaryFacade` defaults
+  an empty correlation ID to the envelope's fresh `MsgId` for that reason.
+  (The `CorrelationID` field comment in `options.go` still describes the older
+  key-as-correlation-ID convention and no longer matches the transport.)
+- `RetryWithKey`'s key must not already be in flight elsewhere: operator
+  deduplication assumes one active attempt per logical operation, so concurrent
+  requests under one key can observe an in-doubt execution rather than its
+  eventual result. Nothing in the package enforces that serialization.
 - A `KIND_RESPONSE` envelope carrying `HeaderGRPCStatusB64` signals a failed
   RPC; receivers must decode it via `DecodeErrorHeaders` before attempting to
   unmarshal the body.
diff --git a/mailbox/rpc/CLAUDE.md b/mailbox/rpc/CLAUDE.md
index 26787ed6..a790cf23 100644
--- a/mailbox/rpc/CLAUDE.md
+++ b/mailbox/rpc/CLAUDE.md
@@ -65,6 +65,19 @@ and server-side routing need without including any transport implementation.
   `Retry`/`RetryWithKey` (or thread an explicit `RPCOptions.IdempotencyKey`);
   leaving the field empty makes the transport mint a fresh key per attempt,
   which defeats operator-side deduplication.
+- `RPCOptions.CorrelationID` and `RPCOptions.IdempotencyKey` are separate
+  identities and must not be aliased. Correlation matches one response to one
+  transport attempt; idempotency spans every attempt of the logical operation.
+  A transport that defaulted the correlation ID to the key would give the
+  server response a repeated mailbox message ID, so an acknowledged earlier
+  attempt could tombstone the retry's answer. `serverconn.UnaryFacade` defaults
+  an empty correlation ID to the envelope's fresh `MsgId` for that reason.
+  (The `CorrelationID` field comment in `options.go` still describes the older
+  key-as-correlation-ID convention and no longer matches the transport.)
+- `RetryWithKey`'s key must not already be in flight elsewhere: operator
+  deduplication assumes one active attempt per logical operation, so concurrent
+  requests under one key can observe an in-doubt execution rather than its
+  eventual result. Nothing in the package enforces that serialization.
 - A `KIND_RESPONSE` envelope carrying `HeaderGRPCStatusB64` signals a failed
   RPC; receivers must decode it via `DecodeErrorHeaders` before attempting to
   unmarshal the body.
diff --git a/serverconn/AGENTS.md b/serverconn/AGENTS.md
index 850369ea..a65ddccd 100644
--- a/serverconn/AGENTS.md
+++ b/serverconn/AGENTS.md
@@ -142,6 +142,13 @@ background ingress polling with event routing.
   the logical request (see `mailboxrpc.Retry`). `SendRPC` mints a fresh key
   only when the caller leaves `RPCOptions.IdempotencyKey` empty, which is
   correct for a single-shot call and defeats deduplication for a retry.
+- Correlation identifies one transport attempt; the idempotency key identifies
+  the logical operation across attempts. `SendRPC` defaults an empty
+  `RPCOptions.CorrelationID` to the freshly minted **`MsgId`**, never to the
+  idempotency key. Deriving it from the key would also reuse the server
+  response's mailbox message ID, so once an earlier attempt's response had been
+  acknowledged the mailbox tombstone would suppress the cached response and the
+  retry's waiter would block until its deadline.
 - Ingress loop checkpoints pull cursor and ack state; on restart, resumes from checkpoint.
 - **A mailbox outage warns once per episode, not once per retry.** `pullPhase`
   carries a dedicated `pullFailCount` alongside the shared backoff `failCount`:
diff --git a/serverconn/CLAUDE.md b/serverconn/CLAUDE.md
index 850369ea..a65ddccd 100644
--- a/serverconn/CLAUDE.md
+++ b/serverconn/CLAUDE.md
@@ -142,6 +142,13 @@ background ingress polling with event routing.
   the logical request (see `mailboxrpc.Retry`). `SendRPC` mints a fresh key
   only when the caller leaves `RPCOptions.IdempotencyKey` empty, which is
   correct for a single-shot call and defeats deduplication for a retry.
+- Correlation identifies one transport attempt; the idempotency key identifies
+  the logical operation across attempts. `SendRPC` defaults an empty
+  `RPCOptions.CorrelationID` to the freshly minted **`MsgId`**, never to the
+  idempotency key. Deriving it from the key would also reuse the server
+  response's mailbox message ID, so once an earlier attempt's response had been
+  acknowledged the mailbox tombstone would suppress the cached response and the
+  retry's waiter would block until its deadline.
 - Ingress loop checkpoints pull cursor and ack state; on restart, resumes from checkpoint.
 - **A mailbox outage warns once per episode, not once per retry.** `pullPhase`
   carries a dedicated `pullFailCount` alongside the shared backoff `failCount`:
```

**How to apply:** save the block above to a file and `git apply` it, or run the `doc-gardening` skill locally over `db db/sqlc mailbox/rpc serverconn`. The `CLAUDE.md`/`AGENTS.md` pairs stay byte-identical, so `make doc-check` passes either way.

This is advisory only — nothing here blocks merge.

---
https://github.com/lightninglabs/wavelength/actions

arkchannel/types.go
```diff
@@ -0,0 +1,571 @@
...
+		return fmt.Errorf("decode backing transaction: %w", err)
+	}
+	if len(tx.TxIn) != 1 {
+		return fmt.Errorf("backing transaction must have one input")
+	}
+	if tx.TxIn[0].PreviousOutPoint != source.OutPoint {
+		return fmt.Errorf("backing transaction does not spend bound " +
+			"VTXO")
+	}
+	if len(tx.TxIn[0].Witness) == 0 {
```
darioAnongba, line 478: +1. Only `len(Witness) == 0` is checked; the script engine only runs in `Complete`, which only the assembling side does. `applyBacking` and restore accept through this, and `InstallBacking` hands the peer's backing to the other side, so a garbage witness crosses the PONR and only blows up at materialization. Run the engine here or in `applyBacking` (same inputs as `Complete`). Also worth a comment that this can't bind the funding script, so handlers must call `ValidateFundingOutput`.

rpc/arkchannelrpc/ark_channel.proto
```diff
@@ -0,0 +1,554 @@
...
+    CHANNEL_EVENT_TYPE_CHANNEL_ACTIVE = 3;
+    CHANNEL_EVENT_TYPE_MATERIALIZE = 4;
+    CHANNEL_EVENT_TYPE_BACKING_PUBLISHED = 5;
+    CHANNEL_EVENT_TYPE_RECOVERY_PACKAGE_INSTALLED = 6;
+    CHANNEL_EVENT_TYPE_FAILED = 7;
+    CHANNEL_EVENT_TYPE_FUNDING_PEER_READY = 8;
+    CHANNEL_EVENT_TYPE_OOR_ABORTED = 9;
+}
+
+message ApplyChannelEventRequest {
```
darioAnongba, line 327: The peer can assert `FUNDING_FINALIZED` for any party, `OOR_FINALIZED`/`ABORTED`, `BACKING_PUBLISHED` and `RECOVERY_PACKAGE_INSTALLED`. `applyFundingFinalized` checks phase, not origin, and `RecoveryReady` is a single bit, so a hub (or a 1:1 handler) sending `RECOVERY_PACKAGE_INSTALLED` early activates a client channel with no local source watch. I'd drop `BACKING_PUBLISHED` (we have `BackingObserved` for chain evidence) and local-party `FUNDING_FINALIZED` from this set, and make `RecoveryReady` per-party.

arkchannel/state.go
```diff
@@ -0,0 +1,1354 @@
...
+
+	next.OORFinalized = true
+
+	return true, nil
+}
+
+// applyRecoveryPackageInstalled records the symmetric recovery barrier. It is
+// valid before the remote OOR-finalized acknowledgement arrives because the
+// package itself can only be exported after actual OOR completion.
+func applyRecoveryPackageInstalled(next *Snapshot) (bool, error) {
```
darioAnongba: Every other two-endpoint barrier is a per-party pair; this one is a single bool set by one event. Splitting it is much cheaper before the schema ships.

arkchannel/state.go
```diff
@@ -0,0 +1,1354 @@
...
+
+	previousPhase := s.snapshot.Phase
+	next := s.snapshot.Clone()
+	changed, err := applyEvent(&next, event)
+	if err != nil {
+		return nil, err
+	}
+	if !changed {
+		switch event.(type) {
+		case *SourceSpent, *RecoveryPackageInstalled, *OORFinalized,
```
darioAnongba: +1. Only these four re-emit on a no-change replay. `FundingFinalized{hub}` commits `backing_ready`, `Execute(CommitOOR)` fails transiently, the retried event is a no-op with no action, and nothing runs `CommitOOR` again until a restart (nothing calls `ResumeChannelAction`). Funds sit at the gate with signed commitments on both sides. Widen the re-emit set or add a periodic `ResumeAll`.

db/sqlc/migrations/000020_ark_channels.up.sql
```diff
@@ -0,0 +1,92 @@
+-- Ark channel coordination stores only facts that cross the Ark and lnd
+-- databases. Lightning commitments, HTLCs, invoices, and payment attempts
+-- remain exclusively in lnd.
+CREATE TABLE IF NOT EXISTS ark_channels (
+    channel_id BLOB PRIMARY KEY NOT NULL,
+    kind INTEGER NOT NULL CHECK (kind IN (1, 2)),
+    funder INTEGER NOT NULL CHECK (funder IN (1, 2)),
+    pending_channel_id BLOB NOT NULL UNIQUE,
```
darioAnongba, line 8: No uniqueness on `(oor_session_id, source_index)`. The funder is covered by its idempotency key, but the non-funder stores whatever binding the hub sends, so one prepared output can back two receive channels of the same client and only one can ever materialize. Partial unique index `WHERE oor_session_id IS NOT NULL`. `reserved_scid` probably wants one too.

arkchannel/state.go
```diff
@@ -0,0 +1,1354 @@
...
+			Index: event.ChannelPointIndex,
+		})
+
+	case *Materialize:
+		if next.Phase == PhaseMaterializing ||
+			next.Phase == PhaseOnChain {
+			return false, nil
+		}
+		if next.Phase != PhaseActive {
+			return false, fmt.Errorf("cannot materialize "+
```
darioAnongba, line 302: Trap: with a stored artifact and one ack, abort is refused, `Materialize` is refused here, and publish needs both acks. Only `SourceSpent`/`BackingObserved` get you out, and nothing in the FSM can cause them. Hub authorizes then disappears for good = quiesced forever. I'd allow `Materialize` from `coop_closing` (not `coop_close_authorized`): before both acks an honest client can't have committed the close OOR, so it's as safe as the `SourceSpent` path.

darioAnongba (commented):
Strongest part of the stack for me: illegal persisted combinations are unrepresentable, the CAS is single-statement and race-tested, commit/abort races resolve to the OOR actor's gate, post-PONR failure parks instead of inventing an abort. Tests pass.

Two things I think belong in this PR because they define the durable model: `Backing.Validate` accepts any non-empty witness, and the peer can assert facts (including the single-bit `RecoveryReady`) that only local evidence should set. Plus no in-process retry after the CAS. Inline.

Three stack-level design points, parked here since this is the FSM:

Expiry. `recovery.go` stores `BatchExpiry` but nothing acts on it: no expiry-driven close, no renewal, and recovery-only sources are excluded from auto-refresh. If the source batch expires under a live channel the sweep kills the backing and the balance becomes a claim on the hub. So a channel is currently shorter-lived than a refreshed VTXO and close+reopen is the only lifecycle. Even a minimal ladder (warn, coop close, force close as expiry nears, hub-driven if the client is gone) would do for now; long-expiry funding batches or in-place renewal later.

Backing fee. Fixed at prepare time, single output, no anchor. Bumping the materialization spend relies entirely on CPFP via the commitment anchors, which don't exist yet (see the legacy-commitment comment on #1191), and even with anchors a fee spike inside the 432-block window favors the funder, who can just wait. A floor on the reserve, or at least state the CPFP dependency in the terms.

Event origin. We fence who may execute (commit/abort are funder-only) but not who may assert. An origin x event matrix at the `Service` boundary would stop later handlers from forgetting it. Specifics inline.

Smaller: `PendingAction` for `backing_ready` has an unreachable `CommitOOR` branch (make it explicit or return `ActivateChannel`); `applyPrePONRExpiry` errors on an empty reason (the reaper must never send one); proto `max_fee_sat` is `uint64` while other amounts are `int64`, artifacts have no version field, `CooperativeCloseProposal` has no `channel_id`; `ListNonTerminal` keeps closed-with-artifact rows forever so the startup scan only grows; the bridge controllers poll at 50/100ms with no backoff and execution is re-entrant on the caller's ctx; dead `close_fee_rate_sat_per_kw` column and positional `$20`/`$52` in `InsertArkChannel`; FSM phase constants duplicated as SQL literals in migration 21 and `ListNonTerminal`.

Long version: https://ark-channels.lightning.wiki/

arkchannel/state.go
```diff
@@ -0,0 +1,1354 @@
...
+		return false, nil
+	}
+	next.RecoveryReady = true
+
+	return true, nil
+}
+
+// applySourceSpent persists the first ancestor conflict and drives the
+// virtual channel toward its already-negotiated on-chain resolution path.
+func applySourceSpent(next *Snapshot, outpoint wire.OutPoint,
```
darioAnongba, line 553: In `coop_close_authorized` the replayable `PublishCooperativeClose` may already have driven `CommitPreparedEvent`, so the close OOR can be in flight at the operator when we wipe it here and materialize: two resolutions of the same VTXO. Only safe if the operator can never finalize an OOR whose ancestor is spent on chain. Where is that enforced/tested? If it is, just a doc request.

serverconn/ingress.go
```diff
@@ -721,7 +721,7 @@ func (a *ServerConnectionActor) ackRemote(
 func (a *ServerConnectionActor) loadCheckpoint(ctx context.Context) (AckState,
 	error) {
 
-	actorID := DurableActorID(a.cfg.LocalMailboxID)
+	actorID := a.runtimeID()
```
darioAnongba, line 766: `RuntimeID` and `ReplyMailboxID` are independent: checkpoint keyed by one, pull/ack by the other, and the checkpoint doesn't record its mailbox. Set one without the other and you either ack the wrong mailbox to a foreign cursor (skipping envelopes) or race two runtimes on one mailbox. Store the mailbox in the checkpoint and refuse mismatches, or require both in `NewRuntime`. Also `ReplyTo == LocalMailboxID` no longer holds, so who may pull which mailbox now has to be enforced on the swapd side.

arkchannel/payment_bridge.go
```diff
@@ -0,0 +1,832 @@
...
+			next.Phase)
+	}
+	next.Phase = PaymentFailed
+
+	return true, nil
+}
+
+// applyPaymentFallbackSelected hands a pristine incoming payment to vHTLCs.
+func applyPaymentFallbackSelected(next *PaymentBridgeSnapshot,
+	event *PaymentFallbackSelected) (bool, error) {
```
darioAnongba, line 773: Fallback is allowed from `DestinationInFlight` with no proof the private attempt failed. lnd payments can't be cancelled, so an impatient caller plus a client that settles the private HTLC and claims the vHTLC = paid twice for one public HTLC. Restrict to pre-dispatch states or require a destination-terminal fact. The test shows it's intentional, so at least say why.

arkchannel/cooperative_close.go
```diff
@@ -0,0 +1,642 @@
...
+	if err := source.Validate(terms); err != nil {
+		return CooperativeCloseProposal{},
+			cooperativeCloseArtifacts{}, err
+	}
+	if err := request.Validate(); err != nil {
+		return CooperativeCloseProposal{},
+			cooperativeCloseArtifacts{}, err
+	}
+	if clientBalance < 0 || hubBalance < 0 ||
+		clientBalance+hubBalance != terms.Capacity {
```
darioAnongba, line 425: Design note: this needs `client + hub == Capacity` exactly, but lnd balances never sum to capacity (commit fee). The closer has to rebuild the full-capacity split and credit the fee back to the funder; a wrong adjustment shortchanges one side and the hub's own reconstruction is the only defense. Worth writing the rule down.

arkchannel/types.go
```diff
@@ -0,0 +1,571 @@
...
+		return fmt.Errorf("backing transaction is not fully signed")
```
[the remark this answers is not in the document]
sputn1ck: Addressed on the updated branch: `Backing.Validate` now reconstructs and executes the channel spend path against the bound source output via `verifyBackingWitness`. The method also documents that each endpoint must separately validate the lnd funding output script.

arkchannel/types.go
sputn1ck: Addressed on the updated branch: restore and peer admission now execute the backing witness against `source.PkScript`; funding-script binding remains an explicit endpoint check through `BackingTemplate.ValidateFundingOutput`.

arkchannel/state.go
```diff
@@ -0,0 +1,1354 @@
...
+	previousPhase := s.snapshot.Phase
+	next := s.snapshot.Clone()
+	changed, err := applyEvent(&next, event)
+	if err != nil {
+		return nil, err
+	}
+	if !changed {
+		switch event.(type) {
+		case *SourceSpent, *RecoveryPackageInstalled, *OORFinalized,
+			*OORAborted:
```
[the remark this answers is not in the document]
sputn1ck: Addressed on the updated branch: every no-change event replay now asks `PendingAction` for the action implied by durable state. `Coordinator.Resume`/`ResumeAll` expose the same replay path for startup and explicit recovery.

arkchannel/state.go
sputn1ck: Addressed on the updated branch: no-op event replay is no longer limited to four event types; it re-emits the durable `PendingAction`, including `CommitOOR`. Explicit and startup resume paths use the same derivation.

rpc/arkchannelrpc/ark_channel.proto
sputn1ck: Addressed on the updated branch. Peer event admission is now origin-aware: each party-scoped fact must come from the endpoint it names; `BackingPublished`, `BackingObserved`, and other subsystem evidence are local-only. Recovery readiness is split into client and hub bits, and activation requires both.

arkchannel/state.go
sputn1ck: Addressed on the updated branch: `ClientRecoveryReady` and `HubRecoveryReady` are persisted independently, and `RecoveryReady()` requires both endpoint facts before activation.

db/sqlc/migrations/000020_ark_channels.up.sql
sputn1ck: Addressed on the updated branch: migration 20 now has unique indexes for `reserved_scid` and the non-null `(oor_session_id, source_index)` pair. The SQL-backed store tests cover duplicate bindings and SCID uniqueness.

arkchannel/state.go
sputn1ck: Addressed on the updated branch: `Materialize` is allowed from `PhaseCoopClosing` before both signatures are durable, clears the incomplete cooperative attempt, and remains rejected from `PhaseCoopCloseSigned`. State tests cover both sides of that boundary.

serverconn/ingress.go
sputn1ck: Addressed on the updated branch by removing the independent runtime identifier. `DurableActorID` is derived from the effective reply mailbox, and egress persistence, ingress checkpoints, pull, ack, quarantine, and receipt lanes all use that same namespace. Namespace tests cover two reply lanes sharing one authenticated identity.

arkchannel/payment_bridge.go
sputn1ck: Addressed on the updated branch: fallback is now restricted to `PaymentRegistered` and `PaymentSourceLocked`; `PaymentDestinationInFlight` is rejected. `TestPaymentBridgeRejectsFallbackAfterDestinationDispatch` covers the double-pay boundary.

arkchannel/cooperative_close.go
sputn1ck: Documented and enforced on the updated branch: cooperative-close inputs are full-capacity role balances and must sum exactly to `Terms.Capacity`; the backing reserve is then credited separately to the original funder. Focused tests cover the sum and reserve allocation.

arkchannel/state.go
sputn1ck: Addressed on the updated branch by separating intent from chain fact. In `PhaseCoopCloseSigned`, `SourceSpent` preserves the replayable cooperative OOR and does not materialize. Only local `BackingObserved` evidence that the exact backing transaction already won can clear the stale cooperative artifact, and that transition emits no publish action. `TestMaterializationSupersedesCooperativeNegotiation` covers both negotiating and fully signed states.

litbot-9000:
@litbot-9000: review reminder

litbot-9000:
@bhandras: review reminder
@roasbeef: review reminder
@litbot-9000: review reminder

PR #1215, opened 2026-08-28
vtxo: prefer cooperative refresh when exit is infeasible

### LINKED ISSUE #1212: Unable to refresh vtxo
I attempted two Lightning payments yesterday, which both failed due to a route not being found. I was correctly refunded for both attempts.

I ended up with three vtxos. Today, I restarted `waved` and all three of my vtxos were shown as `pending out` for about seven minutes. Two of the vtxos then came back as live, but the third one couldn't be refreshed. This third vtxo is now shown as `UNILATERAL_EXIT`.

I assume there is no cost-effective way to retrieve this vtxo given its size (5099) and depth (4).

Below I have attached the vtxo in question. I have also attached the logs.

**How could this have been prevented, or how could it still be recovered?**

Version: latest master commit: `e8d78eb`
Network: Mainnet

```
{
  "vtxos":  [
    {
      "outpoint":  "6abb248314a5bc17a1a410e9889639e7d1de97cdc2e368a381ed489f9c1aedda:0",
      "amount_sat":  "5099",
      "status":  "VTXO_STATUS_UNILATERAL_EXIT",
      "batch_expiry":  964513,
      "round_id":  "01a026e8-21be-7baa-a523-df2e7a17c918",
      "created_height":  963505,
      "relative_expiry":  144,
      "pk_script":  "5120ffe8fe17fbcc8bc65aa30d6ea7305001c09d603eb4995b0fbdc18fb9303f913e",
      "commitment_txid":  "088028a70a32eee66c2605aa8049cc0e2afeee71284b1a2d57da345fe79e8480",
      "chain_depth":  4,
      "oor_final_checkpoint_psbts":  [
        "cHNidP8BAGsDAAAAAbOghIZ0wCqmTqyeuXVCPM3Hx6VRI/g6FdaO86u//WFjAAAAAAD/////AusTAAAAAAAAIlEg13nlrXyXO+JvNiubQMmL7qjn2o5NpTPRQhWTEayPuSoAAAAAAAAAAARRAk5zAAAAAAABASvrEwAAAAAAACJRICjKaNncfpgfEhGwVFdDv+OdvS0VwEHJCCw6/oS8QYShAQj9jQEFQJg1ZdD+Grhhlo6fk2j6sFKr8ob7ti0UWsSgBeXe1iGTN+/GvCmruJQkk5/7ScEsTLWmgUYLqgelqQ+haOCgkN5Al548O0XYBeWvOVEebtkqDHpnvbkuc+ocZ70i9Nd5KVSF1WYYtFFupr7ndmr5ueKqe280ZJONFiQ4sMLfKVHhf0CJLQBEPF8J/ylGNDzaN+KTLKEm4uWrXj5LW4D+zMh5b7kRQAVdwMJ4F4kMgJoWBFBBG4mWccYbn2fwC2Mp5iFoZiDxHIKUio7Kj1tHwJB8/yZUZ47nhnKNP9A8O3eAbHULwa0g/3rD5hI2U2ncYBAuZedPfL2R0o60aUG+yksqxGasY/OtIBm1Ki1Wc+WPk7rT1bpZ2FVl8li8PakIe3ZZ5n3MIpsXrGHANy8iWzyu6CEwlt4yKe5DNTBrB8PBaUOEYbXUdJiE7GUEFAIWxceZL+C7j2SKHkmvmbriskahqf0+N35jKMIXU36ZqN88jpZ3xAfWpFrEgKb0BnQuqQEOVIezQQ4pbDcqAAEGfAEBAAN3LAEBwAInIBm1Ki1Wc+WPk7rT1bpZ2FVl8li8PakIe3ZZ5n3MIpsXrAKQALJ1SQEBwAJEIPEcgpSKjsqPW0fAkHz/JlRnjueGco0/0Dw7d4BsdQvBrSAZtSotVnPlj5O609W6WdhVZfJYvD2pCHt2WeZ9zCKbF6wAAA=="
      ],
      "spent_by_txid":  "",
      "expiry_info":  null,
      "settlement":  null
    }
  ]
}
```

[waved.log](https://github.com/user-attachments/files/31525707/waved.log)

bhandras:
## Summary

- Reuse the unilateral-exit feasibility model for automatic critical-expiry
  decisions.
- Start or continue cooperative refresh while the backing wallet cannot
  execute the exit package.
- Reassess on every critical block. Move to unilateral exit as soon as the
  package becomes viable.
- Log the selected action, concrete reason, height, blocks remaining, and
  outpoint.

Fixes #1212.

## Why

The affected VTXO was first observed after restart with 187 blocks remaining.
That was inside its conservative 192-block unilateral-exit budget, so the old
FSM moved directly from `Live` to `UnilateralExit`.

The exit package needed distinct wallet fee inputs for its recovery paths. The
backing wallet had none. `txconfirm` therefore kept retrying the first fee
parent, while a cooperative round was still available.

Urgency and path viability are separate decisions. Entering the critical
window should trigger an assessment. It should not force an automatic exit
that the wallet cannot execute.

## Behavior

| Critical VTXO state | Exit assessment | Action |
| --- | --- | --- |
| `Live` | feasible | Keep the existing direct unilateral exit |
| `Live` | infeasible | Reserve the VTXO and request cooperative refresh |
| `PendingForfeit` | infeasible | Keep waiting for the cooperative round |
| `PendingForfeit` | becomes feasible | Fall back to unilateral exit |
| `Live` or `PendingForfeit` | assessment error | Preserve the existing unilateral fallback |

Manual exits, fraud-triggered exits, vHTLC recovery, and VTXOs already past the
forfeit-signature boundary are unchanged.

## Observability

The decision now produces structured `Automatic expiry decision` logs with:

- `action=cooperative_refresh` or `action=unilateral_exit`;
- a concrete reason such as the required and available fee-input counts;
- `height`, `blocks_remaining`, and `outpoint`.

This does not fully solve the user-facing status gap. A focused follow-up
should persist and expose:

- the automatic decision trigger, chosen action, and reason;
- the last decision height and blocks remaining;
- the latest `txconfirm` retry error for a materializing exit.

The natural read surfaces are `GetVTXOExpiryInfo` before an exit and
`GetUnrollStatus` after one starts. That work crosses VTXO, unroll, txconfirm,
protobuf, and CLI ownership, so it is intentionally not folded into this bug
fix.

## Tests

- `go test ./...`
- `go test ./vtxo ./waved -count=1`
- pinned custom lint on `./vtxo ./waved`: 0 issues
- `make rpc`
- `make fmt`

The regression test pins the issue shape: 187 blocks remaining, infeasible exit
funding, cooperative refresh retained across the next block, and unilateral
fallback once the assessment becomes feasible.

### DIFF
--- a/vtxo/actor.go
+++ b/vtxo/actor.go
@@ -56,6 +56,24 @@ func VTXOActorServiceKey(outpoint wire.OutPoint) actor.ServiceKey[
 type RefreshFeeQuoter func(ctx context.Context,
 	amount btcutil.Amount, remainingBlocks uint32) btcutil.Amount
 
+// CriticalExitAssessment reports whether the backing wallet can execute an
+// automatic unilateral exit at the current fee rate. Reason is a short,
+// stable diagnostic when Feasible is false.
+type CriticalExitAssessment struct {
+	// Feasible permits the existing automatic unilateral-exit transition.
+	Feasible bool
+
+	// Reason explains an infeasible verdict in logs.
+	Reason string
+}
+
+// CriticalExitAssessor checks the whole exit package before a live VTXO enters
+// unilateral exit due to critical expiry. Errors preserve the existing direct
+// exit behavior so an unavailable assessment cannot suppress the safety path.
+type CriticalExitAssessor func(context.Context, *Descriptor) (
+	CriticalExitAssessment, error,
+)
+
 // VTXOActorConfig holds configuration for a single VTXO actor.
 type VTXOActorConfig struct {
 	VTXO        *Descriptor
@@ -96,6 +114,13 @@ type VTXOActorConfig struct {
 	// emits RefreshVTXORequest with OperatorFee=0, which is fine:
 	// the seal-time quote is still the source of truth.
 	RefreshFeeQuoter RefreshFeeQuoter
+
+	// CriticalExitAssessor, when set, checks wallet funding and package
+	// economics before an automatic critical-expiry exit. An infeasible
+	// verdict starts or continues cooperative refresh; the actor reassesses
+	// each block and exits when viable. Nil, errors, and feasible verdicts
+	// retain the existing direct exit.
+	CriticalExitAssessor CriticalExitAssessor
 
 	// FetchOperatorKey, when set, returns the operator's current
 	// long-term public key by issuing a fresh GetInfo round-trip to
@@ -139,7 +164,7 @@ type VTXOActor struct {
 
 	// autoRefreshRetryHeight is an in-memory maintenance cooldown. It is
 	// deliberately fail-safe on restart: clearing it can cause one earlier
-	// retry, but can never delay the critical unilateral-exit path.
+	// retry, but can never delay the critical path decision.
 	autoRefreshRetryHeight int32
 
 	// autoRefreshCohortLeader owns a manager-forced pending reservation.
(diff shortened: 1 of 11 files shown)


bhandras:
Fable local review: vtxo: prefer cooperative refresh when exit is infeasible

**Head reviewed:** `6b75a92c11591faab9927de721b341e09efb3bd3`
**Verdict:** READY — no blockers.

Local: `go test ./vtxo ./waved -count=1` passes.

## Verified

| Item | Result |
|---|---|
| Decision placement | `preflightCriticalExit` runs after `preflightAutoRefresh` and the cooldown check, only for `BlockEpochEvent` in `Live`/`PendingForfeit` at `ExpiryStatusCritical`. Manual exits, fraud exits, vHTLC recovery, and other states are untouched. Nil assessor, assessment error, or feasible verdict all return the original event, so the existing unilateral path is preserved unchanged. |
| `Live` + infeasible | `criticalRefreshEvent` → `autoRefreshTransition(height, false)`, the same emission as the `NeedsRefresh` path. The operator key is fetched lazily at emission when `refreshKey` is nil (`vtxo/actor.go:230-237`), so skipping the `preflightAutoRefresh` fetch at `Critical` is safe. |
| `PendingForfeit` + infeasible | Stays in state. Feasible → `handleBlockEpoch`, which is the previous escalation code moved verbatim. Past-expiry "keep waiting" behavior is unchanged. |
| Shared feasibility model | `assessUnrollFeasibility` is the same `PlanExitFunding` call manual `preflightUnrollFeasibility` uses; both go through one function so manual admission and automatic path selection cannot disagree. |
| Wiring | `s.rpcServer` is set in `runInner` (`waved/server.go:1433`) before `startWalletReadyServices` → `startWalletDependentActors` → `initVTXOManager` (`:1572`, `:2624`). The method value is bound on a non-nil receiver. |
| Regression test | `TestCriticalUnderfundedExitUsesCooperativeRefresh` pins the 187-block incident: infeasible → `PendingForfeit` with one manager relay and no resolver message; held across the next block; unilateral exit once feasible. |
| Proto | Comment-only change; generated file matches. |

## Notes (non-blocking)

1. **Per-block cost.** The assessor runs `ListWalletUnspent` + fee estimate + lineage resolution synchronously inside each critical VTXO actor's `Receive` on every block. A batch with many VTXOs going critical together means N wallet RPCs per block. Acceptable at current scale; if it becomes a problem, cache the wallet snapshot per height in the manager.
2. **Re-request churn on rollback.** If the cooperative round fails and the VTXO rolls back to `Live` while still critical and infeasible, the cooldown is cleared at `Critical` (`deferAutomaticRefresh`), so it re-requests refresh every block. Churn only, no fund risk, and arguably correct given the exit is infeasible.
3. Permanently infeasible reasons (`ExitUneconomical`, `ExitSweepBelowDust`) now correctly keep the VTXO on cooperative refresh instead of entering an exit that can never pay for itself.

## Findings
Blockers: none.
Follow-ups: none.

PR #1216, opened 2026-08-28
lwwallet: optional SQLite backend for the wallet database

guggero:
Lets `lwwallet` keep btcwallet's wallet database in SQLite instead of
BoltDB, selectable via `wallet.dbbackend` (`bolt`, the unchanged
default, or `sqlite`). Both databases live in the network directory as
before.

## Why

For a daemon embedded in a mobile application, the host application
decides which directory the platform includes in its cloud backup, and
everything in that directory should be one storage engine a backup can
capture consistently. BoltDB's mmap'd, free-list-bearing file is the odd
one out among the daemon's databases: it has to be special-cased by
whatever copies the directory, and it is the reason a daemon that
otherwise speaks only SQL still links BoltDB.

The `walletdb`-over-`database/sql` implementation this uses is not new —
`internal/sqlbase` has been backing the browser wallet store all along.
It was simply tagged `js && wasm`.

## Commits

1. `sqlbase: close the result set and check row errors in ForAll` —
   standalone bug fix. `ForAll` leaked the `*sql.Rows`, and its pooled
   connection, whenever the caller's callback failed, and reported a
   result set truncated by a row error as a fully enumerated bucket.
   Affects the browser build today.
2. `sqlbase: address linter findings in the package` — no functional
   change. The build tags kept the package out of `golangci-lint`, which
   only sees the host platform.
3. `sqlbase: drop the js && wasm build tags` — nothing in the package is
   browser-specific; the tags recorded its first consumer, not a platform
   restriction. Also ends the package's blind spot for `go build ./...`
   and the linter.
4. `lwwallet: extract the SQL wallet database opener` — no functional
   change; moves the platform-independent sqlbase wiring out of the
   browser-only file.
5. `lwwallet: add a SQLite wallet database backend` — the feature.
6. `cmd/waved: group the wallet backend flags into a registrar` — the
   wallet flags were the last group still inline in `newRootCmd`, which
   sat exactly at the `funlen` limit.
7. `waved: add wallet.dbbackend for the lwwallet backend` — config
   option, CLI flag, validation, sample config, CLI guide.

## Notes for review

- **Backend mix-ups fail loudly.** The two backends use different file
  names on purpose, and resolution fails when the other backend's file
  is present. Switching backends on an initialized directory would
  otherwise look exactly like a first start, and btcwallet would create
  a second, empty wallet beside the existing — possibly funded — one.
- **The existence probe has no side effects.** It answers from the
  database file's presence on both backends, as btcwallet's own probe for
  a local database does. Opening the database to look inside would create
  the very database the caller is asking about and hold its write lock.
- **The SQLite DSN sets `_txlock=immediate`.** The wallet reads before it
  writes inside one transaction (deriving the next address reads the
  last-used index, then bumps it), and a deferred transaction that
  upgrades to a writer after another connection committed fails with
  `SQLITE_BUSY_SNAPSHOT`, which bypasses the busy handler and so is not
  absorbed by `busy_timeout`.
- **One database file per store**, rather than sharing the daemon's
  `waved.db`. That keeps the wallet's pool the sole writer of its file,
  which is what makes the hazard above a non-issue.

## Testing

The create/open lifecycle test is parameterized over the backends the
platform offers, so the SQLite store has to satisfy the same contract as
BoltDB rather than only the parts a new test remembers to cover. New
tests cover backend-to-file resolution, the mix-up guard, and that a
SQLite wallet leaves no BoltDB file behind. Default behavior is
unchanged.

### DIFF
--- a/internal/sqlbase/readwrite_bucket.go
+++ b/internal/sqlbase/readwrite_bucket.go
@@ -1,5 +1,3 @@
-//go:build js && wasm
-
 package sqlbase
 
 import (
--- a/internal/sqlbase/readwrite_bucket.go
+++ b/internal/sqlbase/readwrite_bucket.go
@@ -473,6 +473,13 @@ func (b *readWriteBucket) ForAll(cb func(k, v []byte) error) error {
 	}
 	defer cancel()
 
+	// cancel only releases the query's timeout context; the result set
+	// holds a connection from the pool until it is closed explicitly,
+	// which rows.Next() only does for us if it is driven to completion.
+	defer func() {
+		_ = rows.Close()
+	}()
+
 	for rows.Next() {
 		var key, value []byte
 
@@ -487,5 +494,8 @@ func (b *readWriteBucket) ForAll(cb func(k, v []byte) error) error {
 		}
 	}
 
-	return nil
+	// A row error terminates the loop above just like a fully consumed
+	// result set does, so without this check a truncated bucket walk is
+	// indistinguishable from a complete one.
+	return rows.Err()
 }
--- a/internal/sqlbase/readwrite_bucket.go
+++ b/internal/sqlbase/readwrite_bucket.go
@@ -84,7 +84,7 @@ func (b *readWriteBucket) Get(key []byte) []byte {
 	err := row.Scan(&value)
 
 	switch {
-	case err == sql.ErrNoRows:
+	case errors.Is(err, sql.ErrNoRows):
 		return nil
 
 	case err != nil:
@@ -137,7 +137,7 @@ func (b *readWriteBucket) NestedReadWriteBucket(
 	err := row.Scan(&id)
 
 	switch {
-	case err == sql.ErrNoRows:
+	case errors.Is(err, sql.ErrNoRows):
 		return nil
 
 	case err != nil:
@@ -173,7 +173,7 @@ func (b *readWriteBucket) CreateBucket(key []byte) (walletdb.ReadWriteBucket,
 	err := row.Scan(&id, &value)
 
 	switch {
-	case err == sql.ErrNoRows:
+	case errors.Is(err, sql.ErrNoRows):
 	case err == nil && value == nil:
 		return nil, walletdb.ErrBucketExists
 
@@ -229,7 +229,7 @@ func (b *readWriteBucket) CreateBucketIfNotExists(key []byte) (
 	switch {
 	// Bucket does not yet exist, so create it now. Postgres will generate a
 	// bucket id for the new bucket.
-	case err == sql.ErrNoRows:
+	case errors.Is(err, sql.ErrNoRows):
 		row, cancel := b.tx.QueryRow(
 			"INSERT INTO "+b.table+" (parent_id, key) "+
 				"VALUES($1, $2) RETURNING id",
@@ -365,7 +365,7 @@ func (b *readWriteBucket) Delete(key []byte) error {
 	err := row.Scan(&dummy)
 	switch {
 	// No bucket exists, proceed to deletion of the key.
-	case err == sql.ErrNoRows:
+	case errors.Is(err, sql.ErrNoRows):
 	case err != nil:
 		return err
 
@@ -447,7 +447,7 @@ func (b *readWriteBucket) Sequence() uint64 {
 	err := row.Scan(&seq)
 
 	switch {
-	case err == sql.ErrNoRows:
+	case errors.Is(err, sql.ErrNoRows):
 		return 0
 
 	case err != nil:
--- a/lwwallet/walletdb.go
+++ b/lwwallet/walletdb.go
@@ -9,6 +9,25 @@ import (
 )
 
 const (
+	// DBBackendBolt keeps btcwallet's wallet database in the classic
+	// BoltDB file under Config.DBDir. It is the default.
+	DBBackendBolt = "bolt"
+
+	// DBBackendSQLite keeps btcwallet's wallet database in a SQLite
+	// database under Config.DBDir. It lets an embedded daemon hold
+	// all of its state in one storage engine, in a directory the host
+	// application picks, rather than an mmap'd BoltDB file alongside
+	// the SQLite databases every other store already uses.
+	//
+	// Note that this gives up one property BoltDB provides: its file
+	// lock excludes a second OS process, while SQLite in WAL mode
+	// does not. Two daemons on one directory stay safe at the storage
+	// layer but not at the wallet layer, where both would derive the
+	// same addresses and keep their own idea of sync state.
+	DBBackendSQLite = "sqlite"
+)
+
+const (
 	// sqlWalletDBTablePrefix namespaces btcwallet's key/value table
 	// inside the wallet database, so the resulting "walletdb_kv"
 	// table can coexist with tables owned by other stores.
--- a/lwwallet/walletdb.go
+++ b/lwwallet/walletdb.go
@@ -42,9 +42,12 @@ const (
 func openSQLWalletDB(ctx context.Context, driverName,
 	dsn string) (walletdb.DB, error) {
 
-	// The connection set is process-global and initialized exactly
-	// once; a later call with a different limit is a no-op.
-	sqlbase.Init(sqlWalletDBMaxConnections)
+	// The connection set is process-global and its limit is fixed by
+	// the first caller, so this fails rather than hand back a pool
+	// wider than the wallet store's single-writer assumption.
+	if err := sqlbase.Init(sqlWalletDBMaxConnections); err != nil {
+		return nil, err
+	}
 
 	return sqlbase.NewSqlBackend(ctx, &sqlbase.Config{
 		DriverName:      driverName,
--- a/lwwallet/walletdb.go
+++ b/lwwallet/walletdb.go
@@ -1,0 +1,56 @@
+package lwwallet
+
+import (
+	"context"
+	"time"
+
+	"github.com/btcsuite/btcwallet/walletdb"
+	"github.com/lightninglabs/wavelength/internal/sqlbase"
+)
+
+const (
+	// sqlWalletDBTablePrefix namespaces btcwallet's key/value table
+	// inside the wallet database, so the resulting "walletdb_kv"
+	// table can coexist with tables owned by other stores.
+	sqlWalletDBTablePrefix = "walletdb"
+
+	// sqlWalletDBBusyTimeout is how long SQLite waits for a lock held
+	// by another connection to the same database before failing the
+	// query.
+	sqlWalletDBBusyTimeout = 30 * time.Second
+
+	// sqlWalletDBTimeout is the per-query timeout applied to the
+	// wallet database. It has to stay comfortably above
+	// sqlWalletDBBusyTimeout: a query that legitimately waits out the
+	// busy handler would otherwise race its own context deadline, and
+	// the caller would see "context deadline exceeded" instead of the
+	// SQLITE_BUSY that names the actual problem.
+	sqlWalletDBTimeout = 2 * sqlWalletDBBusyTimeout
+
+	// sqlWalletDBMaxConnections bounds the connection pool sqlbase
+	// keeps per DSN. The walletdb emulation funnels its read-write
+	// transactions through a single in-process lock anyway, so a
+	// second connection would only add contention on the database's
+	// own write lock.
+	sqlWalletDBMaxConnections = 1
+)
+
+// openSQLWalletDB opens, creating it if needed, btcwallet's wallet
+// database on the given database/sql driver and DSN. The driver must
+// already be registered by the caller. The returned handle is owned by
+// the caller and must be closed once the wallet has stopped.
+func openSQLWalletDB(ctx context.Context, driverName,
+	dsn string) (walletdb.DB, error) {
+
+	// The connection set is process-global and initialized exactly
+	// once; a later call with a different limit is a no-op.
+	sqlbase.Init(sqlWalletDBMaxConnections)
+
+	return sqlbase.NewSqlBackend(ctx, &sqlbase.Config{
+		DriverName:      driverName,
+		Dsn:             dsn,
+		Timeout:         sqlWalletDBTimeout,
+		TableNamePrefix: sqlWalletDBTablePrefix,
+		WithTxLevelLock: true,
+	})
+}
--- a/lwwallet/walletdb_native.go
+++ b/lwwallet/walletdb_native.go
@@ -3,33 +3,185 @@
 package lwwallet
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
 	"time"
 
+	btcwalletbase "github.com/btcsuite/btcwallet/wallet"
 	"github.com/lightningnetwork/lnd/lnwallet/btcwallet"
+	_ "modernc.org/sqlite" // Register the native SQLite driver.
 )
 
-// newWalletLoaderOptions returns the native btcwallet bbolt loader
-// options. The cleanup func is a no-op: the local database is opened
-// (and closed on failure) by btcwallet's own loader.
+const (
+	// boltWalletDBTimeout is how long the BoltDB backend waits for
+	// the database file lock before giving up.
+	boltWalletDBTimeout = 60 * time.Second
+
+	// sqliteWalletDBDriverName is the database/sql driver registered
+	// by modernc.org/sqlite.
+	sqliteWalletDBDriverName = "sqlite"
+
+	// sqliteWalletDBFileName is the SQLite wallet database's file
+	// name under Config.DBDir. It deliberately differs from
+	// btcwallet's BoltDB file name, so that neither backend can ever
+	// be pointed at a database written by the other.
+	sqliteWalletDBFileName = "wallet.sqlite.db"
+)
+
+// walletDBPath returns the path of the wallet database file the
+// configured backend uses, and fails on an unknown backend so a typo
+// cannot silently resolve to the default.
+//
+// Resolution also fails when the other backend's database file already
+// exists in DBDir. The two backends store the wallet under different
+// names, so without this check switching the backend of an initialized
+// wallet directory looks exactly like a first start: btcwallet would
+// happily create a second, empty wallet next to the existing — and
+// possibly funded — one.
+func walletDBPath(cfg Config) (string, error) {
+	var backend, other string
+	switch cfg.DBBackend {
+	case "", DBBackendBolt:
+		backend, other = DBBackendBolt, DBBackendSQLite
+
+	case DBBackendSQLite:
+		backend, other = DBBackendSQLite, DBBackendBolt
+
+	default:
+		return "", fmt.Errorf("unknown wallet database backend %q, "+
+			"must be %q or %q", cfg.DBBackend, DBBackendBolt,
+			DBBackendSQLite)
+	}
+
+	otherPath := walletDBFilePath(cfg.DBDir, other)
+	otherExists, err := walletDBExists(otherPath)
+	if err != nil {
+		return "", err
+	}
+	if otherExists {
+		return "", fmt.Errorf("wallet database %s exists but the %q "+
+			"database backend is selected; set the backend to %q",
+			otherPath, backend, other)
+	}
+
+	return walletDBFilePath(cfg.DBDir, backend), nil
+}
+
+// walletDBFilePath returns the file the given backend keeps the wallet
+// database in.
+func walletDBFilePath(dbDir, backend string) string {
+	if backend == DBBackendSQLite {
+		return filepath.Join(dbDir, sqliteWalletDBFileName)
+	}
+
+	return filepath.Join(dbDir, btcwalletbase.WalletDBName)
+}
+
+// newWalletLoaderOptions returns the btcwallet loader options for the
+// configured wallet database backend, plus a func that releases
+// whatever the resolution opened.
 func newWalletLoaderOptions(cfg Config) ([]btcwallet.LoaderOption, func(),
 	error) {
 
+	// Resolving the path is also what validates the backend and
+	// rejects a directory that already holds the other backend's
+	// database.
+	dbPath, err := walletDBPath(cfg)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// BoltDB is opened (and closed on failure) by btcwallet's own
+	// loader, so there is nothing for the cleanup func to release.
+	if cfg.DBBackend != DBBackendSQLite {
+		return []btcwallet.LoaderOption{
+			btcwallet.LoaderWithLocalWalletDB(
+				cfg.DBDir, false, boltWalletDBTimeout,
+			),
+		}, func() {}, nil
+	}
+
+	if err := os.MkdirAll(cfg.DBDir, 0700); err != nil {
+		return nil, nil, fmt.Errorf("create wallet database dir: %w",
+			err)
+	}
+
+	// Create the database file ourselves before SQLite gets a chance
+	// to. SQLite would create it with 0666 & ~umask, which under the
+	// common umask 0022 leaves the encrypted seed, the key state and
+	// the full transaction history world-readable; btcwallet passes
+	// 0600 explicitly for its BoltDB file and the two backends should
+	// not differ here. Doing it up front rather than chmod'ing
+	// afterwards also covers the -wal and -shm sidecars, which SQLite
+	// creates with the mode of the main database file.
+	if err := createWalletDBFile(dbPath); err != nil {
+		return nil, nil, err
+	}
+
+	db, err := openSQLWalletDB(
+		context.Background(), sqliteWalletDBDriverName,
+		sqliteWalletDBDSN(dbPath),
+	)
+	if err != nil {
+		return nil, nil, fmt.Errorf("open SQLite wallet database: %w",
+			err)
+	}
+
+	// btcwallet's loader only closes a wallet database it created
+	// itself, so a constructor failure after this point would leak
+	// this handle, and its lock on the database file, for the process
+	// lifetime.
+	cleanup := func() {
+		_ = db.Close()
+	}
+
 	return []btcwallet.LoaderOption{
-		btcwallet.LoaderWithLocalWalletDB(
-			cfg.DBDir, false, 60*time.Second,
-		),
-	}, func() {}, nil
+		btcwallet.LoaderWithExternalWalletDB(db),
+	}, cleanup, nil
 }
 
-// walletExists reports whether a btcwallet bbolt database already
-// exists in the configured directory. For local databases the loader
-// only checks file existence, so this probe does not take the bbolt
-// file lock.
+// walletExists reports whether a wallet has already been created in
+// the configured backend's database.
+//
+// The predicate has to be the same one btcwallet's loader will apply,
+// because New pairs them: it refuses a seed when a wallet exists and
+// lets a seedless open through when one does. btcwallet answers from a
+// marker key committed inside the database once the wallet itself
+// exists, and for the SQL backends that is not the same as the database
+// file being there — openSQLWalletDB creates the file and its schema
+// before btcwallet ever runs, so an interrupted first run leaves an
+// empty database behind. Answering "exists" for that state would refuse
+// the operator's real seed and send the seedless path into
+// btcwallet.New, which creates a wallet from a random seed when handed
+// no seed of its own.
 func walletExists(cfg Config) (bool, error) {
-	opts, _, err := newWalletLoaderOptions(cfg)
+	dbPath, err := walletDBPath(cfg)
 	if err != nil {
 		return false, err
 	}
+
+	// A missing database file settles the question without opening
+	// anything, which is what keeps a probe of a fresh directory from
+	// creating the very database it is asking about.
+	exists, err := walletDBExists(dbPath)
+	if err != nil || !exists {
+		return false, err
+	}
+
+	opts, cleanup, err := newWalletLoaderOptions(cfg)
+	if err != nil {
+		return false, err
+	}
+
+	// A probe must not hand the wallet database on to anyone, so a
+	// handle opened above is closed again here. It would otherwise
+	// hold the database's write lock until the process exits, and the
+	// real open in New would be the one to fail on it.
+	defer cleanup()
 
 	loader, err := btcwallet.NewWalletLoader(
 		cfg.ChainParams, cfg.RecoveryWindow, opts...,
@@ -39,4 +191,100 @@ func walletExists(cfg Config) (bool, error) {
 	}
 
 	return loader.WalletExists()
+}
+
+// sqliteWalletDBDSN returns the modernc.org/sqlite DSN for the wallet
+// database at the given path.
+func sqliteWalletDBDSN(dbPath string) string {
+	pragmas := make(url.Values)
+	for _, pragma := range []string{
+		// The key/value schema nests buckets through a parent_id
+		// self-reference, and relies on the cascade to drop a
+		// deleted bucket's children instead of leaving them
+		// behind as unreachable rows.
+		"foreign_keys=on",
+
+		// WAL lets the wallet's reads proceed against the last
+		// committed snapshot while a write transaction is open,
+		// and keeps a crash from tearing a commit.
+		"journal_mode=WAL",
+
+		// The wallet's seed and key state have no second copy
+		// anywhere, so this database does not follow the
+		// daemon's configurable synchronous=normal default. FULL
+		// is also SQLite's own default; it is set explicitly so
+		// the guarantee does not rest on a driver default.
+		"synchronous=full",
+
+		// fullfsync only matters on Darwin, where a plain fsync
+		// does not reach stable storage. The daemon's own SQLite
+		// store enables it for exactly that reason, and an
+		// embedded daemon on a mobile platform is squarely in
+		// that case, so the wallet database must not end up with
+		// weaker flush semantics than the databases beside it.
+		"fullfsync=true",
+	} {
+		pragmas.Add("_pragma", pragma)
+	}
+
+	// busy_timeout caps how long SQLite waits for a lock held by
+	// another connection to the same file, such as a backup tool,
+	// before failing the query.
+	pragmas.Add(
+		"_pragma",
+		fmt.Sprintf(
+			"busy_timeout=%d",
+			sqlWalletDBBusyTimeout.Milliseconds(),
+		),
+	)
+
+	// Take the write lock when a write transaction begins rather than
+	// on its first write statement. The wallet reads before it writes
+	// within a single transaction — deriving the next address reads
+	// the last-used index and then bumps it — and a deferred
+	// transaction that tries to upgrade to a writer after another
+	// connection committed in the meantime fails with
+	// SQLITE_BUSY_SNAPSHOT. That error bypasses the busy handler, so
+	// busy_timeout above would not absorb it.
+	return fmt.Sprintf("%s?%s&_txlock=immediate", dbPath, pragmas.Encode())
+}
+
+// createWalletDBFile creates the wallet database file with owner-only
+// permissions if it does not exist yet. An existing file is left alone,
+// including its mode: a database created by an older version stays
+// readable rather than being silently tightened underneath a caller
+// that may have chosen its own permissions.
+func createWalletDBFile(dbPath string) error {
+	// The path is the daemon's own wallet database under the
+	// configured data directory, not caller-supplied input.
+	//
+	//nolint:gosec // G304: path is composed from DBDir and a constant
+	f, err := os.OpenFile(dbPath, os.O_CREATE|os.O_WRONLY, 0600)
+	if err != nil {
+		return fmt.Errorf("create wallet database file: %w", err)
+	}
+
+	if err := f.Close(); err != nil {
+		return fmt.Errorf("close wallet database file: %w", err)
+	}
+
+	return nil
+}
+
+// walletDBExists reports whether a wallet database exists at the given
+// path. An unexpected stat failure is returned as an error rather than
+// as a missing database: reading, say, a permission problem as "no
+// wallet here" would send the caller down the create path.
+func walletDBExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	switch {
+	case err == nil:
+		return true, nil
+
+	case errors.Is(err, os.ErrNotExist):
+		return false, nil
+
+	default:
+		return false, err
+	}
 }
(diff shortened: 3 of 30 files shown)


Roasbeef:
This makes a ton of sense! Forgot we have the existing kvdb sqlite version that can slot into walletdb. 

Shooting to get this into the next release.

lwwallet/walletdb_native.go
```diff
@@ -3,40 +3,203 @@
...
+			),
+		}, func() {}, nil
+	}
+
+	if err := os.MkdirAll(cfg.DBDir, 0700); err != nil {
+		return nil, nil, fmt.Errorf("create wallet database dir: %w",
+			err)
+	}
+
+	db, err := openSQLWalletDB(
```
litbot-9000, line 125: **This is my blocking concern.** `openSQLWalletDB` creates the database file (and the `walletdb_kv` schema) *before* `btcwallet.New` runs. But `walletExists` above answers from file presence, while btcwallet's external-DB loader answers from a key inside the database:

```go
// lnd/lnwallet/btcwallet/btcwallet.go
func externalWalletExists(db kvdb.Backend) (bool, error) {
	metaBucket := tx.ReadBucket([]byte(walletMetaBucket))
	...  exists = string(walletReady) == walletReadyKey
}
```

and `btcwallet.New` does `if !walletExists { loader.CreateNewWallet(pubPass, privPass, cfg.HdSeed, ...) }` — with `HdSeed == nil`, btcwallet **generates a random seed**.

So the two predicates disagree on exactly one state — *file present, wallet not inside* — and in that state lwwallet's own invariant check refuses the seed and permits the seedless open, which is the worst possible pairing.

I reproduced it. Creating the DB with this same helper and closing it (i.e. the on-disk state after a kill between here and `btcwallet.New` committing `onWalletCreated`), then calling `New` with `Seed: nil`:

```
PROBE after crash-mid-create: exists=true err=<nil>
OPEN (no seed) => <nil>                      // succeeds!
REAL SEED   identity key: 030e19c748ce39506ef40b6bca0be4f984bd61a4b4e0465b2041179b4c87d48279
HALF-CREATED, no seed:     022ec0a3db3ebd0c87da0c84165cbd7398a583547765c517c11f698f74a060be5b
retry with the REAL seed => wallet database already exists in "...": open it without a seed instead
```

Different chain — a wallet was created from a seed nobody has. And the operator who *does* have the right seed can no longer use it; the directory has to be deleted by hand.

The same run against `bolt` fails loudly instead, because the local loader treats file presence as wallet presence and goes to `OpenExistingWallet`:

```
BOLT OPEN (no seed) => create btcwallet: missing address manager namespace
```

Windows that produce this state: the process dies between this call and `onWalletCreated` committing (an OS-killed app mid-first-run is the headline scenario for this PR); or `btcwallet.New` fails for any reason after the file was created and the operator restarts.

The probe itself is fine as-is — I agree it should not open the database. The gap is that nothing re-checks once the handle *is* open. The cheapest fix I see keeps your no-side-effect probe intact: right here, after `openSQLWalletDB` succeeds, when the caller is on the open path (`cfg.Seed == nil`) consult the same marker btcwallet will consult and return `ErrWalletNotFound` when it is absent — the DB is already open at that point, so it costs one read transaction and no new side effects. Alternatively, create the file under a temporary name and rename it into place only after the wallet-ready key is committed, so file presence and wallet presence can never disagree.

Whatever the shape, a regression test for "database file exists, wallet-ready marker absent" would be worth having, since it is the one state the new `TestWalletDBPath` table does not cover.

lwwallet/walletdb.go
```diff
@@ -0,0 +1,60 @@
...
+	// sqlWalletDBTimeout is the per-query timeout applied to the
+	// wallet database.
+	sqlWalletDBTimeout = 30 * time.Second
+
+	// sqlWalletDBMaxConnections bounds the connection pool sqlbase
+	// keeps per DSN. The walletdb emulation funnels its read-write
+	// transactions through a single in-process lock anyway, so a
+	// second connection would only add contention on the database's
+	// own write lock.
+	sqlWalletDBMaxConnections = 1
```
litbot-9000, line 54: `SetMaxOpenConns(1)` does hold — `sqlbase.Init(1)` is the only `Init` call in the tree, `dbConnSet.Open` applies it, and with `WithTxLevelLock` the RWMutex serializes writers. So *within one process* the single-writer assumption is sound. Worth noting the guarantee is order-dependent though: `Init` is process-global and first-call-wins (the comment in `openSQLWalletDB` acknowledges this), so a future native `sqlbase` consumer initializing with a larger limit first would silently widen the wallet's pool. A `panic`/error on a conflicting re-`Init` would make that non-silent.

The part that isn't in-process, and that I think deserves a conscious decision: **BoltDB's flock excludes a second OS process; SQLite in WAL mode with `locking_mode=normal` does not.** I ran both, two real processes, same file, this exact DSN and config:

```
A: OPEN OK          B: OPEN OK             # sqlbase + modernc sqlite
A: BOLT OPEN OK     B: BOLT OPEN FAILED: timeout   # bbolt, 5s timeout
```

That second line is what `boltWalletDBTimeout = 60 * time.Second` exists for. Today, a stray second `waved` on the same network directory cannot get past it; with `dbbackend=sqlite` it starts cleanly. I grepped `waved/` and `cmd/` and found no pidfile or other process lock that would still catch it.

SQLite will keep the *file* consistent — no corruption, and `_txlock=immediate` plus `busy_timeout` will serialize the two writers correctly. The wallet semantics on top are what break: two daemons independently deriving from the same account will hand out the same address twice, both bump the same `NextSequence`, and both maintain their own idea of sync state.

I don't think this has to block the PR, but the mutual exclusion is a real property being traded away silently. Either take an advisory lock on the directory when the SQLite backend is selected, or say plainly in the docs that `dbbackend=sqlite` gives up the "only one daemon can own this directory" guarantee.

lwwallet/walletdb_native.go
```diff
@@ -3,40 +3,203 @@
...
+	// loader, so there is nothing for the cleanup func to release.
+	if cfg.DBBackend != DBBackendSQLite {
+		return []btcwallet.LoaderOption{
+			btcwallet.LoaderWithLocalWalletDB(
+				cfg.DBDir, false, boltWalletDBTimeout,
+			),
+		}, func() {}, nil
+	}
+
+	if err := os.MkdirAll(cfg.DBDir, 0700); err != nil {
```
litbot-9000, line 108: The directory is correctly `0700`, but the database file inside it is not. SQLite creates it with `0666 & ~umask`, so under the common `umask 0022` the wallet database — encrypted seed, key state, full transaction history — lands world-readable. Measured, same test, same machine:

```
SQLITE  wallet.sqlite.db  mode=-rw-r--r--
BOLT    wallet.db         mode=-rw-------
```

btcwallet passes `0600` explicitly for the bolt file; nothing does the equivalent here. On a phone the sandbox hides it, but on a server or a shared-container deployment it is a straightforward information leak, and it is a silent difference between the two backends rather than a documented one.

A `os.Chmod(dbPath, 0600)` after `openSQLWalletDB` returns would cover it (also worth applying to the `-wal`/`-shm` sidecars, which SQLite creates with the same permissions as the main file — those inherit the main file's mode, so fixing the mode before the first write is the simpler ordering).

lwwallet/walletdb_native.go
```diff
@@ -3,40 +3,203 @@
...
+		// The key/value schema nests buckets through a parent_id
+		// self-reference, and relies on the cascade to drop a
+		// deleted bucket's children instead of leaving them
+		// behind as unreachable rows.
+		"foreign_keys=on",
+
+		// WAL lets the wallet's reads proceed against the last
+		// committed snapshot while a write transaction is open,
+		// and keeps a crash from tearing a commit.
+		"journal_mode=WAL",
```
litbot-9000, line 210: Two things about this pragma set, both concrete rather than stylistic.

**1. Missing durability pragmas relative to the daemon's own SQLite store.** I dumped the effective pragmas from the DSN this function actually builds:

```
journal_mode = wal   synchronous = 2 (FULL)   fullfsync = 0
foreign_keys = 1     busy_timeout = 30000     auto_vacuum = 0
```

`synchronous` lands on FULL via SQLite's own default, so durability is fine there — but `fullfsync=0` is not what the rest of the daemon does. `lnd/sqldb`'s `NewSqliteStore`, which backs `waved`'s other databases, sets `fullfsync=true` with the comment *"used to ensure proper durability for users running on Mac OS; it uses the correct fsync system call to ensure items are fully flushed to disk"*, and `waved` even exposes `db.sqlite.nofullfsync` as an operator knob for it. Given this PR's stated motivation is an embedded daemon on a mobile platform — where the Apple half is exactly where plain `fsync()` does not reach stable storage — the wallet database ending up with *weaker* fsync semantics than every other database in the same directory looks unintentional. `auto_vacuum=incremental` is also set there and not here; that one is cosmetic by comparison.

**2. The backup premise.** I want to push on this gently, because I think the PR is mostly right and the description slightly overstates it. WAL means the running daemon's directory holds three files, which I confirmed:

```
WHILE RUNNING: wallet.sqlite.db, wallet.sqlite.db-shm, wallet.sqlite.db-wal
AFTER STOP:    wallet.sqlite.db   (sidecars cleaned up on clean close)
```

So a naive directory copy taken while the daemon is running still needs to special-case something: it must include `-wal`, must not treat `-shm` as meaningful, and a copy of the main file *without* the matching `-wal` is a stale or unusable wallet. That is not obviously better than special-casing a bolt file.

Where the argument does hold, and I think this is the honest framing: `waved`'s other databases are *already* WAL SQLite, so the backup tool already carries the SQLite+WAL special case. This PR takes the directory from two engines to one rather than from one special case to zero. Worth stating that way in the commit message and the sample config, because "a backup can copy it with a single storage engine's guarantees" reads as though a plain `cp -r` becomes safe, and it does not.

lwwallet/walletdb_native.go
```diff
@@ -3,40 +3,203 @@
...
-	return loader.WalletExists()
+	// Take the write lock when a write transaction begins rather than
+	// on its first write statement. The wallet reads before it writes
+	// within a single transaction — deriving the next address reads
+	// the last-used index and then bumps it — and a deferred
+	// transaction that tries to upgrade to a writer after another
+	// connection committed in the meantime fails with
+	// SQLITE_BUSY_SNAPSHOT. That error bypasses the busy handler, so
+	// busy_timeout above would not absorb it.
+	return fmt.Sprintf("%s?%s&_txlock=immediate", dbPath, pragmas.Encode())
```
litbot-9000, line 249: Checked this reasoning against the pinned driver rather than taking it on trust, and it holds:

- `modernc.org/sqlite@v1.40.0` does support `_txlock` (`sqlite.go:963`, validated against `deferred|immediate|exclusive`), so a typo here would fail loudly rather than silently degrade to deferred.
- It is applied where it matters and only there: `newTx` does `sql := "begin"; if !opts.ReadOnly && c.beginMode != "" { sql = "begin " + c.beginMode }`, and `sqlbase.newReadWriteTx` passes `ReadOnly: readOnly` through. Write transactions get `BEGIN IMMEDIATE`, read transactions stay deferred. That is exactly the right split.
- `url.Values.Encode()` percent-encodes the `=` inside each `_pragma` value, and the driver round-trips it through `url.ParseQuery`, so the pragmas do arrive intact — verified by reading them back.
- And the `SQLITE_BUSY_SNAPSHOT`-bypasses-the-busy-handler premise is right; this is the same reasoning `lnd/sqldb` uses for its own `_txlock=immediate`.

One small mismatch while you are here: `busy_timeout` is 30s and `sqlWalletDBTimeout` is also 30s. A query that genuinely waits out the busy handler will race its own context deadline, so the operator may see `context deadline exceeded` where `SQLITE_BUSY` would have named the actual problem. Making the query timeout comfortably larger than the busy timeout would keep the busy handler's error the one that surfaces.

internal/sqlbase/readwrite_bucket.go
```diff
@@ -487,5 +492,8 @@ func (b *readWriteBucket) ForAll(cb func(k, v []byte) error) error {
 		}
 	}
 
-	return nil
+	// A row error terminates the loop above just like a fully consumed
+	// result set does, so without this check a truncated bucket walk is
+	// indistinguishable from a complete one.
+	return rows.Err()
```
litbot-9000, line 498: Fix is correct, including the bit that is easy to get wrong: `defer rows.Close()` is registered *after* `defer cancel()`, so LIFO runs the close before the context cancel rather than the other way round. It also covers the panic path, which matters here because `Sequence()` and the cursor methods `panic(err)` and `db.catchPanic` only recovers at the transaction boundary — the connection would otherwise be lost on a panicking callback too, not just a failing one. `rows.Err()` after the loop is the right check.

I went looking for siblings and did not find one, which I'll state explicitly since it is a negative result: `ForAll` is the only multi-row `Query` call site in the package. Every other read goes through `readWriteTx.QueryRow`, whose `*sql.Row.Scan` closes the underlying rows itself, and each of those call sites does `defer cancel()`. The two reassigned-`cancel` spots in `CreateBucket` (lines 170/193) are fine — `defer cancel()` captures the function value at statement time, so both cancels run. The cursor is clean too: `First/Last/Next/Prev/Seek/Delete` are all single-row.

Worth noting: `internal/sqlbase` still has **no test files of its own** (`go test ./internal/sqlbase/...` → `[no test files]`), so this fix ships without a regression test and the package's only coverage is indirect, through `lwwallet`'s lifecycle test. That was defensible while it was one browser store behind a build tag; now that it is about to hold a native funds-bearing wallet, a direct test for the bucket/cursor/nested-bucket contract would be worth the commit.

Separately — the commit message says this affects the browser build today. Is a backport to released browser builds needed, or is the aborted-`ForAll` path not reachable there in practice? Worth a sentence in the PR either way.

sample-waved.conf
```diff
@@ -192,6 +192,13 @@
 # Address recovery look-ahead window for lwwallet.
 # wallet.recoverywindow=100
 
+# Database backend for the lwwallet wallet database: bolt or sqlite. Both keep
+# the database in the network data directory. Pick sqlite to keep the directory
+# free of a BoltDB file, for instance so a backup can copy it with a single
+# storage engine's guarantees. The backend cannot be changed once the wallet
```
litbot-9000: Confirmed there is no bolt→sqlite migration anywhere in the tree, and that flipping this on an initialized directory fails with a clear message naming the file it found — which is the right behaviour. The docs mostly say so, and `docs/daemon_cli_guide.md` gets it exactly right with "fixed once the wallet is created".

Two wording suggestions so a reader can't get to the loud failure before the warning:

- "The backend cannot be changed once the wallet has been created" is the single most important sentence in this block and it is currently the last line. Consider leading with it, and saying what "cannot" means operationally: the daemon refuses to start, and the only way across is a fresh directory restored from the seed. A reader who sees `bolt or sqlite` first may well flip it on a funded wallet and only then discover it is one-way.
- "so a backup can copy it with a single storage engine's guarantees" — see my note on the DSN; with WAL enabled a running daemon leaves `-wal`/`-shm` beside the main file, so a plain directory copy is still not automatically consistent. "so the directory holds one storage engine instead of two" would be accurate without implying `cp -r` becomes safe.

The validation in `waved/config.go` is solid, by the way — I checked all four paths and the default really is unchanged: `DefaultConfig()` sets `DBBackendBolt`, the CLI flag defaults off that same value, an absent config key leaves the default in place, the empty string is accepted and resolves to bolt in `walletDBPath`, and an unknown value is rejected at validation time rather than falling back. `TestConfigValidateWalletDBBackend` passes on all four.

litbot-9000 (changes requested):
Really nicely structured series — the commit split is clean, commit 1 is a genuine standalone fix, and the mix-up guard's state table holds up under every combination I could construct (neither file, either file, both files, zero-length, a *directory* wearing the database's name). The probe is genuinely side-effect free on both backends; I confirmed the directory is still empty after `WalletExists` on a fresh dir.

Requesting changes on one thing, which I reproduced rather than reasoned about: **on the SQLite backend, a database file that exists but has no wallet inside it causes btcwallet to silently create a wallet from a random seed.** `walletExists` answers from file presence, but btcwallet's *external*-DB loader answers from the `walletmeta/walletready` key inside the database — and `newWalletLoaderOptions` creates the file before btcwallet ever runs. Any crash or error in that window (an OS-killed mobile app during first-run creation is precisely the deployment this PR targets) leaves a directory where the seeded create path is refused and the seedless open path mints an unbacked-up wallet. BoltDB has no such hole: I checked, an empty `wallet.db` fails loudly with `missing address manager namespace`. Detail and repro inline.

Second, smaller but worth deciding consciously: the SQLite backend drops the single-daemon-per-directory guarantee that bbolt's flock provides today. Two `waved` processes on one wallet directory is currently impossible and would become possible; there is no other process lock in `waved` that I could find.

Also inline: the created database file is `0644` where bolt's is `0600`, and the DSN omits the `synchronous`/`fullfsync` pragmas that the daemon's own SQLite store sets explicitly for Darwin durability.

**What I ran** (worktree on `pr-1216`, Linux/amd64, Go toolchain in-container): `go build ./...` (the one failure, `function main is undeclared in the main package`, reproduces identically on `origin/main` — pre-existing, root package is wasm-only); `go test ./internal/sqlbase/... ./lwwallet/...` → `ok lwwallet 20.6s`, sqlbase reports **no test files**; `go test ./waved/ -run TestConfigValidateWalletDBBackend -v` → all 4 subtests pass; `GOOS=js GOARCH=wasm go build ./lwwallet/... ./internal/sqlbase/...` → clean, so dropping the tags did not break the browser build's compile. I also read `modernc.org/sqlite@v1.40.0`'s DSN parser directly to check the `_txlock` reasoning, and dumped the effective pragmas from the DSN this code actually constructs.

**What I did NOT verify**: `golangci-lint` is not installed here and `make lint` wants Docker, so commit 2's linter claims are unchecked (I did diff-read it; it looks behaviour-preserving); the wasm build compiles but I did not *run* the browser wallet store; and I did not test real iOS/Android backup behaviour, only the on-disk file set.

— claudell ⚡ (not a maintainer; confidence is high on the reproduced items, lower on the pragma/backup judgement calls, which are opinions rather than defects)

lwwallet/walletdb_native.go
guggero: Reproduced, and you are right. Thanks for chasing it down to the two disagreeing
predicates — that is exactly the shape of it.

Worth saying where it came from: the probe answered from the marker in an earlier
revision of this branch, and I changed it to file presence to get `contextcheck`
off my back, without noticing that it decoupled the two predicates. So this is a
hole I dug, not one inherited from the browser store.

I went with a different shape than the one you suggested, because I think the
`cfg.Seed == nil` check fixes only half of it. It does stop the random-seed
wallet, but the operator holding the real seed stays locked out: the invariant
check has already answered "exists" from file presence, so the seeded path is
still refused before anything gets as far as the marker. My repro showed both
halves:

```
OPEN (no seed) => <nil>                                   # minted from a random seed
RETRY with the real seed => wallet database already exists ... open it without a seed instead
```

So instead of reconciling after the fact, I moved the predicate itself:
`walletExists` now shortcuts on a *missing* database file and otherwise opens the
database to read btcwallet's marker. The two predicates cannot disagree any more,
because they are the same one. Both halves fall out of that:

```
OPEN (no seed) => no wallet database found ...: create the wallet with a seed first
RETRY with the real seed => <nil>          # creates the wallet in place
```

The second line is the part I care about most: the directory heals itself, no
manual deletion.

The property you verified is preserved — the probe still never creates the
database it is asking about, because a missing file is answered without opening
anything. `TestSQLiteWalletDBLayout` asserts the directory is still empty after
`WalletExists` on a fresh one.

I did not take the temp-name-and-rename option. It would keep file presence and
wallet presence in step going forward, but it needs a hook on `onWalletCreated`
to know when to rename, and it does nothing for a directory that is already in
the bad state.

Regression test added as `TestSQLiteWalletDBInterruptedCreate`: it builds the
exact on-disk state (open the database through the same helper, close it without
creating a wallet), then asserts the probe reports no wallet, the seedless open
fails with `ErrWalletNotFound`, and the seeded create succeeds in place.

lwwallet/walletdb.go
guggero: **`Init`**: done, in its own commit (`sqlbase: refuse a conflicting
connection-set limit`). It returns an error now; repeating the same limit stays a
no-op, so callers can keep stating the limit they need, and only a genuine
conflict is reported. An error rather than a panic, since this is a library. You
are right that it is about the next consumer rather than the current one — that
is in the commit message.

**Mutual exclusion**: confirmed, and not enforced. It is documented instead, on
the `DBBackendSQLite` constant, in `sample-waved.conf` and in the package
AGENTS.md, in your terms: storage stays consistent, wallet semantics do not,
because both daemons would derive the same addresses.

Reasoning for documenting rather than locking, and I am happy to be overruled:

- A directory lock looks like the daemon's job rather than this backend's. Bolt
  gets it incidentally from `flock`; nothing in `waved` asks for it, which is
  why there is no pidfile for you to find. Making it explicit would cover both
  backends and reads better as its own change than as a rider on this one.
- `syscall.Flock` needs its own build-tagged files to stay portable, which is a
  lot of surface for a guarantee this deployment does not rely on — the embedded
  host is single-process by construction.
- I did consider `locking_mode=EXCLUSIVE`, which the browser store already uses
  and which would restore exclusion through SQLite itself. I decided against it:
  it locks out concurrent readers, including whatever copies the directory, and
  it comes with the pragma-ordering hazard `walletdb_wasm.go` carries that long
  warning about — get the order wrong and the wallet silently lands on a
  rollback journal instead of failing.

Say the word if you would rather have the lock in this PR.

lwwallet/walletdb_native.go
guggero: **`fullfsync`**: added, and `synchronous=full` is now set explicitly rather than
inherited from SQLite's default, so the guarantee does not rest on one. Dumped
the effective pragmas from the DSN the code builds, before and after:

```
before:  synchronous = 2 (FULL)   fullfsync = 0
after:   synchronous = 2 (FULL)   fullfsync = 1
```

**`auto_vacuum`**: not adopted, and I would rather leave it out than match for
symmetry's sake. `incremental` does not reclaim anything on its own — it needs
someone to call `PRAGMA incremental_vacuum`, and nothing here does. Setting it
would be inert while reading as though free-page reclamation were handled. Happy
to add it if you would rather the two DSNs match verbatim.

**The backup premise**: you are right and I have taken your framing. The commit
message now says the gain is that the directory goes from two storage engines to
one, "not from one special case to zero", and spells out that the other
databases are already WAL SQLite so whatever copies the directory already
carries that case. The sample config drops "with a single storage engine's
guarantees" for "so the directory holds one storage engine instead of two", and
adds the sidecars explicitly:

> Note that a running daemon leaves -wal and -shm files beside each SQLite
> database, so a copy taken while it runs has to include them

internal/sqlbase/readwrite_bucket.go
guggero: Two things here, and the first is a correction to my own commit message.

**The connection leak does not happen.** I wrote the test for it — abort `ForAll`
from the callback repeatedly on a one-connection pool, then run a normal read and
require it to complete — and it passes *without* the fix. `ForAll` does leave the
`*sql.Rows` open on the early return, but the transaction is rolled back
immediately afterwards and `database/sql` reaps the rows along with it, so the
pool never wedges. The same holds for the panicking callback: `catchPanic`
recovers, the error propagates, the rollback still runs.

So my commit message was overstating it, and I have rewritten the commit. It is
now `sqlbase: check row errors in ForAll`, the defect is the dropped `rows.Err()`
— a walk truncated by a row error was indistinguishable from a complete one — and
`rows.Close()` is described as hygiene rather than a fix, worth keeping because
the rows and the transaction are opened in different places.

**On the backport**: with the leak gone from the picture I do not think one is
warranted, though it is not my call. There is no hang and no wedged pool; what is
left needs an actual mid-stream driver or scan failure and produces a silently
short bucket walk. Rare, and not a corruption or availability issue.

**On the missing tests**: agreed, and added — `internal/sqlbase/db_test.go`, in
its own commit. It had to land *after* the build-tag removal rather than with the
fix, because until that commit there is no host-platform build of the package to
run tests against. Coverage: the key/value round trip including that an empty
value is a value and not an absent key, nested bucket deletion, `ForAll`'s
ordering and its callback-error path, the cursor's stepping/seeking/deletion
semantics, and that a failed `Update` leaves nothing behind.

One detail from writing it, since it bears on the schema comment: the
nested-bucket test counts rows through a connection of its own, because the
interesting failure is invisible through `walletdb`. `DeleteNestedBucket` removes
only the bucket's own row and leans on `ON DELETE CASCADE` for the children; with
foreign keys off the children survive, but they keep pointing at the deleted
row's id, so a recreated bucket gets a new id and reads empty either way. Only
the row count separates a cascade that fired from one that silently did not — it
goes from 1 to 4 if `foreign_keys` is dropped from the DSN, which is what the
test asserts.

Not covered: the row-error path the first commit fixes. Inducing a mid-stream
driver failure needs fault injection below `database/sql`, which is more
machinery than the remaining paths here justify. Called out in the commit
message.

guggero:
Rebased and addressed all comments.

litbot-9000 (approved):
Approving. The blocking bug is fixed, and fixed in a better shape than the one I suggested.

**`walletExists`.** Collapsing the probe onto btcwallet's own marker is the right call, and strictly better than my proposal. Consulting the marker when `cfg.Seed == nil` layers a guard on top of two predicates that still disagree; moving the predicate removes the second one, so the disagreeing-predicates class stops existing rather than getting detected. You also caught what my version missed — it fixes the random-seed half and leaves the operator holding the real seed locked out. Verified the healing path: seedless open now fails with `ErrWalletNotFound`, and the seeded retry creates in place with no manual deletion.

**The probe.** Verified it lets go of everything. Ten sequential probes against the single-connection pool do not wedge — `dbConnSet`'s refcount reaches zero and the handle genuinely closes — and an independent connection takes `BEGIN IMMEDIATE` on the file 535µs after the last probe, so no write lock survives it. The no-side-effects property holds: a missing file is still answered by `stat` without opening anything. The bolt path is untouched, since `Loader.WalletExists` short-circuits to `fileExists` when `localDB`, so probing does not take the flock.

**File mode.** You fixed this in code without replying, so just confirming it: pre-creating with `os.OpenFile(..., 0600)` is better than the `os.Chmod` I suggested, because it leaves no window where the file exists world-readable, and the sidecars inherit it. Re-ran `TestSQLiteWalletDBPermissions` under `umask 0022` — main file, `-wal` and `-shm` are all `0600`.

**`busy_timeout`.** Also already closed by `sqlWalletDBTimeout = 2 * sqlWalletDBBusyTimeout`.

The rest I raised you either fixed or answered, and I'm taking the answers as they stand: documenting the lost process exclusion rather than locking in this PR (a directory lock is the daemon's job and should cover both backends), leaving out `auto_vacuum` because it would be inert, and no backport now that the leak is out of the picture. `internal/sqlbase/db_test.go` plus `TestSQLiteWalletDBInterruptedCreate` is the right amount of coverage — I agree fault injection below `database/sql` is more machinery than the remaining path justifies. Thanks for the correction to your own commit message on the leak, too; that is the version I'd want in the history.

**What I ran** (fresh worktree at `8b8f0985`, Linux/amd64): `go test ./lwwallet/... ./internal/sqlbase/... ./waved/...` all pass, sqlbase now reports real coverage instead of `[no test files]`; `TestSQLiteWalletDBPermissions` re-run under `umask 0022`; a scratch test for probe handle-release and post-probe write-lock acquisition (not committed); `GOOS=js GOARCH=wasm go build ./lwwallet/... ./internal/sqlbase/...` clean. `go build ./...` still fails with `function main is undeclared in the main package`, identical on `origin/main` — pre-existing, root package is wasm-only. Still unverified: `golangci-lint` (not installed here, `make lint` wants Docker), and real iOS/Android backup behaviour.

— claudell ⚡ (not a maintainer)

lwwallet/walletdb_native.go
litbot-9000: Verified on `8b8f0985`. I read the new `walletExists` and then reproduced the whole sequence myself rather than reading the test.

Built the crash-mid-create state (open through `openSQLWalletDB`, close without creating a wallet), then:

```
probe after crash-mid-create: exists=false err=<nil>
seedless open  => no wallet database found in "...": create the wallet with a seed first
seeded retry   => created in place; reopens seedlessly afterwards
```

The heal-in-place claim is the part I most wanted to check, and "`New` succeeds" is not on its own enough — a wallet created in place could still be the wrong wallet. So I compared the healed wallet's identity key against a clean create from the same seed in a separate directory:

```
reference (clean create with the seed):    024e1688ef...f27eda35
healed in place after interrupted create:  024e1688ef...f27eda35
```

Same key, so it really is the operator's wallet and not another random one. The directory heals itself, confirmed.

The no-side-effect property holds: `WalletExists` on a fresh directory leaves it with zero entries.

On the opening probe itself, which is the new thing your shape introduces — I checked the three ways it could bite, and it doesn't:

- the handle is closed (`defer cleanup()`); repeated probes followed by a real open all succeed, so nothing is left holding the write lock;
- a corrupt file errors instead of reading as "no wallet": `open SQLite wallet database: file is not a database (26)`;
- an unreadable one likewise: `create wallet database file: ...: permission denied`.

Those last two matter for the same reason the original bug did: answering "false" there would send the caller down the create path.

`TestSQLiteWalletDBInterruptedCreate` asserts all three properties — the probe, `ErrWalletNotFound`, and the seeded create in place — not just some. It stops short of the identity-key comparison above; that is the only thing I would have added, and it is not worth a round trip.

Good call rejecting the `cfg.Seed == nil` shape I suggested. You are right that it fixes only the half that mints the random-seed wallet and leaves the operator holding the real seed locked out.

— claudell ⚡

lwwallet/walletdb.go
litbot-9000: **`Init`**: verified. A conflicting limit errors, a repeat of the same limit stays a no-op, and I measured the wallet's pool rather than trusting the plumbing — `dbConns.maxConnections == 1` and `sql.DB.Stats().MaxOpenConnections == 1` on the DSN the wallet store opens. `Init(2)` and `Init(25)` are both refused with "already initialized". `SetMaxOpenConns(1)` still holds.

**Directory lock** — answering your question directly: no, do not add it in this PR. A lock covering *both* backends as its own change is the better shape, for the reason you gave: bolt gets exclusion incidentally from `flock` and the daemon never asked for it anywhere, so putting it here would leave the guarantee attached to a backend choice rather than to the daemon. Your other two reasons hold up too — build-tagged `flock` files are real surface for a property this deployment does not lean on, and `locking_mode=EXCLUSIVE` trading away concurrent readers (including whatever copies the directory) plus the pragma-ordering hazard `walletdb_wasm.go` warns about is a bad deal. Documenting is the right call.

The documentation landed in all three places you said, and says the right thing in each:

- the `DBBackendSQLite` constant — "its file lock excludes a second OS process, while SQLite in WAL mode does not ... safe at the storage layer but not at the wallet layer, where both would derive the same addresses";
- `sample-waved.conf` — "sqlite gives up BoltDB's file lock, which is what stops a second daemon from opening the same directory";
- `lwwallet/AGENTS.md` — "gives up the one-daemon-per-directory guarantee. The storage layer stays consistent; the wallet layer does not."

All three name the storage-consistent / wallet-inconsistent split, which is the part an operator actually needs. Point closed.

— claudell ⚡

lwwallet/walletdb_native.go
litbot-9000: **Pragmas**: dumped from the DSN the code actually builds on this head, same way as before:

```
journal_mode = wal      synchronous = 2 (FULL)   fullfsync = 1
foreign_keys = 1        busy_timeout = 30000     auto_vacuum = 0
```

`fullfsync` is on and `synchronous=full` no longer rests on the driver default. That is the fix.

**`auto_vacuum`**: your reasoning is right, and I would rather you left it out. It is inert without something calling `PRAGMA incremental_vacuum`, and that makes it worse than absent — it reads as though free-page reclamation were handled when nothing reclaims anything. Do not add it for symmetry's sake; declining it is the better answer than the one I offered.

**The backup premise**: verified in both places. The commit message now carries "the directory goes from two storage engines to one, not from one special case to zero", with the already-WAL-SQLite justification spelled out. And `sample-waved.conf` reads "Pick sqlite so the directory holds one storage engine instead of two. Note that a running daemon leaves -wal and -shm files beside each SQLite database, so a copy taken while it runs has to include them" — "with a single storage engine's guarantees" is gone from the tree entirely (grepped). That is the accurate framing.

While I was in there: the config block also took the other suggestion and now leads with "This cannot be changed once the wallet has been created", with what that means operationally. Thanks — a reader cannot get to the loud failure before the warning now.

— claudell ⚡

internal/sqlbase/readwrite_bucket.go
litbot-9000: You are right and my comment overstated it. Thanks for testing the claim instead of taking it from me.

I reproduced your result independently rather than reading yours: 50 callback-error aborts plus 50 panicking-callback aborts on the one-connection pool, calling the concrete `readWriteBucket.ForAll` so it could not quietly be exercising `kvdb.ForAll`'s `ForEach` fallback, then a normal read. All prompt. Then I deleted `defer rows.Close()` from `ForAll` and ran exactly the same thing:

```
with    rows.Close(): 100 aborts, then normal read => pool not wedged
without rows.Close(): 100 aborts, then normal read => pool not wedged
```

So the connection leak does not happen, and the fix is not what makes it not happen. I also checked the structural reason, not just the empirical one: `View` and `Update` both go through `executeTransaction`, which always lands on commit-or-rollback, and `BeginReadTx`/`BeginReadWriteTx` have no callers outside the package — so there is no path that returns from an aborted `ForAll` without a rollback behind it to reap the rows.

`sqlbase: check row errors in ForAll` is the accurate description, and more accurate than what I wrote: the dropped `rows.Err()` is the defect, a truncated walk being indistinguishable from a complete one, and `rows.Close()` is hygiene worth the two lines. Agreed on no backport — no hang, no wedged pool, and what is left needs a mid-stream driver failure.

**On the tests**: `internal/sqlbase/db_test.go` exists, passes, and does land after the build-tag removal as you said (`823814ff` then `c67adf22`; I ran the package at `c67adf22` and it is green there). I read it against your list and it covers what you claimed: empty-value-is-a-value-not-an-absent-key, the cascade counted through `countRows`'s own connection with the 1-vs-4 distinction, `ForAll` ordering with deliberately out-of-order inserts plus the callback-error path and the abort-then-read follow-up, the cursor's stepping/seeking/off-both-ends/deletion semantics, and `TestUpdateRollback`. The uninduced row-error path being out of scope is called out in the commit message, which is the right place for it.

— claudell ⚡

lwwallet/walletdb_native.go
litbot-9000: Fixed. This thread had no reply, and it is the one I would have blocked on, so I verified it by measurement rather than by reading the diff.

`createWalletDBFile` creates the file at `0600` before SQLite ever opens the path — better ordering than the `os.Chmod` I suggested, since the sidecars inherit it rather than being chased afterwards. Measured on `8b8f0985` with the wallet running, so `-wal` and `-shm` are still on disk:

```
wallet.sqlite.db      mode=-rw-------
wallet.sqlite.db-wal  mode=-rw-------
wallet.sqlite.db-shm  mode=-rw-------
```

Matches bolt now, and `TestSQLiteWalletDBPermissions` asserts the same three so it cannot regress silently.

One branch I checked because the comment calls it out explicitly: an existing file keeps its mode. That is fine — `newWalletLoaderOptions` always goes through `createWalletDBFile` before `openSQLWalletDB`, and no released build writes this file at all, so there is no in-tree path that can produce a `0644` SQLite wallet database for the deliberate leave-it-alone branch to preserve. Nothing outstanding here.

— claudell ⚡

lwwallet/walletdb_native.go
litbot-9000: Also fixed, and since this thread has no reply I will note where: `sqlWalletDBTimeout` is now `2 * sqlWalletDBBusyTimeout` rather than an independent 30s, with the reasoning in the constant's own comment and repeated in `lwwallet/AGENTS.md`. Measured from the built DSN and the constants: `busy_timeout = 30000`, query timeout `1m0s`.

So a query that legitimately waits out the busy handler now surfaces `SQLITE_BUSY` instead of racing its own deadline into `context deadline exceeded`. That was the whole of the nit; nothing left here.

The rest of the `_txlock=immediate` reasoning I checked against the pinned driver last time still holds unchanged on this head.

— claudell ⚡

litbot-9000 (approved):
All six of my earlier points are either fixed or consciously declined with reasoning I checked and accept. Details are in the individual threads; the short version of what I re-verified on `8b8f0985` and how:

**Reproduced, not read.** The disagreeing-predicates bug: built the crash-mid-create on-disk state and confirmed the probe reports no wallet, the seedless open fails with `ErrWalletNotFound` instead of minting a random-seed wallet, and the seeded retry creates in place. I did not take "the seeded create succeeds" as sufficient for the heal-in-place claim, so I compared the healed wallet's identity key against a clean create from the same seed in a separate directory — same key, so the recovered wallet is the operator's. Also checked the new opening probe closes its handle (repeated probes then a real open all succeed), and that a corrupt or unreadable file errors rather than reading as "no wallet".

**Measured.** File modes with the wallet running: `wallet.sqlite.db`, `-wal` and `-shm` all `-rw-------`. Pragmas dumped from the DSN the code builds: `synchronous = 2 (FULL)`, `fullfsync = 1`, `journal_mode = wal`, `foreign_keys = 1`, `busy_timeout = 30000`. Wallet pool: `sql.DB.Stats().MaxOpenConnections == 1`, with `Init(2)`/`Init(25)` refused and `Init(1)` a no-op. Query timeout now `1m0s` over the 30s busy timeout.

**Corrected.** I was wrong about the severity of the `ForAll` connection leak — guggero tested it and I confirmed his result independently by deleting `defer rows.Close()` and re-running 100 aborts on a one-connection pool: the pool does not wedge either way, because the rollback reaps the rows. The rewritten commit message is more accurate than my comment was; the dropped `rows.Err()` is the real defect.

**Declined and accepted:** the directory lock (documented rather than enforced — answered in-thread: a lock covering both backends as a follow-up is the better shape, don't add it here) and `auto_vacuum=incremental` (inert without an `incremental_vacuum` caller, and misleading for it).

**Docs verified** in all three claimed places for the mutual-exclusion caveat, and the backup framing in both the commit message and `sample-waved.conf`.

**Build and tests:** `go build ./...` clean apart from the pre-existing wasm-only root-package failure that reproduces on `origin/main`. `go test ./internal/sqlbase/... ./lwwallet/... ./waved/...` all pass uncached. `GOOS=js GOARCH=wasm go build ./lwwallet/... ./internal/sqlbase/...` compiles. All nine commits build native and wasm individually, and `internal/sqlbase` is green at `c67adf22` where its tests land.

**Not verified:** the row-error path `ForAll` now checks has no test, which the commit message says and I agree is the right trade; and I did not run the test suites per-commit beyond `internal/sqlbase` at `c67adf22`.

— claudell (⚡)

PR #1218, opened 2026-08-28
waved: fix seed recovery missing live VTXOs

GeorgeTsagk:
## Problem

A wallet restored from seed reports zero balance for funds the operator still holds, and cannot spend them once found:

- `recoverIndexedVTXOs` scans the indexer under the VTXO-owner key family only. A VTXO settled into a round after an OOR receive stays on its OOR receive script, which is never queried. The mailbox replay pass rebuilds spent history but never learns what is still live, so recovery returns `recovered_vtxos: 0` as success.
- `recoveryDescriptorFromIndexer` rebuilds the tapscript from the indexer's `relative_expiry`, which is the batch's relative expiry, not the script's exit delay (1008 vs 144 in practice). The witness program does not hash to the script the operator holds, so every join naming a recovered VTXO is rejected with a script validation failure.

## Fix

Scan both key families, and build the tapscript from `terms.VTXOExitDelay`, the delay the matched pk script was derived with.

## Testing

Both tests fail without their fix. The stub indexer only answers for the exact script queried, so a scan that skips the OOR script cannot pass. The descriptor test reruns the operator's join script validation.

Note: the OOR family is now walked by both the indexed scan and `recoverOORReceiveScripts`, and the scan registers receive scripts for every index in the window. Folding the live-VTXO query into `recoverOORReceiveScripts` would avoid that at the cost of a larger diff; happy to switch if preferred.

### DIFF
--- a/waved/wallet_recovery.go
+++ b/waved/wallet_recovery.go
@@ -41,6 +41,16 @@ const (
 	recoveryIndexerRetryMax      = 5 * time.Second
 )
 
+// recoveryIndexedVTXOFamilies lists the key families seed recovery scans
+// against the operator's indexer. OOR receive scripts must be included: a
+// VTXO settled into a round after an OOR receive stays on its OOR receive
+// script, and the mailbox replay pass never learns whether the operator
+// still holds it.
+var recoveryIndexedVTXOFamilies = []keychain.KeyFamily{
+	libtypes.VTXOOwnerKeyFamily,
+	oorReceiveKeyFamily,
+}
+
 // WalletRecoveryResult reports the state discovered by one idempotent wallet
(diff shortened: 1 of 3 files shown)


darioAnongba (commented):
nice thanks! My only comment would be that we can greatly reduce the verbosity of the PR. A single commit with a concise description of what the problem was and the fix. 

Same for comments in the code. The code already describes what is being done, comments are only useful to add context when something is unclear or describe background effects, etc.

Some agents (ahem... claude), will write more way too much: PR description, commit messages and comments in code.

GeorgeTsagk:
Squashed into one commit and trimmed the description, commit message, and code comments.

darioAnongba (approved):
LGTM

PR #1227, opened 2026-09-01
rpc: add scoped macaroon baking

### LINKED ISSUE #1194: rpc: add scoped macaroons for least-privilege payment clients
## Motivation

Wavelength currently generates an `admin.macaroon` with full RPC access and a `readonly.macaroon` for query-only access. There does not appear to be a way to bake a macaroon with a custom subset of RPC permissions.

This creates a least-privilege gap for applications that need to initiate payments. A payment client currently needs the admin macaroon even if it only needs to inspect wallet state and call `Send`.

## Use case

I'm building an automated application that uses Wavelength as a user-owned payment wallet.

The application has its own policy layer that decides whether a payment is authorized based on things like amount, destination, and user-granted permissions. Once authorized, it needs to:

- inspect wallet/balance state;
- inspect payment/activity state;
- initiate an authorized Lightning payment;
- observe the result.

It should not be able to perform unrelated administrative operations simply because it has permission to send a payment.

In particular, I don't want to give an automated process possession of `admin.macaroon` when the only state-changing capability it needs is payment.

## Desired flow

Ideally there would be a first-class way to create a macaroon containing only the permissions required by a particular integration, similar conceptually to `lncli bakemacaroon`.

For example, a payment integration could create a credential with read access to the wallet/payment state plus the RPC permission required to `Send`, while excluding unrelated write/admin RPCs.

This could be exposed through a `wavecli` command, RPC, or predefined role such as `pay-only`; I'm not particularly attached to the interface as long as an application can receive a narrowly scoped credential instead of `admin.macaroon`.

## What would unblock this integration

Any supported mechanism that lets me provision a credential that can:

- perform the read operations required to inspect wallet/payment state; and
- initiate a Lightning payment;

without granting the rest of admin.macaroon's permissions would solve the immediate problem.

Longer term, arbitrary RPC-scoped macaroon baking would seem useful for other integrations as well.

darioAnongba:
Closes #1194.

## Summary

- expose local authenticated `BakeMacaroon` and `ListPermissions` RPCs, with credential creation protected by the admin-only `macaroon:generate` permission
- accept registered entity/action pairs and exact `uri:/package.Service/Method` grants so payment clients can avoid unrelated methods that share a broader write permission
- add `wavecli bakemacaroon`, including secure binary file output, root-key selection, and raw JSON request input

### DIFF
--- a/cmd/wavecli/waveclicommands/cmd_macaroon.go
+++ b/cmd/wavecli/waveclicommands/cmd_macaroon.go
@@ -1,0 +1,274 @@
+package waveclicommands
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"math"
+	"os"
+	"strings"
+
+	"github.com/lightninglabs/wavelength/waverpc"
+	"github.com/lightningnetwork/lnd/macaroons"
+	"github.com/spf13/cobra"
+	"gopkg.in/macaroon.v2"
+)
+
+// bakedMacaroonResult is the command result. A saved macaroon reports only its
+// path so the credential is not also copied into terminal history.
+type bakedMacaroonResult struct {
+	Macaroon string `json:"macaroon,omitempty"`
+	Path     string `json:"path,omitempty"`
+}
+
+// newBakeMacaroonCmd creates the scoped macaroon baking command.
+func newBakeMacaroonCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "bakemacaroon <entity:action>...",
+		Short: "Bake a scoped daemon macaroon",
+		Long: "Bakes a daemon macaroon with known entity/action " +
+			"permissions or exact " +
+			"uri:/package.Service/Method scopes. " +
+			"For example: wavecli bakemacaroon " +
+			"uri:/waverpc.DaemonService/GetInfo.",
+		RunE: bakeMacaroon,
+	}
+
+	cmd.Flags().String(
+		"save-to", "", "save the binary macaroon to this path",
+	)
+	cmd.Flags().Uint64(
+		"root-key-id", 0, "numeric macaroon root key ID",
+	)
+	cmd.Flags().Uint64(
+		"valid-for", 0, "validity in seconds (zero means no expiry)",
+	)
+
+	return cmd
+}
+
+// bakeMacaroon validates CLI permission tuples and calls the daemon baker.
+func bakeMacaroon(cmd *cobra.Command, args []string) error {
+	rawJSON, err := cmd.Flags().GetString("request-json")
+	if err != nil {
+		return err
+	}
+	if rawJSON != "" {
+		switch {
+		case len(args) != 0:
+			return invalidArgs(
+				fmt.Errorf("positional permissions cannot " +
+					"be combined with --request-json"),
+			)
+
+		case cmd.Flags().Changed("root-key-id"):
+			return invalidArgs(
+				fmt.Errorf("--root-key-id cannot be " +
+					"combined with --request-json; set " +
+					"rootKeyId in JSON"),
+			)
+		}
+	}
+
+	req := &waverpc.BakeMacaroonRequest{}
+	err = parseRequest(cmd, req, func() error {
+		if len(args) == 0 {
+			return invalidArgs(
+				fmt.Errorf("specify at least one " +
+					"entity:action permission"),
+			)
+		}
+
+		permissions, err := parseMacaroonPermissions(args)
+		if err != nil {
+			return invalidArgs(err)
+		}
+
+		rootKeyID, err := cmd.Flags().GetUint64("root-key-id")
+		if err != nil {
+			return err
+		}
+
+		req.Permissions = permissions
+		req.RootKeyId = rootKeyID
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	outputPath, outputFile, err := openMacaroonOutput(cmd)
+	if err != nil {
+		return err
+	}
+	keepOutput := false
+	if outputFile != nil {
+		defer func() {
+			if keepOutput {
+				return
+			}
+
+			_ = outputFile.Close()
+			_ = os.Remove(outputPath)
+		}()
+	}
+
+	client, conn, err := getMacaroonClient(cmd)
+	if err != nil {
+		return err
+	}
+	if conn != nil {
+		defer func() {
+			_ = conn.Close()
+		}()
+	}
+
+	ctx, cancel := rpcContext(cmd)
+	defer cancel()
+
+	resp, err := client.BakeMacaroon(ctx, req)
+	if err != nil {
+		return fmt.Errorf("BakeMacaroon RPC failed: %w", err)
+	}
+
+	macBytes, macHex, err := constrainBakedMacaroon(
+		cmd, resp.GetMacaroon(),
+	)
+	if err != nil {
+		return err
+	}
+
+	result := bakedMacaroonResult{
+		Macaroon: macHex,
+	}
+	if outputFile != nil {
+		if _, err := outputFile.Write(macBytes); err != nil {
+			return fmt.Errorf("write baked macaroon: %w", err)
+		}
+		if err := outputFile.Close(); err != nil {
+			return fmt.Errorf("close baked macaroon: %w", err)
+		}
+		keepOutput = true
+
+		result.Macaroon = ""
+		result.Path = outputPath
+	}
+
+	return printBakedMacaroonResult(cmd, result)
+}
+
+// openMacaroonOutput reserves a new private output file before an RPC can
+// mint a credential. Exclusive creation prevents accidental replacement of an
+// existing file or symlink.
+func openMacaroonOutput(cmd *cobra.Command) (string, *os.File, error) {
+	savePath, err := cmd.Flags().GetString("save-to")
+	if err != nil {
+		return "", nil, err
+	}
+	if savePath == "" {
+		return "", nil, nil
+	}
+
+	expandedPath, err := expandCLIPath(savePath)
+	if err != nil {
+		return "", nil, err
+	}
+
+	//nolint:gosec // G304: the destination is supplied by the CLI user.
+	file, err := os.OpenFile(
+		expandedPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600,
+	)
+	if err != nil {
+		return "", nil, fmt.Errorf("create baked macaroon: %w", err)
+	}
+
+	return expandedPath, file, nil
+}
+
+// constrainBakedMacaroon validates the daemon response and applies optional
+// client-side restrictions before the credential is displayed or saved.
+func constrainBakedMacaroon(cmd *cobra.Command, macHex string) ([]byte, string,
+	error) {
+
+	macBytes, err := hex.DecodeString(macHex)
+	if err != nil {
+		return nil, "", fmt.Errorf("decode baked macaroon: %w", err)
+	}
+	if len(macBytes) == 0 {
+		return nil, "", fmt.Errorf("daemon returned an empty macaroon")
+	}
+
+	rawMacaroon := &macaroon.Macaroon{}
+	if err := rawMacaroon.UnmarshalBinary(macBytes); err != nil {
+		return nil, "", fmt.Errorf("decode baked macaroon: %w", err)
+	}
+
+	validFor, err := cmd.Flags().GetUint64("valid-for")
+	if err != nil {
+		return nil, "", err
+	}
+	if validFor > math.MaxInt64 {
+		return nil, "", invalidArgs(
+			fmt.Errorf(
+				"--valid-for exceeds %d", int64(math.MaxInt64),
+			),
+		)
+	}
+	if validFor != 0 {
+		rawMacaroon, err = macaroons.AddConstraints(
+			rawMacaroon,
+			macaroons.TimeoutConstraint(
+				int64(validFor),
+			),
+		)
+		if err != nil {
+			return nil, "", fmt.Errorf("constrain baked "+
+				"macaroon: %w", err)
+		}
+
+		macBytes, err = rawMacaroon.MarshalBinary()
+		if err != nil {
+			return nil, "", fmt.Errorf("encode baked macaroon: %w",
+				err)
+		}
+	}
+
+	return macBytes, hex.EncodeToString(macBytes), nil
+}
+
+// parseMacaroonPermissions parses entity:action tuples without splitting URI
+// actions on their path separators.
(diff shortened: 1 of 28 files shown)


waved/rpc_macaroons.go
```diff
@@ -0,0 +1,171 @@
...
+		op := bakery.Op{
+			Entity: permission.GetEntity(),
+			Action: permission.GetAction(),
+		}
+		if op.Entity == "" || op.Action == "" {
+			return nil, fmt.Errorf("permission %d requires "+
+				"non-empty entity and action", idx)
+		}
+
+		if op.Entity == macaroons.PermissionEntityCustomURI {
```
litbot-9000, line 152: **Privilege escalation: a scoped macaroon can bake a more powerful one.**

This branch accepts any `uri:` action that is a key in the permission map — and `/waverpc.MacaroonService/BakeMacaroon` is one of those keys. lnd's `macaroons.Service.CheckMacAuth` falls back to checking `bakery.Op{Entity: "uri", Action: fullMethod}` when the broad entity/action check fails, so a macaroon carrying only that URI grant authorizes `BakeMacaroon` itself. Nothing in `validatePermissions` intersects the requested ops with the caller's ops, so that macaroon can then mint `macaroon:generate` plus every write op in the map.

I confirmed this on this branch with a temporary test:

```go
// bake a "scoped" mac whose only op is uri:/waverpc.MacaroonService/BakeMacaroon
resp, err := server.BakeMacaroon(ctx, &waverpc.BakeMacaroonRequest{
    Permissions: []*waverpc.MacaroonPermission{{
        Entity: macaroons.PermissionEntityCustomURI,
        Action: waverpc.MacaroonService_BakeMacaroon_FullMethodName,
    }},
})
require.NoError(t, err)                                 // accepted
require.NoError(t, authService.CheckMacAuth(            // authorizes BakeMacaroon
    ctx, macBytes, wavedRPCPermissions[bake], bake,
))
// ...and that credential mints macaroon:generate + onchain:write + vtxo:write,
// which CheckMacAuth then accepts for /waverpc.DaemonService/SendOnChain.
```

Both steps pass. The same hole exists for a plain `macaroon:generate` grant, which the entity/action branch below also accepts.

The operational shape matters: the PR's whole pitch is "hand a payment client a narrowly scoped credential". An operator who includes the bake URI — or who scripts a provisioning agent with `macaroon:generate` so it can rotate its own tokens — believes they granted one method and has actually granted mainnet spend authority. There is no revocation path to undo it.

Suggested fix, cheapest version — reject self-referential grants outright in `validatePermissions`:

```go
if op.Entity == entityMacaroon {
    return nil, fmt.Errorf("permission %d: %s permissions cannot be "+
        "delegated through baking", idx, entityMacaroon)
}
if op.Entity == macaroons.PermissionEntityCustomURI &&
    strings.HasPrefix(op.Action, "/"+macaroonServiceName+"/") {

    return nil, fmt.Errorf("permission %d: %s methods cannot be "+
        "delegated through baking", idx, macaroonServiceName)
}
```

That keeps `admin.macaroon` (which holds `macaroon:generate` from `extractPerms`) as the only thing that can ever bake, which I think is the intended model. If instead you want delegation to be possible, it needs a genuine subset check against the caller's macaroon ops via `RawMacaroonFromContext`. Either way, whichever you pick should be stated in the `BakeMacaroon` proto comment — right now it promises "scoped permissions" and says nothing about the fact that a baked credential may be able to escalate.

waved/rpc_macaroons.go
```diff
@@ -0,0 +1,171 @@
...
+			"macaroon authentication is disabled",
+		)
+	}
+
+	ops, err := manager.validatePermissions(req.GetPermissions())
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	rootKeyID := []byte(strconv.FormatUint(req.GetRootKeyId(), 10))
```
litbot-9000, line 72: **`root_key_id` is unvalidated, and the default is the admin root key.**

`req.GetRootKeyId()` defaults to 0, which formats to `"0"` — exactly `macaroons.DefaultRootKeyID`, the key backing `admin.macaroon` and `readonly.macaroon`. So every credential baked without an explicit `--root-key-id` shares a root key with the daemon's own admin credential. Since the only way to revoke a macaroon is to destroy its root key, revoking a leaked payment macaroon also locks the operator out of their own daemon. That is the opposite of what "least privilege, easy to rotate" implies, and neither the proto comment nor `docs/daemon_cli_guide.md` warns about it.

There is also no validation on the value at all: I confirmed `0`, `1`, and `1<<63` are all accepted, each silently creating a new root key row on demand via the store's `RootKey()`. A caller with bake access can therefore create unbounded root-key rows.

Two suggestions:
- Reject `root_key_id == 0` (or at minimum have `wavecli bakemacaroon` default to a non-zero id), so operator credentials are never entangled with baked ones. lnd guards the equivalent case by refusing `DeleteMacaroonID` on the default id; here the guard is more useful at bake time since there is no delete at all.
- Document the semantics on the proto field: root key ids are namespaces for revocation, and picking the same id for two clients means they can only ever be revoked together.

waved/rpc_macaroons.go
```diff
@@ -0,0 +1,171 @@
...
+	mac, err := manager.authService.NewMacaroon(
+		ctx, rootKeyID, ops...,
```
litbot-9000, line 74: **Baked macaroons are permanent and unrevocable — please say so explicitly.**

`Service.NewMacaroon` calls `Oven.NewMacaroon(ctx, bakery.LatestVersion, nil, ops...)` — caveats are always `nil`, and `BakeMacaroonRequest` has no field to supply any. The daemon registers `IPLockChecker`/`IPRangeLockChecker` in `rpc_security.go`, so the machinery for `ipaddr`/`ipnet` caveats is already present and unused here; `time-before` would come free from the bakery's standard checkers.

Combined with the absence of any revocation surface (I grepped: no `DeleteMacaroonID`, no `ListMacaroonIDs`, no root-key rotation RPC anywhere in the repo), a credential minted here is valid for the lifetime of the daemon's macaroon DB, full stop. On a daemon that moves mainnet funds, an operator handing a payment macaroon to a third-party integration will reasonably assume they can time-box or withdraw it.

I don't think caveat support has to land in this PR, but the API shape should not foreclose it and the limitation should be documented. Concretely:
- Add `int64 expiry_seconds = 3;` (or a `repeated string caveats`) to `BakeMacaroonRequest` now, even if unimplemented — retrofitting field 3 later is fine, but agreeing the shape while the message is new is cheaper.
- Add a line to the `bakemacaroon` section of `docs/daemon_cli_guide.md`: baked macaroons never expire and can only be invalidated by destroying their root key, which currently requires direct DB surgery.

cmd/wavecli/waveclicommands/cmd_macaroon.go
```diff
@@ -0,0 +1,159 @@
...
+		expandedPath, err := expandCLIPath(savePath)
+		if err != nil {
+			return err
+		}
+
+		macBytes, err := hex.DecodeString(resp.GetMacaroon())
+		if err != nil {
+			return fmt.Errorf("decode baked macaroon: %w", err)
+		}
+		if err := os.WriteFile(
```
litbot-9000: **`--save-to` silently clobbers an existing credential.**

`os.WriteFile` opens with `O_CREATE|O_WRONLY|O_TRUNC`. Point `--save-to` at a path that already holds a live macaroon — a typo'd `admin.macaroon`, or re-running the documented `--save-to ~/.waved/payment.macaroon` example — and the old credential is destroyed with no prompt and no way to recover it, since there is no listing or re-export RPC. For a file whose whole purpose is to be an irreplaceable secret, refusing to overwrite is the right default.

Two smaller things in the same block:
- If the path already exists, `os.WriteFile` does **not** apply the `0600` mode argument — it keeps the existing mode. The `os.Chmod` on the next line does fix it, but only *after* the secret has already been written, so a pre-existing `0644` file is world-readable for that window. `O_EXCL` removes both problems and makes the `Chmod` unnecessary.
- A failed write leaves a truncated file that still looks like a macaroon to the next reader.

```go
f, err := os.OpenFile(
    expandedPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600,
)
if err != nil {
    return fmt.Errorf("write baked macaroon: %w", err)
}
defer f.Close()
if _, err := f.Write(macBytes); err != nil {
    return fmt.Errorf("write baked macaroon: %w", err)
}
```

Add `--force` if overwriting needs to stay possible. Worth noting `cmd_macaroon_test.go:156` asserts the `0600` mode on a fresh path only — a case that writes to a pre-existing `0644` file would have caught the ordering issue.

Separately, the hex macaroon is still returned in the RPC response and only *suppressed* from the JSON output after the file write succeeds. That is fine for stdout, but it does mean the credential passes through `--request-json`-style debug logging and any gRPC request/response logging enabled on the client. Not blocking, just worth knowing before the docs call this path "secure".

waverpc/daemon.yaml
```diff
@@ -3,6 +3,11 @@ config_version: 3
 
 http:
   rules:
+    - selector: waverpc.MacaroonService.BakeMacaroon
+      post: /v1/daemon/macaroon
```
litbot-9000, line 7: Does credential minting need to be on the REST gateway?

Auth does hold here — I traced it: `macaroonHeaderMatcher` forwards the macaroon header into gRPC metadata and the proxied call goes through the same interceptor, so `POST /v1/daemon/macaroon` is not an unauthenticated bypass. Good.

But it does widen the blast radius for no stated benefit. `gateway.BrowserHeaders` will emit `Access-Control-Allow-Origin: *` when `AllowedOrigins` contains `*`, and the response body carries the raw hex macaroon over HTTP where it is far more likely to end up in a proxy log or a browser devtools history than the gRPC path is. Any XSS in an allowed origin that already holds the admin macaroon escalates from "read the balance" to "mint a permanent credential".

`ListPermissions` over REST is fine and useful. I'd drop the `BakeMacaroon` rule unless there's a concrete consumer, and if there is one, note in the PR description why a browser-reachable minting endpoint is acceptable. The proto comment on `MacaroonService` says it is "registered on the authenticated local gRPC/REST listener only", which reads as if REST exposure were the conservative choice rather than an addition.

waved/rpc_auth.go
```diff
@@ -70,6 +73,7 @@ var wavedEntities = []string{
 	entityRecovery,
 	entityFees,
 	entityActivity,
+	entityMacaroon,
```
litbot-9000, line 76: Adding `entityMacaroon` here means `wavedReadOnlyPermissions()` now mints `macaroon:read` into every `readonly.macaroon`. Harmless today — no method requires `macaroon:read`, and `rpc_security_test.go` correctly proves readonly cannot reach `BakeMacaroon`. 

The hazard is that `wavedReadOnlyPermissions` grants `read` on every entity *automatically*, so the day someone adds a `macaroon:read` method (a `ListMacaroonIDs`, say — the natural next step once revocation exists), every previously distributed read-only macaroon silently gains it, with no diff to this function to review. Credential metadata is exactly the entity where that auto-inclusion is wrong.

Suggest excluding `entityMacaroon` from the read-only set explicitly, with a comment saying why, rather than relying on there never being a read action in that domain.

systest/scoped_macaroon_test.go
```diff
@@ -0,0 +1,248 @@
...
+			return err
+		},
+		"recovery": func() error {
+			_, err := daemonClient.ArmVHTLCRecovery(
+				ctx, &waverpc.ArmVHTLCRecoveryRequest{},
+			)
+
+			return err
+		},
+		"administration": func() error {
```
litbot-9000, line 230: This suite is genuinely good — the `deniedCalls` table is a real negative test and it covers the cases I'd have asked for, including the `administration` case proving a URI-scoped payment macaroon cannot re-bake. Credit where due; a lot of macaroon PRs ship with only the allow-path asserted.

The one case it is missing is the one that actually breaks: this proves a macaroon *without* a bake grant is refused, not that a macaroon *with* one cannot escalate. Once the `rpc_macaroons.go` fix lands, please add the inverse — bake a macaroon whose only op is `uri:/waverpc.MacaroonService/BakeMacaroon`, and assert the daemon refuses to mint it (or, if you go the subset-check route, that it can only re-mint that same op and not `onchain:write`). A unit test in `rpc_macaroons_test.go` is enough; it doesn't need the Docker harness.

Note for the record: I could not run this file — the harness needs Docker and `dial unix /var/run/docker.sock` fails in my container. I read it rather than executed it.

litbot-9000 (changes requested):
**Request changes.** The permission wiring itself is solid — `registeredRPCPermissions` now fails closed at startup with auth enabled, the lnd interceptor rejects any method missing from the map, exact-URI grants are validated against the *registered* method set (not the static one), and `TestScopedPaymentMacaroon` has real negative assertions rather than a happy path. That part I'd merge.

The blocker is a privilege-escalation loop. `validatePermissions` never intersects the requested ops with the caller's own ops, and it happily accepts `uri:/waverpc.MacaroonService/BakeMacaroon` because that key exists in the permission map. lnd's `CheckMacAuth` treats `uri:<fullMethod>` as an alternative to the required entity/action, so a macaroon whose only op is that URI grant authorizes `BakeMacaroon` and can then mint `macaroon:generate` + every write op. I wrote a throwaway test against this branch and it passes — details inline on `rpc_macaroons.go`. Same hole via a plain `macaroon:generate` grant. It needs an explicit denylist in `validatePermissions`, or a real subset check.

The other two I'd want fixed before merge: `wavecli --save-to` silently truncates an existing file (clobbering a live credential), and the default `root_key_id = 0` puts every baked macaroon on the same root key as `admin.macaroon` and `readonly.macaroon`, so a leaked scoped credential can't be revoked without destroying the daemon's own admin access. There is no revocation surface at all — no `DeleteMacaroonID`, no `ListMacaroonIDs`, no caveat support — so these credentials are permanent. Worth stating that in the docs even if it stays the design.

**Verified:** `go build ./...` and `go vet` clean on the touched packages (the root `main is undeclared` note is pre-existing). `go test ./waved/` and `go test ./cmd/wavecli/...` pass. Read the full hand-written surface plus the existing `rpc_security.go`, `gateway_server.go`, `gateway.go`, and the vendored `lndclient/macaroon_service.go`, `lnd/macaroons/service.go` and `lnd/rpcperms/interceptor.go` to confirm what the interceptor does with the `uri` entity and with an unlisted method. Proved the escalation and the unvalidated root-key-id by adding a temporary test to `waved/` (removed before reporting).

**Not verified:** the systests. `go test -tags systest ./systest/ -run TestScopedPaymentMacaroon` fails at harness setup with `dial unix /var/run/docker.sock: no such file or directory` — no Docker in my container, so I read that test rather than ran it. I also did not exercise the REST gateway end to end.

— claudell ⚡

PR #1264, opened 2026-09-04
sdk/swaps: Add durable two-vHTLC refresh swaps

sputn1ck:
## Summary

- add an age cap and exact managed-input reservation to SendOOR, enforced against each VTXO backing batch creation height
- add the CreateRefreshSwap protocol RPC and a durable Go SDK RefreshSession that coordinates equal-value vHTLC legs under one payment hash
- resume role-scoped refresh workers after restart and expose exact completed custom-input evidence for reconciliation

## Protocol and recovery

The client creates and retains the preimage. Its old VTXO funds the first vHTLC, while the replacement vHTLC uses the same payment hash and must satisfy the requested maximum backing-batch age. Exact input sets and the age cap are part of the replay identity, while reservation epochs remain durable fencing metadata.

Refresh initiation is exposed through the Go SDK. Legacy daemon summary RPCs omit refresh settlements because their public enum cannot represent that settlement kind.

Recovery is intentionally conservative: if an accepted replacement leg remains ambiguous and neither live, spent, nor authoritative absence evidence appears, the session waits rather than risking a duplicate funding attempt. The original leg still retains unilateral recovery.

## Related work

This composes with the existing custom-vHTLC refresh and recovery machinery. It does not replace indexed vHTLC recovery or automatic in-round refresh policy.

## Testing

- RPC and SQL generated artifacts verified with the repository-pinned generator versions
- make fmt-changed
- make lint-changed-local
- make tidy-module-check
- make build
- go test ./... -count=1
- go test -tags=swapruntime ./swapclientserver -count=1
- go test -race ./sdk/swaps -run Refresh -count=1


litbot-9000:
@bhandras: review reminder
@roasbeef: review reminder
@darioanongba: review reminder

litbot-9000:
@bhandras: review reminder
@roasbeef: review reminder
@darioanongba: review reminder

PR #1269, opened 2026-09-04
unroll: Reissue staged work after route errors

### LINKED ISSUE #1268: unroll: reissue staged work after a live route failure
## The problem

The unroll actor stages planner state before routing txconfirm work. That ordering protects the on-chain side effect, but a same-process retry can lose the route operation.

A concrete path is:

1. A txconfirm confirmation notification arrives as a durable `Tell`.
2. The notification unblocks the next proof transaction.
3. Unroll stages the next transaction in `InFlightTxids`.
4. The `EnsureConfirmedReq` for that transaction returns a transient error.
5. The durable actor nacks and redelivers the notification.
6. Reapplying the confirmation is idempotent. The planner now sees the next transaction as already in flight, so it emits no new ensure request.
7. The notification is acknowledged while the txconfirm subscription remains unarmed.

A daemon restart recovers because `ResumeEvent` explicitly reissues every in-flight transaction. The same live process does not use that path.

## Why it matters

The unroll can remain stuck until restart after a transient txconfirm setup or transport error. The durable checkpoint says the proof transaction is in flight, but txconfirm may never have registered or broadcast it.

This is pre-existing. A caller that converts setup failures into request errors makes the path easier to reach, but does not create the replay gap.

## Proposed fix

When outbox routing fails after the checkpoint was staged, remember that the live actor owes a reissue. Before the next turn applies its message, drive the existing `ResumeEvent` path against the staged state. Clear the marker only after the turn commits.

Keep restart behavior unchanged. Txconfirm already deduplicates repeated ensures by txid.

## Acceptance criteria

- [ ] A durable confirmation notification unblocks a proof transaction.
- [ ] The first ensure request for that transaction fails once.
- [ ] The same actor process redelivers the notification and reissues the same txid.
- [ ] The retry can receive confirmation and advance durable planner state.
- [ ] Already-confirmed ancestor work is not resubmitted.
- [ ] No distinct transaction or duplicate unroll job is created.
- [ ] Restart reissue behavior remains unchanged.

bhandras:
## Summary

Fixes #1268.

Reissue staged unroll work after a live outbox route failure. This prevents a transient txconfirm error from leaving a proof transaction recorded as in flight without an active txconfirm subscription.

## Problem

Unroll stages its next planner state before routing actor-boundary IO. A confirmation notification can therefore:

1. unblock the next proof transaction;
2. stage that transaction in `InFlightTxids`;
3. fail while asking txconfirm to ensure it;
4. nack and redeliver the durable notification.

Reapplying the notification is idempotent. The planner sees the child transaction as already in flight and emits no second ensure request. The notification then commits, and recovery waits for a daemon restart because only `ResumeEvent` reissues the in-flight set.

## Fix

When route IO fails after Stage, the behavior records an in-memory pending reissue. The next actor turn drives the existing `ResumeEvent` path before applying its own message. The marker clears only after the turn commits.

Restart behavior stays unchanged. The marker does not need persistence because restored actors already enter through `ResumeUnrollRequest`. Repeated ensure requests keep the same txid and use txconfirm's existing deduplication boundary.

## Test

`TestUnrollTellRetryReissuesStagedInFlight` uses the real SQLite durable mailbox:

- start a linear unroll and ensure the root once;
- confirm the root through a durable txconfirm notification;
- fail the newly unblocked target ensure once;
- let the same actor process redeliver the notification;
- assert the same target txid is reissued exactly once;
- confirm it and assert the durable checkpoint records confirmation;
- assert the root was not resubmitted and no distinct work was created.

Local validation:

- `go test ./unroll -count=3`
- `go test -race ./unroll -run '^TestUnrollTellRetryReissuesStagedInFlight$' -count=3`
- `make fmt-changed`
- `make lint-changed-local`
- Go documentation audit for both changed Go files

## Relationship to #1243

This fixes a pre-existing unroll replay gap and has no code dependency on #1243. #1243 makes transient txconfirm setup failures surface as request errors, which is one way to reach this existing path.

### DIFF
--- a/unroll/actor.go
+++ b/unroll/actor.go
@@ -195,6 +195,13 @@ type behavior struct {
 	// so a policy-gated sweep is not reported as imminent. None until the
 	// sweep phase resolves the policy.
 	requiredLockTime fn.Option[uint32]
+
+	// routeRetryPending records that the current FSM state was durably
+	// staged before an outbox effect failed. The next turn replays the
+	// checkpoint's in-flight work through ResumeEvent before applying its
+	// own message. This is in-memory only because process restart already
+	// enters through ResumeUnrollRequest.
+	routeRetryPending bool
 }
 
 // unrollTx is the transaction-scoped store handed to the unroll behavior inside
@@ -265,12 +272,20 @@ func (b *behavior) Receive(ctx context.Context, msg Msg,
 	// Run the FSM pipeline. Every checkpoint write inside is a short,
 	// lock-releasing Stage and the slow txconfirm IO runs with no writer
 	// transaction held; dispatch never commits.
+	if b.routeRetryPending {
+		if err := b.reissueStagedRoute(ctx, ax); err != nil {
+			return fn.Err[Resp](err)
+		}
+	}
+
 	res := b.dispatch(ctx, ax, msg)
 	if res.IsErr() {
 
-		// The behavior did not commit, so the framework nacks the
-		// message; it is redelivered and replayed against the durably
-		// Staged state.
+		// The behavior did not commit. Durable Tells are nacked for
+		// redelivery; Asks return the error to their caller. If route
+		// IO failed after Stage, routeRetryPending makes the next live
+		// turn reissue the checkpoint's in-flight work before it
+		// applies its own message.
 		return res
 	}
 
@@ -281,8 +296,26 @@ func (b *behavior) Receive(ctx context.Context, msg Msg,
 	if err := b.commitAck(ctx, ax); err != nil {
 		return fn.Err[Resp](err)
 	}
+	b.routeRetryPending = false
 
 	return res
+}
+
+// reissueStagedRoute replays every outbox effect implied by the current staged
+// checkpoint. A failed route has already advanced the FSM state, so applying
+// the original event again cannot rediscover work now classified as in-flight;
+// ResumeEvent is the explicit reissue path for that state.
+func (b *behavior) reissueStagedRoute(ctx context.Context,
+	ax actor.Exec[unrollTx]) error {
+
+	state, err := b.currentState()
+	if err != nil {
+		return err
+	}
+
+	return b.driveEvent(ctx, ax, &ResumeEvent{
+		Height: stateHeight(state),
+	})
 }
 
 // dispatch maps one durable message onto the FSM event surface and runs the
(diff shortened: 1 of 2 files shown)


bhandras:
## Post-merge review: no PR-introduced blocker found

Posted on behalf of @bhandras. These are **fresh AI-assisted reviews performed on 2026-09-07, after merge**, not retrospective pre-merge approval or a formal Gateway review.

Reviewed merge: `d53fac84b36ee1dfd3703e09c4696e4b7cc71acc`, against first parent `c193cb7ab5355469f44e7b81f90f652777dddf10`. Original PR head: `1255879cf4461e1ac8f2ffab780e5d7f2f0ebd2f`. This merged tree already includes #1243 and #1242.

### Review evidence

- **Independent Claude CLI review:** `claude -p`, returned model `claude-fable-5-1`, completed 09:51 UTC. Reviewed the full diff and surrounding unroll, registry, planner, txconfirm, and durable actor code. Verdict: no PR-introduced blocker; one pre-existing follow-up.
- **Separate Codex review:** independently inspected the same paths, ran the package/regression tests below, and added local-only fault-injection diagnostics for repeated route failure and commit failure. Same verdict.
- **Correction to the earlier review status:** the three green Gateway workflow runs were unsupported-event no-ops (`no_op_event_unsupported_for_apps`), not completed reviews. They should not be counted as Gateway approval.

### What we checked

The failure starts after a checkpoint marks a transaction in flight but the request to txconfirm fails. Replaying the original notification does not rediscover that transaction because it is already in flight.

This fix marks that route failure and invokes the existing Resume path on the next live actor turn. Resume reissues the staged work. The marker clears only after commit succeeds.

- Reissue uses the same proof transactions. txconfirm deduplicates by txid and checks matching parameters. Stable subscriber identity avoids accumulating duplicate subscriptions.
- A repeated route failure or commit failure retains the marker. Successful commit clears it.
- Existing persisted sweep reuse and restart Resume recovery are unchanged.
- The new SQLite regression confirms through the successful reissue's subscriber, checks durable confirmation, and checks exact request counts.

### Fresh tests: passed

- Entire `./unroll` package, three repetitions.
- Focused same-process retry, Resume, and confirmation regressions under `-race`, three repetitions. Claude independently ran this group too.
- Entire `./txconfirm` and `./baselib/actor` packages.
- Eight focused staging, sweep reuse, lease-loss replay, and Resume regressions, ten repetitions.
- Two extra **local-only** fault-injection tests under `-race`, three repetitions: three consecutive route failures followed by confirmation; and commit failure after successful reissue, followed by recovery and marker clearing.

The repeated-failure diagnostic initially exceeded its 20-second test wait. Logs showed the existing retry plus mailbox-poll backoff was still progressing. Increasing only that diagnostic wait to 60 seconds made all repetitions pass. No production timeout changed.

### Follow-up, not a blocker: failed start followed by re-admission

A separate, pre-existing path remains: an initial root-submission error can stop the child via [`failAdmittedChild`](https://github.com/lightninglabs/wavelength/blob/d53fac84b36ee1dfd3703e09c4696e4b7cc71acc/unroll/registry.go#L781), leaving its staged in-flight checkpoint. Re-admission restores that checkpoint but sends Start. The [non-idle Start handling](https://github.com/lightninglabs/wavelength/blob/d53fac84b36ee1dfd3703e09c4696e4b7cc71acc/unroll/fsm_logic.go#L105) does not request reissue, so the re-admitted exit can stall until restart Resume recovery.

This requires a failed initial submission and later re-admission. It is delayed recovery, not demonstrated fund loss. #1269 neither introduces nor worsens it: this fix handles a live child, while that path stops the child. Next step should be a focused registry regression and a bounded fix at re-admission. Changing Start semantics needs admission-lifecycle coverage. This path was code-traced, not reproduced end to end in these tests.

**Disposition:** no corrective change or revert recommended for #1269 based on these reviews. The follow-up should be handled separately. This was scoped local validation, not a fresh full system-test run or production validation. No production code was changed during review.

PR #1333, opened 2026-09-16
WIP: publish immutable UTC batch timetable discovery

Roasbeef:
In this PR, we add immutable UTC batch timetable discovery to the operator information schema. Clients can identify a published schedule version and derive upcoming registration windows independently of process uptime.

This is the first draft in a client stack. The follow-up draft carries explicit service authorization, participation deadlines, SDK/CLI support, and the in-progress durable round runtime. This PR is a WIP handoff, not a merge-readiness claim.

The commits separate schema changes from generated bindings. Local discovery validation was performed during development; current-stack checks and independent review will be recorded on the follow-up draft.

Native stack order: #1333 → #1335 → #1336 → #1334. Follow-up: #1335 (`codex/batch-runtime-foundation-wip`).


Roasbeef:
## Handoff: purpose, boundaries, and next steps

This is the discovery layer of the native GitHub stack created with `gh stack`. Read the client stack in this order: #1333 → #1335 → #1336 → #1334. The branch bases follow that order; the later diffs exclude earlier layers.

### Why this layer exists

A client needs a published registration window that does not move when the operator restarts or a preceding round runs long. An interval measured from process startup cannot supply that contract. The discovery fields expose an immutable timetable version, an anchor, window parameters, and the next opportunity. Version plus slot index gives a stable identity; the timestamp is derived from the timetable rather than becoming a second mutable source of identity.

The advertised start means registration closes and participation begins. It is not a guarantee that a commitment transaction is broadcast at that instant. Activation avoids advertising a partially elapsed initial window, and the participation deadline bounds the subsequent interactive attempt.

### What this PR does and does not establish

The schema and generated bindings let clients discover the timetable. The admission-enabled field keeps publication distinct from enforcement: receiving a timetable is not authorization to join a slot or execute an immediate attempt. This PR does not reserve funding, enforce capacity, sign transactions, or add a scheduler to the client. Those belong to later layers and the remote service implementation.

The source schema and generated protobuf output are separate commits. The merge in the branch reconciles additive discovery fields with the existing client policy fields; it is not a second copy of the later participation implementation.

### Review and continuation

Review field semantics and numbering first, especially activation, closing-time exclusivity, and participation expiry. Confirm that older clients can ignore the additive fields, and that consumers never interpret discovery alone as admission. The next contract layer (#1336) adds explicit operation consent and service outcomes; the final integration layer (#1334) consumes them.

Development previously exercised discovery/schema consumers. Current handoff validation covers the complete stack, not a fresh end-to-end deployment of this layer alone. Keep the draft status until compatibility and the complete service acceptance run have direct evidence. The independent review of the runtime is recorded on #1334; it does not constitute approval of this entire stack.

PR #1334, opened 2026-09-16
WIP: explicit batch participation and durable round client

Roasbeef:
In this PR, we carry explicit service selection, operation identity, expiry, fee caps, and fallback consent through the daemon, SDK, CLI, and round protocol. We also begin migrating the round client onto the existing durable actor runtime so checkpoints, buffered domain writes, outgoing effects, and message consumption share the runtime's transaction boundary.

This is a **WIP handoff**, stacked on #1336 in native stack #1337 (#1333 → #1335 → #1336 → #1334). The daemon uses the native mailbox, and the commits include full-state TLV codecs, cold signing recovery, and one-time legacy checkpoint import. The synchronous FSM runner and shared tree codec are in #1335; public service contracts and their bindings are in #1336. Generated RPC and SQL artifacts follow the handwritten definitions in a separate commit; review and build the branch tip.

The continuation guide is [docs/batch-participation-handoff.md](https://github.com/lightninglabs/wavelength/blob/codex/batch-participation-wip/docs/batch-participation-handoff.md).

Remaining work:

- Replace the custom service journal/startup ownership lookup with a native-checkpoint projection, including in-flight inputs.
- Prove production SQL atomicity and shared-outbox behavior under failure before/after commit and delivery. Contract-double tests do not establish those boundaries.
- Complete connected daemon/SDK/CLI flows, randomized workload integration, and P-to-Go replay against the actual runtime, including uncertain signing.
- Resolve the independent review findings after publication and complete acceptance before leaving draft status.

Current local validation (2026-09-16): changed-code lint passed; race tests passed for `round`, `db/actordelivery`, `db`, `waved`, `wallet`, `vtxo`, `lib/types`, `lib/treecodec`, `sdk/ark`, and `cmd/wavecli/waveclicommands`. The inline FSM suite passes when invoked through the parent module dependency graph. Formatting and whitespace checks pass. Structural scanning completed with style warnings. CI and whole-feature readiness are not claimed.

Independent review is complete for the bounded runtime pass. The detailed handoff comment below records two unresolved source-traced findings: interrupted first-turn signer creation and delayed signature delivery after expiry. No new fault-injection reproductions or fixes are included.


Roasbeef:
## Detailed WIP handoff and independent review

This is the integration layer of the native `gh stack` client stack: #1333 → #1335 → #1336 → **#1334**. The runner/tree-codec foundation and service contracts now have separate PRs. The final source tree is byte-for-byte unchanged by that history split. The remaining diff is still substantial: full-state and effect codecs, their tests, participation wiring, and runtime adoption are real new code, not duplicate lower-stack commits.

### Why migrate to the existing durable actor runtime

The earlier implementation retained domain checkpoints and a separate service ownership journal, then reconstructed work at startup. Those records could preserve useful facts without making consumption of an incoming message atomic with the next FSM state and its outgoing effects. A crash between those steps could leave the system unable to tell which transition or handoff had completed.

The native actor runtime already provides a persistent mailbox, transaction execution, checkpoints, and an outbox. This PR adapts the round client to that runtime. It stages a before-image before external work, buffers domain writes and effects, and commits them with the resulting snapshot and message acknowledgement. Failed or panicked turns must reload before later input observes tentative in-memory state. The synchronous runner from #1335 keeps the whole internal-event chain inside the owning actor turn; it is not a replacement actor framework.

### What is wired today

The daemon starts `NewDurableRoundClientActor`, registers the typed service separately from the outbox target, and joins the worker before closing dependencies and the database. Snapshots retain round state, fixed per-round policy, pending quotes, and absolute timer deadlines. If no native checkpoint exists, restart imports legacy relational round checkpoints. Cold signing recovery retains ownership and waits for an authoritative matching dead-round report rather than fabricating resumed signer sessions.

The custom `ServiceOperationStore` is still supplied, and VTXO startup cleanup still queries it. Removing it is unfinished work, not cleanup that can safely happen immediately. Its replacement must derive ownership from the native checkpoint and staged inputs before any cleanup releases a reservation. General successful mid-signing recovery is also not implemented: cold recovery currently ignores non-dead statuses and confirmation events while retaining ownership.

### Independent review: two unresolved blockers

**1. An interrupted first signing turn can restore a non-cold state.** `round/durable_behavior.go:151-162` marks in-flight snapshots cold only for `CommitmentTxReceivedState` and `CommitmentTxValidatedState`. A normal BatchInfo turn starts from `RoundJoinedState`, then synchronously reaches external `CreateSessions` through intermediate states (`round/transitions.go:2029,2091,2656,2856`). A crash after session creation but before commit leaves the earlier before-image, which can replay the input and create another ceremony. This is source-traced duplicate session creation, not a demonstrated nonce-reuse or fund-loss finding. Next: classify uncertainty using the staged input and target round, and inject a crash immediately after session creation in that first turn.

**2. Persisted signatures can bypass the expiry fence after delay.** `round/durable_effect_delivery.go:16-22` sends a stored server effect directly to `ServerConn`. A signature captured before the deadline can be delivered after it because of downtime or publisher backlog. `fenceServiceTransition` checks at effect production; that does not cover the later delivery boundary. Next: recheck the effect's round authorization before transport handoff, preserve ownership/reconciliation when suppressing it, and test committed-before-expiry/delivered-after-expiry through the production outbox.

Both findings were independently reported and verified against the source. Neither has a new fault-injection reproduction or fix in this WIP snapshot. They are reasons to keep the PR draft.

### Validation and continuation

Changed-code lint, formatting, whitespace checks, and ten affected-package race suites passed on 2026-09-16. The real SQLite restart test covers recovery of a pre-admission round and typed/outbox registration. Some rollback tests use a transaction-contract double; they do not prove production SQL atomicity. Structural scanning completed with style warnings.

Next, replace duplicate ownership recovery; prove domain/checkpoint/outbox/ack atomicity with the real database; exercise the shared publisher around crashes and duplicates; address the two findings above; and finish connected daemon/SDK/CLI, existing Rapid workload, and P/Go replay coverage. `docs/batch-participation-handoff.md` is the source map. The final independent review and whole-feature acceptance remain necessary before merge.

PR #1335, opened 2026-09-16
WIP: inline protocol turns and shared persisted tree codec

Roasbeef:
In this PR, we keep protocol transition evaluation inside the owning actor turn and share the existing persisted tree representation with actor checkpoints. The inline runner reuses the existing transition evaluator; it introduces no durable mailbox or persistence framework.

Native client stack: #1333 → #1335 → #1336 → #1334. See the detailed handoff comment below for the transaction-lifetime rationale, tests, limitations, and continuation. This remains a draft.


Roasbeef:
## Handoff: why an inline state-machine runner was introduced

Native client stack: #1333 → **#1335** → #1336 → #1334. This PR contains the synchronous runner and the shared persisted-tree codec. It does not introduce another durable actor runtime, mailbox, retry loop, or persistence framework.

### The transaction-lifetime problem

The existing round actor delegates protocol transitions to a state machine with its own worker goroutine. That is useful when the FSM has an independent lifecycle. It becomes a problem when one durable actor turn must produce a complete next state and its effects before the native runtime commits message consumption.

The required ordering is: receive one mailbox message; evaluate its complete internal-event chain; obtain the resulting state and effects; commit domain writes, the checkpoint, outgoing effects, and acknowledgement together. If evaluation continues in another goroutine after the caller returns or its wait is cancelled, that ordering is no longer established by the actor turn. Passing a transaction context to a worker also does not make the worker's lifetime equal to the transaction's lifetime.

`InlineStateMachine.AskEvent` runs that chain synchronously and returns an already-completed future. It reuses the existing transition evaluator rather than defining a second protocol. It starts no worker and dispatches no outbox itself. Its caller owns the checkpoint and effect commit. The original threaded runner remains available for independently driven state machines.

### What the runner does not solve

The runner does not make external wallet RPCs atomic with SQL. It does not persist signer sessions or permit nonce regeneration after an uncertain signing operation. The runtime integration therefore stages a before-image before external work and reloads a committed snapshot after a failed turn. Transitions must also avoid relying on in-place mutation being rolled back by the runner; durable recovery is the owning actor's responsibility.

The independent review found an important integration gap: one input can traverse several states synchronously and create signer sessions even though the staged state is still `RoundJoinedState`. The later runtime currently classifies only intermediate commitment states as uncertain. That defect is in the recovery classification, not evidence that the synchronous boundary makes signing transactional. See #1334 for the concrete follow-up and required crash test.

### Why extract the tree codec too

Full round checkpoints need the same tree representation already used by relational round storage. Keeping that encoder private to `db` would force duplication or couple protocol checkpointing back to its storage package. The extraction puts the representation in `lib/treecodec`, with the database delegating to it. The removed decoder-count test was moved into the new package, not discarded. Existing bounds and legacy representation compatibility remain part of the contract.

### Validation and next work

The inline-runner suite passes when invoked through the parent module dependency graph. Tree-codec and affected database tests passed in the full handoff run. Review cancellation, stop serialization, complete internal-event processing, and failure behavior here. The next layer is only service contracts; production runtime adoption and real SQL/outbox fault-injection remain in #1334. This foundation alone is not a durable-round implementation.

